PR #10, opened 2021-05-04
psp-698 Tenant Configuration

Fosol:
Simple tenant configuration provides a way to present different information based on the environment variable selecting a tenant.

Add unit tests
Fix vs-code Jest-Runner `Run` plugin

- I couldn't resolve the Jest-Runner `Debug` plugin regrettably.  Although I did find a way to make it run, it would never step into the code.  Additionally if it was configured to "work", the `Run` button would stop and it would also impact the normal command line testing...

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -56,6 +56,27 @@
       "type": "coreclr",
       "request": "attach",
       "processId": "${command:pickProcess}"
+    },
+    {
+      "name": "Debug Frontend Tests",
+      "type": "node",
+      "request": "launch",
+      "runtimeExecutable": "${workspaceRoot}/frontend/node_modules/.bin/react-scripts",
+      "args": [
+        "test",
+        "${fileBasenameNoExtension}",
+        "--runInBand",
+        "--no-cache",
+        "--watchAll=false"
+      ],
+      "cwd": "${workspaceRoot}/frontend",
+      "protocol": "inspector",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "env": {
+        "CI": "true"
+      },
+      "disableOptimisticBPs": true
     }
   ]
 }
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -59,7 +59,6 @@
     "yup": "0.28.3"
   },
   "devDependencies": {
-    "@babel/preset-typescript": "7.8.3",
     "@react-leaflet/core": "1.0.2",
     "@testing-library/dom": "7.22.1",
     "@testing-library/jest-dom": "4.2.4",
@@ -101,7 +100,6 @@
     "@types/yup": "0.26.32",
     "axios-mock-adapter": "1.17.0",
     "cross-env": "7.0.2",
-    "node-sass": "4.14.1",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.2",
     "enzyme-to-json": "3.4.4",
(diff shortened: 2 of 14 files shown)


.vscode/launch.json
```diff
@@ -56,6 +56,27 @@
       "type": "coreclr",
       "request": "attach",
       "processId": "${command:pickProcess}"
+    },
+    {
+      "name": "Debug Frontend Tests",
```
Fosol, line 61: This allows us to debug the currently open test file.  Regrettably it does run all the tests in the file instead of only the one you want.

frontend/package.json
```diff
@@ -59,7 +59,6 @@
     "yup": "0.28.3"
   },
   "devDependencies": {
-    "@babel/preset-typescript": "7.8.3",
```
Fosol, line 62: Didn't appear to be required.

frontend/src/components/layout/Header/ErrorModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * ErrorModal component that displays an error message and information in a modal.
+ * @param param0 ErrorModal component properties.
+ * @param param0.errors An array of errors.
+ * @param [param0.show] Whether to show the model.
+ * @param param0.setShow A function to set the show value.
+ * @returns ErrorModal component.
+ */
+export const ErrorModal = ({ errors, show, setShow }: IErrorModalProps) => {
```
Fosol, line 23: Moved this into its own file.  I would have preferred to have the `show, setShow` as internal properties.  But I couldn't figure out a good way to control when to show/hide it if I did that.

devinleighsmith: While I think this is a good change to have this in its own file, does this have anything to do with this PR? This doesn't seem related to tenant configuration.

frontend/src/components/layout/Header/ErrorModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  show?: boolean;
+  // Change the visible state.
+  setShow: (show: boolean) => void;
+}
+
+/**
+ * ErrorModal component that displays an error message and information in a modal.
+ * @param param0 ErrorModal component properties.
+ * @param param0.errors An array of errors.
+ * @param [param0.show] Whether to show the model.
```
devinleighsmith, line 19: is something weird going on with your autogenerated comments?

devinleighsmith (approved):

frontend/src/components/layout/Header/ErrorModal.tsx
Fosol: I manually added the second two based on jsdoc.  Thinking about it now, I probably don't need to as they will inherit the documentation from the interface automatically.

frontend/src/components/layout/Header/ErrorModal.tsx
Fosol: Not related to tenant, but some tech-debt that is small enough and touches the components I'm working on.  I could have done two PRs, one for the config and then one for the tech-debt, but that would be more effort.  Not sure the value of the extra effort is critical in this case.  The only functional change is to dynamically display the appropriate title based on the tenant config.

frontend/src/components/layout/Header/ErrorModal.tsx
devinleighsmith: what is with the []?

Fosol: optional parameter

PR #11, opened 2021-05-04
Update keycloak, react-keycloak. Includes fixes for breaking changes.

devinleighsmith:
as part of PSP-663.

releasing a commit for just this change to facilitate rollback if necessary.

Completed manual testing of changes in local environment.
1) Tested normal login/logout process.
2) Tested token timeout logic.

Need to test further in dev environment.


devinleighsmith:
Likely need to add a test to the eventhandler as well, but we'll see.

Fosol (approved):
Seems like the upgrade went fairly smooth

PR #14, opened 2021-05-06
PSP-676 Add Tenant Logos and text

Fosol:
When an environment uses the MOTI tenant they will now see the appropriate text, title, and logos.


frontend/src/components/common/BCGovLogo.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import BClogoUrl from 'assets/images/logo-banner.svg';
+
+/**
+ * BC Gov Logo image.
+ * @param props BCGovLogo properties.
+ * @returns BCGovLogo component.
+ */
+export const BCGovLogo = (
```
Fosol, line 9: Reused component extracted into its own file.

frontend/src/features/account/Login.tsx
```diff
@@ -104,4 +95,14 @@ const Login = () => {
   );
 };
 
+//check to see if user is using Internet Explorer
+//as their browser
+const usingIE = () => {
```
Fosol, line 100: Moved to bottom of file so that the component in this file is the first and primary function.

frontend/src/tenants/Logo.tsx
```diff
@@ -0,0 +1,30 @@
...
+  // Whether to include the logo with text.
+  withText?: boolean;
+}
+
+/**
+ * A Logo component provides an image for the current tenant.
+ * @param param0 Logo component properties.
+ * @returns Image containing the
+ */
+export const Logo = ({ withText = false, ...rest }: ILogoProps) => {
```
Fosol, line 18: A new component to simplify using the appropriate tenant logo.

jdtoombs (approved):

frontend/src/features/account/Login.tsx
```diff
@@ -31,6 +25,8 @@ const Login = () => {
   const activated = useSelector<RootState, IGenericNetworkAction>(
     state => (state.network as any)[actionTypes.ADD_ACTIVATE_USER] as IGenericNetworkAction,
   );
+  const { login } = tenant();
```
devinleighsmith, line 28: premature optimization, but having to re-evaluate tenant every time it is used isn't ideal given that the environment variables shouldn't change when the application is running. This could be memoized or changed to a globally executed function and only store/export the result.

frontend/src/components/layout/Header/EmptyHeader.tsx
```diff
@@ -21,25 +14,29 @@ const EmptyHeader = () => {
...
         </a>
         <VerticalBar />
-        <img className="pims-logo" src={PIMSlogo} height="50" alt="PIMS logo" />
+        <Logo />
       </Navbar.Brand>
       <Nav className="title mr-auto">
         <Nav.Item>
-          <h1 className="longAppName">Property Inventory Management System</h1>
+          <h1 className="longAppName">{tenant().title}</h1>
           <h1 className="shortAppName">PIMS</h1>
```
devinleighsmith, line 25: this is correct?

frontend/src/features/account/Login.tsx
Fosol: I thought of this, but wasn't sure the best approach in this case.  I see what I can do to improve this.

devinleighsmith (approved):

frontend/src/features/account/Login.tsx
Fosol: Take another look at my implementation.  I think it is much better now.

frontend/src/components/layout/Header/EmptyHeader.tsx
Fosol: I've added both to the config now.

frontend/src/App.tsx
```diff
@@ -13,13 +13,15 @@ import OnLoadActions from 'OnLoadActions';
...
 import FilterBackdrop from 'components/maps/leaflet/FilterBackdrop';
+import { setTenant } from 'tenants';
 
 export const store = configureStore();
 
 const App = () => {
   const keycloakWrapper = useKeycloakWrapper();
   const keycloak = keycloakWrapper.obj;
   const dispatch = useDispatch();
+  React.useMemo(() => setTenant(), []);
```
Fosol, line 24: Run once

frontend/src/tenants/CITZ/config.ts
```diff
@@ -1,16 +1,21 @@
...
-    favicon: '/tenants/CITZ/favicon-16x16.ico',
-    image: '/tenants/CITZ/logo_only.png',
-    imageWithText: '/tenants/CITZ/logo_with_text.png',
-  },
-  login: {
-    title: 'Search and visualize government property information',
-    heading: 'PIMS enables you to search properties owned by the Government of British Columbia',
-    body:
-      'The data provided can assist your agency in making informed, timely, and strategic decisions on the optimal use of real property assets on behalf of the people and priorities of British Columbia.',
+  ...defaultTenant,
```
Fosol, line 5: Inherit from the default

frontend/src/tenants/tenant.ts
```diff
@@ -18,20 +19,25 @@ export const defaultTenant = {
   },
 };
 
+// This ensures the tenant information is set once.
+export let tenant: ITenantConfig = defaultTenant;
```
Fosol, line 23: global variable that is configured through the `setTenant()` function

PR #17, opened 2021-05-06
Lint

Fosol:
For some reason our linting isn't catching everything.  The pipeline caught this and won't deploy because of it.


openshift/4.0/templates/jenkins/jenkins.yaml
```diff
@@ -0,0 +1,139 @@
+kind: Template
```
Fosol, line 1: Adding jenkins deployment config to repo to make it easy to change the resources.

devinleighsmith (approved):

PR #18, opened 2021-05-07
PSP_659 replace access request redux components with slice. (#2)

devinleighsmith:
Update action creators and other consumers to use slice.
Add tests for new access request reducer.
remove all old access request redux components.
update store generation to allow more concise typings of selector.


frontend/src/constants/reducerTypes.ts
```diff
@@ -7,7 +7,7 @@ export const PARCEL = 'parcel';
 export const LOOKUP_CODE = 'lookupCode';
 
 // access requests
-export const ACCESS_REQUEST = 'accessRequest';
+export const ACCESS_REQUEST = 'accessRequests';
```
devinleighsmith, line 10: May need another solution for this. The redux repo advocates for just naming slices with a string and then using that exact string when setting up each reducer in the store. This doesn't seem as safe as using a static string in both places.

frontend/src/store/slices/accessRequests/accessRequestsActionCreator.spec.tsx
```diff
@@ -0,0 +1,305 @@
...
+const mockStore = configureMockStore([thunk]);
+const getStore = (values?: any) => {
+  currentStore = mockStore(values ?? {});
+  return currentStore;
+};
+const getWrapper = (store: any) => ({ children }: any) => (
+  <Provider store={store}>{children}</Provider>
+);
+
+describe('useAccessRequests functionality', () => {
```
devinleighsmith, line 36: update the action creators test to consume the actions when formatted as a hook. Also increase the resolution of test expectations by checking the result of the action being executed against the store, rather than just the number of dispatch calls.

frontend/src/store/slices/accessRequests/accessRequestsReducer.test.tsx
```diff
@@ -0,0 +1,128 @@
+import { mockAccessRequest } from 'mocks/filterDataMock';
+import { accessRequestsSlice, initialState } from 'store/slices/accessRequests/accessRequestsSlice';
+
+const accessRequestReducer = accessRequestsSlice.reducer;
+describe('access request reducer functionality', () => {
```
devinleighsmith, line 5: Created these reducer tests to validate that no functionality was changed between the old and new reducer by:
1) create these tests, execute against the old reducer.
2) create the new reducer.
3) swap these tests to point to the new reducer.
4) delete the old reducer.

frontend/src/store/slices/accessRequests/accessRequestsSlice.ts
```diff
@@ -0,0 +1,90 @@
...
+  pageSize: MAX_ACCESS_RESULTS_PER_PAGE,
+  pageIndex: 0,
+};
+/**
+ * The access request slice manages the current access request, a list of all active access requests, and methods to sort and filter this list.
+ */
+export const accessRequestsSlice = createSlice({
+  name: ACCESS_REQUEST,
+  initialState: initialState,
+  reducers: {
```
devinleighsmith, line 46: immer based version of the old access request reducer.

frontend/src/store/slices/accessRequests/useAccessRequests.ts
```diff
@@ -0,0 +1,166 @@
...
+} from './accessRequestsSlice';
+
+export const useAccessRequests = () => {
+  const dispatch = useAppDispatch();
+  /**
+   * Get the fetchCurrent function with returns the current access request for the current user.
+   * @returns The dispatchable action which will return the access request if one exists, or 204 if one doesn't
+   */
+  const fetchCurrent = React.useCallback(async (): Promise<IAccessRequest> => {
+    dispatch(request(actionTypes.GET_REQUEST_ACCESS));
```
devinleighsmith, line 25: Note that these dispatches are still calling the old network reducer and so appear in the old format still. Will need to update this when that reducer is converted to the new format.

devinleighsmith:
Note that I haven't updated the root reducer yet, that'll be the last step when all of the existing reducers are converted into slices.

devinleighsmith:
As mentioned in the title of this PR, this is only ONE of the reducers being converted into a slice. As a result, large portions of the app as well as this PR still have some reducers written in the traditional way. Those will be slowly corrected as I publish more PRs.

frontend/src/constants/reducerTypes.ts
Fosol: Redux toolkit does this in a much cleaner way.  Removing the need to maintain these odd little files with strings.  Just move the value directly into your slice.

frontend/src/store/slices/accessRequests/accessRequestsSlice.ts
```diff
@@ -0,0 +1,90 @@
...
+  selections: [],
+  accessRequest: null,
+  pageSize: MAX_ACCESS_RESULTS_PER_PAGE,
+  pageIndex: 0,
+};
+/**
+ * The access request slice manages the current access request, a list of all active access requests, and methods to sort and filter this list.
+ */
+export const accessRequestsSlice = createSlice({
+  name: ACCESS_REQUEST,
```
Fosol, line 44: Why not just put the string value here?  What value is there in having a separate file full of strings that shouldn't ever be referenced by anything else anyways?

frontend/src/reducers/rootReducer.ts
```diff
@@ -24,12 +23,13 @@ import filterSlice from './filterSlice';
 import agencyReducer from './agencyReducer';
 import agencyDetailReducer from './agencyDetailReducer';
 import parcelLayerDataSlice from './parcelLayerDataSlice';
+import { accessRequestsSlice } from 'store/slices/accessRequests';
 
 export const reducerObject = {
```
Fosol, line 28: Use the `configureStore({reducer: { name: reducer, name: reducer } })` pattern.  Remove the need to have a separate string file.

frontend/src/store/slices/accessRequests/useAccessRequests.ts
Fosol: I suspect this answers my questions above about removing the extra string file.  It's in progress.

Fosol (approved):

jdtoombs (approved):

goemen (approved):

PR #21, opened 2021-05-10
PSP-665 Update Nuget Packages

Fosol:
Packages have been updated to the latest versions without upgrading to .NET5


devinleighsmith (approved):

Fosol:
Nuget package manager does a good job of most of the work.  Also the language and tools are more mature than javascript related code.  For example `obsolete` code is tagged and allows for slower transition.

PR #24, opened 2021-05-11
PSP-687 Add API version component

Fosol:
The App footer will now display the API version information stored with the compiled metadata.

- Update backend version information
- Add API health endpoints hook

image


backend/api/Controllers/HealthController.cs
```diff
@@ -25,10 +24,8 @@ public class HealthController : ControllerBase
         /// Creates a new instances of a HealthController class, initializes it with the specified arguments.
         /// </summary>
         /// <param name="environment"></param>
-        /// <param name="logger"></param>
-        public HealthController(IWebHostEnvironment environment, ILogger<HealthController> logger)
```
devinleighsmith, line 29: I'm assuming this is unrelated. What is the reason for this change?

backend/dal.keycloak/Pims.Dal.Keycloak.csproj
```diff
@@ -3,8 +3,8 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <OutputType>Library</OutputType>
-    <Version>0.1.0.0-alpha</Version>
-    <AssemblyVersion>0.1.0.0</AssemblyVersion>
+    <Version>1.0.0.0</Version>
```
devinleighsmith, line 6: everything else is 1.10.0.0 but this is 1.0.0.0?

Fosol: It would be nice to automate this in a build some how, but I doubt we'll have time to dedicate to it.

backend/api/Controllers/HealthController.cs
Fosol: It's related in the sense this is the controller for the endpoint for the version information.  I've cleaned up the code to remove unused variables.

frontend/src/components/layout/Footer/Footer.tsx
```diff
@@ -1,19 +1,60 @@
...
+      <a href="http://www.gov.bc.ca/gov/content/home/copyright">Copyright</a>
+      <ApiVersionInfo />
+    </FooterStyled>
   );
 }
 
+/**
+ * Styled footer component.
+ */
+const FooterStyled = styled('div')`
```
devinleighsmith, line 24: I think eventually we want to try and promote reuse a bit more by putting splitting of some of these components and placing them in a shared location (ie, the a portion). Not necessarily required here however.

frontend/src/hooks/pims-api/useApiHealth.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    mockAxios.onGet('health/env').reply(200, defaultVersion);
+    mockAxios.onGet('health/live').reply(200, defaultLive);
+    mockAxios.onGet('health/ready').reply(200, defaultReady);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('Has endpoints', () => {
```
devinleighsmith, line 19: I'm not a big fan of tests like this as you are essentially testing the structure of the code and not the functionality, but personal preference maybe.

frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * Provides a way to display the API version information.
+ * Makes an AJAX request to the API for the version information.
+ * @returns ApiVersionInfo component.
+ */
+export const ApiVersionInfo = () => {
+  const api = useApiHealth();
+  const [version, setVersion] = React.useState<IApiVersion>();
+
+  React.useEffect(() => {
+    api.getVersion().then(response => {
```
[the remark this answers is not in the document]
Fosol: I think changing to `<div className="version">{`v${version?.version}`}</div>` would solve the loading issue.

Are you suggesting caching the api endpoint?  Technically there is little value in doing so.  Or do you mean using redux or context to store the value so that it's only called in one place?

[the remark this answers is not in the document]
devinleighsmith: I don't think caching, redux or context needs to be used as long as we are sure this is the only place we will be using this component/the version api call.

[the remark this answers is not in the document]
devinleighsmith: Less concerned about the above than I am about the missing error handling logic.

frontend/src/hooks/pims-api/useApiHealth.test.ts
```diff
@@ -0,0 +1,62 @@
+import { renderHook } from '@testing-library/react-hooks';
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+import { useApiHealth } from '.';
+
+const mockAxios = new MockAdapter(axios);
+
+describe('useApiHealth testing suite', () => {
```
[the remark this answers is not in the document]
devinleighsmith: in summary - nothing needs to be done in this case, just want to make the intended use of the tool clear.

frontend/src/components/common/ApiVersionInfo.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const result = renderer.create(<ApiVersionInfo />).toJSON();
+    expect(result).toMatchSnapshot();
+  });
+
+  it('Displays version information', async () => {
+    await act(async () => {
+      const { findByTestId } = render(<ApiVersionInfo />);
+      const element = await findByTestId('version');
+      expect(element).toContainHTML(`v${defaultVersion.version}`);
+      expect(mockGetVersion).toHaveBeenCalledTimes(4);
```
Fosol: @devinleighsmith This gets called 4 times because there are two tests in this suite.  I thought `mockClear()` would resolve this, but it doesn't appear to work.  If I use `mockReset()` then the whole thing fails to work.
Additionally, it shouldn't be getting called 2 times each test either.  For some reason the `useApi` hook fires twice, even though it uses empty array `[]`.  I'm not sure how to debug.

frontend/src/components/layout/Footer/__snapshots__/Footer.test.tsx.snap
```diff
@@ -1,40 +1,91 @@
...
-    onKeyDown={[Function]}
   >
     Copyright
   </a>
-</nav>
+  <div
+    className="version"
+    data-testid="version"
+  >
+    vundefined
```
Fosol, line 88: For some reason this is undefined which is incorrect as the mock would have returned a value.

frontend/src/components/common/ApiVersionInfo.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      expect(result).toMatchSnapshot();
+    });
+  });
+
+  it('Displays version information', async () => {
+    await act(async () => {
+      const { findByTestId } = render(<ApiVersionInfo />);
+      const element = await findByTestId('version');
+      expect(element).toContainHTML(`v${defaultVersion.version}`);
+      expect(mockGetVersion).toHaveBeenCalledTimes(2);
```
Fosol, line 38: This shouldn't have been called 2 times.  But for some reason `useApi` is, even though it has empty array `[]`.

frontend/src/components/common/__snapshots__/ApiVersionInfo.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ApiVersionInfo suite Displays version component 1`] = `null`;
```
Fosol, line 3: I don't get this snapshot?

frontend/src/layouts/__snapshots__/EmptyLayout.test.tsx.snap
```diff
@@ -74,34 +131,36 @@ exports[`Empty Layout renders 1`] = `
...
           href="http://www.gov.bc.ca/gov/content/home/copyright"
         >
           Copyright
         </a>
-      </nav>
+        <div
+          class="version"
+          data-testid="version"
+        >
+          vundefined
```
Fosol, line 161: I don't get why this is undefined?

frontend/src/components/common/ApiVersionInfo.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      expect(result).toMatchSnapshot();
+    });
+  });
+
+  it('Displays version information', async () => {
+    await act(async () => {
+      const { findByTestId } = render(<ApiVersionInfo />);
+      const element = await findByTestId('version');
+      expect(element).toContainHTML(`v${defaultVersion.version}`);
+      expect(mockGetVersion).toHaveBeenCalledTimes(2); // TODO: Figure out why this is getting called twice.
```
Fosol, line 38: I don't get why this gets called twice

Fosol: @devinleighsmith any insight into this one?

frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const getVersion = async () => {
+      const response = await api.getVersion();
+      setVersion(response.data);
+    };
+    getVersion();
+  }, [api]);
+
+  return (
+    <div className="version" data-testid="version">
+      {version?.version ? `v${version?.version ?? ''}` : 'unavailable'}
```
Fosol, line 23: It now will indicate `unavailable` if the api call fails.

frontend/src/components/common/__snapshots__/ApiVersionInfo.test.tsx.snap
devinleighsmith: yeah, this doesn't seem right

devinleighsmith: once this gets cleaned up I think I'm good.

devinleighsmith (approved):

jdtoombs (approved):

PR #27, opened 2021-05-13
PSP-690 API Response Logging

Fosol:
To assist with debugging issues response will now be logged.

Additionally, the body of both requests and responses will be logged if the logging level is set at Debug.


backend/api/Pims.Api.csproj
```diff
@@ -40,6 +40,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Identity.Core" Version="5.0.5" />
     <PackageReference Include="Microsoft.Extensions.PlatformAbstractions" Version="1.1.0" />
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="2.0.0" />
```
Fosol, line 43: Used to deal with memory leaks with the request and response streams.

PR #28, opened 2021-05-14
replace users reducer with slice. remove unused code.

devinleighsmith:


Fosol (approved):
Looks better every day

PR #29, opened 2021-05-14
Psp 659 parcels

devinleighsmith:
replace all parcel, building, property reducers with a slice.
remove unused code
move all interfaces defined in parcelsactioncreator to the root /interfaces folder
required test updates


frontend/src/components/maps/hooks/useActiveFeatureLayer.test.tsx
```diff
@@ -27,7 +27,7 @@ let findOneWhereContains = jest.fn();
 
 const mockStore = configureMockStore([thunk]);
 const history = createMemoryHistory();
-const getStore = (values?: any) => mockStore(values ?? { parcel: { draftParcels: [] } });
+const getStore = (values?: any) => mockStore(values ?? { properties: { draftParcels: [] } });
```
Fosol, line 30: Happy to see little name changes to be more aligned.

frontend/src/features/mapSideBar/SidebarContents/subforms/BuildingReviewPage.test.tsx
```diff
@@ -25,17 +25,115 @@ const lCodes = {
 
 const store = mockStore({
   [reducerTypes.LOOKUP_CODE]: lCodes,
-  [reducerTypes.PARCEL]: { parcels: [] },
+  [reducerTypes.PROPERTIES]: { parcels: [] },
 });
 
+export const mockDetails = [
```
Fosol, line 31: No change required for this PR.
I see these types of mock datasets everywhere.  I think we probably should come up with a strategy to put them all in one place and make them generic/modifiable.

frontend/src/interfaces/address.ts
```diff
@@ -0,0 +1,9 @@
+export interface IAddress {
```
Fosol, line 1: I generally lean towards naming the file the same as the primary export, but it isn't necessary.

frontend/src/interfaces/parcel.ts
```diff
@@ -0,0 +1,21 @@
+import { IAddress, IBuilding, IEvaluation, IFiscal, IProperty } from 'interfaces';
+
+export interface IParcel extends IProperty {
+  pid?: string;
+  pin?: number | '';
```
Fosol, line 5: I'd love to find a way where we don't have to muddy our interfaces to support Formik/inputs `number | ''`

Fosol (approved):
Big one!

PR #32, opened 2021-05-15
Psp 659 network reducer replaced with slice

devinleighsmith:


Fosol (approved):
Happy to see us moving away from separate string constant files that are just used for key naming.  Also happy to see the switch to styled components!

PR #35, opened 2021-05-18
1) Remove label on images causes the agent templates in the configmap…

devinleighsmith:
…s from being overwritten

2) Remove hardcoded sonar token, prefer the use of secrets
3) Remove exec shell which was causing build failures.

### DIFF
--- a/openshift/4.0/scripts/oc-sonar-dotnet.sh
+++ b/openshift/4.0/scripts/oc-sonar-dotnet.sh
@@ -31,7 +31,6 @@ TARGET=${TARGET:-dev}
 SONAR_PROJECT_KEY=${SONAR_PROJECT_KEY:-pims-api-${TARGET}}
 SONAR_PROJECT_NAME=${SONAR_PROJECT_NAME:-PIMS Backend [${TARGET}]}
 SONAR_URL=${SONAR_URL:-https://sonarqube-3cd915-tools.apps.silver.devops.gov.bc.ca}
-SONAR_TOKEN=${SONAR_TOKEN:-872c1fef72c4d4fff20bb98aacdd0df95e6cc43b}
 
 BACKEND_DIR="${BACKEND_DIR:-../../../backend}"
 TEST_DIR="${TEST_DIR:-${BACKEND_DIR}/tests/unit}"
@@ -114,4 +113,3 @@ display_helper "${CMD_SONAR_BEGIN}" "${CMD_BUILD}" "dotnet test" "${CMD_SONAR_EN
 #
 # The final `coverage.opencover.xml` is the one we want
 # -----------------------------------------------------------------------------------------
-exec $SHELL;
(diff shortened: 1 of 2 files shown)


openshift/4.0/scripts/oc-sonar-dotnet.sh
```diff
@@ -113,5 +112,4 @@ display_helper "${CMD_SONAR_BEGIN}" "${CMD_BUILD}" "dotnet test" "${CMD_SONAR_EN
 #   - copy results to common folder "../TestResults"
 #
 # The final `coverage.opencover.xml` is the one we want
-# -----------------------------------------------------------------------------------------
-exec $SHELL;
```
devinleighsmith, line 117: @sookeke Is it alright if I remove this? this was causing the build to fail with:
**This account is currently not available.**

openshift/4.0/scripts/oc-sonar-dotnet.sh
```diff
@@ -31,7 +31,6 @@ TARGET=${TARGET:-dev}
 SONAR_PROJECT_KEY=${SONAR_PROJECT_KEY:-pims-api-${TARGET}}
 SONAR_PROJECT_NAME=${SONAR_PROJECT_NAME:-PIMS Backend [${TARGET}]}
 SONAR_URL=${SONAR_URL:-https://sonarqube-3cd915-tools.apps.silver.devops.gov.bc.ca}
-SONAR_TOKEN=${SONAR_TOKEN:-872c1fef72c4d4fff20bb98aacdd0df95e6cc43b}
```
devinleighsmith, line 34: The rocketchat and sonar token secrets were not set up correctly in tools which is possibly why this was introduced. I'd rather not have this as I don't like exposing secrets content in github as this repository is not private. Also the build probably SHOULD fail if the secrets aren't set up properly.

openshift/4.0/templates/jenkins-slaves/jenkins-slave-generic-template.yaml
```diff
@@ -14,7 +14,6 @@ objects:
     metadata:
       labels:
         build: "${NAME}"
-        role: jenkins-slave
```
devinleighsmith, line 17: This role causes Jenkins to overwrite the config map agent configuration with empty defaults. We don't want this behaviour since we are using config maps (default resources cause our build to fail), so I see no issues removing this.

openshift/4.0/scripts/oc-sonar-dotnet.sh
sookeke: @devinleighsmith  Yes, I remember making this changes and hardcoding the Sonar key. This however should be store as OC secret. Go ahead and make the changes where needed. Thank you

sookeke:
@devinleighsmith  I can not access the full coverage on codecov. Maybe I do not have the right access. Says
Oops, something went wrong

Fosol (approved):

PR #37, opened 2021-05-19
Sort imports, add eslint based sort plugin.

devinleighsmith:


devinleighsmith:
I realized that this will create nightmare merges for anyone working on the FE. Planning on merging this when we are closer to the cutoff for this sprint.

frontend/package.json
```diff
@@ -101,6 +101,7 @@
     "enzyme-to-json": "3.6.2",
     "eslint-config-prettier": "8.3.0",
     "eslint-plugin-prettier": "3.4.0",
+    "eslint-plugin-simple-import-sort": "^7.0.0",
```
Fosol, line 104: ^7.0.0

frontend/src/components/Table/ColumnSort.tsx
```diff
@@ -1,8 +1,9 @@
+import variables from '_variables.module.scss';
```
Fosol, line 1: No change required for this PR.  This looks like something that could be replaced by the styled theme variables now.

Fosol (approved):
Why did all the imports get sorted?  Will they be sorted automatically going forward?  Does it require a vscode extension?

devinleighsmith:
> Why did all the imports get sorted? Will they be sorted automatically going forward? Does it require a vscode extension?

Not sure how to answer the first question, the entire purpose of this PR is to sort the imports. Maybe the question is why is that important? The obvious answer is neatness and consistency. Less obvious is the import of external scss needs to be handled in a consistent order otherwise webpack will throw an error.

With this change imports should be sorted by eslint automatically, no extension required.

PR #39, opened 2021-05-19
PSP-694 Improve coverage

Fosol:
For some reason coverage is complaining about code comments.  Not sure how to resolve this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,4 +138,11 @@ convert: ## Convert Excel files to JSON
 	@echo "$(P) Convert Excel files to JSON..."
 	@cd tools/converters/excel; dotnet build; dotnet run;
 
-.PHONY: local setup restart refresh up down stop build rebuild clean client-test server-test pause-30 server-run db-migrations db-add db-update db-rollback db-remove db-clean db-drop db-seed db-refresh npm-clean npm-refresh keycloak-sync convert
+backend-coverage: ## Generate coverage report for backend
+	@echo "$(P) Generate coverage report for backend"
+	@cd backend; coverlet ./tests/unit/api/bin/Debug/netcoreapp3.1/Pims.Api.Test.dll --target "dotnet" --targetargs "test ./ --no-build" -o "./tests/TestResults/coverage.json" --exclude "[*.Test]*" --exclude "[*]*Model" --exclude-by-attribute "CompilerGenerated" -f json
+	@cd backend; coverlet ./tests/unit/dal/bin/Debug/netcoreapp3.1/Pims.Dal.Test.dll --target "dotnet" --targetargs "test ./ --no-build" -o "./tests/TestResults/coverage.xml" --exclude "[*.Test]*" --exclude "[*]*Model" --exclude-by-attribute "CompilerGenerated" --merge-with "tests/TestResults/coverage.json" -f cobertura
+	@cd backend; reportgenerator "-reports:./tests/TestResults/coverage.xml" "-targetdir:./tests/TestResults/Coverage" -reporttypes:Html
+	@cd backend; start ./tests/TestResults/Coverage/index.htm
+
+.PHONY: local setup restart refresh up down stop build rebuild clean client-test server-test pause-30 server-run db-migrations db-add db-update db-rollback db-remove db-clean db-drop db-seed db-refresh npm-clean npm-refresh keycloak-sync convert backend-coverage
--- a/Makefile
+++ b/Makefile
@@ -138,11 +138,26 @@ convert: ## Convert Excel files to JSON
 	@echo "$(P) Convert Excel files to JSON..."
 	@cd tools/converters/excel; dotnet build; dotnet run;
 
+backend-test: ## Run backend unit tests
+	@echo "$(P) Run backend unit tests"
+	@cd backend; dotnet test;
+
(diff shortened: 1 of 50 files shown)


Makefile
```diff
@@ -138,4 +138,11 @@ convert: ## Convert Excel files to JSON
 	@echo "$(P) Convert Excel files to JSON..."
 	@cd tools/converters/excel; dotnet build; dotnet run;
 
-.PHONY: local setup restart refresh up down stop build rebuild clean client-test server-test pause-30 server-run db-migrations db-add db-update db-rollback db-remove db-clean db-drop db-seed db-refresh npm-clean npm-refresh keycloak-sync convert
+backend-coverage: ## Generate coverage report for backend
```
Fosol, line 141: Quick way to generate the backend coverage reports.

backend/notifications/Models/EmailResponse.cs
```diff
@@ -10,16 +10,31 @@ namespace Pims.Notifications.Models
     public class EmailResponse
     {
         #region Properties
```
Fosol, line 12: After I added comments to this file coverage is complaining I don't have coverage for these comments... Which is idiotic...

backend/tests/core/Entities/AccessRequestHelper.cs
```diff
@@ -7,7 +7,6 @@ namespace Pims.Core.Test
     /// <summary>
     /// EntityHelper static class, provides helper methods to create test entities.
     /// </summary>
-    [ExcludeFromCodeCoverage]
```
Fosol, line 10: Moved to `EntityHelper` file.

backend/tests/core/Entities/EntityHelper.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    [ExcludeFromCodeCoverage]
+    public static partial class EntityHelper
+    {
+        /// <summary>
+        /// Generate a random rowversion.
+        /// </summary>
+        /// <param name="data">Data to convert into row version.</param>
+        /// <returns></returns>
+        public static byte[] GenerateRowVersion(string data)
```
Fosol, line 19: Simple way to generate a row version instead of using the same one everywhere.  This is useful for testing the templates as they use the rowversion to ensure the compiled template is unique.

backend/core/Extensions/GeometryExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static class GeometryExtensions
         /// <returns></returns>
         public static Polygon ToPolygon(this Envelope envelope)
         {
-            if (envelope != null && envelope.MinX <= envelope.MaxX && envelope.MinY <= envelope.MaxY)
```
devinleighsmith, line 17: what's going on here?

backend/dal/Migrations/20200722193504_Initial.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System;
 using Pims.Dal.Helpers.Migrations;
 using Microsoft.EntityFrameworkCore.Migrations;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Pims.Dal.Migrations
 {
+    [ExcludeFromCodeCoverage]
```
devinleighsmith, line 8: oh wow, of course this would get included, I suppose we'll get notified of missing coverage when we push PR's so hopefully we'll remember to do this in the future.

backend/core/Extensions/GeometryExtensions.cs
Fosol: After testing I determined it's impossible to have those conditions.  An envelope automatically reorganizes it's X,Y coords so that they don't conflict.  I tried a few dozen attempts to see if I could reach this path, but was unable.

backend/dal/Migrations/20200722193504_Initial.cs
Fosol: I've updated the script to auto include them for us `make db-add n=v01.11.00`

backend/notifications/Models/EmailResponse.cs
jdtoombs: Guess they really want to check our grammar 😛

jdtoombs (approved):

backend/core/Extensions/StringExtensions.cs
```diff
@@ -56,7 +56,7 @@ public static string FormatAsPostal(this string postal)
         /// <returns></returns>
         public static string LowercaseFirstCharacter(this string word)
         {
-            if (!String.IsNullOrWhiteSpace(word) && !char.IsUpper(word[0]))
+            if (!String.IsNullOrWhiteSpace(word) && char.IsUpper(word[0]))
```
Fosol, line 59: Fixed bug

scripts/db-migration.sh
```diff
@@ -6,11 +6,16 @@ FILE1=./Migrations/$(basename ./Migrations/*_$MNAME.cs);
 echo "Updating migration '$FILE1'";
 
 sed -i "2iusing Pims.Dal.Helpers.Migrations;" $FILE1;
+sed -i "3iusing System.Diagnostics.CodeAnalysis;" $FILE1;
```
Fosol, line 9: Exclude migrations from test coverage.

PR #40, opened 2021-05-19
PSP-660 update access request action creator, split out api. correspo…

devinleighsmith:
This is a proof of concept PR for the first stage of the refactor to pims api calls to separate the api calls from the async redux actions.

future PR's will cover the remainder of the async action creators in this style.

Separately, I'm planning on updating all of the async action creators to use either interceptors or a wrapped promise to make them less verbose.


frontend/src/store/slices/accessRequests/useAccessRequests.ts
```diff
@@ -1,31 +1,36 @@
 import { useAppDispatch } from './../../hooks';
 import { showLoading, hideLoading } from 'react-redux-loading-bar';
-import CustomAxios from 'customAxios';
-import { ENVIRONMENT } from 'constants/environment';
 import { AxiosResponse, AxiosError } from 'axios';
 import { IAccessRequest } from 'interfaces';
 import * as actionTypes from 'constants/actionTypes';
 import * as API from 'constants/API';
 import React from 'react';
+import { useApiAccessRequests } from './../../../hooks/pims-api/useApiAccessRequests';
```
Fosol, line 8: I think you can just go `hooks/pims-api`

Fosol (approved):
I like it.

frontend/src/store/slices/accessRequests/useAccessRequests.ts
devinleighsmith: Good point, I have this in a follow up PR.

PR #44, opened 2021-05-25
Psp 660 agencies

devinleighsmith:


frontend/src/hooks/pims-api/useApi.ts
```diff
@@ -1,13 +1,23 @@
...
-  return React.useMemo(() => CustomAxios({ baseURL: ENVIRONMENT.apiUrl }), []);
+export const useApi = (
+  axiosOptions: {
+    lifecycleToasts?: LifecycleToasts;
+    selector?: Function;
+    envelope?: typeof defaultEnvelope;
+    baseURL?: string;
+  } = {},
+) => {
+  return useDeepCompareMemo(
```
Fosol, line 17: Can you explain why this is needed?  Is it because we need a different memo for each axios options?

frontend/src/store/slices/agencies/useAgencies.ts
```diff
@@ -1,29 +1,22 @@
+import { useApiAgencies } from './../../../hooks/pims-api/useApiAgencies';
```
Fosol, line 1: Can be `hooks/pims-api`

frontend/src/store/slices/agencies/useAgencies.ts
```diff
@@ -1,29 +1,22 @@
...
 import { IPagedItems } from './../../../interfaces/pagedItems';
```
Fosol, line 2: Can be `interfaces/pagedItems`?

frontend/src/hooks/pims-api/useApi.ts
devinleighsmith: axiosOptions is an object, which will cause a render loop. Either need to make sure all callers only pass memoized objects or use value comparison here (instead of reference comparison).

PR #45, opened 2021-05-25
PSP-694 Add unit tests

Fosol:


backend/entities/Address.cs
```diff
@@ -103,7 +103,7 @@ public Address(string address1, string address2, string administrativeArea, Prov
         /// <returns></returns>
         public override string ToString()
         {
-            return String.Join(", ", new[] { this.Address1, this.Address2, this.AdministrativeArea, this.ProvinceId, this.Postal }.Where(s => String.IsNullOrWhiteSpace(s)));
+            return String.Join(", ", new[] { this.Address1, this.Address2, this.AdministrativeArea, this.ProvinceId, this.Postal }.Where(s => !String.IsNullOrWhiteSpace(s)));
```
Fosol, line 106: Fixed bug

backend/entities/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -19,11 +19,6 @@ public static class ProjectExtensions
         /// <returns></returns>
         private static List<string> AddProjectNumber(IEnumerable<string> projectNumbers, string projectNumber)
         {
-            if (projectNumber == null)
```
Fosol, line 22: Impossible to get this path with current implementation

backend/entities/Models/ProjectProperty.cs
```diff
@@ -40,6 +40,8 @@ public ProjectProperty(Views.Property property)
             Postal = property.Postal;
             Location = property.Location;
 
+            ProjectNumbers = property.ProjectNumbers;
```
Fosol, line 43: Added to enable cleaner testing.

backend/tests/core/Entities/BuildingHelper.cs
```diff
@@ -22,7 +23,40 @@ public static partial class EntityHelper
         /// <param name="agency"></param>
         public static Entity.Building CreateBuilding(int id, string projectNumber = null, string name = null, int lat = 0, int lng = 0, Entity.Agency agency = null)
         {
-            return CreateBuilding(id, projectNumber, name, lat, lng, agency);
+            projectNumber ??= $"p{id}";
```
Fosol, line 26: Fixed infinite loop bug

backend/tests/core/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    [ExcludeFromCodeCoverage]
+    public static class IEnumerableExtensions
+    {
+        /// <summary>
+        /// Lazy loads the enumerable to enable testing "IEnumerator IEnumerable.GetEnumerator()"
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns></returns>
+        public static IEnumerable AsWeakEnumerable(this IEnumerable source)
```
Fosol, line 17: Needed to test some paths.

backend/tests/unit/dal/Entities/AccessRequestAgencyTest.cs
```diff
@@ -0,0 +1,73 @@
...
+    [Trait("group", "entity")]
+    [ExcludeFromCodeCoverage]
+    public class AccessRequestAgencyTest
+    {
+        #region Tests
+        [Fact]
+        public void AccessRequestAgency_Id_Constructor()
+        {
+            // Arrange
+            var accessRequest = EntityHelper.CreateAccessRequest(5);
```
devinleighsmith, line 22: I'm generally nervous around using the same id for two different values in a test like this. What if we had some code that accidentally assigned an agency id to an access request id or vice versa? Nice to have unique ids for each object to prevent that.

backend/tests/unit/dal/Entities/Extensions/ParcelExtensionsTest.cs
```diff
@@ -0,0 +1,412 @@
...
+        #region Tests
+        #region GetMostRecentEvaluation
+        [Fact]
+        public void GetMostRecentEvaluation_Value()
+        {
+            // Arrange
+            var date = DateTime.UtcNow;
+            var building = new Parcel();
+            building.Evaluations.Add(new ParcelEvaluation(building, date, EvaluationKeys.Assessed, 111.11m));
+            building.Evaluations.Add(new ParcelEvaluation(building, date.Subtract(new TimeSpan(1, 0, 0)), EvaluationKeys.Assessed, 222.22m));
```
devinleighsmith, line 27: date syntax with the timespan is a bit harder to read then the date.addYears used in the building extension tests.

backend/tests/unit/dal/Entities/Extensions/ParcelExtensionsTest.cs
```diff
@@ -0,0 +1,412 @@
...
+    public class ParcelExtensionsTest
+    {
+        #region Tests
+        #region GetMostRecentEvaluation
+        [Fact]
+        public void GetMostRecentEvaluation_Value()
+        {
+            // Arrange
+            var date = DateTime.UtcNow;
+            var building = new Parcel();
```
devinleighsmith, line 25: This appears to be copy pasta

backend/tests/unit/dal/Entities/Extensions/ProjectExtensionsTest.cs
```diff
@@ -0,0 +1,390 @@
...
+
+            // Act
+            var result = projectProperty.UpdateProjectNumbers("5");
+
+            // Assert
+            result.ProjectNumbers.Should().Be("[\"1\",\"2\",\"3\",\"4\",\"5\"]");
+        }
+
+        [Fact]
+        public void UpdateProjectNumbers_Building()
```
devinleighsmith, line 94: Curios, why does the land example add a new project number but the building example just overwrites an existing one?

devinleighsmith (approved):

backend/tests/unit/dal/Entities/AccessRequestAgencyTest.cs
Fosol: Technically it is meaningless what the ids are.  The only thing important is that the code applies the id of the appropriate object.  Humanly speaking though I think it makes sense to use different ids.  I'll update.

backend/tests/unit/dal/Entities/Extensions/ParcelExtensionsTest.cs
Fosol: Nice catch

backend/tests/unit/dal/Entities/Extensions/ParcelExtensionsTest.cs
Fosol: I agree, I'll update.

PR #48, opened 2021-05-26
Psp 660 lookup codes

devinleighsmith:


frontend/src/store/slices/tenants/useTenants.ts
```diff
@@ -26,7 +26,7 @@ export const useTenants = () => {
       dispatch(storeSettings(response.data));
       return response;
     } catch (error) {
-      dispatch(error(tenantsSlice.name, error?.response?.status, error));
+      dispatch(logError({ name: tenantsSlice.name, status: error?.response?.status, error }));
```
devinleighsmith, line 29: This was causing the application to crash on load.

frontend/package.json
```diff
@@ -139,6 +139,7 @@
   },
   "jest": {
     "resetMocks": false,
+    "clearMocks": true,
```
devinleighsmith, line 142: force jest to clear mocks between test suites, reduces the risk of a test being introduced that doesn't clean up after itself properly affecting an unrelated test that is executed later.

Fosol (approved):

PR #49, opened 2021-05-26
Psp 660 users

devinleighsmith:
separate users api from async action creator

### DIFF
--- a/frontend/src/store/slices/tenants/useTenants.test.tsx
+++ b/frontend/src/store/slices/tenants/useTenants.test.tsx
@@ -47,7 +47,7 @@ describe('useTenant slice hook', () => {
   });
 
   afterAll(() => {
-    jest.resetAllMocks();
+    jest.restoreAllMocks();
   });
 
   it('getSettings reducer + api hook', () => {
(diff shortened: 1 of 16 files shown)


frontend/src/store/slices/tenants/useTenants.test.tsx
```diff
@@ -47,7 +47,7 @@ describe('useTenant slice hook', () => {
   });
 
   afterAll(() => {
-    jest.resetAllMocks();
+    jest.restoreAllMocks();
```
Fosol, line 50: I'm curious why this was needed?

frontend/src/store/slices/users/useUsers.ts
```diff
@@ -1,39 +1,31 @@
+import { useApiUsers } from './../../../hooks/pims-api/useApiUsers';
```
Fosol, line 1: Could use `hooks/api-pims`

Fosol (approved):

frontend/src/store/slices/tenants/useTenants.test.tsx
devinleighsmith: have a look here for the differences
https://stackoverflow.com/questions/58151010/difference-between-resetallmocks-resetmodules-resetmoduleregistry-restoreallm

PR #50, opened 2021-05-26
Psp 660 properties

devinleighsmith:
separate property api from async action creators.

### DIFF
--- a/frontend/src/hooks/pims-api/useApiProperties.ts
+++ b/frontend/src/hooks/pims-api/useApiProperties.ts
@@ -1,0 +1,77 @@
+import { IPropertySearchParams } from 'constants/API';
+import { IBuilding, IPagedItems, IParcel } from 'interfaces';
+import React from 'react';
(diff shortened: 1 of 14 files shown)


frontend/src/hooks/pims-api/useApiProperties.ts
```diff
@@ -0,0 +1,77 @@
...
+  errorToast: pimsToasts.parcel.PARCEL_DELETING_ERROR,
+};
+
+const parcelUpdatingToasts: LifecycleToasts = {
+  loadingToast: pimsToasts.parcel.PARCEL_UPDATING,
+  successToast: pimsToasts.parcel.PARCEL_UPDATED,
+  errorToast: pimsToasts.parcel.PARCEL_UPDATING_ERROR,
+};
+
+const buildingDeletingToasts: LifecycleToasts = {
```
Fosol, line 33: I wonder if there is a friendlier way to support different toasts without the repetitive structure here?

Fosol (approved):

PR #51, opened 2021-05-26
only prompt for version if the flag is enabled.

devinleighsmith:
as requested by @Fosol, this will remove the version prompt unless it is explicitly enabled.


Fosol (approved):
Thanks

PR #54, opened 2021-05-28
PSP-997 Remove Home Button from Access Request Page

asanchezr:
* Removed Home button for new users without any roles (who can only request access to PIMS)
* Added unit tests to cover new scenarios

-------

image

### DIFF
--- a/frontend/src/components/layout/AppNavBar/AppNavBar.tsx
+++ b/frontend/src/components/layout/AppNavBar/AppNavBar.tsx
@@ -15,15 +15,12 @@ import { FeatureToggle } from 'tenants/FeatureToggle';
  * Nav bar with role-based functionality.
  */
 function AppNavBar() {
-  const history = useHistory();
   return (
     <Navbar variant="dark" className="map-nav" expand="lg">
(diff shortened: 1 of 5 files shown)


jdtoombs (approved):

frontend/src/components/layout/AppNavBar/AppNavBar.tsx
```diff
@@ -40,6 +37,19 @@ function AppNavBar() {
   );
 }
 
+function HomeButton() {
```
Fosol, line 40: I generally lean towards separate files for components, however when the component isn't used anywhere the rule gets blurry.

Fosol (approved):

frontend/src/components/layout/AppNavBar/AppNavBar.tsx
asanchezr: I also lean towards "one-file one-component" rule. However, I noticed this file had multiple components in it so I did not want to break the pattern there.

Can be refactored later for sure. Extracting all sub-components into separate files would be trivial

PR #55, opened 2021-05-31
update content security policy to be more restrictive.

devinleighsmith:
This should address the following issues from sonarqube:
https://sonarqube-3cd915-tools.apps.silver.devops.gov.bc.ca/project/extension/zap/report_page?id=pims-frontend-test&qualifier=TRK

1) CSP: Wildcard Directive
2) CSP: script-src unsafe-inline
3) CSP: style-src unsafe-inline


devinleighsmith:
Tested this locally and in dev.

PR #57, opened 2021-06-01
resolve some zap security warnings.

devinleighsmith:


frontend/package.json
```diff
@@ -123,7 +123,7 @@
   "scripts": {
     "preinstall": "npx npm-force-resolutions",
     "start": "react-scripts start",
-    "build": "cross-env CI=true react-scripts build",
+    "build": "cross-env CI=true INLINE_RUNTIME_CHUNK=false react-scripts build",
```
devinleighsmith, line 126: by default CRA apps inject a script tag into the index.html at production build time. This flag disables this behaviour, which allows us to use a content security policy that does not allow inline scripts.

frontend/public/index.html
```diff
@@ -6,7 +6,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Property Inventory Management System" />
+    <!-- Any updates to this meta tag must be reflected in the applications nginx.conf.template for these changes to take effect in openshift-->
```
devinleighsmith, line 9: the index.html here allows us to fail fast during development, so that if someone adds a new call and forgets to update the CSP the call will be blocked. unfortunately I had to include a unsafe-inline here to get the css to show up in webpack, would be nice to get around this. Will circle back if I have time.

frontend/public/index.html
```diff
@@ -6,7 +6,13 @@
...
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self'; frame-ancestors 'self'; img-src 'self' data: https://openmaps.gov.bc.ca/ https://*.tile.openstreetmap.org/ https://server.arcgisonline.com/; frame-src 'self' https://*.oidc.gov.bc.ca http://localhost:8080; connect-src 'self' https://*.oidc.gov.bc.ca/ http://localhost:8080/ https://openmaps.gov.bc.ca/ https://server.arcgisonline.com/; form-action 'self'; style-src 'self' 'unsafe-hashes' 'unsafe-inline';"
```
devinleighsmith, line 12: this csp also include localhost so that the app will run in development mode.

frontend/public/index.html
devinleighsmith: Would also be lovely if the same csp used in nginx could be used during development, but I don't see an easy way of doing that.

openshift/4.0/pipelines/Jenkinsfile.cicd
```diff
@@ -159,7 +159,7 @@ pipeline {
       steps {
         script {
           sh 'mkdir -p zap-output'
-          def return_code = sh returnStatus: true, script: "/zap/zap-baseline.py -x zap-report.xml -r zap-report.html -t $VANITY_URL"
+          def return_code = sh returnStatus: true, script: "/zap/zap-baseline.py -x zap-report.xml -r zap-report.html -t $VANITY_URL -j"
```
devinleighsmith, line 162: -j flag turns on the optional web spider functionality.

openshift/4.0/scripts/oc-sonar-scan.sh
```diff
@@ -52,6 +52,7 @@ CMD_TEST="npm run coverage"
 # Begin analysis
 #
 CMD_SONAR_SCAN="sonar-scanner \
+  -D'sonar.projectName=pims-frontend[${TARGET}]' \
```
devinleighsmith, line 55: the project name was always defaulting to pims-frontend[dev] this should fix that.

openshift/s2i/nginx-runtime/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use the offical nginx (based on debian)
-FROM nginx:stable
+FROM docker.io/nginx:stable
```
devinleighsmith, line 2: Updated in an attempt to mitigate docker pull rate issue.

openshift/s2i/nginx-runtime/nginx.conf.template
```diff
@@ -40,7 +40,7 @@ http {
         ignore_invalid_headers off;
 
         # add in most common security headers
-        add_header Content-Security-Policy "default-src * data: blob: filesystem: 'unsafe-inline' 'unsafe-eval'";
+        add_header Content-Security-Policy "default-src 'self'; frame-ancestors 'self'; img-src 'self' data: https://openmaps.gov.bc.ca/ https://*.tile.openstreetmap.org/ https://server.arcgisonline.com/; frame-src 'self' https://*.oidc.gov.bc.ca; connect-src 'self' https://*.oidc.gov.bc.ca/ https://openmaps.gov.bc.ca/ https://server.arcgisonline.com/; form-action 'self'; style-src 'self' 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=';";
```
devinleighsmith, line 43: This is the actual CSP that runs when the app is deployed. Notice the sha1 has on the style-src attribute. Not ideal, but it appears that something within our app (I speculate a dependency) is using inline styles. I attempted to modify style tags within our application, and that did not result in a change to this hash.

Similarly, hoping to circle back to this if I get time. Would be nice to isolate the source of the inline script and remove it.

devinleighsmith:
A summary of the remaining issues ( will provide a more detailed report on the JIRA issue)

https://sonarqube-3cd915-tools.apps.silver.devops.gov.bc.ca/project/extension/zap/report_page?id=pims-frontend-fake&qualifier=TRK

issue 1, Hash Disclosure - Mac OSX salted SHA-1
I have analyzed the so called Max OSX salted hash it is complaining about, and determined that it is just some TZ related information contained in moment. As a result, this is a false positive.

issue 2, Cookie Without SameSite Attribute
To the best of my knowledge/research keycloak currently does not support returning a cookie with an attribute other then 'None' see PSP-984 for more details.

issue 3, Modern Web Application
I enabled the web spider and this error remains. This may be by design as this is an informational warning. https://www.zaproxy.org/docs/alerts/10109/

issue 4, Timestamp Disclosure - Unix
Triggered by information bundled with moment that is setting off ZAP. not related to our server at all so this is also a false positive.

issue 5, Information Disclosure - Suspicious Comments
I read through the code block in the warning and am not seeing any comments that would help an attacker, appears to be a false postive.

openshift/s2i/nginx-runtime/Dockerfile
Fosol: Isn't there issues if we pull directly from docker.io?  Since all requests come through the same IP people are getting timeouts, as Docker limits the number of requests allowed.  I think this will lead to issues.

openshift/s2i/nginx-runtime/nginx.conf.template
Fosol: I don't understand the sha statement?  Somewhere there is a request for `<link rel="stylesheet" src="sha...".>`?

Fosol (approved):

openshift/s2i/nginx-runtime/nginx.conf.template
asanchezr: Yes Jeremy that's what I read from Devin's comment

asanchezr (approved):

openshift/s2i/nginx-runtime/Dockerfile
devinleighsmith: yeah, just using nginx:stable also pulls from docker.io.

see the *Docker Auth* section of this article

not sure if the above actually resolves the issue however, as the docker.io part is in the dockerfile not the build config. may need to cache the nginx:stable issue instead.

openshift/s2i/nginx-runtime/nginx.conf.template
devinleighsmith: No, not quite. Let me provide some background links

devinleighsmith: This is one of the articles I used to better understand this issue:
https://www.uriports.com/blog/creating-a-content-security-policy-csp/

Essentially to allow an inline style block, the csp must have a hash explicitly specified that corresponds to running a hash function on that block of inlined code.

devinleighsmith: What that means is that if that inlined styling changes for any reason, the hashes will no longer match and the inlined style will be blocked. Assuming I'm right and the inlined styling is coming from a dependency, this would make upgrading packages more dangerous. As a result, I'd like to find and document the source of these inline styles, but I also have other stories this sprint and am not convinced this issue is the highest priority.

PR #59, opened 2021-06-03
PSP-722 Disable Agency dropdown for MOTI tenant

asanchezr:
## CITZ
image

## MOTI
image


frontend/src/utils/YupSchema.ts
```diff
@@ -14,9 +14,13 @@ Yup.addMethod(Yup.string, 'optional', function optional() {
...
-  agency: Yup.number()
-    .min(1, 'Invalid Agency')
-    .required('Required'),
+  showAgency: Yup.boolean(),
+  agency: Yup.number().when('showAgency', {
+    is: true,
+    then: Yup.number()
+      .min(1, 'Invalid Agency')
+      .required('Required'),
+  }),
```
asanchezr, line 23: Had to introduce a `showAgency` form field (hidden from the UI) to control whether `agency` is a required field. 

* MOTI: no agency (optional)
* CITZL agency is required

frontend/src/features/admin/access-request/AccessRequestPage.MOTI.test.tsx
```diff
@@ -132,52 +149,64 @@ describe('AccessRequestPage functionality', () => {
...
+  it('displays enabled roles', () => {
+    const { queryByText } = testRender();
+    expect(queryByText('roleVal')).toBeVisible();
+  });
 
-    it('Does not display private roles', () => {
-      const { queryByText } = testRender();
-      expect(queryByText('privateRole')).toBeNull();
-    });
+  it('does not display disabled roles', () => {
```
devinleighsmith, line 180: is there a test for the agency not being visible? I suppose technically the snapshot would cover it. Might be better to make it explicit. What do you think?

devinleighsmith: Kind of covered by the showAgency functionality, up to you.

devinleighsmith (approved):

frontend/src/features/admin/access-request/AccessRequestPage.MOTI.test.tsx
asanchezr: I'm okay adding a text to explicitly check for non-existence. My thoughts were that the snapshot covers that but you are right that snapshot could be updated and we would loose that check

PR #60, opened 2021-06-04
Refactor UI feature toggles

asanchezr:
This PR refactors the **FeatureToggle** component into two separate variants:

* FeatureVisible
* FeatureHidden

This simplifies the logic and makes it quite explicit to show/hide content based on tenants

```jsx
// Hide content for a specific tenant only - note that you need to pass the "hide" property
<FeatureToggle tenant="MOTI" hide>
  some content here...
</FeatureToggle>
```
```jsx
// Now simpler with
<FeatureHidden tenant="MOTI">
  this won't be rendered for MOTI tenant
</FeatureHidden>
```


frontend/src/tenants/FeatureHidden/FeatureHidden.tsx
```diff
@@ -0,0 +1,28 @@
...
+interface Props {
+  tenant: string;
+}
+
+/**
+ * Component for hiding content for specific tenants.
+ * @param param0 Component properties.
+ * @returns If the current tenant matches the `tenant` property it will *not* render the content at all.
+ */
+const FeatureHidden: React.FC<Props> = ({ tenant, children }) => {
```
Fosol, line 16: Might be nice to support multiple tenants.  Don't change for this, as we don't currently have a need to do this.

Fosol (approved):

devinleighsmith (approved):

PR #63, opened 2021-06-08
PSP_971 add api calls for ltsa, as well as tests.

devinleighsmith:
Just double checking the test coverage on codecov, still need to add openshift secrets as well as connection doc.


backend/ltsa/Models/Altos1ChargeCorrection.cs
```diff
@@ -0,0 +1,116 @@
+/*
```
devinleighsmith, line 1: All models are generated by swagger, but many files needed to be hand edited to correct errors in the swagger spec. Also, the generated files contained a large number of unused utility methods that I've removed (to prevent a code coverage nightmare).

backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="titleNumber">the title number to create the order for</param>
+        /// <param name="landTitleDistrictCode">the land title district code</param>
+        /// <returns>The order created within LTSA</returns>
+        [HttpPost("titleOrder")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.OrderWrapper<Model.TitleOrder>), 200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "tools-ltsa" })]
+        [HasPermission(Permissions.PropertyEdit)]
+        public async Task<IActionResult> PostTitleOrderAsync(string titleNumber, string landTitleDistrictCode)
```
Fosol, line 76: Seems odd to call it a Post since the Http method would already infer this.  Why not `SendTitleOrderAsync`?  Again, just a naming preference that isn't required.

backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -0,0 +1,134 @@
...
+        #region Endpoints
+        /// <summary>
+        /// Make a request to Ltsa for title summaries that match the specified `search`.
+        /// </summary>
+        /// <param name="pid">the parcel identifier to search for</param>
+        /// <returns>An array of title summary matches.</returns>
+        [HttpGet("titleSummaries")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.TitleSummariesResponse), 200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
```
Fosol, line 55: Can we receive a 401 or 403? If so might be nice to include it in the attributes.

backend/api/Helpers/Middleware/ErrorHandlingMiddleware.cs
```diff
@@ -175,6 +175,15 @@ private async Task HandleExceptionAsync(HttpContext context, Exception ex)
...
                 _logger.LogError(ex, "CHES unhandled exception.");
             }
+            else if (ex is LtsaException)
+            {
+                var exception = ex as LtsaException;
+                code = exception.StatusCode.Value;
+                message = exception.Message;
+                details = exception.Detail;
+
+                _logger.LogError(ex, "Ltsa unhandled exception.");
```
Fosol, line 185: Spelling `LTSA` I think make more sense here.

backend/api/Startup.cs
```diff
@@ -113,7 +115,7 @@ public void ConfigureServices(IServiceCollection services)
                 options.PropertyNameCaseInsensitive = jsonSerializerOptions.PropertyNameCaseInsensitive;
                 options.PropertyNamingPolicy = jsonSerializerOptions.PropertyNamingPolicy;
                 options.WriteIndented = jsonSerializerOptions.WriteIndented;
-                options.Converters.Add(new JsonStringEnumConverter());
+                options.Converters.Add(new JsonStringEnumMemberConverter());
```
Fosol, line 118: Does this break CHES or Geocoder using different enum converter?

backend/api/Startup.cs
```diff
@@ -321,10 +324,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
                 options.RoutePrefix = this.Configuration.GetValue<string>("Swagger:RoutePrefix");
             });
 
-            app.UseMiddleware<ErrorHandlingMiddleware>();
             app.UseMiddleware<ResponseTimeMiddleware>();
             app.UseMiddleware<LogRequestMiddleware>();
             app.UseMiddleware<LogResponseMiddleware>();
+            app.UseMiddleware<ErrorHandlingMiddleware>();
```
Fosol, line 330: Did you create a PR for CITZ PIMS with this change?

backend/core/Http/Models/TokenModel.cs
```diff
@@ -8,6 +9,23 @@ namespace Pims.Core.Http.Models
     /// </summary>
     public class TokenModel
     {
+
+        public TokenModel()
+        {
+        }
+        public TokenModel(string jwtToken, string refreshToken)
```
Fosol, line 16: I'd rename the first arg to `accessToken` for consistency.

backend/ltsa/Configuration/LtsaOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// LtsaOptions class, provides a way to configure the LtsaService.
+    /// </summary>
+    public class LtsaOptions
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - The URI to the Ltsa API service.
+        /// </summary>
+        public string HostUri { get; } = "https://tdtest3-x42b.ltsa.ca/titledirect/search/api";
```
Fosol, line 12: Seems like the default is the same as what you have in the configurations.  This one has a path included.

backend/ltsa/Extensions/UriExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Linq;
+using System.Text;
+
+namespace Pims.Ltsa.Extensions
+{
+    public static class UriExtensions
+    {
+        public static string AppendToURL(this string baseURL, params string[] segments)
+        {
+            return string.Join("/", new[] { baseURL.TrimEnd('/') }.Concat(segments.Select(s => s.Trim('/'))));
```
Fosol, line 13: I wonder if it's better to use one of the existing libraries to append paths.  Wouldn't this rip off the last path in an existing URL?  `https://some.com/test` would result in losing `test`.  Might want to document this to make it clear for what its purpose is.

backend/ltsa/LtsaService.cs
```diff
@@ -0,0 +1,286 @@
...
+
+        #region methods
+
+        /// <summary>
+        /// Send a request to the specified endpoint.
+        /// </summary>
+        /// <typeparam name="TR"></typeparam>
+        /// <param name="method"></param>
+        /// <returns></returns>
+        private async Task<TR> SendAsync<TR>(string url, HttpMethod method)
```
Fosol, line 64: Seeing the duplication we're incurring on multiple different api wrappers, wondering if it would be a good time to abstract this common code so that we don't have to copy+paste each time.

backend/ltsa/LtsaService.cs
```diff
@@ -0,0 +1,286 @@
...
+                }
+                catch (HttpClientRequestException ex)
+                {
+                    await handleHttpClientRequestException(ex, url);
+                }
+            }
+            throw new LtsaException("max retries exceeded");
+        }
+
+        private async Task handleHttpClientRequestException(HttpClientRequestException ex, string url)
```
Fosol, line 121: Naming is normally Pascal Case within C#.  `HandleHttpClientRequestException`

backend/ltsa/LtsaService.cs
```diff
@@ -0,0 +1,286 @@
...
+        /// <summary>
+        /// Retrieve all of the title summaries for the given pid, and create a title order for each one. Return all of the title orders and the parcel info order for the given pid
+        /// </summary>
+        /// <param name="pid"></param>
+        /// <returns></returns>
+        public async Task<LtsaOrders> PostLtsaFields(string pid)
+        {
+            int parcelIdentifier = ConvertPID(pid);
+            var titleSummariesTask = GetTitleSummariesAsync(parcelIdentifier);
+            var PostParcelInfoOrderTask = PostParcelInfoOrder(pid);
```
Fosol, line 261: Variable name inconsistent.  Normally CamelCase. `postParcelInfoOrderTask`

backend/ltsa/LtsaService.cs
```diff
@@ -0,0 +1,286 @@
...
+            var titleSummaries = await titleSummariesTask;
+
+            ICollection<Task<OrderWrapper<TitleOrder>>> titleOrderTasks = new List<Task<OrderWrapper<TitleOrder>>>();
+            foreach (TitleSummary titleSummary in titleSummaries.TitleSummaries)
+            {
+                titleOrderTasks.Add(PostTitleOrder(titleSummary.TitleNumber, titleSummary.LandTitleDistrictCode.ToString()));
+            }
+            var titleOrders = await Task.WhenAll(titleOrderTasks);
```
Fosol, line 269: Having multiple requests, can we result in a timeout in the original request that executed these?

backend/ltsa/Models/IntegratorCredentials.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Pims.Ltsa.Models
+{
+    public class IntegratorCredentials
+    {
+        public string IntegratorPassword { get; set; }
```
Fosol, line 9: Nice to have documentation on this.

backend/tests/unit/api/Controllers/Tools/LtsaControllerTest.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            var service = helper.GetService<Mock<ILtsaService>>();
+            service.Setup(m => m.GetTitleSummariesAsync(It.IsAny<Int32>())).ReturnsAsync(response);
+
+            // Act
+            var result = await controller.FindTitleSummariesAsync("123-456-789");
+
+            // Assert
+            JsonResult actionResult = Assert.IsType<JsonResult>(result);
+            var results = Assert.IsAssignableFrom<Model.TitleSummariesResponse>(actionResult.Value);
```
Fosol, line 51: Might want to verify the `GetTitleSummariesAsync` was called once.

openshift/4.0/templates/api/deploy-alt-routes.yaml
```diff
@@ -100,6 +100,23 @@ parameters:
     description: "The Common Hosted Email Service password."
     required: true
 
+  - name: MY_LTSA_USERNAME
+    displayName: "My Ltsa Username."
+    description: "The LTSA username."
+    required: true
```
Fosol, line 106: For CITZ PIMS, I wonder if we should not make these required.  It will just throw credential errors if they attempt to use the endpoints for LTSA.

backend/ltsa/LtsaService.cs
devinleighsmith: potentially, but that would have to be handled on the client side would it not?

backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -0,0 +1,134 @@
...
+        [SwaggerOperation(Tags = new[] { "tools-ltsa" })]
```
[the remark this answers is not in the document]
devinleighsmith: this is the same convention that is being used for the other tools controllers

backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+        #endregion
+
+        #region Endpoints
+        /// <summary>
+        /// Make a request to Ltsa for title summaries that match the specified `search`.
+        /// </summary>
+        /// <param name="pid">the parcel identifier to search for</param>
+        /// <returns>An array of title summary matches.</returns>
+        [HttpGet("titleSummaries")]
```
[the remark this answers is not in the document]
devinleighsmith: updated.

backend/ltsa/Models/IntegratorCredentials.cs
devinleighsmith: added

openshift/4.0/templates/api/deploy-alt-routes.yaml
devinleighsmith: Is that a realistic concern?

PR #64, opened 2021-06-08
DotNet 5.0 BuildConfig for API

Fosol:
Update the BuildConfig for the API to be able to build with dotnet 5.0.


backend/Dockerfile.ocp
```diff
@@ -0,0 +1,40 @@
+ARG BUILD_CONFIGURATION=Release
```
Fosol, line 1: This Dockerfile is so that we can build the API with dotnet 5.0.  It is essentially identical to our local build image.  Which is probably better in that we will be developing with the same image.  The main difference is that I have pushed a copy of the images to OpenShift registry to speed up the build process.

openshift/4.0/templates/api/build.yaml
```diff
@@ -50,53 +50,22 @@ parameters:
...
-    description: The s2i image which is used to build the code.
-    required: true
-    value: dotnet-31-rhel7
-  - name: DOTNET_BUILDER_TAG
-    displayName: .NET builder tag
-    description: The s2i image tag which is used to build the code.
-    required: true
-    value: "3.1"
-
-  - name: DOTNET_STARTUP_PROJECT
```
Fosol, line 65: All of this was specific to the original RedHat image.  It is no longer required.

openshift/4.0/templates/base-images/dotnet50/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.redhat.io/ubi8/dotnet-50
```
Fosol, line 1: Attempts to build dotnet 5.0 with Red Hat images.  So far not successful, but I'd like to solve this eventually.

openshift/4.0/templates/base-images/dotnet50/Dockerfile.ocp3
```diff
@@ -0,0 +1,12 @@
+FROM registry.redhat.io/ubi8/dotnet-50
```
Fosol, line 1: Attempts to build dotnet 5.0 with Red Hat images.  So far not successful, but I'd like to solve this eventually.  This file is for OCP3.

devinleighsmith: if this is for ocp3 it seems weird to have this in the openshift/4.0 directory.

Fosol: It's technically in this folder because it required building the base image in OCP3 and then manually copying to OCP4...

jdtoombs (approved):

PR #65, opened 2021-06-08
Disable Jenkins Pipeline Testing and SonarQube

Fosol:
Temporarily commenting out the Jenkins pipeline from running the tests and providing the SonarQube scan.  In the next sprint we'll try and add the GitHub Action to submit the SonarQube scan.

I'm still debugging why the Jenkins Slave for DotNet 5.0 doesn't work.  For some reason it cannot restore the Nuget packages in the API.

### DIFF
--- a/openshift/4.0/pipelines/Jenkinsfile.cicd
+++ b/openshift/4.0/pipelines/Jenkinsfile.cicd
@@ -18,7 +18,7 @@ pipeline {
   agent any
   options {
     disableResume()
-	disableConcurrentBuilds() 
+	disableConcurrentBuilds()
     buildDiscarder(logRotator(numToKeepStr: '5')) // keep 5 builds only
   }
   environment {
@@ -65,7 +65,7 @@ pipeline {
     // To enable pipeline verbose debug output set to "true"
     DEBUG_OUTPUT = sh(script: 'echo "${DEBUG_OUTPUT:-false}"', returnStdout: true).trim()
   }
-  
+
   stages {
     stage('Initialize') {
       steps {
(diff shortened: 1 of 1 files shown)


jdtoombs (approved):

openshift/4.0/pipelines/Jenkinsfile.cicd
```diff
@@ -130,24 +130,24 @@ pipeline {
...
-            }
-            env.SKIP_BUILD = 'false'
-          } catch (err) {
-            echo '[ci/cd]  No code changes. Skipping the Build stage...'
-            env.SKIP_BUILD = 'true'
-          }
-        }
-      }
-    }
+    // stage('Force Build?') {
```
devinleighsmith, line 133: not sure if removing this in addition to the sonarqube changes was intentional or not.

openshift/4.0/pipelines/Jenkinsfile.cicd
```diff
@@ -174,57 +174,57 @@ pipeline {
...
-              echo 'Frontend lint checks and tests passed'
-            }
-          }
-          post {
-            failure {
-              error '*** Frontend lint checks and tests failed ***'
-            }
-          }
-        }
-        stage('Test Backend') {
```
devinleighsmith, line 210: unless I'm mistaken, this is the only part that needs to be disabled, the frontend tests should be fine.

devinleighsmith:
Have you tested this? I'm under t

devinleighsmith:
AH! sorry, misclick

devinleighsmith (approved):

openshift/4.0/pipelines/Jenkinsfile.cicd
Fosol: I'm thinking this causes us more grief than it's worth.  I'd rather just have the pipeline run if we start it and not ask the question.

openshift/4.0/pipelines/Jenkinsfile.cicd
Fosol: I've removed both as we should be able to use the GitHub Action to do the SonarQube scan.  This will speed up the pipeline and we already do the tests in GitHub.

PR #67, opened 2021-06-10
Code cleanup.

devinleighsmith:
replace Obj x = new Obj(...) with Obj x = new (...) where appropriate
replace Obj x = default(Obj) with Obj x = default where appropriate
ran roslyn based code cleanup on all files.
updated code to use array.empty instead of empty array creator
switched all ltsa models from StringEnumConverter to StringEnumMemberConverter, as it has greater support for parsing spaces, special characters.


Fosol (approved):
Are you trying to find a reason to break down and cry?  The GitHub Action branch delta coverage is going to be screaming at you lol.

PR #70, opened 2021-06-11
PSP-1252 Fix background image style

Fosol:
The background image was having issues with cropping and scrolling in some browser sizes, or odd window sizes.

During testing I had inconsistent results that hopefully don't show up. However it appears now it scales better.


jdtoombs (approved):

devinleighsmith:
One of us should follow up on this by testing in DEV.

devinleighsmith (approved):

PR #71, opened 2021-06-11
PSP-1146 Add SonarQube GitHub Action

Fosol:
Instead of submitting SonarQube scans with Jenkins we will submit them through GitHub Actions.

The SonarQube scans will only be run on a `push` (i.e. successful merge of a PR).

We cannot currently run the SonarQube scan during a `pull_request` because secrets will not be provided for forks for security reasons.

### DIFF
--- a/.github/workflows/api-dotnetcore.yml
+++ b/.github/workflows/api-dotnetcore.yml
@@ -2,32 +2,43 @@ name: API (.NET 5)
 
 on:
   push:
-    branches: [master, dev, dev-alpha]
+    branches: [master, dev, test]
   pull_request:
-    branches: [master, dev, dev-alpha]
+    branches: [master, dev, test]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       working-directory: ./backend
-      codeCov-token: ${{ secrets.CodeCov }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Extract Branch Name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+
       - name: Setup .NET 5
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: 5.0.300
+
       - name: Install dependencies
         run: dotnet restore
         working-directory: ${{env.working-directory}}
+
       - name: Install coverlet for code coverage
         run: dotnet tool install -g coverlet.console --version 1.7.2
         working-directory: ${{env.working-directory}}
+
       - name: Build
         run: dotnet build --configuration Release --no-restore
         working-directory: ${{env.working-directory}}
+
       - name: Test
         run: dotnet test --no-restore --verbosity normal
         working-directory: ${{env.working-directory}}
@@ -70,13 +81,14 @@ jobs:
           mv TestResults/*/* ../TestResults/
           cd ..
           head TestResults/coverage.opencover.xml
+
       - name: Codecov
         uses: codecov/codecov-action@v1.5.0
         with:
           # User defined upload name. Visible in Codecov UI
           name: PIMS
           # Repository upload token - get it from codecov.io. Required only for private repositories
-          token: ${{env.codeCov-token}}
+          token: ${{ secrets.CodeCov }}
           # Path to coverage file to upload
           file: ${{env.working-directory}}/tests/unit/TestResults/coverage.opencover.xml
           # Flag upload to group coverage metrics (e.g. unittests | integration | ui,chrome)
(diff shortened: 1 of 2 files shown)


.github/workflows/api-dotnetcore.yml
```diff
@@ -2,32 +2,42 @@ name: API (.NET 5)
...
-    branches: [master, dev, dev-alpha]
+    branches: [master, dev, test]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       working-directory: ./backend
-      codeCov-token: ${{ secrets.CodeCov }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
devinleighsmith, line 14: oh, so we are dropping codecov in favour of sonar?

.github/workflows/api-dotnetcore.yml
```diff
@@ -70,13 +80,14 @@ jobs:
...
           head TestResults/coverage.opencover.xml
+
       - name: Codecov
         uses: codecov/codecov-action@v1.5.0
         with:
           # User defined upload name. Visible in Codecov UI
           name: PIMS
           # Repository upload token - get it from codecov.io. Required only for private repositories
-          token: ${{env.codeCov-token}}
+          token: ${{ secrets.CodeCov }}
```
devinleighsmith, line 90: Oh, I see, just changing the format of the secret's usage.

.github/workflows/api-dotnetcore.yml
```diff
@@ -85,3 +96,20 @@ jobs:
...
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        with:
+          dotnetBuildArguments: ${{env.working-directory}}
+          dotnetTestArguments: ${{env.working-directory}} --logger trx --collect:"XPlat Code Coverage" -- DataCollectionRunSettings.DataCollectors.DataCollector.Configuration.Format=opencover
+          # Optional extra command arguments the the SonarScanner 'begin' command
+          sonarBeginArguments: /d:sonar.cs.opencover.reportsPaths="**/TestResults/**/coverage.opencover.xml" -d:sonar.cs.vstest.reportsPaths="**/TestResults/*.trx"
+          # The key of the SonarQube project
+          sonarProjectKey: 483bc59d-f9ef-428c-866e-581d9ea69eab
+          # The name of the SonarQube project
+          sonarProjectName: PIMS-API
```
devinleighsmith, line 113: I suppose this means that we are moving to just one sonar project for the api, instead of the one-per-env setup that we currently have (ie pims-api[dev])

devinleighsmith (approved):

.github/workflows/api-dotnetcore.yml
Fosol: I thought about this.  It doesn't make sense to me to have multiple projects for different branches.  It's one project.  We don't want to have to apply rules multiple times to deal with the different branches.  Any inconsistencies between branches should show up in the report history.
The only annoying outcome possibly would be if someone started pushing changes to other branches without adding them to the `dev` branch.

.github/workflows/api-dotnetcore.yml
```diff
@@ -2,32 +2,42 @@ name: API (.NET 5)
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
       - uses: actions/checkout@v2
+
+      - name: Extract Branch Name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+
```
asanchezr, line 23: Do we need this step at all? I don't see variable "branch" being used anywhere else

.github/workflows/app-react.yml
```diff
@@ -47,3 +57,17 @@ jobs:
...
           # Specify whether or not CI build should fail if Codecov runs into an error during upload
           fail_ci_if_error: true
+
+      - name: SonarQube Scan
+        uses: sonarsource/sonarqube-scan-action@master
+        if: ${{ github.event_name == 'push' }}
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_URL }}
+          PROJECT_KEY: 21faa627-fe70-4a67-89e3-1262a32d2e2c
```
asanchezr, line 67: This is fine for now but maybe we should have that as a secret?

asanchezr (approved):

.github/workflows/api-dotnetcore.yml
Fosol: Not anymore.  I was using for the different SonarQube projects, but I don't think that's a good way to go.  I've left it in because it's useful.

.github/workflows/app-react.yml
Fosol: It's technically not a secret.

.github/workflows/api-dotnetcore.yml
Fosol: No, just clean up of the config.

.github/workflows/api-dotnetcore.yml
devinleighsmith: The problem with it from my perspective is timing I guess, if we are about to push to PROD, and Stanley goes to check if we have any outstanding vulnerabilities, and he sees that we have a whole bunch from an unstable build just pushed to DEV he is unlikely to want to release to PROD, even though the code being released to PROD may not have those issues.

devinleighsmith: Same is possible in reverse, we have a whole bunch of vulnerabilities in PROD, someone fixes them in DEV, Stanley looks at the report and everything looks good, and so he releases code with known vulnerabilities to PROD.

Fosol: I think most of those kind of issues can be resolved through review of the SonarQube Activities.  Each report should be reviewable.

Fosol: Additionally at a future date we could find a way to make a failure on merge show up in GitHub.

Fosol: We can also review and split these up at a later date if required.  It would only require a modification to these actions and the creation of the projects in SonarQube.

PR #75, opened 2021-06-15
PSP-1140 Property Information Form

devinleighsmith:
Add WIP form ui for the tabs and the property form.
Add search by pid logic
Add search by marker logic
Add search by address logic
Add ltsa api integration


frontend/src/interfaces/ltsaModels.ts
```diff
@@ -0,0 +1,2384 @@
+/**
```
devinleighsmith, line 1: this is all generated code.

frontend/public/index.html
```diff
@@ -9,7 +9,7 @@
     <!-- Any updates to this meta tag must be reflected in the applications nginx.conf.template for these changes to take effect in openshift-->
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'self'; frame-ancestors 'self'; img-src 'self' data: https://openmaps.gov.bc.ca/ https://*.tile.openstreetmap.org/ https://server.arcgisonline.com/; frame-src 'self' https://*.oidc.gov.bc.ca http://localhost:8080; connect-src 'self' https://*.oidc.gov.bc.ca/ http://localhost:8080/ https://openmaps.gov.bc.ca/ https://server.arcgisonline.com/; form-action 'self'; style-src 'self' 'unsafe-hashes' 'unsafe-inline';"
+      content="default-src 'self'; frame-ancestors 'self'; img-src 'self' data: https://openmaps.gov.bc.ca/ https://*.tile.openstreetmap.org/ https://server.arcgisonline.com/; frame-src 'self' https://*.oidc.gov.bc.ca http://localhost:8080; connect-src 'self' https://*.oidc.gov.bc.ca/ http://localhost:8080/ https://openmaps.gov.bc.ca/ https://server.arcgisonline.com/; form-action 'self'; style-src 'self' 'unsafe-hashes' 'unsafe-inline'; script-src 'self' 'sha256-yt+SNVxRkIi6H6yb7ndFuZM1esMX9esg3UpRHaTsyVk=';"
```
devinleighsmith, line 12: fixes an issue where the UI would freeze when saving modifications when developing locally.

backend/api/Mapping/Converters/ParcelConverter.cs
```diff
@@ -15,5 +18,10 @@ public static int ConvertPID(string pid)
             int.TryParse(pid?.Replace("-", "") ?? "", out int value);
             return value;
         }
+
+        public static string ConvertPIDToDash(string pid)
+        {
+            return String.Join("-", Regex.Split(pid.Replace("-", "").PadLeft(9, '0'), @"(?<=\G.{3})(?!$)"));
```
devinleighsmith, line 24: add some fault tolerance to the backend for pids in an invalid format.

Fosol: Add some documentation for the function.

frontend/public/index.html
Fosol: Can you explain how this works?  For some reason I think I remember SonarQube complaining about a sha256 like this before.

frontend/src/features/moti/mapSideBar/InventoryTabs.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+/**
+ * Tab wrapper, provides styling and nests form components within their corresponding tabs.
+ * @param param0 object containing all react components for the corresponding tabs.
+ */
+const InventoryTabs: React.FunctionComponent<IInventoryTabsProps> = ({ PropertyForm }) => {
+  return (
+    <Tabs defaultActiveKey={InventoryTabNames.property}>
+      <Tab eventKey={InventoryTabNames.property} title="Property">
+        {PropertyForm}
```
Fosol, line 24: Is there any immediately value in passing in form components this way?

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.test.tsx
```diff
@@ -1,29 +1,78 @@
...
+import VisibilitySensor from 'react-visibility-sensor';
+import leafletMouseSlice from 'store/slices/leafletMouse/LeafletMouseSlice';
 import TestCommonWrapper from 'utils/TestCommonWrapper';
+import { fillInput } from 'utils/testUtils';
 
 import MotiInventoryContainer from './MotiInventoryContainer';
 
 enableFetchMocks();
 
+jest.mock(
```
Fosol, line 24: How do you clean up all the mocks?

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.tsx
```diff
@@ -1,27 +1,113 @@
...
+import { FormikProps, FormikValues, getIn } from 'formik';
+import { FeatureCollection, GeoJsonProperties, Geometry } from 'geojson';
+import { useApiLtsa } from 'hooks/pims-api/useApiLtsa';
+import { IGeocoderResponse } from 'hooks/useApi';
+import { ParcelInfo } from 'interfaces/ltsaModels';
+import { LatLngLiteral } from 'leaflet';
 import * as React from 'react';
+import { toast } from 'react-toastify';
 
+import PropertyForm from '../forms/PropertyForm';
```
Fosol, line 16: Makes me think we should add an `index.ts` to the `../forms` path so we can pull in a multiple forms in a single line.

frontend/src/features/moti/mapSideBar/forms/PropertyForm.tsx
```diff
@@ -0,0 +1,22 @@
+import { Formik } from 'formik';
+import { noop } from 'lodash';
+import * as React from 'react';
+import { Form } from 'react-bootstrap';
+
+interface IPropertyFormProps {
+  formikRef: any;
```
Fosol, line 7: Do you have to pass the ref?  Is there a way to extract it from the hook?

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.tsx
```diff
@@ -1,27 +1,113 @@
...
       hidePolicy={true}
     >
-      <SubmitPropertySelector addProperty={noop} />
+      {!sidebarContext || sidebarContext === SidebarContextType.ADD_PROPERTY_TYPE_SELECTOR ? (
+        <SubmitPropertySelector
+          addProperty={() => setShowSideBar(true, SidebarContextType.ADD_BARE_LAND, 'wide', true)}
+        />
+      ) : (
+        <>
+          <PropertySearchForm
```
Fosol, line 101: The PropertySearchForm provides two features, one for when PID changes, the other when Civic Address changes?  Makes me think there may be a more generic implementation that will support all different types of changes.  Then let the container act on the change based on the type.

frontend/src/features/moti/mapSideBar/forms/PropertySearchForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+const SearchHeader = styled.h5`
+  text-align: left;
+`;
+
+interface ISearchFormProps {
+  /** used for determining nameSpace of field */
+  nameSpace?: string;
+  /** handle the population of Geocoder information */
+  handleGeocoderChanges: (data: IGeocoderResponse, nameSpace?: string) => Promise<void>;
```
Fosol, line 48: Why is `handlePidChange` singular, but `handleGeocoderChanges` plural?

frontend/src/store/slices/leafletMouse/LeafletMouseSlice.ts
```diff
@@ -1,11 +1,17 @@
...
 export const clearClickLatLng = createAction('clearLatLng');
 
+export interface ISerializableLeafletMouseEvent {
+  latlng: { lat: number; lng: number };
+  originalEvent: {
+    timeStamp: number | undefined;
+  };
+}
+
 export interface ILeafletMouseSlice {
```
Fosol, line 13: Is the purpose of this slice to simply keep all the Leaflet mouse events in memory?

Fosol (approved):

jdtoombs (approved):

frontend/public/index.html
devinleighsmith: all of the scripts used by the application are packaged up by webpack into a single merged file used by the dev application. that file can then be hashed to the above value. Note, this does not seem to include styled components scripts, just static included js, primarily from 3rd party components.

The idea behind including this here is that if something is introduced that breaks these hashes, it'll fail the application in the developer's local environment. Hopefully this will clue them into the fact that they will need to update the actual content-security-policy that is provided by nginx.

frontend/src/features/moti/mapSideBar/InventoryTabs.tsx
devinleighsmith: This is one method to avoiding prop drilling.

frontend/src/store/slices/leafletMouse/LeafletMouseSlice.ts
devinleighsmith: the purpose of the slice is to allow other components to react to leaflet mouse events.

frontend/src/features/moti/mapSideBar/forms/PropertySearchForm.tsx
devinleighsmith: fixed in follow-up pr.

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.tsx
devinleighsmith: What you are talking about is a common react pattern, where you replace x function event handlers with a single reducer based event handler. Personally I think that pattern is better used where x >=3 due to the overhead involved.

frontend/src/features/moti/mapSideBar/forms/PropertyForm.tsx
devinleighsmith: The formikRef is coming from the MotiSideBarContainer, not a hook. Remember that hooks don't preserve state between multiple invocations, so hooks aren't an effective way of sharing state between components.

What you may be thinking of is the pattern of creating a context with a hook that manages access to the context. That is certainly an option, but I guess my question is, why do you not want to pass the ref? Passing refs from parents to children is also a react pattern to allow parent->child communication.

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.tsx
devinleighsmith: Yes, I should get more in the habit of doing this. I'll add an index to my follow up PR.

frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.test.tsx
devinleighsmith: jest.restoreAllMocks();

backend/api/Mapping/Converters/ParcelConverter.cs
devinleighsmith: Added in follow-up pr.

PR #76, opened 2021-06-16
Update Docker compose configuration

Fosol:
Supporting multiple instances of PIMS is currently impossible due to ports and naming conventions.  This update will allow more flexibility for each developer to control their environment.

This will require you to update your containers.

```bash
make down
make up
```

You can now control your environment ports with `.env`.  However, note that you will need to update other `.env` files to point to the selected ports if you don't use the defaults.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   ####################### Keycloak #######################
   keycloak:
     restart: on-failure:1
-    container_name: pims-keycloak
+    container_name: psp-keycloak
     build:
       context: ./auth/keycloak
     env_file:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -4,28 +4,28 @@ services:
   ####################### Keycloak #######################
   keycloak:
     restart: on-failure:1
-    container_name: pims-keycloak
+    container_name: psp-keycloak
```
asanchezr, line 7: I like this change so it can co-exist locally with CITZ pims. 
Note that you would need to also change the volume names at the end of this file so they don't override each other (PSP and CITZ pims)

GitHub is not letting me add a comment on the specific lines but I mean these:

```yaml
####################### Volumes Definition #######################
volumes:
  api-db-data:
    name: pims-api-db-data  <--- need to be changed to "psp-api-db-data" or similar
  frontend-node-cache:
    name: pims-frontend-node-cache

```

asanchezr (changes requested):

docker-compose.yml
Fosol: Yes I have a PR in CITZ PIMS for this.

asanchezr: well we should change this here to PSP right?

asanchezr: Like CITZ pims can keep the original volume names but maybe change them for PSP?

asanchezr: Otherwise you could end up using CITZ pims database and node_modules in a PSP local docker deployment

Fosol: No, I submitted a PR in CITZ to rename things.  It impacts more people if I change it here for MOTI.  The change I've submitted here will allow user preferences for ports and will allow you all to keep you db.

asanchezr (approved):

PR #79, opened 2021-06-17
DAL Unit Tests

jdtoombs:
Splitting off some more tests. 

Also working on going through the tests that are > 0% but < 100% but have a few failures I am trying to figure out. 

Will try and keep these PRs smaller since I know they are fun to read 😄

### DIFF
--- a/backend/tests/unit/dal/Services/Admin/ProjectRiskServiceTest.cs
+++ b/backend/tests/unit/dal/Services/Admin/ProjectRiskServiceTest.cs
@@ -21,6 +21,7 @@ namespace Pims.Dal.Test.Services.Admin
 {
     [Trait("category", "unit")]
     [Trait("category", "dal")]
+    [Trait("area", "admin")]
     [Trait("group", "project")]
     [ExcludeFromCodeCoverage]
     public class ProjectRiskServiceTest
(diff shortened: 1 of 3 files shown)


asanchezr (approved):

backend/tests/unit/dal/Services/Admin/ProjectRiskServiceTest.cs
```diff
@@ -0,0 +1,191 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Xunit;
+using Entity = Pims.Dal.Entities;
+
+namespace Pims.Dal.Test.Services.Admin
+{
+    [Trait("category", "unit")]
+    [Trait("category", "dal")]
```
devinleighsmith, line 23: It seems like a lot of the other /Admin tests have Trait("admin", "area"). Confirm that you don't need that here and in the next text.

devinleighsmith (approved):

backend/tests/unit/dal/Services/Admin/ProjectRiskServiceTest.cs
jdtoombs: Good catch, thank you

PR #84, opened 2021-06-22
PIMS-1248 Update API version to v01.11.00

Fosol:


asanchezr (approved):

backend/api/Pims.Api.csproj
```diff
@@ -4,8 +4,8 @@
     <TargetFramework>net5.0</TargetFramework>
     <UserSecretsId>0ef6255f-9ea0-49ec-8c65-c172304b4926</UserSecretsId>
     <LangVersion>8.0</LangVersion>
-    <Version>1.10.5.0</Version>
-    <AssemblyVersion>1.10.5.0</AssemblyVersion>
+    <Version>1.11.0.0</Version>
+    <AssemblyVersion>1.11.0.0</AssemblyVersion>
```
asanchezr, line 8: I wonder if there's a way we could automate this. Are the release numbers dictated by MOTI? e.g. why 1.10.5 -> 1.11.0 instead of 1.10.6? Or we could create a tool that does this for us?

Fosol: "Everything" is automatable, but I doubt we have time to automate this presently.  Numbers are somewhat arbitrary.  In the past a release would be a [Minor] version bump, unless we were just fixing bugs. [Major].[Minor].[Bug/Build]

PR #87, opened 2021-06-23
update csp to include a wider match for gov.bc.ca whitelist

devinleighsmith:
the current expectation is that all oidc domains are prefixed with the environment. IE:
dev.oidc.gov.bc.ca
test.oidc.gov.bc.ca

this is not the case for prod which is oidc.gov.bc.ca

The usage of * in csps is limited to subdomains of the non-starred section. As a result *.oidc.gov.bc.ca does not match prod (also *oidc.gov.bc.ca is invalid). 

My recommendation is we move up the matcher to *.gov.bc.ca
Another option would be to add an explicit matcher for oidc.gov.bc.ca in addition to *.oidc.gov.bc.ca

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -9,7 +9,7 @@
     <!-- Any updates to this meta tag must be reflected in the applications nginx.conf.template for these changes to take effect in openshift-->
     <meta
       http-equiv="Content-Security-Policy"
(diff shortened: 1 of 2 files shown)


devinleighsmith:
updated to just add oidc.gov.bc.ca, leaving *.oidc.gov.bc.ca as it was before
dev,test,uat -> *.oidc.gov.bc.ca
prod -> oidc.gov.bc.ca

Fosol (approved):

asanchezr (approved):
Thanks for fixing this Devin!

PR #90, opened 2021-06-23
PSP_1144 duplicate pid and populate by layer functionality.

devinleighsmith:


frontend/src/features/moti/mapSideBar/containers/MotiInventoryContainer.tsx
```diff
@@ -178,11 +187,11 @@ export const MotiInventoryContainer: React.FunctionComponent = () => {
           ltsaTitleResponse?.data,
           geocoderResponse,
         );
+        searchRef?.current?.resetForm();
```
devinleighsmith, line 190: reset the search fields after the search completes, to prepare the search fields for the next search.

frontend/src/features/properties/components/GeocoderAutoComplete.tsx
```diff
@@ -52,30 +51,32 @@ export const GeocoderAutoComplete: React.FC<IGeocoderAutoCompleteProps> = ({
   const { handleBlur } = useFormikContext<any>();
   const errorTooltip = error && touch && displayErrorTooltips ? error : undefined;
 
-  const search = useCallback(
-    (val: string, abort: boolean) =>
-      debounce(async (val: string, abort: boolean) => {
+  const debouncedSearch = React.useRef(
```
devinleighsmith, line 54: The previous format of this debounce function was not working.

frontend/src/hooks/useCallbackAppSelector.ts
```diff
@@ -15,11 +14,13 @@ export function useCallbackAppSelector<TState = RootState, TSelected = unknown>(
...
 
-  useDeepCompareEffect(() => {
+  useEffect(() => {
     if (isInitialMount.current) {
       isInitialMount.current = false;
     } else {
       callbackFn && callbackFn(selected);
     }
-  }, [callbackFn, selected]);
+    // TODO: currently useApi is causing this to refresh when the jwt token expires, that hook needs to be re-written
```
devinleighsmith, line 23: the reason for this requirement is stated above. the functions returned by useApi will trigger useEffect hooks as they are regenerated whenever the axios instance is updated.

frontend/src/hooks/useApi.ts
```diff
@@ -48,7 +49,8 @@ export interface PimsAPI extends AxiosInstance {
 
 export const useApi = (): PimsAPI => {
   const dispatch = useDispatch();
-  const axios = useMemo(() => CustomAxios() as PimsAPI, []);
```
devinleighsmith, line 51: this causes massive performance issues on the frontend, had to remove it.

frontend/src/hooks/useCallbackAppSelector.ts
Fosol: I'm very curious about this, as I was seeing this behaviour when it was created.  But I couldn't get my head around moving this down the pipe so that it only impacts relevant components.  How is it that you have solved it?

devinleighsmith: I've only partially solved it, see my comments on useApi for more details.

frontend/src/hooks/useApi.ts
```diff
@@ -212,5 +214,7 @@ export const useApi = (): PimsAPI => {
     [],
   );
 
-  return useDeepCompareMemo(() => axios, [axios]);
+  // The below memo is only intended to run once, at startup. Or when the jwt is updated.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  return useMemo(() => axios, [jwt]);
```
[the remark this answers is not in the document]
devinleighsmith: I have an idea of how to fix this without having to rebuild it completely, will update this PR with that when complete, tested.

frontend/src/hooks/useApi.ts
```diff
@@ -48,32 +46,35 @@ export interface PimsAPI extends AxiosInstance {
...
-      return config;
-    },
-    error => {
-      dispatch(hideLoading());
-      return Promise.reject(error);
-    },
-  );
-
-  axios.isPidAvailable = useCallback(
+  const getAxios = useCallback(() => {
```
devinleighsmith, line 50: Alright, going to go through some of my decision making on this:
1) Original case, axios instance is recreated every time as it was not memoized. This causes any memoizing dependent on functions in this hook to recreate every render.
2) Ideal case, have one instance of axios, re-use that instance for all requests. All of the functions added to the axios object should remain the same as they are wrapped in a callback. **This caused major performance issues, also jwt token was not renewed correctly**
3) This Implementation. What I've ended up going with here is a getter that is invoked whenever an instance of axios is required. This is also more in line with our other api hooks that call customAxios() directly. The downside of this is that the axios instance is being rebuilt for each request. However, based on testing this method is much more performant. Also updated the returned object to just include the function. We have no use case for returning the actual axios instance here.

frontend/src/hooks/useCallbackAppSelector.ts
devinleighsmith: see this comment: https://github.com/bcgov/PSP/pull/90#discussion_r657514359

frontend/src/hooks/useApi.ts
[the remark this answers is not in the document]
devinleighsmith: https://github.com/bcgov/PSP/pull/90#discussion_r657514359

frontend/src/features/properties/components/GeocoderAutoComplete.tsx
Fosol: Can you explain why?  To me it looks like you're now calling the debounce function immediately rather than as a function.

Fosol (approved):
Great work

frontend/src/features/properties/components/GeocoderAutoComplete.tsx
devinleighsmith: Need to look into why the previous version wasn't working, I was under the impression that pattern worked.
This pattern stores the debounced function in a global ref that isn't updated after every render. (note that calling debounce()) on a function doesn't execute that function, it just wraps that function and returns the wrapped function). All subsequent calls to debouncedSearch are guaranteed to execute against the same debounced wrapper, which ensures that the wrapper has a chance to debounce all of the calls appropriately.

PR #91, opened 2021-06-24
PSP-1254 DB Schema to align with MOTI Standards

Fosol:
# Summary

To adhere to MOTI DB schema standards the whole solution required significant refactoring and sections needed to be rewritten.

Features such as Projects and Workflows have been entirely removed.

These changes do not support migrating from the prior version.
A new `Initial` migration has been generated and will require you drop all objects from your database.

There are more changes than I can remember to refactor PIMS to support MOTI DB schema standards.

## Development Environments

This update is dramatic and destructive.  You will need to delete all of your containers and rebuild from the ground up.

```bash
make up n=keycloak
# Wait until Keycloak is up and running fully.
docker logs psp-keycloak --follow
make up
# Wait until the database is up and running fully.
make db-update
make keycloak-sync
```

There are now only three default users, **admin, rem, rea**.  They all have the same password `password`.

When importing properties use the query parameter override to direct all properties to a specified agency. 
Update your `.env` with the following.

```conf
Api__ImportUrl=/tools/import/properties?addToAgency=BCTFA
```

Now you can run the import if you have a JSON file containing properties.

```bash
make db-seed
```

## Schema Changes

- Table names
- Column names
- Constrain names
- Foreign key names
- Primary keys use Sequences not Identity
- Do not support DATETIME2
- Concurrency
- RowVersion datatype changed to BIGINT
- Do not support multi-primary keys
- Primary keys must be BIGINT
- History triggers
- Insert triggers
- Update triggers
- Changes to foreign keys
- Changes to common app audit columns and which tables they must be included in
- Additional hidden audit columns


backend/api/Models/User/AccessRequestAgencyModel.cs
```diff
@@ -1,8 +1,13 @@
 namespace Pims.Api.Models.User
 {
-    public class AccessRequestAgencyModel : CodeModel<int>
+    public class AccessRequestAgencyModel : CodeModel
     {
         #region Properties
+        /// <summary>
+        /// get/set - Primary key to identify agency.
```
devinleighsmith, line 7: agency -> access request

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -117,56 +115,80 @@ public async Task<Entity.User> GetUserAsync(Guid id)
...
 
             if (user.Username != kuser.Username) throw new InvalidOperationException($"Cannot change the username from '{kuser.Username}' to '{user.Username}'.");
 
-            var addRoles = user.Roles.Except(euser.Roles, new UserRoleRoleIdComparer()).ToArray();
-            var removeRoles = euser.Roles.Except(user.Roles, new UserRoleRoleIdComparer()).ToArray();
-            var addAgencies = user.Agencies.Except(euser.Agencies, new UserAgencyAgencyIdComparer()).ToArray();
-            var removeAgencies = euser.Agencies.Except(user.Agencies, new UserAgencyAgencyIdComparer()).ToArray();
+            IEnumerable<long> addRoleIds;
+            IEnumerable<long> removeRoleIds;
+            if (user.Roles.Any() || euser.Roles.Any())
```
devinleighsmith, line 125: So if a user or entity user has no roles they may still have RolesManyToMany populated?

backend/dal/Configuration/AccessRequestAgencyConfiguration.cs
```diff
@@ -1,28 +1,45 @@
...
     /// AccessRequestAgencyConfiguration class, provides a way to configure AccessRequest Agencies in the database.
     /// </summary>
-    public class AccessRequestAgencyConfiguration : BaseEntityConfiguration<AccessRequestAgency>
+    public class AccessRequestAgencyConfiguration : BaseAppEntityConfiguration<AccessRequestAgency>
     {
         #region Methods
         public override void Configure(EntityTypeBuilder<AccessRequestAgency> builder)
         {
-            builder.ToTable("AccessRequestAgencies");
+            builder.ToMotiTable().HasAnnotation("ProductVersion", "2.0.0");
```
devinleighsmith, line 16: so this annotation is the version that this database entity was added in (and therefor shouldn't change)?

backend/dal/Migrations/Initial/Up/PostUp/01-AdministrativeAreas.sql
```diff
@@ -0,0 +1,352 @@
...
+('Agassiz','5',NULL),
+('Ahousaht','6',NULL),
+('Albert Canyon','7',NULL),
+('Aldergrove','8',NULL),
+('Alert Bay','9','Regional District of Mount Waddington'),
+('Alexis Creek','10',NULL),
+('Aleza Lake','11',NULL),
+('Allison Pass','12',NULL),
+('Anahim Lake','13',NULL),
+('Anmore','Anmore','Metro Vancouver Regional District'),
```
devinleighsmith, line 17: seems odd that some of the abbreviations are just the name again, especially when the name in question is kinda long.

backend/dal/Migrations/Initial/Up/PostUp/01-Agencies.sql
```diff
@@ -1,556 +1,88 @@
...
-    , [IsDisabled]
-    , [SendEmail]
-) VALUES (
-    70
-    , 4
-    , 'CMB'
-    , 'Capital Management Branch'
-    , 0
-    , 0
+MERGE INTO dbo.[PIMS_AGENCY] dest
```
devinleighsmith, line 3: why are some of these straight inserts and some merges? isn't everything here Initial and therefore only an insert is required?

backend/dal/Services/Admin/Concrete/BuildingOccupantTypeService.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        /// <param name="entity"></param>
+        public override void Remove(BuildingOccupantType entity)
+        {
+            entity.ThrowIfNull(nameof(entity));
+
+            var buildingConstructionType = this.Context.BuildingOccupantTypes.Find(entity.Id);
+            if (buildingConstructionType == null) throw new KeyNotFoundException();
+
+            this.Context.Entry(buildingConstructionType).CurrentValues.SetValues(entity);
```
devinleighsmith, line 70: is this just to get the audit information to update?

backend/api/Models/User/AccessRequestAgencyModel.cs
Fosol: Actually it is supposed to be the primary key of the agency here.

devinleighsmith (approved):

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
Fosol: This is something we'll need to clean up later.  In the DB, no.  However depending on what we `.Include(...)` in our EF queries, yes.

I've had to keep the ManyToMany reference in the entity for the following reasons.

- MOTI wants to include specific columns for auditing on the many-to-many tables
- EF won't let me not include the property/collection on the parent objects
- All of our code was originally created with the ManyToMany relationships, which means it's easier to continue using it in most places

I would like to make a general rule moving forward that we don't use the ManyToMany tables unless we have to in the future.  Slowly dealing with technical debt as we go.

backend/dal/Configuration/AccessRequestAgencyConfiguration.cs
Fosol: This is one of the annoying things I need to remove.  EF Core 5 doesn't correctly apply all annotations, which means this does nothing.  However the purpose was to identify when the table was added.

backend/dal/Migrations/Initial/Up/PostUp/01-AdministrativeAreas.sql
Fosol: I didn't change this, it came from the original PIMS project.  I suspect the data isn't perfect.

backend/dal/Migrations/Initial/Up/PostUp/01-Agencies.sql
Fosol: I think in this case I was just copy+paste and kept it in.

backend/dal/Services/Admin/Concrete/BuildingOccupantTypeService.cs
Fosol: I don't understand your question.  
I believe I had to add this service to get the import to work with the new schema requirements.

devinleighsmith: My question was, why do we need to update the entity if we are deleting it from the DB?

Fosol: Ah shortcut to apply RowVersion.  Not a good practice to use `SetValues(...)` however, so I should change it.

Fosol: I would have just copied this from another service regrettably.

devinleighsmith:
@Fosol it seems like the property type id is no longer in sync within the frontend, I'm seeing a lot of subdivisions

devinleighsmith:
Also, we will want to notify the other developers that they will need to update their property import script to only include one of the available agencies. The current json causes the import script to fail.

devinleighsmith:
Yes, the property list view parcels view is showing buildings, and the buildings view is showing subdivisions.

devinleighsmith:
Probably unrelated to this PR, but I thought you fixed this in a previous PR?

image

devinleighsmith:
I'm also getting errors from keycloak-sync, investigating.

Fosol:
@devinleighsmith What was the issue with Help Desk?

I've fixed the Inventory List View.  Primary keys have been changed to support MOTI schema.

I have also included more information in the PR description so everyone knows how to get everything working again.

devinleighsmith:
@Fosol the help desk is very narrow, which causes it to extend below the bottom of the screen on a 1920x1080 monitor when viewing the mapview's filter help information.

Fosol:
I don't recall changing the size of the help desk, so for now I'll leave it as is.

PR #93, opened 2021-06-24
PSP-1392 DevOps improvements

asanchezr:
Various improvements to devops tooling based on stakeholder requests:

* Nginx base images updated appropriately
* UAT pipeline asks for manual confirmation before kicking off deployment to UAT
* Devops documentation updated


openshift/4.0/pipelines/Jenkinsfile.cicd
```diff
@@ -117,6 +121,20 @@ pipeline {
       }
     }
 
+    stage('Approval Prompt') {
```
Fosol, line 124: Is there a way to only show a stage if it's configured to be used?  I think this stage would show up in the UI as completed even if it wasn't required.

Fosol (approved):

devinleighsmith (approved):

openshift/4.0/pipelines/Jenkinsfile.cicd
asanchezr: Don't know. In any case it doesn't hurt functionality

PR #96, opened 2021-06-28
Psp 1421 add reverse geocoder lookup api and frontend POC with unit tests.

devinleighsmith:


backend/api/Areas/Tools/Controllers/GeocoderController.cs
```diff
@@ -83,6 +83,44 @@ public async Task<IActionResult> FindPidsAsync(Guid siteId)
...
+        [HttpGet("nearest")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.AddressModel), 200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "tools-geocoder" })]
+        [HasPermission(Permissions.PropertyEdit)]
+        public async Task<IActionResult> FindNearestAddressAsync(string point)
+        {
+            var parameters = this.Request.QueryString.ParseQueryString<NearestParameters>();
+            parameters.Point = point;
```
Fosol, line 101: I would have though line 100 would have set the `Point` as it would have been part of the query.

backend/geocoder/Extensions/ParameterExtensions.cs
```diff
@@ -68,7 +68,7 @@ public static T ParseQueryString<T>(this Dictionary<string, StringValues> parame
                     {
                         props[key].SetValue(result, Boolean.Parse(p.Value));
                     }
-                    else if (props[key].PropertyType == typeof(double))
+                    else if (props[key].PropertyType == typeof(double) || props[key].PropertyType == typeof(double?))
                     {
                         props[key].SetValue(result, Double.Parse(p.Value));
```
Fosol, line 73: Not sure it's required, but this would never return a `null` value.

Fosol (approved):

asanchezr (approved):

PR #99, opened 2021-06-29
PSP-1433 Default access request agency

Fosol:
All access requests will default to TRAN.


jdtoombs (approved):

asanchezr (approved):
The change looks good but you need to adjust the typing here because TS is complaining and it won't build the frontend (see GH action output)

devinleighsmith (approved):

Fosol:
Don't need this anymore

PR #102, opened 2021-07-06
Correct 404 errors when zooming too far into the new map.

devinleighsmith:
this was introduced with the basemap change. Fix is to set the max zoom level at a value supported by the basemap.


Fosol (approved):

asanchezr (approved):
Thanks for fixing this. I thought I had done that when I added the new basemap

PR #103, opened 2021-07-06
Help desk tests

jdtoombs:
Have some more tests in my stash that I can put up as well (was working through the common components folder), just need to go through them and split off the ones that are still in PSP!

### DIFF
--- a/backend/tests/unit/dal/Services/Admin/BuildingServiceTest.cs
+++ b/backend/tests/unit/dal/Services/Admin/BuildingServiceTest.cs
@@ -180,6 +180,51 @@ namespace Pims.Dal.Test.Services.Admin
             Assert.Equal(building, result, new ShallowPropertyCompare());
             Assert.True(result.IsSensitive);
         }
+
+        [Fact]
+        public void Get_BuildingByPid_NotAuthorized()
+        {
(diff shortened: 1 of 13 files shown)


backend/tests/unit/dal/Services/Admin/BuildingServiceTest.cs
```diff
@@ -180,6 +180,51 @@ public void Get_Sensitive()
             Assert.Equal(building, result, new ShallowPropertyCompare());
```
jdtoombs, line 180: Found this in my stash..

jdtoombs:
I also have another test buried somewhere for close/submit for the HelpModal just trying to find it

frontend/src/utils/TestCommonWrapper.tsx
```diff
@@ -31,17 +31,19 @@ const TestCommonWrapper: React.FunctionComponent<TestProviderWrapperParams> = ({
   agencies,
   history,
 }) => {
-  if (!!roles || !!agencies) {
```
jdtoombs, line 34: Fixing this - did not mean to remove

devinleighsmith (approved):

PR #105, opened 2021-07-07
Ensure that updated jwt token used for all requests.

devinleighsmith:


frontend/src/customAxios.ts
```diff
@@ -37,10 +37,10 @@ export const CustomAxios = ({
     baseURL,
     headers: {
       'Access-Control-Allow-Origin': '*',
-      Authorization: `Bearer ${store.getState().jwt}`,
```
devinleighsmith, line 40: This is only called during instantiation, the interceptor is called for every request.

asanchezr (approved):

PR #108, opened 2021-07-08
PSP-1444 Fix Agency Administration - Concurrency

devinleighsmith:
Subsequent updates to the agency form will no longer cause concurrency issues.

Note that the bomb icon will still be displayed for actual concurrency issues. I'm not convinced that this needs to be fixed given that we also display a friendly error message for this case.


Fosol (approved):
Much appreciated!

PR #109, opened 2021-07-08
psp-1004 dropdown does not scale.

devinleighsmith:
also removed the roles section if empty, and added an alignment correction to prevent cutoff on small devices.


devinleighsmith:
image

Fosol (approved):
A nice improvement

PR #111, opened 2021-07-09
Test tsc corrections.

devinleighsmith:
now run tsc before tests, to ensure that tests don't have tsc issues (since jest-ts ignores ts warnings).
Fix all current test typescript issues.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -126,7 +126,7 @@
     "preinstall": "npx npm-force-resolutions",
     "start": "react-scripts start",
     "build": "cross-env CI=true INLINE_RUNTIME_CHUNK=false react-scripts build",
-    "test": "react-scripts test --ci --watchAll=false --runInBand",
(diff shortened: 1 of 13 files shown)


frontend/package.json
```diff
@@ -126,7 +126,7 @@
     "preinstall": "npx npm-force-resolutions",
     "start": "react-scripts start",
     "build": "cross-env CI=true INLINE_RUNTIME_CHUNK=false react-scripts build",
-    "test": "react-scripts test --ci --watchAll=false --runInBand",
+    "test": "tsc && react-scripts test --ci --watchAll=false --runInBand",
```
devinleighsmith, line 129: tsc runs quite quickly, shouldn't cause an issue for ci.

devinleighsmith:
Note that the tests will not run at all now if there are typescript failures.

Fosol (approved):
Thanks

PR #113, opened 2021-07-12
Correct concurrency error when updating users.

devinleighsmith:
Enforce consistent order of returned agencies,
to ensure that the first agency is a parent if a parent agency is present in the list.

### DIFF
--- a/backend/api/Areas/Admin/Mapping/AccessRequest/AccessRequestMap.cs
+++ b/backend/api/Areas/Admin/Mapping/AccessRequest/AccessRequestMap.cs
@@ -1,5 +1,4 @@
 using Mapster;
-using System.Linq;
 using Entity = Pims.Dal.Entities;
 using Model = Pims.Api.Areas.Admin.Models.AccessRequest;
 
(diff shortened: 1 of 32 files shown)


backend/api/Areas/Admin/Mapping/AccessRequest/AccessRequestMap.cs
```diff
@@ -11,7 +12,7 @@ public void Register(TypeAdapterConfig config)
             config.NewConfig<Entity.AccessRequest, Model.AccessRequestModel>()
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.Status, src => src.Status)
-                .Map(dest => dest.Agencies, src => src.AgenciesManyToMany)
+                .Map(dest => dest.Agencies, src => src.AgenciesManyToMany.OrderBy(a => a.Agency.ParentId))
```
devinleighsmith, line 15: The frontend displays the first agency if the backend returns more then one. This order by ensures that agencies with null parent ids (ie. parent agencies) are listed first. This prevents a bug where if an agency is changed on the frontend to a parent agency, the frontend would display one of that parent agencies children instead.

frontend/src/store/slices/users/usersSlice.ts
```diff
@@ -47,6 +47,7 @@ export const usersSlice = createSlice({
         ),
       ];
       state.pagedUsers.items = items;
+      state.userDetail = action.payload;
```
devinleighsmith, line 50: this piece is what actually makes the rowversion get updated properly.

frontend/src/store/slices/users/useUsers.ts
```diff
@@ -105,7 +105,7 @@ export const useUsers = () => {
    * @return the updated user.
    */
   const update = useCallback(
-    async (updatedUser: IUserDetails): Promise<AxiosResponse<IUserDetails>> => {
+    async (updatedUser: IUserDetails): Promise<IUserDetails> => {
```
Fosol, line 108: Seems like this shouldn't have even worked before.

Fosol (approved):

PR #116, opened 2021-07-16
psp-1478 initial ETL proof of concept migrations.

devinleighsmith:


devinleighsmith:
Unfortunately there is no devops process to synchronize these with the deployed packages in ssistst.th.gov.bc.ca.
Perhaps that is something our ETL developer will work on.

Fosol (approved):

PR #121, opened 2021-07-22
psp-862

devinleighsmith:


Fosol (approved):

devinleighsmith:
going forward, we should only prioritize resolving these if they show up on npm audit --production. We don't really care if dev dependencies have vulnerabilities @Fosol @asanchezr

Fosol:
While I generally agree, I know there was a past issue were dev dependencies aren't isolated as much as developers thought they were.  In fact I believe Create React library no longer uses dev dependencies because apparently they add little value.  I am not knowledgeable enough to know the best decision however.

PR #122, opened 2021-07-23
add gh action based dev pipeline.

devinleighsmith:


devinleighsmith:
This is an attempt at a like-for-like pipeline that has the same functionality (and shares sh code with) our existing dev pipeline.

Fosol (approved):
Can you provide some additional information like a README for this?  It looks like you're building and deploying from GitHub directly to OpenShift.  Does this mean the Jenkins pipeline is no longer used?

.github/workflows/dev-pipeline.yml
```diff
@@ -0,0 +1,90 @@
...
+
+          ./openshift/4.0/player.sh deploy api $DESTINATION -apply
+          ./openshift/4.0/player.sh deploy app $DESTINATION -apply
+  notify:
+    needs: deploy
+    if: always()
+    name: notify-teams
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Log in to OpenShift
```
asanchezr, line 85: Not sure the name is appropriate here

asanchezr (approved):
I love this direction. When I created the sh scripts that was the end goal - to get rid of Jenkins and replace with GH actions.

asanchezr:
My only comment with this PR (and the others currently open) is that we are still seeing GH action running for Node 10. Wasn't expecting to see that once the PR to update to Node 14 got merged - ideas?

image

PR #123, opened 2021-07-23
Psp 1652

devinleighsmith:


.github/workflows/zap-scan.yml
```diff
@@ -0,0 +1,34 @@
+name: Zap Scan
+
+on:
+  push:
```
devinleighsmith, line 4: If we end up merging the dev pipeline action, I can update this to run after the pipeline, and so for the first time we'll actually be running the zap scan against the deployed code instead of the code before the site is deployed (current state)

Fosol (approved):

asanchezr (approved):
I like this move to GH actions. I think this one should always run after the dev pipeline to check the deployed code. I know this deviates from our current state but I rather check the new deployed code for vulnerabilities than check the last deployed code before a new build is kicked off.

In an ideal world we would have dynamic environments created per PR and **then** it would make sense to check that environment prior to deploying to DEV. But until we have that, better to have feedback on the code just deployed. Thoughts?

devinleighsmith:
@asanchezr I agree, but I don't see a way to make this run after the jenkins pipeline unless I merge the other PR (which completely replaces the jenkins pipeline with a Github action).

PR #125, opened 2021-07-23
PSP-1488 create a storage that accept binary upload via http or https

sookeke:
- parametize dc,bc and rolebinding template
- Added timestamp in PTD
- use openSSL to send to S3
- Added connections for both S3 and azure blob


openshift/4.0/templates/Logging/Dockerfile
```diff
@@ -7,14 +7,17 @@ ARG RUN_DEPS='bash curl ca-certificates gettext'
...
     tar xzvf /tmp/oc.tar.gz -C /usr/local/bin/ && \
     rm -rf /tmp/oc.tar.gz 
 
 RUN mkdir -p ${APP_ROOT} && touch ${APP_ROOT}/liveness && chmod a+r ${APP_ROOT}/liveness && touch ${APP_ROOT}/readiness && chmod a+r ${APP_ROOT}/readiness
-COPY ./sidecar-logging.sh ${APP_ROOT}
-RUN chmod a+rx ${APP_ROOT}/sidecar-logging.sh
+
+COPY . ${APP_ROOT}
+
+RUN rm -Rf ${APP_ROOT}/*.yaml && rm -Rf ${APP_ROOT}/*.env
```
asanchezr, line 19: This works but created an unnecessary docker layer with the files in it, which depending on who inspects the generated docker image it might be possible to extract the **.env**  file and all its secrets. 

Best practice for docker is to create a **.dockerignore** file so those files are NOT sent to docker

See a sample in the backend folder: https://github.com/bcgov/PSP/blob/dev/backend/.dockerignore

openshift/4.0/templates/Logging/Dockerfile
```diff
@@ -7,14 +7,17 @@ ARG RUN_DEPS='bash curl ca-certificates gettext'
...
     rm -rf /tmp/oc.tar.gz 
 
 RUN mkdir -p ${APP_ROOT} && touch ${APP_ROOT}/liveness && chmod a+r ${APP_ROOT}/liveness && touch ${APP_ROOT}/readiness && chmod a+r ${APP_ROOT}/readiness
-COPY ./sidecar-logging.sh ${APP_ROOT}
-RUN chmod a+rx ${APP_ROOT}/sidecar-logging.sh
+
+COPY . ${APP_ROOT}
+
+RUN rm -Rf ${APP_ROOT}/*.yaml && rm -Rf ${APP_ROOT}/*.env 
+RUN chmod a+rx ${APP_ROOT}/sidecar-logging.sh && chmod a+rx ${APP_ROOT}/bin/s3-put && chmod a+rx ${APP_ROOT}/lib/s3-common.sh
```
asanchezr, line 20: I don't see a command here to add execution permission to **s3-get** (which is part of this PR)
I see that it is not being used at the moment so would recommend to either add execution permissions to it or delete it for now until needed.

openshift/4.0/templates/Logging/Dockerfile
sookeke: resolved

openshift/4.0/templates/Logging/Dockerfile
sookeke: resolved

PR #127, opened 2021-07-27
Frontend Unit Tests

jdtoombs:
More tests I had in stash that still apply after the removal of various frontend components

Updates: Just adding new ones I've been creating as well to this PR

### DIFF
--- a/frontend/src/components/common/form/ParentSelect.test.tsx
+++ b/frontend/src/components/common/form/ParentSelect.test.tsx
@@ -9,9 +9,9 @@ import { ParentSelect } from './ParentSelect';
 import { SelectOptions } from './Select';
 
 const testOptions: SelectOptions = [
-  { value: 1, label: 'parent' },
+  { value: 1, label: 'parent', parent: '' },
   { value: 2, label: 'child', parentId: 1, parent: 'parent' },
-  { value: 3, label: 'other' },
+  { value: 3, label: 'other', parent: '' },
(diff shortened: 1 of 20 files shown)


frontend/src/components/common/form/ParentSelect.test.tsx
```diff
@@ -9,9 +9,9 @@ import { ParentSelect } from './ParentSelect';
 import { SelectOptions } from './Select';
 
 const testOptions: SelectOptions = [
-  { value: 1, label: 'parent' },
+  { value: 1, label: 'parent', parent: '' },
```
jdtoombs, line 12: Was causing a console warning for the typeahead component as the `filterBy` prop always gets casted to a `string` and parent/label were passed to `filterBy`

frontend/src/features/admin/agencies/AgencyFilterBar.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  keycloak: {
+    userInfo: {
+      agencies: [1],
+      roles: [],
+    },
+    subject: 'test',
+  },
+});
+
+it('renders correctly...', () => {
```
jdtoombs, line 24: As agencies are nota  big part of this application I am leaving the test at a snapshot for now in case of removal

asanchezr (approved):
Thanks for this. Great work!

PR #139, opened 2021-08-10
pims-logging workflow

sookeke:
- workflow is being triggered by a pull request from a based repository not a forked repo.
- secrets are not passed to the runner when a workflow is triggered from a forked repository.
- Therefore, this workflow should run on test and master branch and will be triggered by based repo (dev, test)


sookeke:
merging this as there's not a lot of change

PR #145, opened 2021-08-13
Treat keycloak as the source of truth for usernames. Ignore any usern…

devinleighsmith:
…ame changes from PIMS.


jdtoombs (approved):
Ahh I see what was happening, thanks for fixing this!

PR #151, opened 2021-08-17
psp-1722

devinleighsmith:
-extend layers to add auth header when making requests, dynamically inject img tag with result.
-add proxy controller.


frontend/src/components/maps/leaflet/LayersControl/wmsHeaders.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Override the base leaflet WMS functionality to handle adding an auth token to the requests.
+ */
+class WmsHeaders extends L.TileLayer.WMS {
+  createTile(coords: any, done: any) {
+    const url = this.getTileUrl(coords);
+    const img = document.createElement('img');
+    const axios = CustomAxios();
+    axios.get(url, { responseType: 'blob' }).then((response: AxiosResponse) => {
```
devinleighsmith, line 13: @asanchezr this will likely fail unless the cors policy on the geoserver is set up to allow requests from PSP. Maybe we can talk to Peter Spry about this tomorrow.

backend/api/Areas/Tools/Controllers/GeoServerController.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <returns>the GeoServer response.</returns>
+        [HttpGet("{**rest}")]
+        [SwaggerOperation(Tags = new[] { "{**rest}" })]
+        public async Task<HttpResponseMessage> Proxy(string rest)
+        {
+            var queryString = this.Request.QueryString.Value;
+            var geoServerProxyUrl = _configuration.GetValue<string>("GeoServerProxyUrl");
+            var proxiedUrl = $"{geoServerProxyUrl}{rest}{queryString}";
+
+            return await _proxyRequestClient.ProxySendAsync(Request, proxiedUrl);
```
devinleighsmith, line 46: This doesn't work currently, even with the Authorization header in place, the request just 302s to https://logontest7.gov.bc.ca/

devinleighsmith: However, I have tested that the auth header is present in the forwarded request.

jdtoombs (approved):

devinleighsmith:
@asanchezr it seems like nginx can connect to vip-dev-geo.th.gov.bc.ca locally when on vpn, but within openshift all connections result in **1816 recv() failed (104: Connection reset by peer) while reading response header from upstream**

Not nsp related as there is currently no knp support for egress policy (also curling google works fine).
Current theory is that there is some kind of firewall issue between openshift and geoserver, Peter is looking into that.

openshift/s2i/nginx-runtime/nginx.conf.template
```diff
@@ -55,6 +55,10 @@ http {
             proxy_pass          %API_URL%;
         }
 
+        location /ogs-internal {
+            proxy_pass          %GEOSERVER_URL%;
+        }
+
```
asanchezr, line 61: Putting my comment here because GitHub doesn't let me comment on the relevant line (43)
I believe we need to add "blob"  to the CSP here too right? just like `index.html`?

Which brings me to my next suggestion. I was thinking it would be useful to use an env var for the CSP (like the API_URL, etc). That way we could customize it without having to commit to git and create a new build in OpenShift. Just change the env var in OCP and it would re-apply. What do you think? 

Given time constraints this improvement could come in a future PR....

devinleighsmith:
converting this back to draft, until openshift can connect to geoserver there is no benefit in merging this.

PR #156, opened 2021-08-20
fix keycloak sync tool, as well as some issues with the user api.

devinleighsmith:


backend/dal/Configuration/PersonConfiguration.cs
```diff
@@ -34,7 +34,6 @@ public override void Configure(EntityTypeBuilder<Person> builder)
                 .HasMaxLength(200)
                 .HasComment("Person's middle names.");
             builder.Property(m => m.NameSuffix)
-                .IsRequired()
```
Fosol, line 37: I should add this to my PR.

backend/dal/Migrations/Initial/Up/PreUp/01-History.sql
```diff
@@ -815,7 +815,7 @@ CREATE TABLE [dbo].[PIMS_USER_ORGANIZATION_HIST] (
     [USER_ORGANIZATION_ID] BIGINT NOT NULL,
     [ORGANIZATION_ID] BIGINT NOT NULL,
     [USER_ID] BIGINT NOT NULL,
-    [ROLE_ID] BIGINT NOT NULL,
```
Fosol, line 818: According to Doug's scripts this should remain as `NOT NULL`

backend/dal/Services/UserService.cs
```diff
@@ -257,6 +273,7 @@ public User Get(Guid keycloakUserId)
         /// <returns></returns>
         public User Add(User add)
         {
+            add.IssueOn = DateTime.Now;
```
Fosol, line 276: `UtcNow`

backend/entities/ContactMethod.cs
```diff
@@ -75,16 +75,15 @@ public class ContactMethod : BaseAppEntity
         /// <param name="organization"></param>
         /// <param name="methodType"></param>
         /// <param name="value"></param>
-        public ContactMethod(Person person, Organization organization, ContactMethodType methodType, string value)
+        public ContactMethod(Person person, Organization organization, string methodTypeId, string value)
```
Fosol, line 78: Create an override in another constructor to support this.

backend/entities/ContactMethodType.cs
```diff
@@ -38,4 +38,13 @@ public ContactMethodType(string id, string description) : base(id, description)
         }
         #endregion
     }
+
+    public static class ContactMethodTypes
```
Fosol, line 42: You could use an Enum here instead even with string values.  We had originally some nice extensions to do string comparisons with them.

backend/entities/User.cs
```diff
@@ -64,6 +64,12 @@ public class User : BaseAppEntity
         [Column("IS_DISABLED")]
         public bool IsDisabled { get; set; }
 
+        /// <summary>
+        /// get/set - information following users full name.
+        /// </summary>
+        [Column("NAME_SUFFIX")]
+        public string NameSuffix { get; set; }
```
Fosol, line 71: This doesn't look correct to me

backend/entities/UserOrganization.cs
```diff
@@ -42,7 +42,7 @@ public class UserOrganization : BaseAppEntity
         /// get/set - The foreign key to the role.
         /// </summary>
         [Column("ROLE_ID")]
-        public long RoleId { get; set; }
+        public long? RoleId { get; set; }
```
Fosol, line 45: Incorrect based on Doug's scripts

Fosol (commented):
I'll submit a PR with a new Initial.  I don't think some of your schema changes are correct based on Doug's scripts.

backend/dal/Migrations/Initial/Up/PreUp/01-History.sql
devinleighsmith: Yeah, this isn't supported by the frontend, so I'm going to try and get him to change this.

backend/entities/ContactMethod.cs
devinleighsmith: Could do, important to realize that this way causes errors during inserts.

backend/entities/UserOrganization.cs
devinleighsmith: see my other comment

backend/dal/Migrations/Initial/Up/PreUp/01-History.sql
Fosol: I'll let you schedule a discussion.  I can update my PR once we have confirmation.
You could always just assume the role the user asking for is the same role that would be applied here.

backend/entities/ContactMethod.cs
Fosol: Most likely errors related to not passing a ContactMethodType from the active Context.

backend/entities/ContactMethodType.cs
devinleighsmith: For this usage (essentially just grouping some constants) I don't see the benefit of an enum.

backend/dal/Migrations/Initial/Up/PreUp/01-History.sql
devinleighsmith: Right, it just gets complicated:
1) when there are multiple roles.
2) when the user updates their organization.

Of course it is possible to make this work, I just would rather not spend time supporting database functionality that we aren't using.

backend/entities/ContactMethod.cs
devinleighsmith: right, I was trying to create the contact method during mapping which wasn't possible using the above.

devinleighsmith: Regardless, I'll create the overload.

backend/dal/Services/UserService.cs
```diff
@@ -80,10 +80,10 @@ public User Activate()
                 this.Logger.LogInformation($"User Activation: key:{key}, email:{email}, username:{username}, first:{givenName}, surname:{surname}");
 
                 var person = new Person(surname, givenName);
-                var contactMethodType = this.Context.ContactMethodTypes.First(cmt => cmt.Id == "EMAIL");
-                var contactMethod = new ContactMethod(person, organization, contactMethodType, email);
+                var contactMethod = new ContactMethod(person, organization, ContactMethodTypes.WorkEmail, email);
```
devinleighsmith, line 83: Even though I could revert this since I'm adding the previous constructor back, this is still more concise.

backend/dal/Migrations/Initial/Up/PostUp/01-Roles.sql
```diff
@@ -65,6 +65,19 @@ VALUES
     , ''
     , 'migration'
     , ''
+),
+(
+    5
+    , '57726C45-BEE0-4C60-9E93-98E84B4AD136'
+    , 'Undetermined'
```
Fosol, line 72: We really want this lol

backend/entities/UserOrganization.cs
```diff
@@ -42,7 +42,7 @@ public class UserOrganization : BaseAppEntity
         /// get/set - The foreign key to the role.
         /// </summary>
         [Column("ROLE_ID")]
-        public long RoleId { get; set; }
+        public long RoleId { get; set; } = 5; // this is the default value for the undefined role.
```
Fosol, line 45: Generally an odd practice, but okay I guess

Fosol (approved):

backend/dal/Migrations/Initial/Up/PostUp/01-Roles.sql
devinleighsmith: This was the compromise I got out of Doug

PR #157, opened 2021-08-20
PSP-1784 Replace Initial DB Migration

Fosol:
Replacing `Initial` DB Migration.

Adding properties
- User.Position
- User.Note
- User.LastLogin
- AccessRequest.Note

Changing properties
- Property.Location to GEOMETRY type 
- Property.Boundery to GEOMETRY type


devinleighsmith (approved):

devinleighsmith:
looks good. I think you should push first, because I want to make sure the sync works on some of the fields that you've added here.

PR #158, opened 2021-08-20
Replacing Initial DB Migration

Fosol:
Replacing `Initial` DB Migration.

Updating seed scripts and fixing Address nullable properties.


devinleighsmith (approved):

backend/core/Extensions/ObjectExtensions.cs
```diff
@@ -30,6 +31,8 @@ public static class ObjectExtensions
                 .Where(p => p.PropertyType.IsPrimitive
                     || p.PropertyType.IsEnum
                     || p.PropertyType == typeof(string)
+                    || p.PropertyType == typeof(Point)
```
Fosol, line 34: It will now copy these values from one object to another.

PR #159, opened 2021-08-24
Psp 1786 access request dev fixes.

devinleighsmith:


backend/dal/Configuration/AccessRequestConfiguration.cs
```diff
@@ -35,8 +34,8 @@ public override void Configure(EntityTypeBuilder<AccessRequest> builder)
             builder.HasOne(m => m.Status).WithMany(u => u.AccessRequests).HasForeignKey(m => m.StatusId).OnDelete(DeleteBehavior.Cascade).HasConstraintName("PIM_ARQSTT_PIM_ACRQST_FK");
 
             builder.HasMany(m => m.Organizations).WithMany(m => m.AccessRequests).UsingEntity<AccessRequestOrganization>(
-                m => m.HasOne(m => m.Organization).WithMany(m => m.AccessRequestsManyToMany).HasForeignKey(m => m.OrganizationId),
-                m => m.HasOne(m => m.AccessRequest).WithMany(m => m.OrganizationsManyToMany).HasForeignKey(m => m.AccessRequestId)
+                m => m.HasOne(m => m.Organization).WithMany(m => m.AccessRequestsManyToMany).HasForeignKey(m => m.OrganizationId).OnDelete(DeleteBehavior.ClientCascade),
```
Fosol, line 37: Are these foreign keys named correctly?

backend/dal/Services/AccessRequest.cs
```diff
@@ -189,10 +197,12 @@ public AccessRequest Update(AccessRequest update)
             // Copy values into entity.
             accessRequest.RoleId = update.RoleId;
             accessRequest.RowVersion = update.RowVersion;
```
Fosol, line 199: Make sure concurrency works.  I know just setting the value like this normally does not.

backend/entities/AccessRequestStatusType.cs
```diff
@@ -38,4 +38,15 @@ public AccessRequestStatusType(string id, string description) : base(id, descrip
         }
         #endregion
     }
+
+    public static class AccessRequestStatusTypes
```
Fosol, line 42: Generally bad practice to hide a **public** class in another's file.

Fosol (approved):

Fosol:
Seemed to be a lot of places the URL needed to be changed.  Makes me wonder if we are not centralizing the api correctly.

backend/entities/AccessRequestStatusType.cs
devinleighsmith: It is good practice to co-locate related functionality, such as a static constants class that is intended for use with the sibling class.

devinleighsmith:
The url needed to be changed a lot of places for two reasons:
1) some of the controllers were incorrectly appending duplicate information on top of a base route. IE: /access/requests/access/requests
2) there are tests written to explicitly check that individual controller methods have specific routes. My initial attempt to place all of a controllers routing information at the top level route caused these tests to fail.

backend/dal/Configuration/AccessRequestConfiguration.cs
devinleighsmith: PIM_ACRQST_PIM_ACRQOR_FK and PIM_ORG_PIM_ACRQOR_FK

PR #161, opened 2021-08-24
Initial to dev.

devinleighsmith:
This is a PR to merge initial to DEV, as most major breaking functionality has been resolved.


devinleighsmith:
Note the 404 on the map screen is still present. I'm ok with keeping that in for dev for now as we aren't planning on releasing to TEST without having geoserver functionality.

Temporarily disabling that functionality seems unnecessary if the 404 only affects dev.

Fosol (approved):
Let's do this and move on.

PR #187, opened 2021-09-02
Corrections from dev testing of psp-1370

devinleighsmith:


frontend/src/App.scss
```diff
@@ -1,46 +1,47 @@
 @import './colors.scss';
-.App {
-  .error {
-    color: $danger-color;
-  }
 
-  ul.pagination li {
-    cursor: pointer;
-  }
+.modal-dialog.modal-xl {
```
devinleighsmith, line 3: the following css elements are not children of the app component.

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -52,9 +53,38 @@ export const convertToProperty = (
...
   if ([PropertyTypes.Land, PropertyTypes.Subdivision].includes(property.propertyTypeId)) {
-    return property;
-  } else if (property.propertyTypeId === PropertyTypes.Building) {
-    return property;
+    return {
+      pid: property.PID,
+      latitude: latitude,
+      longitude: longitude,
+      propertyTypeId: property.propertyTypeId,
+      address: {
```
devinleighsmith, line 61: The current layer doesn't expose the address.

jdtoombs (approved):

Fosol (approved):
Thanks

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -52,9 +53,38 @@ export const convertToProperty = (
   longitude?: number,
 ): IProperty | null => {
   if ([PropertyTypes.Land, PropertyTypes.Subdivision].includes(property.propertyTypeId)) {
```
Fosol, line 55: Wonder if the PointClusterer should be generic and the conversion of a property should be based on a separate abstraction.

Fosol (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
asanchezr: Agree. Prob a tech-debt ticket

backend/dal/Migrations/Initial/Up/PostUp/02_PIMS_PROPERTY_LOCATION_VW.sql
```diff
@@ -40,7 +40,6 @@ SELECT PROP.PROPERTY_ID
      , PROP.DESCRIPTION   
      , PROP.REGION_CODE
      , PROP.DISTRICT_CODE
-     , PROP.BOUNDARY.STCentroid() AS GEOMETRY
```
asanchezr, line 43: Curious as to why this was removed from the view?

asanchezr (approved):

backend/dal/Migrations/Initial/Up/PostUp/02_PIMS_PROPERTY_LOCATION_VW.sql
Fosol: I didn't even notice this view.  You'll also have to manually run it in DEV otherwise it won't be included.

devinleighsmith: That doesn't seem to be the case, I tested that this is added automatically when the migration runs.

devinleighsmith: Oh, this column was duplicated. Doug removed it.

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: Yeah, there are a number of tech-debt issues that I've logged with this ticket.

PR #203, opened 2021-09-08
Add migration so views are added without requiring db refresh

devinleighsmith:
The view scripts were added to the initial migration, which means the new views are only added if the database is reset.
This change adds a new migration for those scripts, so that the views are created automatically as the IS is merged to each env.


backend/dal/Migrations/20210908232309_v0.2.0.9.1.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
Fosol, line 1: The only issue I see with this migration is that when I do one for the leases I think mine would be v0.02.09.0. But I'm not certain.  All I know is we probably want the one that goes to each environment to match the release version.

Fosol: Just realized I think we're in `v0.2.10.0`, so that should make my migration work fine.

Fosol (approved):

asanchezr:
Question - why is this targeting the **test** branch instead of **dev**?

Fosol:
Ah good catch Alejandro.  This will cause issues if we don't have consistent migrations in the DB.

PR #206, opened 2021-09-09
PSP-1854 Add DB Migration for IS10

Fosol:
Added DB Migration `v0.2.0.10.0`
Added the identified **Leases** entities to the data model

> Please note that the DB scripts provided altered most tables in the database, but cursory glance they don't appear to have been actually changed...  Additionally the only location seed scripts are provided is in the full build folder, which could imply there are additional change to prior seed values.  I am not going to manually scan every line.

> Additionally, please note that the DB scripts provided do not follow the standardized naming conventions or rules originally provided and expected.  I have taken the liberty to continue following the original rules/standards so that this DB Migration doesn't take another week to refactor.

> A particularly risky part of this process is copying the triggers over to the migration.  I have no way to verify there are valid.

Here is the DB schema scripts that were provided - [link](https://github.com/bcgov/PSP/tree/dev/database/mssql/scripts/dbscripts/PSP_PIMS_S10_00)

## IMPORTANT

Devin's other PR with a DB migration must go first or this migration will block his when EF attempts to update your DB.  Additionally Devin's migration needs to be generated before this one.  If it has been regenerated and now has a newer date, this PR must be updated first.

## UNVERIFIED

As we don't have functional components or requirements to validate this schema, there is a risk that this migration will need to be redone before it is released to TEST.  Which would also require DEV DB to be rolled back before getting the updated migration.


backend/dal/Configuration/LeaseActivityConfiguration.cs
```diff
@@ -0,0 +1,54 @@
...
+                .HasMaxLength(40)
+                .HasComment("A comment related to the activity");
+
+            builder.HasOne(m => m.Lease).WithMany(m => m.Activities).HasForeignKey(m => m.LeaseId).OnDelete(DeleteBehavior.Restrict).HasConstraintName("PIM_LEASE_PIM_LSACTV_FK");
+            builder.HasOne(m => m.LeaseType).WithMany(m => m.Activities).HasForeignKey(m => m.LeaseTypeId).OnDelete(DeleteBehavior.Cascade).HasConstraintName("PIM_LSTYPE_PIM_LSACTV_FK");
+            builder.HasOne(m => m.Subtype).WithMany(m => m.Activities).HasForeignKey(m => m.SubtypeId).OnDelete(DeleteBehavior.Cascade).HasConstraintName("PIM_LSSTYP_PIM_LSACTV_FK");
+            builder.HasOne(m => m.Period).WithMany(m => m.Activities).HasForeignKey(m => m.PeriodId).OnDelete(DeleteBehavior.Restrict).HasConstraintName("PIM_LSACPR_PIM_LSACTV_FK");
+
+            builder.HasIndex(m => m.PeriodId).HasDatabaseName("LSACTV_LEASE_ACTIVITY_PERIOD_ID_IDX");
+            builder.HasIndex(m => m.LeaseId).HasDatabaseName("LSEACT_LEASE_ID_IDX");
```
devinleighsmith, line 46: Think these are all LSACTV* not LSEACT

Fosol: Amazing eyes!  I'll update

backend/dal/Migrations/v0.2.10.0/Up/PostUp/02-LeasePaymentFrequencyType.sql
```diff
@@ -0,0 +1,8 @@
+PRINT N'Inserting [PIMS_LEASE_PMT_FREQ_TYPE]'
```
devinleighsmith, line 1: Nice to see we already have these.

devinleighsmith (approved):

PR #214, opened 2021-09-13
Standardize all units within the app to rem, set base font-size to 10px.

devinleighsmith:
This should probably merge at the end of the sprint to avoid conflicts. Low priority, but nice to have.


Fosol (approved):
I like

sookeke:
@asanchezr  I have added the credential scan as require check status. see above conflicts

asanchezr:
@sookeke the conflicts are not due to credential scans. This PR was created before that action was put in place so it has never run for this PR. I expect when @devinleighsmith updates this PR to rebase this PR then the scan will run.

jdtoombs (approved):

PR #218, opened 2021-09-14
PSP-1848 Update Property Search Bar

Fosol:
image

## Property Search Bar - Map

Created a new component called `SearchToggle` which displays two buttons that switch the view from Map to Property List View.

> Testing is currently impossibly locally for the Map and Property List View.  There is no guarantee that search works at all.  I will need to test in DEV.
> The Property List View page throws errors and is outside of this story scope
> The Property List View styling of the filter bar is different than the Map, this is also outside of this story scope


frontend/src/components/common/form/PlusButton.tsx
```diff
@@ -15,7 +15,7 @@ interface IPlusButtonProps extends ButtonProps {
  * PlusButton displaying a plus button, used to add new items.
  * @param param0
  */
-const PlusButton: React.FC<IPlusButtonProps> = ({ toolId, toolText, ...props }) => {
+export const PlusButton: React.FC<IPlusButtonProps> = ({ toolId, toolText, ...props }) => {
```
Fosol, line 18: Make it nicer to add multiple components

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -407,14 +408,15 @@ const PropertyListView: React.FC = () => {
   return (
     <Container fluid className="PropertyListView">
       <Container fluid className="filter-container border-bottom">
-        <Container className="px-0">
+        <Container fluid className="px-0 map-filter-container">
```
Fosol, line 411: Fixing some of the styling issues.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ...props
+}) => {
+  const history = useHistory();
+  const handleClick = () => {
+    if (toggle === SearchToggleOption.Map) history.push('/properties/list');
+    else history.push('/mapview');
+  };
+
+  return (
+    <StyledToggle>
```
Fosol, line 40: Not a big fan of the design, as one button will always be disabled.  It should just be a single button that shows one or the other... but I'm just doing what is in the story.

devinleighsmith: better to bring up these issues in the design meeting.

devinleighsmith: I get the frustration of communicating during some meetings but presumably that doesn't extend to Bobbi.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+const StyledToggle = styled('div')`
+  display: flex;
+`;
+
+const StyledNav = styled('div')`
+  flex: 1;
+  width: 100%;
+`;
+
+const StyledLink = styled('div')`
```
devinleighsmith, line 70: you may also need flex-shrink: 0 on the svgs.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+  toggle = SearchToggleOption.Map,
+  toolId,
+  ...props
+}) => {
+  const history = useHistory();
+
+  return (
+    <StyledToggle>
+      <StyledNav>
+        <StyledLink style={{ cursor: toggle === SearchToggleOption.List ? 'pointer' : 'auto' }}>
```
devinleighsmith, line 38: I'd recommend using a class to toggle these styles, That way you don't have to repeat the fill: toggle === SearchToggleOption.List conditional.
so: 
1) do the conditional **once** to set the class.
2) use the class (or absence of) to set the desired css attributes.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+              style={{ fill: toggle === SearchToggleOption.Map ? '#003366' : '#909090' }}
+            />
+          </TooltipWrapper>
+        </StyledLink>
+      </StyledNav>
+    </StyledToggle>
+  );
+};
+
+const StyledToggle = styled('div')`
```
devinleighsmith, line 61: Presumably we'll make these common styled components at some point, fine if this is left for now.

frontend/src/components/common/form/SearchToggle.tsx
devinleighsmith: of course, these things get missed in the design meeting sometimes too, Bobbi is pretty good about ad-hoc design meetings during sprint as well.

devinleighsmith (approved):

frontend/src/components/common/form/SearchToggle.tsx
Fosol: The comment was to explain why there are two buttons that are acting as a toggle, nothing more.  Sprint is done today, so having a design change that has no impact to functionality isn't something I would recommend.  Instead I've provided feedback in the PR that may spur further conversation between the dev team and the UI/UX.  Up to you guys if you want to chase it down in the future... But for now I'm just doing what's in the story.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,96 @@
...
+ */
+export const SearchToggle: React.FC<ISearchToggleProps> = ({
+  toggle = SearchToggleOption.Map,
+  toolId,
+  ...props
+}) => {
+  const history = useHistory();
+
+  return (
+    <StyledToggle toggle={toggle}>
```
Fosol, line 36: Moved repetitive logic into styled component properties.  Makes the implementation here cleaner.

frontend/src/components/common/form/SearchToggle.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+interface ISearchToggleStyleProps {
+  toggle?: SearchToggleOption;
+}
+
+const StyledToggle = styled('div')<ISearchToggleStyleProps>`
+  display: flex;
+
+  div:nth-child(${props => (props.toggle === SearchToggleOption.Map ? 1 : 2)}) {
```
Fosol, line 62: Was hoping to find a way to make this a little cleaner, but this is easy enough to read.

PR #226, opened 2021-09-15
support searching by pid/pin with leading 0's.

devinleighsmith:
The view is expecting numerical pids/pins, and doesn't appear to support leading zeros on the geoserver side.

### DIFF
--- a/frontend/src/components/layout/Header/Header.test.tsx
+++ b/frontend/src/components/layout/Header/Header.test.tsx
@@ -1,5 +1,5 @@
 import { useKeycloak } from '@react-keycloak/web';
-import { cleanup, fireEvent, render } from '@testing-library/react';
+import { cleanup, fireEvent, render, waitFor } from '@testing-library/react';
 import * as API from 'constants/API';
 import { createMemoryHistory } from 'history';
 import renderer from 'react-test-renderer';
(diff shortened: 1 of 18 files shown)


frontend/src/components/layout/Header/Header.test.tsx
```diff
@@ -54,29 +54,35 @@ describe('Header tests', () => {
     process.env = OLD_ENV;
   });
 
-  it('Header renders correctly', () => {
+  it('Header renders correctly', async () => {
```
devinleighsmith, line 57: These tests were periodically leaking into the next test due to async actions running in the background.

frontend/src/components/layout/SideNavBar/SideNavbar.test.tsx
```diff
@@ -19,117 +19,139 @@ const renderComponent = (props?: IRenderProps) =>
   );
 
 describe('SideNavbar display and logic', () => {
-  it('renders', () => {
+  it('renders', async () => {
```
devinleighsmith, line 22: same issue here, tests not handling async logic correctly.

frontend/src/hooks/pims-api/useApiUsers.ts
```diff
@@ -4,7 +4,7 @@ import { LifecycleToasts } from 'customAxios';
 import { IPagedItems, IUser } from 'interfaces';
 import React from 'react';
 
-import { useApi } from '.';
+import { useAxiosApi } from '.';
```
devinleighsmith, line 7: rename to avoid name collision, there is no only one useApi hook.

frontend/src/hooks/pims-api/useApi.ts
```diff
@@ -6,7 +6,7 @@ import useDeepCompareMemo from 'hooks/useDeepCompareMemo';
  * Common hook to make requests to the PIMS APi.
  * @returns CustomAxios object setup for the PIMS API.
  */
-export const useApi = (
+export const useAxiosApi = (
```
Fosol, line 9: What's with the rename?  Just for clarity?

Fosol (approved):

PR #232, opened 2021-09-16
support customizable layer urls in TEST.

devinleighsmith:
I was mistaken when I said the customizable layers are not required, there are now two sets of layers used by the application.
1) PSP layers, hosted at psp_env/ows endpoints
2) external layers, hosted at /ows endpoints

I've tested this change on my local use psp test keycloak and it appears to work, allowing both the highways layer and properties layer to be active simultaneously (as was working before the _env changes).


frontend/public/tenants/tenant.json
```diff
@@ -10,5 +10,7 @@
     "title": "",
     "heading": "",
     "body": ""
-  }
+  },
+  "layers" : [],
```
asanchezr, line 14: On a different PR I noticed the auto-formatting "fixed" this for me. Maybe fix it here to avoid other further conflicts? thanks!

If not obvious this should be:

```json
"layers": []  # no space before colon
```

asanchezr (approved):

PR #234, opened 2021-09-17
Update README

asanchezr:
Simple PR to test the secret scanning action.


asanchezr:
@sookeke as you can see from this PR. The secret scanning tool works as expected and it leaves a comment in the PR with the scan result.

sookeke:
@asanchezr  yes, that what we want. thank you

devinleighsmith (approved):

Fosol (approved):

PR #242, opened 2021-09-19
PSP-1381 Property List View

asanchezr:
image


backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
```diff
@@ -9,19 +9,41 @@ public class PropertyMap : IRegister
...
         {
             config.NewConfig<Entity.Property, Model.PropertyModel>()
-                .Map(dest => dest.PropertyType, src => src.PropertyTypeId)
-                .Map(dest => dest.Classification, src => src.ClassificationId)
+                .Map(dest => dest.Id, src => src.Id)
+                .Map(dest => dest.RowVersion, src => src.RowVersion)
+                .Map(dest => dest.PID, src => src.ParcelIdentity)
+                .Map(dest => dest.PIN, src => src.PIN)
+                .Map(dest => dest.PropertyTypeId, src => src.PropertyTypeId)
+                .Map(dest => dest.PropertyType, src => src.PropertyType.Description)
```
devinleighsmith, line 17: Can you confirm that none of these dereferences can result in an NPE? IE, are any of these relationships nullable.

frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -1,261 +1,140 @@
...
-      expect(getByTestId('edit-icon')).toBeInTheDocument();
-      fireEvent(
-        getByTestId('edit-icon'),
-        new MouseEvent('click', { bubbles: true, cancelable: true }),
-      );
-      await waitFor(() => expect(getByTestId('save-changes')).toBeInTheDocument(), {
-        timeout: 500,
-      });
-    });
+    // default table message when there is no data to display
```
devinleighsmith, line 107: does this comment make sense?

frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -1,261 +1,140 @@
...
-          />
-          <PropertyListView />
-          <ToastContainer />
-        </Router>
-      </Provider>
-    </TenantProvider>,
-  );
+  return {
+    component,
+    findSpinner: () => component.queryByRole('status'),
```
devinleighsmith, line 50: this is nice.

frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -1,261 +1,140 @@
...
+    } = setup();
 
-    const { container } = renderPage();
+    // wait for table to finish loading
+    await waitFor(async () => expect(findSpinner()).not.toBeInTheDocument());
 
-    await waitFor(async () => expect(container.querySelector('.spinner-border')).toBeNull());
-    const cells = container.querySelectorAll('.td.clickable');
-    fireEvent.click(cells[0]);
+    const tableCells = container.querySelectorAll('.td.clickable');
```
devinleighsmith, line 136: you made a function for this above.

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -278,12 +130,16 @@ const PropertyListView: React.FC = () => {
...
+      pageSize,
+      sort && !isEmpty(sort) ? generateMultiSortCriteria(sort) : undefined,
+      filter,
+    );
     return dispatch(
       download({
-        url: getAllFields
-          ? getAllFieldsPropertyReportUrl({ ...query, all: true, propertyType: undefined })
-          : getPropertyReportUrl({ ...query, all: true, propertyType: undefined }),
+        url: getAllFields ? getAllFieldsPropertyReportUrl(query) : getPropertyReportUrl(query),
```
devinleighsmith, line 142: Not sure if this getAllFields flag makes sense any more, did they mention this in the requirements?

frontend/src/features/properties/list/components/DirtyRowsTracker.tsx
```diff
@@ -0,0 +1,33 @@
...
+  assessedBuilding?: boolean;
+  netBook?: boolean;
+  market?: boolean;
+}
+
+/**
+ *  Component to track edited rows in the formik table
+ * @param param0 {setDirtyRows: event listener for changed rows}
+ */
+export const DirtyRowsTracker: React.FC<{ setDirtyRows: (changes: IChangedRow[]) => void }> = ({
```
devinleighsmith, line 17: Are you using this? If not we can just delete this.

frontend/src/features/properties/list/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { IPaginateProperties } from 'constants/API';
```
devinleighsmith, line 1: Putting url generation in a utils file doesn't really match up with what we've been doing with our other api calls. Mostly we've been centralizing urls in src\hooks\pims-api\use*

tools/keycloak/sync/realm.json
```diff
@@ -556,4 +556,4 @@
       }
     ]
   }
-}
\ No newline at end of file
+}
```
devinleighsmith, line 559: ?

asanchezr: This auto-formatting changes. Last PR from Stanley removed the newline at end of file. He doesn't run prettier automatically like we do on each PR

frontend/src/features/properties/list/components/DirtyRowsTracker.tsx
asanchezr: No I am not. Happy to delete. I didn't in case we needed but we have the history in git. Will remove

devinleighsmith:
noticed the pid/pin requires an exact match. This appears to be consistent with previous functionality but there is now a delta between the map search and the list view for that functionality. I still think we leave as is for now however.

frontend/src/features/properties/list/PropertyListView.tsx
asanchezr: Nobody has mentioned this but was not sure what to take out without breaking existing functionality. If you know more please provide context here and will be happy to remove/change. Not sure 100% if export endpoints work to be hones

frontend/src/features/properties/list/PropertyListView.test.tsx
asanchezr: yeah - miss that

backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
asanchezr: Not sure. I copied this from the other mapping from Jeremy's work. How do we check for nullable?

devinleighsmith:
Allowing the user to click a property row is sort of an anti-feature for now since the map doesn't support that functionality (the detail view used to handle that and it has been removed. The slide out could be updated to handle this but doesn't currently.

My recommendation is to remove the on click event and don't show rows as clickable for now. If that conflicts with the story then maybe confirm with Julian.

frontend/src/features/properties/list/utils.ts
asanchezr: This was hardcoded at the top of the PropertyListView file. I just extracted to a separate file. I don't think a function that generates a URL warrants a custom "hook" - or does it?

devinleighsmith:
what is the default searchby param according to your story? on the mapview it was the pid/pin. (Note, the url param controls the searchby as well, to see the true default, remove the &searchBy=pid from the url and navigate to the remaining part of the URL).

asanchezr:
@devinleighsmith - will remove the row onclick handler. As said, there was much to remove and it was hard to determine what to leave. The requirements were "should work as before but with these columns instead" so hard to take that and make decisions on what to take out

asanchezr:
My story has no default for search param. I can use whatever is used for the map search. I thought they were using the "default" filter

devinleighsmith:
I know you said you were waiting for css fixes, just want to make sure you were aware of this one:

image
it looks like the list view + the header and footer are exceeding 100vh, as the footer is getting cut off, and the side bar is being stretched past its max-height.

asanchezr:
That is not exclusive of the list view. Yes I saw it on manage user etc if you make the window smaller

asanchezr:
I think we need to file a ticket against the container of the sidebar

devinleighsmith:
sure, go ahead.

asanchezr:
image

I haven't done any research into the sidebar CSS - but I did notice it. Will file it

backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
devinleighsmith: src.PropertyType != null ? src.PropertyType.Description : null
or write an extension for the property entity, something like getPropertyDescription(), which can then just do src.PropertyType?.Description.

frontend/src/features/properties/list/PropertyListView.tsx
devinleighsmith: There is this comment:
@param {boolean} getAllFields Enable this field to generate report with additional fields. For SRES only.

So it sounds like some fields were locked to the SRES role. I don't think we have any hidden fields that are role locked currently, so probably can remove this.

frontend/src/features/properties/list/utils.ts
devinleighsmith: strictly speaking, no. This was a convention that Jeremy started when he refactored parts of the application. We don't have to keep using it if it isn't providing value.

At a minimum I think I'd still like to see utils renamed to something more specific. 

We have been working to centralize API calls as well, so moving this out of the feature specific area would be nice (optional).

devinleighsmith:
Re: the sidebar issue, please just make sure the issue is not coming from the listview itself (ie if the listview's height is wrong, and it is missing up the sidebar and the footer, please fix that as part of this story).

if it is the sidebar that is pushing the header down, feel free to log the issue.

frontend/src/features/properties/list/PropertyListView.tsx
devinleighsmith: based on testing they seem to work ok. Not sure if the list of columns is correct, but if you have no requirements on that I'd leave it as is.

backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
asanchezr: happy to make that change but two questions:
- then what do we do with the other mapping that Jeremy created `Mapping/Property/PropertyMap.cs` that one has the same "faulty" code and it is used in the property info slideout
- How do we determine what is a required field (ie not null on DB) on the entities? 

I ask because by looking at Doug sql file, I see two things:
1. on PIMS_PROPERTY table "PROPERTY_TYPE_CODE"                 	nvarchar(20) NOT NULL," so that tells me prop type code cannot be null (this is mapped in EF to PropertyTypeId)
2. When I search for the FK constraint for the Property Type ("PIM_PRPTYP_PIM_PRPRTY_FK") I see that:
```sql
ALTER TABLE "dbo"."PIMS_PROPERTY"
	ADD CONSTRAINT "PIM_PRPTYP_PIM_PRPRTY_FK"
	FOREIGN KEY("PROPERTY_TYPE_CODE")
	REFERENCES "dbo"."PIMS_PROPERTY_TYPE"("PROPERTY_TYPE_CODE")
...
```
What I understand for this is that the FK will never be null because it uses the type code which cannot be null on PIMS_PROPERTY. What could I be missing here? 
Again, just trying to understand how we determine when things can be nullable in EF

frontend/src/features/properties/list/PropertyListView.test.tsx
asanchezr: The comment was there from whoever created the tests but I can remove as it is very clear what the test is asserting

asanchezr: nvm this is a copy pasta mistake LOL

devinleighsmith (approved):

backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
devinleighsmith: As long as the foreign key is not nullable, and references a non-nullable pk the null checks are not necessary. you can either check that in the ef fluent configuration or in the actual database itself. Would recommend not looking at Doug's scripts as they don't necessarily reflect the actual database.

asanchezr: I looked at the latest EF designer for the current migration **20210909194500_v0.2.10.0.Designer.cs** and this is what I see:

```csharp
            modelBuilder.Entity("Pims.Dal.Entities.Property", b =>
                {
                    b.HasOne("Pims.Dal.Entities.Address", "Address")
                        .WithMany("Properties")
                        .HasForeignKey("AddressId")
                        .HasConstraintName("PIM_ADDRSS_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyAreaUnitType", "AreaUnit")
                        .WithMany("Properties")
                        .HasForeignKey("AreaUnitId")
                        .HasConstraintName("PIM_ARUNIT_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyClassificationType", "Classification")
                        .WithMany("Properties")
                        .HasForeignKey("ClassificationId")
                        .HasConstraintName("PIM_PRPCLT_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyDataSourceType", "DataSource")
                        .WithMany("Properties")
                        .HasForeignKey("DataSourceId")
                        .HasConstraintName("PIM_PIDSRT_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.District", "District")
                        .WithMany("Properties")
                        .HasForeignKey("DistrictId")
                        .HasConstraintName("PIM_DSTRCT_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyType", "PropertyType")
                        .WithMany("Properties")
                        .HasForeignKey("PropertyTypeId")
                        .HasConstraintName("PIM_PRPTYP_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.Region", "Region")
                        .WithMany("Properties")
                        .HasForeignKey("RegionId")
                        .HasConstraintName("PIM_REGION_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyStatusType", "Status")
                        .WithMany("Properties")
                        .HasForeignKey("StatusId")
                        .HasConstraintName("PIM_PRPSTS_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();

                    b.HasOne("Pims.Dal.Entities.PropertyTenureType", "Tenure")
                        .WithMany("Properties")
                        .HasForeignKey("TenureId")
                        .HasConstraintName("PIM_PRPTNR_PIM_PRPRTY_FK")
                        .OnDelete(DeleteBehavior.Restrict)
                        .IsRequired();
```

I think then there is no need for the null checks here as the FK constrains are marked required - right?

asanchezr:
@devinleighsmith - I have addressed most of the feedback from your PR review. Feel free to check the changes.

A couple notes:
- I have moved the report export/download functionality to the **useProperties** hook - I agree this is cleaner than having it inlined in the Property List View. However I have not done the same for other list views that might have export functionality (Manage Users, etc)
  - Tech-debt could be filed to improve those areas but I want to push this forward instead of refactor other major areas of the frontend
- CSS issues will have to be addresses in another PR. I feel like I could spend a lot of time debugging CSS while this functionality holds back other features that depend on the property list, testing etc. We can file additional sub-tasks against the original story to track CSS/styling work

devinleighsmith:
@asanchezr thanks for doing this, changes look good.

PR #253, opened 2021-09-22
psp-1899 add missing default values to database.

devinleighsmith:


devinleighsmith:
Manually went through Doug's scripts and compared default values against the ef generated DB as well as *configuration files. In all cases, missing defaults appear to have been caused by missing fluent notations, therefore this is a human error issue and not an issue with EF.

asanchezr (approved):

PR #258, opened 2021-09-23
PSP-1381 Fix unit tests for `useProperties` hook

asanchezr:
Noticed that the unit tests were passing regardless of assertions due to assertions being inside a promise that would always resolve (or fail) after tests had completed.

We will need to go over other hook tests and fix accordingly. More tech-debt to file


devinleighsmith:
Did you file a story for the tech debt then?

frontend/src/store/slices/properties/useProperties.test.tsx
```diff
@@ -39,250 +43,204 @@ const getWrapper = (store: any) => ({ children }: any) => (
...
+    const url = RegExp(`/reports/properties?.*`);
+    it('Request successful, dispatches success with correct response', async () => {
+      const mockResponse = 'foo bar baz - this would be binary content for a csv file';
+      mockAxios.onGet(url).reply(200, mockResponse);
+      const filter: IPaginateProperties = { page: 1, quantity: 10 };
+
+      const { exportProperties } = setup();
+      await exportProperties(filter);
+
+      expect(find(currentStore.getActions(), { type: 'network/logRequest' })).toBeDefined();
```
devinleighsmith, line 229: could you use toMatchObject instead?

devinleighsmith (approved):

devinleighsmith:
actually, https://github.com/bcgov/PSP/pull/260

PR #260, opened 2021-09-23
correct renderhook tests such that failures are reported properly.

devinleighsmith:


frontend/src/hooks/pims-api/useApi.test.ts
```diff
@@ -39,18 +39,19 @@ describe('useApi testing suite', () => {
...
-        await api.get('failure');
-      } catch (error) {
-        expect(error.response.status).toBe(400);
-        expect(error.response.data).toBe('failure');
-        expect(mockAxios.history.get).toHaveLength(1);
-      }
+    let api = {} as any;
+    renderHook(() => {
+      api = useAxiosApi();
     });
```
asanchezr, line 47: This probably works okay but I think best practice is to use the return value from `renderHook`:

```ts
const { result } = renderHook(useAxiosApi); // no need to create a callback function if hook doesn't take any params
const api = result.current;

// do whatever with api
```

frontend/src/store/slices/accessRequests/accessRequestsActionCreator.spec.tsx
```diff
@@ -26,78 +27,57 @@ const getWrapper = (store: any) => ({ children }: any) => (
...
-            }),
-        {
-          wrapper: getWrapper(getStore()),
-        },
-      );
+      const { fetchCurrentAccessRequest } = setup();
+      fetchCurrentAccessRequest();
+      await waitFor(async () => {
+        expect(find(currentStore.getActions(), { type: 'network/logRequest' })).not.toBeNull();
+        expect(find(currentStore.getActions(), { type: 'network/logError' })).not.toBeNull();
```
asanchezr, line 67: This assertion can lead to false-positives because:

1. lodash [find](https://lodash.com/docs/4.17.15#find) returns **undefined** if it cannot find the element
2. **undefined** is not null so it will pass

I changed these types of assertions to **not.toBeUndefined()** or **toBeDefined()** in my previous PR to solve that

PR #262, opened 2021-09-23
fixes psp-1936, psp-1370 pid search bug.

devinleighsmith:


devinleighsmith:
@asanchezr note, this would be a time where we want to keep the above commits. Since they represent distinct functionality.

devinleighsmith:
Performed minimal testing due to missing local geoserver:
1) Clicked on map, verified popup displays for parcel layer.
2) verified that pid/pin searches by numeric pid/pin value.

asanchezr (approved):
Looks good

PR #266, opened 2021-09-23
allow pidformatter to handle number or string types.

devinleighsmith:


frontend/src/utils/propertyUtils.ts
```diff
@@ -6,7 +6,7 @@ import { IAddress } from 'interfaces';
  */
 export const pidFormatter = (pid?: string) => {
   if (!!pid) {
-    let result = pid.padStart(9, '0');
+    let result = pid.toString().padStart(9, '0');
```
devinleighsmith, line 9: toString() is a no-op if already a string, but now this supports numeric pids as well.

Logged follow up task to fix the fact that the type of pid is changing within the form after the first click.

jdtoombs (approved):

frontend/src/utils/propertyUtils.ts
jdtoombs: Thank you!

PR #271, opened 2021-09-26
PSP-1261 Add geoserver docker container

devinleighsmith:


.gitignore
```diff
@@ -13,3 +13,4 @@ bash.exe.stackdump
 auth/keycloak/config/hsperfdata_jboss/
 auth/keycloak/config/*.log
 trufflehog_report.*
+datastore.xml
```
devinleighsmith, line 16: this is a workaround.

geoserver supports env var substitution in its datastore files:
https://docs.geoserver.org/master/en/user/datadirectory/configtemplate.html

however, I could not get this to work in the datastore.xml for the mssql connection, possibly because this datastore is specific to the optional sql-server-plugin and not part of the core geoserver functionality

devinleighsmith: so instead, ignore this file as each developer will have their own unique connection string. Instead, add this file to the files generated by gen-env-files.sh

Makefile
```diff
@@ -210,5 +210,9 @@ frontend-coverage: ## Generate coverage report for frontend
 	@echo "$(P) Generate coverage report for frontend"
 	@cd frontend; npm run coverage;
 
+env: ## Generate env files
```
devinleighsmith, line 213: add this script to the core make file.

docker-compose.yml
```diff
@@ -68,7 +68,21 @@ services:
     env_file: ./frontend/.env
     networks:
       - psp
-
+  geoserver:
+    image: kartoza/geoserver:2.19.2
+    volumes:
+      - ./geoserver/geoserver_data:/opt/geoserver/data_dir/workspaces
```
devinleighsmith, line 74: mount the workspace config so that the property layer works automatically.

devinleighsmith:
NOTE: the container won't return any properties if you have no properties with location information in your local development database. 

I was able to get properties by dev just be exporting them.

asanchezr (approved):

asanchezr (approved):

PR #272, opened 2021-09-27
added image scan security 

sookeke:
- Add secops for for security analysis for PSP
-Added  readme for set up and documentation
- Turn on quality gate for static code analysis scan on API and APP
- Push or PR will fail when sonar quality gate fails
- Added security Image scan for test and master branch push and pr (all base images and api/app)
- Critical or High vulnerability with Aqua trivy will block merge on test and master branch
-  Turn on fail_action for ZAP scan
- Block merge when zap scan fail


sookeke:
@asanchezr @devinleighsmith  have a look before I merge

.github/workflows/image-scan-analysis.yml
```diff
@@ -1,61 +1,705 @@
...
+      with:
+        registry: ${{secrets.OPENSHIFT_REGISTRY}}
+        username: ${{ secrets.OC_USERNAME_SA }}
+        password: ${{ secrets.OPENSHIFT_TOKEN }}
+    
+    - name: Pull nodejs Base Image
+      env:
+        OPENSHIFT_NAMESPACE: 3cd915-tools
+      run: |
+        #docker login -u ${{env.OPENSHIFT_USER}} -p $(oc whoami -t) ${{env.OPENSHIFT_REGISTRY}}
```
devinleighsmith, line 59: probably can remove these comments?

.github/workflows/image-scan-analysis.yml
```diff
@@ -1,61 +1,705 @@
...
+      OPENSHIFT_NAMESPACE: 3cd915-tools
+      output-filename: nodejs14.txt
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download artifact
+        id: artifact
+        uses: actions/download-artifact@master
+        with:
+         name: ${{env.image-name}}_scan_report
+         path: ./${{env.image-name}}
```
devinleighsmith, line 101: if the purpose of this step is to download the file created by the trivy scan, should this not be: 
./${{env.output-filename}} to match line 79?

secops/README.md
```diff
@@ -0,0 +1,418 @@
...
+- [Sonarque](https://docs.sonarqube.org/latest/analysis/github-integration/)
+- [CodeQL](https://github.com/github/codeql-action)
+
+
+#### Static Code Analysis for React Frontend
+
+The Project uses Sonarque as a Static Code Analysis and Quality Assurance Tool to collect and analyses our source code and provide reports for the code quality of our project
+
+**Requirement:**
+- [Sonarque Server](https://github.com/BCDevOps/sonarqube) (community version - free)
```
devinleighsmith, line 21: Sonarque -> Sonarqube?

.github/workflows/image-scan-analysis.yml
```diff
@@ -1,61 +1,705 @@
...
-    branches: [master, test, dev]
+    branches: [master, test]
   schedule:
-    - cron: '42 15 * * 5'
+    - cron: '30 5 14 * *' # Run this scan every sprint release cycle 14th of every month at 5:30 UTC
 
 jobs:
 
-  build_frontend:
+  nodejs-base-image:
```
devinleighsmith, line 13: This looks like a number of repeated sections with just a few variables changing. This may be a good candidate for GHA matrixes.

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix

sookeke:
Good point

devinleighsmith (approved):

.github/workflows/api-dotnetcore.yml
```diff
@@ -98,18 +101,61 @@ jobs:
...
+          SONAR_HOST_URL: ${{ secrets.SONAR_URL }}
+          PROJECT_KEY: ${{secrets.PROJECT_KEY_API}}
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            QUALITY GATE STATUS FOR .NET 5: FAILED.
+            
+            [View and resolve details on][1]
+            [1]: ${{env.SONAR_HOST_URL}}dashboard?id=${{env.PROJECT_KEY}}
+          reactions: confused
```
asanchezr, line 144: @sookeke - unfortunately creating/updating comments do not work on pull requests triggered from forks. See https://github.com/peter-evans/create-or-update-comment. This will fail future PRs. This is consistent with what I saw with my auto-version action. The easiest solution is to use an admin token or personal access token - [PAT](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token)

sookeke:
I know already fix this to target unforked repo

.github/workflows/api-dotnetcore.yml
```diff
@@ -98,18 +101,61 @@ jobs:
...
+      - name: Check Quality Gate and Update Comment
+        if: failure() && steps.scan.outcome == 'failure' && github.event_name == 'pull_request' && steps.fc.outputs.comment-id != '' && github.event.pull_request.head.repo.full_name == github.repository
+        uses: peter-evans/create-or-update-comment@v1
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
asanchezr, line 149: Question - why would we be passing the SONAR_TOKEN to the create-comment action? I don't think it needs it or does anything with it. Am I wrong here?

sookeke:
No, must be a typo

.github/workflows/image-scan-analysis.yml
asanchezr: I think this could be removed as oc login is used before. Or remove the oc login step and use docker login instead. Either should work

.github/workflows/image-scan-analysis.yml
```diff
@@ -1,61 +1,705 @@
...
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: ${{secrets.OPENSHIFT_REGISTRY}}/${{env.image-name}}:${{env.tag}} vulnerability report
+          
+      - name: Comment on PR
+        if: steps.fc.outputs.comment-id == '' && steps.get-comment-body.outcome != 'failure'
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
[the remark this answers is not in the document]
asanchezr: The only drawback of PAT is that they are tied to the **user account** that creates it - e.g. if @sookeke you create it then it is as if you are the one making these PR comments. If you (or whoever creates the PAT) leave the project then that token is no longer valid and all actions would start failing. We need to discuss how to best approach this

sookeke:
PAT or Deployment Key which one is better?

sookeke:
Yes, I will create a deployment key instead?

This can be used as a secret

asanchezr:
@sookeke - I'm not very familiar with deployment keys but if you say they can be used instead of GITHUB_TOKEN in actions we can try. 
All I could find in the documentation is how to create a "machine user" that has that deployment key attached and then a repo admin (you or james payer) would have to add that user to this repo. But I don't know what the policy is about creating users in **bcgov** as there might be extra fees per account

asanchezr:
@sookeke - I like the idea of a deploy key but it seems to fail as well

image

asanchezr:
https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

image

sookeke:
We cannot create organizational secrets only bcgov repo admin can

asanchezr:
I would not merge this PR until we have a way forward that works. Otherwise it will fail all builds

PR #295, opened 2021-09-28
Secops2

sookeke:


.github/workflows/image-scan-analysis.yml
```diff
@@ -276,7 +272,7 @@ jobs:
           reactions: confused 
 
   build_frontend:
-    if: always() && github.event.pull_request.head.repo.full_name == github.repository
+    #if: always() && github.event.pull_request.head.repo.full_name == github.repository
```
devinleighsmith, line 275: hmm, can you explain the motivation behind commenting all of these out? did this not work as you were expecting?
Ideally I'd rather we just removed lines instead of pushing configuration with commented out code, assuming that we are confident that we don't need them.

FYI, it should be possible for you to test all of these actions on your fork, that way you don't need to test on the actual dev branch.

.github/workflows/image-scan-analysis.yml
```diff
@@ -1,10 +1,8 @@
 name: Image Security Scan
 
 on:
-  push:
-    branches: [master, test, dev]
-  pull_request:
-    branches: [master, test, dev]
+  pull_request_target:
+    types: [assigned, opened, synchronize, reopened]
```
devinleighsmith, line 5: What was the motivation behind this? wouldn't this run the action once if a PR was opened, and then again when it was assigned?

devinleighsmith:
@sookeke can you provide a brief description on why you created this PR? is something in the image scan action not working?

PR #296, opened 2021-09-28
PSP-1700 Fix automatic versioning action

asanchezr:
Made a few adjustments to the auto-increment action:
- It runs only when pull requests are merged (as opposed to any **push** to dev)
- It uses a deploy key with write access to the repo to act as an "admin" user with power to bypass the "no-push" restriction so it can commit the new version number back into the repo automatically

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -1,14 +1,19 @@
 name: Bump version
 
 on:
-  push:
+  pull_request_target:
+    types: [closed]
     branches: [dev]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/version.yml
```diff
@@ -1,14 +1,19 @@
 name: Bump version
 
 on:
-  push:
+  pull_request_target:
```
devinleighsmith, line 4: ah, interesting!

devinleighsmith (approved):

PR #310, opened 2021-09-29
psp-2016 correct display of user management page

devinleighsmith:


frontend/src/layouts/AuthLayout.tsx
```diff
@@ -27,6 +27,7 @@ const AuthLayout: React.FC = ({ children }) => {
 
 const Content = styled.div`
   grid-area: content;
+  overflow: auto;
```
devinleighsmith, line 30: the content region should never overflow, as this could push the footer out. Essentially the height dimensions of the header, footer, and navbar are fixed, and the content takes up the remaining space. If there is not enough space to display the content, scroll.

frontend/src/features/admin/users/ManageUsers.tsx
```diff
@@ -100,7 +100,7 @@ export const ManageUsers = () => {
   useEffect(() => {
     fetchUsers(
       toFilteredApiPaginateParams<IUsersFilter>(
-        pageIndex,
+        pageIndex ?? 0,
```
devinleighsmith, line 103: The api doesn't like a null value here.

asanchezr (approved):

sookeke:
@devinleighsmith  @asanchezr You will need to update your forked repo to use the latest Image security scan workflow. your repo is using the old one which will block you from merging. Please update your repo.

see https://github.com/bcgov/PSP/pull/311

PR #326, opened 2021-10-04
Revert "psp-1899 add missing default values to database. (#253)"

devinleighsmith:
This reverts commit 652e1aaf281c42510d17216609259860a8b4993f.

Based on testing EF generated migrations for adding a default to an existing column throws an error when running multiple migrations. Recommend removing this for now until I can come up with a fix.


devinleighsmith:
My suggestion is to right a database script to handle this migration instead.

asanchezr (approved):

PR #330, opened 2021-10-04
PSP-2018 Property List View pagination bugfix

asanchezr:
Other updates:
- new **make start-infra** command to facilitate starting containers required for local dev (database, keycloak and geoserver)
  - is just a shortcut for **make start n=keycloak**, **make start n=database**, etc
- Updated some styles to be consistent with new style guide.
- Removed unnecessary (and error prone) CSS calc for container heights by properly applying grid and flex values

### DIFF
--- a/frontend/src/components/Table/Table.tsx
+++ b/frontend/src/components/Table/Table.tsx
@@ -671,7 +671,7 @@ const Table = <T extends IIdentifiedObject, TFilter extends object = {}>(
       {!props.hideToolbar && (
         <div className="table-toolbar">
           {props.pageSize !== -1 && <TablePagination<T> instance={instance} />}
-          {!props.lockPageSize && props.data.length > 0 && !props.lockPageSize && (
+          {!props.lockPageSize && props.data.length > 0 && (
             <TablePageSizeSelector
               options={props.pageSizeOptions || DEFAULT_PAGE_SELECTOR_OPTIONS}
               value={props.pageSize || DEFAULT_PAGE_SIZE}
--- a/frontend/src/components/maps/leaflet/styles.ts
+++ b/frontend/src/components/maps/leaflet/styles.ts
@@ -5,6 +5,7 @@ export const MapContainer = styled.div`
 `;
 
 export const MapGrid = styled.div`
+  width: 100%;
   display: grid;
   grid: ${props => props.theme.css.mapfilterHeight} 1fr / min-content 1fr;
   grid-template-areas:
--- a/frontend/src/features/admin/users/ManageUsers.tsx
+++ b/frontend/src/features/admin/users/ManageUsers.tsx
@@ -10,7 +10,7 @@ import find from 'lodash/find';
 import isEmpty from 'lodash/isEmpty';
 import queryString from 'query-string';
 import { useCallback, useEffect, useMemo } from 'react';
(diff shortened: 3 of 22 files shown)


frontend/src/components/Table/Table.tsx
```diff
@@ -671,7 +671,7 @@ const Table = <T extends IIdentifiedObject, TFilter extends object = {}>(
       {!props.hideToolbar && (
         <div className="table-toolbar">
           {props.pageSize !== -1 && <TablePagination<T> instance={instance} />}
-          {!props.lockPageSize && props.data.length > 0 && !props.lockPageSize && (
```
asanchezr, line 674: No need to check twice for the same thing

frontend/src/components/maps/leaflet/styles.ts
```diff
@@ -5,6 +5,7 @@ export const MapContainer = styled.div`
 `;
 
 export const MapGrid = styled.div`
+  width: 100%;
```
asanchezr, line 8: We need to set the width so it takes all available space

devinleighsmith:
the toast messages still seem very small:

image

frontend/src/features/admin/users/ManageUsers.tsx
```diff
@@ -149,60 +131,62 @@ export const ManageUsers = () => {
...
-              sort={sort}
-              onPageSizeChange={size => dispatch(setUsersPageSize(size))}
-              loading={!(users && !users.isFetching)}
-              clickableTooltip="Click IDIR/BCeID link to view User Information page"
-            />
-          </TableContainer>
-        </>
-      }
-    </div>
+    <Styled.ListView fluid className="users-management-page">
```
asanchezr, line 134: The **users-management-page** doesn't seem to be set anywhere but didn't want to delete it to avoid breakage due to the current mix of SCSS and styled components in the app. I felt this falls more into a cleanup tech-debt story. I already fixed a few stylings as I saw them but didn't wanted to expand the scope of this ticket (which is mainly about property list view not manage users list)

devinleighsmith (approved):

frontend/src/layouts/styles.ts
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 export const AppGridContainer = styled.div`
   padding: 0;
   text-align: center;
-  grid: ${props => props.theme.css.headerHeight} auto ${props => props.theme.css.footerHeight} / min-content 1fr;
+  grid: ${props => props.theme.css.headerHeight} 1fr ${props => props.theme.css.footerHeight} / min-content 1fr;
```
asanchezr, line 6: We need **1fr** here so the content area in the grid takes all available space between header and footer. Auto does not do that but instead takes whatever space needed by the children.

We want it to take all available space so list views don't show "short" scrolls bar when using 5 items per page like shown here (in DEV now)

image

asanchezr:
@devinleighsmith - I couldn't find how to modify that toast message font-size. Happy to change it if you point me in the right direction. If not, can be addresses in a future PR

devinleighsmith:
@asanchezr looks like that is being controlled via global bootstrap scss:

image
so you can override toast globally either in a new file or in App.scss

PR #335, opened 2021-10-07
psp-2074 IS12 schema changes.

devinleighsmith:


backend/dal/Helpers/PimsMigrationSqlGenerator.cs
```diff
@@ -19,10 +19,6 @@ public class PimsMigrationSqlGenerator : SqlServerMigrationsSqlGenerator
     {
```
devinleighsmith, line 19: Note that this file was the cause of the previous strange behaviour from EF with defaults.

there are two issues with the original implementation that I've noted below.

backend/dal/Helpers/PimsMigrationSqlGenerator.cs
```diff
@@ -56,113 +47,44 @@ public PimsMigrationSqlGenerator(MigrationsSqlGeneratorDependencies dependencies
...
-        protected override void ColumnDefinition(
-            string schema,
-            string table,
-            string name,
-            ColumnOperation operation,
-            IModel model,
-            MigrationCommandListBuilder builder)
-        {
-            _currentEntityType = GetEntityType(table);
-            _currentTableName = table;
```
devinleighsmith, line 103: This isn't thread safe, so there may be race conditions where the _currentTableName gets used innapropriately.

backend/dal/Helpers/PimsMigrationSqlGenerator.cs
```diff
@@ -56,113 +47,44 @@ public PimsMigrationSqlGenerator(MigrationsSqlGeneratorDependencies dependencies
...
-        Type GetEntityType(string table)
+        protected override void Generate(CreateTableOperation operation, IModel model, MigrationCommandListBuilder builder, bool terminate = true)
         {
-            return table switch
+            Dictionary<string, string> defaultValues = new Dictionary<string, string>();
+            foreach (var column in operation.Columns)
             {
-                "__EFMigrationsHistory" => null,
-                "PIMS_MIGRATION_HISTORY" => null,
-                _ => _tableTypes.GetValueOrDefault(table) ?? throw new InvalidOperationException($"Entity model is required for table {table}")
```
devinleighsmith, line 81: This will always throw this error in case of deleted entities. In practice, this causes all old migrations to fail when an entity is deleted, not good.

backend/dal/Helpers/Extensions/LeaseExtensions.cs
```diff
@@ -25,7 +25,8 @@ private static IQueryable<Entity.Lease> GenerateCommonLeaseQuery(this IQueryable
             filter.ThrowIfNull(nameof(user));
 
             if (!String.IsNullOrWhiteSpace(filter.TenantName)) {
-                query = query.Where(l => l.Tenant != null && EF.Functions.Like(l.Tenant.Surname + ", " + l.Tenant.FirstName + ", " + l.Tenant.MiddleNames, $"%{filter.TenantName}%"));
+                query = query.Where(l => l.Persons.Any(person => person != null && EF.Functions.Like(person.Surname + ", " + person.FirstName + ", " + person.MiddleNames, $"%{filter.TenantName}%"))
```
devinleighsmith, line 28: This ensures that updates to the schema do not break the pre-existing L&L search functionality.

PR #353, opened 2021-10-13
psp-1686 add common lease header.

devinleighsmith:


frontend/src/features/leases/detail/LeaseContainer.tsx
```diff
@@ -55,15 +64,19 @@ export const LeaseContainer: React.FunctionComponent<ILeaseAndLicenseContainerPr
...
+    <>
+      <LeaseLayout>
+        <LeaseBreadCrumb
+          leaseId={props.match.leaseId}
+          leasePage={leasePage}
+          onClickManagement={onClickManagement}
+        />
+        <LeaseHeader lease={lease} />
+        <BackToSearchButton />
+        <LeaseIndex currentPageName={leasePageName} leaseId={lease?.id}></LeaseIndex>
```
asanchezr, line 76: I see 3 ways to get the lease Id in this section:

- props?.match?.params?.leaseId
- props.match.leaseId
- lease?.id

Any change we could just pick one? I understand **lease** is undefined until we find it by the route param but once we find it we not use that instead of the route param?

devinleighsmith: Will correct in follow up PR, thanks.

PR #363, opened 2021-10-14
Psp-1689

devinleighsmith:
tenant page for lease and licenses.
re-org some files based on category.


backend/dal/Migrations/20211013173510_v0.2.12.1.cs
```diff
@@ -0,0 +1,23 @@
...
+using Pims.Dal.Helpers.Migrations;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v02121 : SeedMigration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
```
devinleighsmith, line 11: need a migration to handle missing seed values

frontend/src/features/leases/detail/LeasePages/details/DetailTermInformationBox.tsx
```diff
@@ -48,7 +47,7 @@ const StyledDate = styled.div`
 
 const StyledTermInformationBox = styled.div`
   width: 25.5rem;
-  background-color: ${props => props.theme.css.primaryColor};
+  background: rgba(0, 33, 66, 0.73);
```
asanchezr, line 50: This is okay for now but would love to see it added to the **colors.scss** file so we can re-use it when needed

frontend/src/utils/formUtils.ts
```diff
@@ -7,3 +7,23 @@
...
+ * The phoneFormatter is used to format the specified phone number value
+ * @param {string} phoneNumber This is the target phone number to be formatted
+ */
+export const phoneFormatter = (phoneNumber?: string) => {
+  if (!!phoneNumber) {
+    let result = phoneNumber;
+    const regex =
+      phoneNumber.length === 10
+        ? /(\d\d\d)[\s-]?(\d\d\d)[\s-]?(\d\d\d\d)/
+        : /\d(\d\d\d)[\s-]?(\d\d\d)[\s-]?(\d\d\d\d)/;
```
asanchezr, line 21: If we are going to support country codes in the phone number, this regex only allows for single digit country codes., e.g. USA/CAN (1). Other countries, say UK (44) would not be matched here.

Not sure it's worth supporting phone numbers outside of CAN but worth mentioning here.

https://countrycode.org/

frontend/src/features/leases/detail/LeasePages/details/DetailTermInformationBox.tsx
devinleighsmith: This is just the primary color with an alpha filter applied, I'm not convinced that deserves its own entry.

PR #455, opened 2021-11-02
Lease Surplus Declaration view - Psp 1949

FuriousLlama:


frontend/public/manifest.json
```diff
@@ -20,7 +20,7 @@
       "type": "image/webp"
     },
     {
-      "src": "tenants/moti/pims-logo-lg.png",
```
FuriousLlama, line 23: Noticed that this file did not exist on the public folder, causing errors

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
```diff
@@ -24,7 +24,7 @@ export const LeasePageForm: React.FunctionComponent<ILeasePageFormProps> = ({
     <StyledLeasePage>
       <StyledLeasePageHeader>
         <Styled.LeaseH2>{leasePage.title}</Styled.LeaseH2>
-        <p>{leasePage.description}</p>
+        {leasePage.description && <p>{leasePage.description}</p>}
```
FuriousLlama, line 27: if there is no description, do not render the paragraph

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      lease: { ...defaultFormLease, persons: [mockUser], organizations: [mockOrganization] },
+    });
+    expect(result.asFragment()).toMatchSnapshot();
+  });
+  it('Snapshot matches', () => {
+    const result = setup({
+      lease: { ...defaultFormLease, persons: [mockUser], organizations: [mockOrganization] },
+    });
+    expect(result.asFragment()).toMatchSnapshot();
+  });
```
asanchezr, line 39: This test seems to be a duplicate of the one above - right?

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+  it('Table row count is equal to the number of properties', () => {
+    const result = setup({
+      lease: {
+        ...defaultFormLease,
+        properties: mockProperties,
+        persons: [mockUser],
+        organizations: [mockOrganization],
+      },
+    });
+    const rows = result.container.querySelectorAll('[role="row"]');
```
asanchezr, line 50: React Testing Library provides **getByRole**, **getAllByRole** that should be used whenever possible.
https://testing-library.com/docs/queries/byrole/

```tsx
const rows = result.getAllByRole('row'); 
...
```

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.tsx
```diff
@@ -0,0 +1,86 @@
...
+  {
+    Header: 'Declaration Comments',
+    accessor: 'comments',
+    minWidth: 150,
+  },
+];
+
+const Surplus: React.FunctionComponent<{}> = () => {
+  const { values } = useFormikContext<ILease>();
+  const organizations: IProperty[] = getIn(values, withNameSpace(undefined, 'properties')) ?? [];
```
asanchezr, line 55: I think here `withNameSpace()` shouldn't take undefined but instead receive it via props like **Tenant.tsx** (line 21).
The namespace values allow to create sub-forms from a big complex Formik object. 
A quick scan of related code tell me that so far we are not nesting the forms (but instead consuming form fields from the main form) but if the other sub-components of the whole Lease page are build with namespacing, this one should follow in my opinion.

Any thoughts @devinleighsmith ?

asanchezr:
@FuriousLlama
Please run `make test`  in the **frontend/** folder to update Jest snapshots

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.tsx
FuriousLlama: Yeah, chatted about that one and although the other components are using the property, they are mostly set to undefine since the parent does not set it. My opinion would be to set it to undefined if is not currently set on the parent.

FuriousLlama:
> @FuriousLlama Please run `make test` in the **frontend/** folder to update Jest snapshots

I ran the 'update-snapshots' job and that seemed to do the trick. Is there a snapshot you think is missing?

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.test.tsx
FuriousLlama: that actually sounds way clearer! Will update the tests

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.test.tsx
FuriousLlama: oph yeah, totally. I think I was tinkering with the asFragment method and forgot to remove it

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.tsx
asanchezr: I'd rather add it to the parent or remove altogether on every sub-form. Namespacing adds complexity to the code and if it will not be used (because all fields will be on the main formik form) then better to remove `withNameSpace()` from the new forms

asanchezr:
```tsx
withNameSpace(undefined, 'properties') === 'properties'
```
So better to just use the string instead

asanchezr:
This is the build error from github actions:
image

I'm assuming snapshots are outdated. **make test** or **npm test -- -u**

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.tsx
FuriousLlama: updated

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.test.tsx
FuriousLlama: updated the top level. The subquery was needed since it was selecting inside an html element

frontend/src/features/leases/detail/LeasePages/surplus/Surplus.tsx
```diff
@@ -0,0 +1,85 @@
...
+  {
+    Header: 'Declaration Comments',
+    accessor: 'comments',
+    minWidth: 150,
+  },
+];
+
+const Surplus: React.FunctionComponent<{}> = () => {
+  const { values } = useFormikContext<ILease>();
+  const organizations: IProperty[] = getIn(values, 'properties') ?? [];
```
asanchezr, line 54: /nit: sorry just noticed this. var name seems off - shouldn't it be **properties** instead?

asanchezr (approved):

PR #468, opened 2021-11-08
PSP-2261 Correct NPE on L&L list view

devinleighsmith:
Loading many-to-many navigation properties in the previous manner is not safe. Update to explicitly include the many-to-many table instead of directly including the navigation property.


backend/dal/Migrations/v0.2.13.3/Up/PostUp/01-History.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE PIMS_LEASE_HIST ALTER COLUMN LEASE_PAY_RVBL_TYPE_CODE NVARCHAR(20) NULL
```
devinleighsmith, line 1: There is no down for this since this is essentially a bug in the original schema. The _HIST tables should have the same null constraints as the underlying table.

jdtoombs (approved):

PR #475, opened 2021-11-09
PSP-2250 Lease Improvements

asanchezr:
Screenshot 2021-11-09 115853


backend/api/Areas/Leases/Models/Lease/LeaseModel.cs
```diff
@@ -139,6 +139,26 @@ public class LeaseModel
...
+
+        /// <summary>
+        /// get/set - Whether this improvement contains a building that is subject to RTA.
+        /// </summary>
+        public bool IsSubjectToRta { get; set; }
+
+        /// <summary>
+        /// get/set - Whether this improvement contains a commercial building.
+        /// </summary>
+        public bool IsCommBldg { get; set; }
```
FuriousLlama, line 156: nit: variable is not that long when is not abbreviated and makes it more readable "IsCommBldg -> isCommercialBuilding" thoughts?

frontend/src/components/common/Grid/Grid.tsx
```diff
@@ -0,0 +1,83 @@
+import styled from 'styled-components';
+
+import GridItem from './GridItem';
+
+const formatAreas = (areas: string[]) => areas.map(area => `'${area}'`).join(' ');
+
+const frGetter = (value: string | number) =>
+  typeof value === 'number' ? `repeat(${value}, 1fr)` : value;
```
FuriousLlama, line 8: nit: frGetter sounds kind of obscure since this is generating a string. Some docs or what does fr mean wold help readability

frontend/src/components/common/Grid/Grid.tsx
```diff
@@ -0,0 +1,83 @@
...
+`;
+
+type GridType = typeof StyledGrid & {
+  Item: typeof GridItem;
+};
+
+const Grid = StyledGrid as GridType;
+Grid.Item = GridItem;
+
+export default Grid;
```
FuriousLlama, line 83: Is this new object used in this PR? couldn't find an example of its usage

backend/api/Areas/Leases/Models/Lease/LeaseModel.cs
asanchezr: I'm reusing the same name as the one coming from the DB/Entity. Happy to rename but just know we have tons of these

frontend/src/components/common/Grid/Grid.tsx
asanchezr: yeah - as I mentioned to Devin this PR is low on comments due to time constraints to get it out the door. Can add more comments later

asanchezr: Also "fr" is standard CSS grid terminology:

https://css-tricks.com/introduction-fr-css-unit/
https://www.digitalocean.com/community/tutorials/css-css-grid-layout-fr-unit

asanchezr: But you are right, the intent of this helper function should be documented

frontend/src/components/common/Grid/Grid.tsx
```diff
@@ -0,0 +1,83 @@
...
+ *      <MySidebar />
+ *    </Grid.Item>
+ *
+ *    <Grid.Item area="content">
+ *      <MyContentArea />
+ *    </Grid.Item>
+ *
+ *    <Grid.Item area="ads">
+ *      <MyAdsBanner />
+ *    </Grid.Item>
```
FuriousLlama, line 55: q: Is this component based on a react component? What are the benefits of it over the bootstrap <Row> <Col>?

asanchezr: 1. React-Bootstrap doesn't provide a Grid component (that I know of)
2. We have been moving to CSS Grid from Flexbox in newer components - however we are copy/pasting a lot when we could create a generic Grid component
3. You are right that is not being used because I though I would in my code but ended up not doing it to avoid refactoring existing grids (such as FormGrid)
4. Sets the table for such refactoring (if needed). If not, you can use it if you find it simpler to mucking with CSS/styled components directly. Up to the dev

asanchezr: Also note this would be for simple cases. I did not include all available CSS Grid properties here.
You can always add you custom grid prop like so:

```tsx
const MyGrid = styled(Grid)`
  some-grid-prop: some-value;
`
```

frontend/src/components/common/Grid/Grid.tsx
asanchezr: Answered above. If not desired now will delete but seem it could be useful (and plan to use it in my next PR)

frontend/src/components/common/form/TextArea.tsx
```diff
@@ -9,6 +9,8 @@ import { Input, InputProps } from './Input';
 export type TextProps = InputProps & {
   /** use FastInput instead of Input */
   fast?: boolean;
+  rows?: number;
+  cols?: number;
```
asanchezr, line 13: Needed to support standard HTML **textarea** properties - `rows` and `cols`

frontend/src/features/leases/detail/LeaseContainer.tsx
```diff
@@ -83,7 +84,7 @@ export const LeaseContainer: React.FunctionComponent<ILeaseAndLicenseContainerPr
     <>
       <LeaseLayout>
         <LeaseBreadCrumb
-          leaseId={props.match.leaseId}
+          leaseId={lease?.id}
```
asanchezr, line 87: small improvement - we already have a **lease** var here. No need to go back to the URL params

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
```diff
@@ -44,6 +44,7 @@ export const ViewEditToggleForm = styled(Form)`
   &#leaseForm {
     text-align: left;
     input:disabled,
+    select:disabled,
```
asanchezr, line 47: support "disabled" styling (ie no borders) on SELECT inputs as well

frontend/src/components/common/Grid/Grid.tsx
FuriousLlama: sorry, I think the last part of my question didn't render `<Row> <Col>`.
It is not being used, should we hold on to it? As it is right now is dead code. We could re-introduce it if necessary. Also, a demo would be cool!

FuriousLlama (approved):

frontend/src/features/leases/detail/styles.ts
```diff
@@ -11,6 +11,7 @@ export const LeaseH1 = styled.h1`
 
 export const LeaseH2 = styled.h2`
   font-size: 3.2rem;
+  line-height: 4.2rem;
```
asanchezr, line 14: These changes affect **all** Lease forms so they follow the specs from UXpin

frontend/src/components/common/Grid/Grid.tsx
asanchezr: I see, `<Row>` and `<Col>` are based on Flexbox. We have already tons of forms on Flexbox but decided in a dev meeting to start migrating towards CSS grid because:

1. It's less "verbose" and more expressive (ie you don't need a ton of Rows, Cols to achieve complex layouts)
2. It's supported by major browsers now so it's kinda safe to use CSS grid

frontend/src/components/common/Grid/Grid.tsx
devinleighsmith: Why not just use the package instead of copying/pasting code?
https://www.npmjs.com/package/@react-css/grid?activeTab=readme

frontend/src/components/common/Grid/Grid.tsx
devinleighsmith: Why not use the package instead of just copying the source? https://www.npmjs.com/package/@react-css/grid?activeTab=readme

frontend/src/components/common/Grid/Grid.tsx
asanchezr: We could but would be one more dep. Happy to switch (when we need it)

asanchezr: Oh I remembered why now. That one is **not** a styled component, so re-styling it with styled might not be straightforward

asanchezr: This one is a styled component (but I liked the API from the other package better) - https://github.com/azz/styled-css-grid

frontend/src/components/common/Grid/Grid.tsx
asanchezr: Will remove and replace with npm package when needed

asanchezr:
Removed grid component and renamed model fields for clarity

PR #480, opened 2021-11-10
psp-2167 prevent leading zeros from being stripped from pid search.

devinleighsmith:


backend/dal/Migrations/20211110022507_v1.0.14.2.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
FuriousLlama, line 1: Are these related to the bug? I am guessing the migrations are being piggybacked.

FuriousLlama (approved):

backend/dal/Migrations/20211110022507_v1.0.14.2.cs
devinleighsmith: This is related to the bug yes, the newer view provides the pid in the expected format.

devinleighsmith: Previously the view only exposed the pid as an int, and there isn't a lot you can do in cql to get around that.

PR #484, opened 2021-11-10
force refresh local geoserver layer.

devinleighsmith:
I did not realize this, but it seems like geoserver caches exposes layers, so my update to the location layer in my last PR will not be visible within geoserver. I was able to resolve this locally by recreating the layer, hopefully by propagating that configuration your local dev geoserver's will display the new layer. if not I recommend restarting the container.


devinleighsmith:
@asanchezr @FuriousLlama

FuriousLlama (approved):

PR #506, opened 2021-11-17
Dev Sprint 15.01 Database Schema

dfilteau:
PSP_PIMS S15.01 Sprint 15 2021-Nov-15
- Added tables:
  - ETL_DATA_SOURCE_TRACE
- Altered tables:
  - PIMS_PROPERTY_DATA_SOURCE_TYPE renamed to PIMS_DATA_SOURCE_TYPE
  - PIMS_LEASE
  - PIMS_SECURITY_DEPOSIT_RETURN
- Added views:
  - PIMS_CONTACT_MGR_VW


database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Alter Up/02_PSP_PIMS_S15_00_Alter.sql
```diff
@@ -0,0 +1,1086 @@
...
+PRINT N'Alter table dbo.PIMS_SECURITY_DEPOSIT'
+GO
+EXEC dbo.sp_rename N'dbo.PIMS_SECURITY_DEPOSIT.TOTAL_AMOUNT' , N'OTHER_DEP_HOLDER_TYPE_DESC', 'COLUMN'
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
+ALTER TABLE [dbo].[PIMS_SECURITY_DEPOSIT] ALTER COLUMN [OTHER_DEP_HOLDER_TYPE_DESC] nvarchar(100) NULL
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
```
asanchezr, line 376: This table modification will only work (I think) if there is not pre-existing data in the table. 
I'm not a SQL expert but would expect the tool to generate something like:

```sql
ALTER TABLE ... DROP COLUMN "X"
ALTER TABLE ... ADD COLUMN "Y"
```

Instead it is reusing the **same** column by renaming it and then changing the data type. Since the original column is for decimal values I think any existing data would cause this script to throw an error.

Again, I'm just highlighting potential issues in the future - I think this time it would be fine since there's no security deposit data at the moment

@devinleighsmith thoughts?

database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Alter Up/02_PSP_PIMS_S15_00_Alter.sql
```diff
@@ -0,0 +1,1086 @@
...
+PRINT N'Alter table dbo.PIMS_SECURITY_DEPOSIT_HIST'
+GO
+EXEC dbo.sp_rename N'dbo.PIMS_SECURITY_DEPOSIT_HIST.TOTAL_AMOUNT' , N'OTHER_DEP_HOLDER_TYPE_DESC', 'COLUMN'
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
+ALTER TABLE [dbo].[PIMS_SECURITY_DEPOSIT_HIST] ALTER COLUMN [OTHER_DEP_HOLDER_TYPE_DESC] nvarchar(100) NULL
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
```
asanchezr, line 410: Same issue as described above

asanchezr (approved):

database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Alter Up/02_PSP_PIMS_S15_00_Alter.sql
dfilteau: Definitely interesting.  I tested the Alter Up and had no problem but at the time of testing the table was empty.

dfilteau: As it turns out, decimal (money in this case) is an easy redefinition of the character representation of a DECIMAL value  to VARCHAR.  The DECIMAL (or MONEY) is stored internally as character numbers and the DECIMAL representation is basically a set of behaviours imposed on the DECIMAL value.  At least that's my understanding of it.

This is not the case for FLOAT values but the cure for that is something like (where Phonenumber is defined as FLOAT):
  ALTER TABLE Customer ALTER COLUMN Phonenumber decimal(25, 0);
  ALTER TABLE Customer ALTER COLUMN Phonenumber varchar(25);

Magic!

database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Alter Up/02_PSP_PIMS_S15_00_Alter.sql
dfilteau: Same answer as above.

database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Test Data/04_DML_LEASE_LOAD.sql
```diff
@@ -0,0 +1,318 @@
+/* -----------------------------------------------------------------------------
+Insert leases for testing
+. . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . 
+Author        Date         Comment
+------------  -----------  -----------------------------------------------------
+Doug Filteau  2021-Aug-19  Initial version
+----------------------------------------------------------------------------- */
+
+-- Insert the lease data
+INSERT INTO [dbo].[PIMS_LEASE]([LEASE_PURPOSE_TYPE_CODE], [LEASE_PURPOSE_SUBTYPE_CODE], [LEASE_STATUS_TYPE_CODE], [LEASE_PMT_FREQ_TYPE_CODE], [LEASE_PROGRAM_TYPE_CODE], [TENANT_ID], [TFA_FILE_NO], [PS_FILE_NO], [START_DATE], [RENEWAL_DATE], [EXPIRY_DATE], [LEASE_AMOUNT], [INSURANCE_START_DATE], [INSURANCE_END_DATE], [SECURITY_START_DATE], [SECURITY_END_DATE], [INSPECTION_DATE], [INSPECTION_NOTES], [LEASE_NOTES], [UNIT], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_DIRECTORY], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_DIRECTORY])
```
devinleighsmith, line 10: It looks like PSP_PIMS_S15_00 inadvertently overwrote all of my updates to these LEASE test data scripts again. This appears to be an older version of this script that won't run against the current schema. 

How can I better communicate those changes to you? I brought this up in a previous PR but it still seems to be a problem.

My changes are in PSP_PIMS_S13_00

devinleighsmith: @dfilteau

dfilteau: The Test Data folder was in there for reference and I didn't expect it to be used currently.  We should extract the test data you're using and capture it in this folder so it can be easily rolled out anywhere.

PR #511, opened 2021-11-18
PSP-2306 Lease Deposits

asanchezr:
Screenshot 2021-11-16 151713


backend/dal/Services/LeaseService.cs
```diff
@@ -120,6 +120,13 @@ public Lease Get(int id)
                     .ThenInclude(i => i.BctfaRiskManagementContact)
                     .ThenInclude(p => p.ContactMethods)
 
+                .Include(l => l.SecurityDeposits)
+                .ThenInclude(s => s.SecurityDepositHolderType)
```
FuriousLlama, line 124: nit: since the ThenInclude corresponds to the previous statement we could tab them to make it easier to follow

asanchezr: I assumed auto-format on save would take care but seems it won't. Sure I can indent it

frontend/src/components/Table/helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+import { formatMoney, formatNumber, prettyFormatDate } from 'utils';
+
+/**
+ * These helper methods below provide ways to render common values on a Table cell;
+ *  - dates
+ *  - money
+ *  - percentages
+ */
+
+export const renderDate = <T extends object>({ cell: { value } }: CellProps<T, string>) =>
```
FuriousLlama, line 11: Nit: it would help readability to put it under a static class or namespace so it reads **TableUtils.renderDate()** instead of just **renderDate()**

backend/entities/SecurityDeposit.cs
```diff
@@ -77,7 +77,7 @@ public class SecurityDeposit : BaseAppEntity
         /// get/set - The annual interest rate.
         /// </summary>
         [Column("ANNUAL_INTEREST_RATE")]
-        public int AnnualInterestRate { get; set; }
+        public decimal AnnualInterestRate { get; set; }
```
asanchezr, line 80: This should really be a decimal - e.g. 2.1% interest (not an int)

backend/api/Areas/Leases/Models/Lease/SecurityDepositModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        /// <summary>
+        /// get/set - Foreign key to the security deposit holder.
+        /// </summary>
+        public string SecurityDepositHolderTypeId { get; set; }
+
+        /// <summary>
+        /// get/set - Security deposit holder.
+        /// </summary>
+        public string SecurityDepositHolderType { get; set; }
```
FuriousLlama, line 26: instead of passing the type as a string, it could be an object. that way the id (type code) and description are encapsulated

backend/api/Areas/Leases/Models/Lease/SecurityDepositReturnModel.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// get/set - the deposit total, less any claims against.
+        /// </summary>
+        public decimal ReturnAmount { get; set; }
+
+        /// <summary>
+        /// get/set - The date when the deposit was returned.
+        /// </summary>
+        public DateTime ReturnDate { get; set; }
+
+        public string ChequeNumber { get; set; }
```
FuriousLlama, line 53: docs missing compared to other fields

backend/api/Areas/Leases/Mapping/Lease/SecurityDepositMap.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config.NewConfig<Entity.SecurityDeposit, Model.SecurityDepositModel>()
+                .Map(dest => dest.Id, src => src.Id)
+                .Map(dest => dest.RowVersion, src => src.RowVersion)
+                .Map(dest => dest.SecurityDepositHolderTypeId, src => src.SecurityDepositHolderTypeId)
+                .Map(dest => dest.SecurityDepositHolderType, src => src.SecurityDepositHolderType.Description)
+                .Map(dest => dest.SecurityDepositTypeId, src => src.SecurityDepositTypeId)
+                .Map(dest => dest.SecurityDepositType, src => src.SecurityDepositType.Description)
```
FuriousLlama, line 17: The types could be mapped directly having the model like:
`public TypeModel<string> SecurityDepositType{ get; set; }`
`.Map(dest => dest.SecurityDepositType, src => src.SecurityDepositType)`

frontend/src/components/Table/helpers.ts
asanchezr: I can but that is debatable. The pattern you mention (static classes to hold util methods) is widespread in NET but not javascript. Since every module (file) is kinda of it's own static module there is no need to create an additional static class for that. 

What I mean with this is that every time you "import something from './somemodule'" it gets cached and it is the same thing

backend/api/Areas/Leases/Models/Lease/SecurityDepositReturnModel.cs
asanchezr: I thought we agreed only to provide comments where they would add value. Happy to add comment saying "this is the cheque number" but kinda obvious

backend/api/Areas/Leases/Models/Lease/SecurityDepositModel.cs
asanchezr: I'm following the same pattern as other models that were written before me. Let's discuss best practice moving forward and create a tech debt ticket. what do you think?

backend/api/Areas/Leases/Mapping/Lease/SecurityDepositMap.cs
asanchezr: See my previous comment. Been following an existing pattern. Happy to discuss as a team what approach is best

backend/api/Areas/Leases/Models/Lease/SecurityDepositModel.cs
FuriousLlama: I think it works either way. The only problem I see is that the type information is not encapsulated, leaving the naming to show the relationship. Non blocking either way

FuriousLlama (approved):

backend/dal/Services/LeaseService.cs
FuriousLlama: I dont think is smart enough to know that the chaining is working on different layers of the object :(

frontend/src/components/Table/helpers.ts
FuriousLlama: I see. My comment is only about the readability of the file that is using the util methods. Also, it could avoid namespace collision if using multiple methods with the same name (not sure if that is possible right now).

asanchezr: In JS land if you want a "static class" you would do this in the consumer code:

```tsx
import * as TableUtils from '../../helpers';

...

TestUtil.renderDate()
```

asanchezr: This is kinda related as to whether to use **React.useEffect()** in code or import **useEffect()** directly. I prefer to use the direct method/helper until a collision is found - then we could namespace it

backend/api/Areas/Leases/Models/Lease/SecurityDepositModel.cs
asanchezr: Agree and happy to switch to an object but would prefer to merge the PR and then agree on all the instances where we are using this pattern moving forward.

PR #523, opened 2021-11-24
PSP-2308: Align tenant notes with rest of page

jdtoombs:
Added left margin to the note body to align with rest of tenant information under lease and licenses.


frontend/src/features/leases/detail/styles.ts
```diff
@@ -86,6 +87,10 @@ export const FormDescriptionLabel = styled(Form.Label)`
   font-weight: 700;
 `;
 
+export const TenantNotes = styled(TextArea)`
+  margin-left: 95px;
```
devinleighsmith, line 91: Use rem instead of px for everything except for media queries.

devinleighsmith: I'm also not convinced adding a margin here is the best solution. I'll try and have a closer look at this tmrw.

jdtoombs: Sounds good, will change to rem in the meantime

devinleighsmith (approved):

PR #527, opened 2021-11-24
add convenience script for deployment.

devinleighsmith:


database/mssql/scripts/clear-objects.sql
```diff
@@ -1,116 +1,55 @@
...
-
-WHILE @name IS NOT NULL
-BEGIN
-    SELECT @SQL = 'DROP SEQUENCE [dbo].[' + RTRIM(@name) +']'
-    EXEC (@SQL)
-    PRINT N'Dropped SEQUENCE: ' + @name
-    SELECT @name = (SELECT TOP 1 [name] FROM  sys.sequences WHERE [name] > @name ORDER BY [name])
-END
-GO
+declare @n char(1)
```
devinleighsmith, line 1: script was targetting non-app related schema (etl, pmbc) and then failing to remove those elements, causing the script to fail.

This update attempts to correct that @dfilteau

database/mssql/scripts/dbscripts/deploy.sh
```diff
@@ -0,0 +1,25 @@
+eval $(grep -v '^#' ../../.env | xargs)
```
devinleighsmith, line 1: requires sqlcmd to be installed.
provides an easier way to run all of the scripts in a build/alter folder against a target db.

jdtoombs (approved):

PR #533, opened 2021-11-25
PSP-1825 release automation

asanchezr:
This PR introduces two new GitHub actions:

- `workflows/tag.yml` to automatically tag all PRs merged into **master** with the appropriate version number
- `workflows/release.yml` to automatically create a GitHub release whenever a new tag gets pushed

Also of note is that release notes will be generated automatically based on labels assigned to PRs:

- enhancements
- bugs
- breaking-changes

All of the above is driven by configuration file `.github/release.yml`


.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*"
```
asanchezr, line 6: We only create tags against code landed in master by convention. So this action runs whenever a tag against master is created.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,40 @@
+name: Pre-release Tag
+
+on:
+  pull_request_target:
+    types: [closed]
+    branches: [master]
```
asanchezr, line 6: Pull request target and is safe here because we don't build any code so there no chance secrets could be leaked to a 3rd party. We only create a tag once PRs are merged

Makefile
```diff
@@ -33,14 +33,58 @@ help:
 
 .PHONY: help
 
+# these are various tools required by make commands below
+GH := $(shell command -v gh 2> /dev/null)
+PYTHON := $(shell command -v python 2> /dev/null)
+JQ := $(shell command -v jq 2> /dev/null)
```
asanchezr, line 39: These are tools required to be present in development environments for make commands to work. This change checks for said tools in advance to fail early and prompt devs to install what's needed. 

GitHub action environments come with those tools pre-installed.

Makefile
```diff
@@ -33,14 +33,58 @@ help:
...
+ifndef JQ
+	$(error "jq is not available please install from https://stedolan.github.io/jq")
+endif
+	$(info gh found...)
+	$(info python3 found...)
+	$(info jq found...)
+	@exit 0
+
+.PHONY: check-github-auth
+check-github-auth: ## Checks GITHUB_TOKEN has been set
```
asanchezr, line 62: GitHub cli needs either GITHUB_TOKEN set in the environment or devs can login via command `gh auth login`

Makefile
```diff
@@ -49,17 +93,41 @@ ifndef $(ARGS)
...
 
+CURRENT_VERSION := $(shell build/version.sh)
+CURRENT_DATE := $(shell date +'%b %d, %Y')
+tag = v$(CURRENT_VERSION)
+owner := $(shell gh repo view --json owner --jq '.owner.login')
+repo := $(shell gh repo view --json name --jq '.name')
+branch = HEAD
+release_notes = Release $(tag) ($(CURRENT_DATE))
+
+.PHONE: tag
```
asanchezr, line 104: Oops typo. Will fix

Makefile
```diff
@@ -49,17 +93,41 @@ ifndef $(ARGS)
...
 ##############################################################################
 # python is required for DevSecOps tools
-PYTHON := $(shell command -v python 2> /dev/null)
-JQ := $(shell command -v jq 2> /dev/null)
 
 .PHONY: devops-install
 devops-install: ## Installs software required by DevSecOps tooling (e.g. python, etc)
-	@if [ -z $(PYTHON) ]; then echo "Python could not be found. See See https://docs.python.org/3/"; exit 2; fi
-	@if [ -z $(JQ) ]; then echo "JQ could not be found. See See https://stedolan.github.io/jq/"; exit 2; fi
+	@if [ -z "$(PYTHON)" ]; then echo "Python not found. Install from "; exit 2; fi
```
devinleighsmith, line 129: from where?

devinleighsmith (approved):

Makefile
asanchezr: Good catch. Will fix

PR #550, opened 2021-11-25
Fix auto-versioning action

asanchezr:
Small PR to fix minor issues with the auto-versioning GitHub action

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -14,21 +14,25 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ssh-key: ${{ secrets.DEPLOY_KEY }} # with this we can bypass branch protection rules here (no-push)
+
       - uses: actions/setup-node@v2
         with:
-          node-version: "14"
-      - name: Bump version number
+          node-version: "16"
+
+      - name: Setup git
         run: |
           git config user.name "github-actions"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
 
+      - name: Bump version number
+        run: |
           # this just prints the new version number - will not update any files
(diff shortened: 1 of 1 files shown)


.github/workflows/version.yml
```diff
@@ -14,21 +14,25 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ssh-key: ${{ secrets.DEPLOY_KEY }} # with this we can bypass branch protection rules here (no-push)
+
       - uses: actions/setup-node@v2
         with:
-          node-version: "14"
-      - name: Bump version number
+          node-version: "16"
```
asanchezr, line 20: We should use Node 16

.github/workflows/version.yml
```diff
@@ -14,21 +14,25 @@ jobs:
...
 
+      - name: Bump version number
+        run: |
           # this just prints the new version number - will not update any files
-          NEW_VERSION=$(node build/bump-version.js --print-version)
+          NEW_VERSION=$(build/version-next.sh)
 
           # auto-increment version number upon merging pull requests
-          node build/bump-version.js --apply
+          build/bump.sh --apply
```
asanchezr, line 33: Use wrapper shell script instead of long node command (convenience)

.github/workflows/version.yml
```diff
@@ -14,21 +14,25 @@ jobs:
...
 
           # create commit
-          git add -A
-          git commit -m "chore: bump version to v${NEW_VERSION}"
+          git add frontend/package.json backend/api/Pims.Api.csproj
```
asanchezr, line 36: **Security:** Ensure we only add relevant files to the repo. We wouldn't want to add other files that might have been modified (by a 3rd party potentially)

devinleighsmith (approved):

PR #573, opened 2021-12-01
Re-applied filter and cleaned up legacy filter options for property.

FuriousLlama:
Fixes psp-2181 and cleans up filter logic and models

### DIFF
--- a/backend/api/Areas/Leases/Models/Search/LeaseFilterModel.cs
+++ b/backend/api/Areas/Leases/Models/Search/LeaseFilterModel.cs
@@ -11,7 +11,7 @@ namespace Pims.Api.Areas.Lease.Models.Search
         /// <summary>
         /// get/set - The unique identifier for titled property, either pid or pin.
         /// </summary>
-        public string PidOrPin { get; set; }
+        public string PinOrPid { get; set; }
 
         /// <summary>
         /// get/set - The value of the tenant name.
@@ -48,7 +48,7 @@ namespace Pims.Api.Areas.Lease.Models.Search
             // We want case-insensitive query parameter properties.
             var filter = new Dictionary<string, Microsoft.Extensions.Primitives.StringValues>(query, StringComparer.OrdinalIgnoreCase);
 
-            this.PidOrPin = filter.GetStringValue(nameof(this.PidOrPin));
+            this.PinOrPid = filter.GetStringValue(nameof(this.PinOrPid));
             this.TenantName = filter.GetStringValue(nameof(this.TenantName));
             this.LFileNo = filter.GetStringValue(nameof(this.LFileNo));
             this.Programs = filter.GetStringArrayValue(nameof(this.Programs));
@@ -68,7 +68,7 @@ namespace Pims.Api.Areas.Lease.Models.Search
                 Page = model.Page,
                 Quantity = model.Quantity,
 
-                PidOrPin = model.PidOrPin,
+                PinOrPid = model.PinOrPid,
                 TenantName = model.TenantName,
                 LFileNo = model.LFileNo,
                 Programs = model.Programs,
@@ -86,7 +86,7 @@ namespace Pims.Api.Areas.Lease.Models.Search
         public override bool IsValid()
         {
             return base.IsValid()
-                || !String.IsNullOrWhiteSpace(this.PidOrPin)
+                || !String.IsNullOrWhiteSpace(this.PinOrPid)
                 || !String.IsNullOrWhiteSpace(this.TenantName)
                 || !String.IsNullOrWhiteSpace(this.LFileNo);
         }
--- a/backend/dal/Helpers/Extensions/LeaseExtensions.cs
+++ b/backend/dal/Helpers/Extensions/LeaseExtensions.cs
@@ -4,6 +4,7 @@ using Microsoft.EntityFrameworkCore;
 using Pims.Core.Extensions;
 using Pims.Dal.Entities;
 using Entity = Pims.Dal.Entities;
+
 namespace Pims.Dal.Helpers.Extensions
 {
     /// <summary>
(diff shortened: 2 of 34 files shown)


backend/api/Areas/Leases/Models/Search/LeaseFilterModel.cs
```diff
@@ -11,7 +11,7 @@ public class LeaseFilterModel : PageFilter
         /// <summary>
         /// get/set - The unique identifier for titled property, either pid or pin.
         /// </summary>
-        public string PidOrPin { get; set; }
+        public string PinOrPid { get; set; }
```
FuriousLlama, line 14: cosmetic change to align with the naming everywhere else

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -97,25 +85,5 @@ public static IQueryable<Entity.PimsProperty> GeneratePropertyQuery(this PimsCon
 
             return query;
         }
-
-        /// <summary>
-        /// Return the pid (if valued) or pin of the property.
-        /// </summary>
-        /// <param name="property"></param>
-        /// <returns></returns>
-        public static int? GetPidOrPin(this Entities.PimsProperty property)
```
FuriousLlama, line 106: removed methods not used anywhere

backend/dal/Helpers/Extensions/LeaseExtensions.cs
```diff
@@ -27,13 +28,10 @@ private static IQueryable<Entities.PimsLease> GenerateCommonLeaseQuery(this IQue
...
-            if (!string.IsNullOrWhiteSpace(filter.PidOrPin))
+            if (!string.IsNullOrWhiteSpace(filter.PinOrPid))
             {
-                var pidOrPinValue = filter.PidOrPin.Replace("-", "").Trim();
-                if (Int32.TryParse(pidOrPinValue, out int pidOrPin))
-                {
-                    query = query.Where(l => l.PimsPropertyLeases.Where(p => p != null && p.Property != null).Select(p => p.Property).Count(p => p.Pid == pidOrPin || p.Pin == pidOrPin) > 0);
-                }
+                var pinOrPidValue = filter.PinOrPid.Replace("-", "").Trim();
+                query = query.Where(l => l.PimsPropertyLeases.Any(pl => pl != null && (pl.Property.Pid.ToString().Contains(pinOrPidValue) || pl.Property.Pin.ToString().Contains(pinOrPidValue))));
```
asanchezr, line 34: I think we have been using `EF.Functions.Like(...)` for partial matches. See line 27 above.
I would expect something like:

```cs
  ... && EF.Functions.Like(pl.Property.Pid.ToString(), $"%{pinOrPidValue}%") && ...
```

I'm unsure if using `X.ToString().Contains(Y)` is equivalent. Where you able to test partial matches?

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -30,40 +30,28 @@ private static IQueryable<Entity.PimsProperty> GenerateCommonPropertyQuery(this
...
-                query = query.Where(p => EF.Functions.Like(p.Name, $"%{filter.Name}%"));
 
-            if (!String.IsNullOrWhiteSpace(filter.PID))
+            if (!string.IsNullOrWhiteSpace(filter.PinOrPid))
             {
-                var pidValue = filter.PID.Replace("-", "").Trim();
-                if (Int32.TryParse(pidValue, out int pid))
-                    query = query.Where(p => p.Pid == pid || p.Pin == pid);
+                var pinOrPidValue = filter.PinOrPid.Replace("-", "").Trim();
+                query = query.Where(l => l.PimsPropertyLeases.Any(pl => pl != null && (pl.Property.Pid.ToString().Contains(pinOrPidValue) || pl.Property.Pin.ToString().Contains(pinOrPidValue))));
```
asanchezr, line 40: Same comment as above re: `EF.Functions.Like()`

backend/dal/Helpers/Extensions/LeaseExtensions.cs
FuriousLlama: Partial matches worked fine using Contains since that is translated to the correct SQL by EF:
```
WHERE ((@__pinOrPidValue_0 LIKE N'') OR 
          	(CHARINDEX(@__pinOrPidValue_0, CONVERT(VARCHAR(11), [p].[PID])) > 0)) OR 
          ((@__pinOrPidValue_0 LIKE N'') OR 
          	(CHARINDEX(@__pinOrPidValue_0, CONVERT(VARCHAR(11), [p].[PIN])) > 0))
```

FuriousLlama: Using like produces a similar one with the conversion
```
WHERE (CONVERT(VARCHAR(11), [p].[PID]) LIKE @__Format_1) OR (CONVERT(VARCHAR(11), [p].[PIN]) LIKE @__Format_1)
```

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -94,9 +92,11 @@ const whitelistedFilterKeys = ['PID', 'PIN', 'STREET_ADDRESS_1', 'LOCATION'];
...
  */
 const getQueryParams = (filter: IPropertyFilter): IGeoSearchParams => {
+  // The map will search for either identifier.
+  const pinOrPidValue = filter.pinOrPid ? filter.pinOrPid?.replace(/-/g, '') : undefined;
   return {
-    PID: filter.pid ? filter.pid?.replace(/-/g, '') : undefined,
-    PIN: filter.pin ? filter.pin?.replace(/-/g, '') : undefined,
+    PID: pinOrPidValue,
+    PIN: pinOrPidValue,
     STREET_ADDRESS_1: filter.address,
```
asanchezr, line 100: Did you test this with a live GeoServer? (dev container is fine)
I ask because there is a bit of change in the logic for the map here.

**Original:**
- Only one of PID or PIN is captured in the form so the query params would only have either PID=123 or PIN=123

**Updated:**
- We now send **two** query params with same value: `&PID=123&PIN=123`

This would likely affect the CQL query we send to GeoServer (@devinleighsmith  might be more familiar with that than I). See the code I'm talking about in `mapUtils.tsx` (lines 373-385)

```tsx
/**
 * Convert any object to a cql filter string, assuming the object's keys should be used as CQL filter properties.
 * AND all object keys together within the generated cql filter string.
 * @param object an object to convert to a cql filter string.
 */
export const toCqlFilter = (object: any) => {
  const cql: string[] = [];
  Object.keys(object).forEach((key: string) => {
    if (object[key]) {
      if (key === 'PID' && object[key]) {
        cql.push(`PIN ilike '%25${object[key]}%25' OR PID ilike '%25${object[key]}%25'`);
      } else {
        cql.push(`${key} ilike '%25${object[key]}%25'`);
      }
    }
  });
  return cql.length ? `cql_filter=${cql.join(' AND ')}` : '';
};

```

asanchezr: Effectively we would be sending an additional "AND" CQL query to GeoServer (i.e. `http://geoserver/...?cql_filter="PID ilike 123 AND PIN ilike 123"`) which would likely return no results (unless there is a property with same partial PID and PIN at the same time

asanchezr (approved):

frontend/src/components/maps/leaflet/Map.tsx
FuriousLlama: Ah, good observation. Previously only the pin parameter would be filled since the filter was mapping the pin/din entry to pin only.

I tested it on my local machine and it seemed to work

PR #577, opened 2021-12-02
Psp 2114

devinleighsmith:


Makefile
```diff
@@ -254,7 +254,7 @@ db-clean: ## create a new, clean database using the script file in the database.
 
 db-seed: ## create a new, database seeded with test data using the script file in the database. defaults to using the folder specified in database/mssql/.env, but can be overriden with n=PSP_PIMS_S15_00.
 	@echo "$(P) Seed the database with test data. n=FOLDER_NAME (PSP_PIMS_S15_00)"
-	TARGET_SPRINT=$(n) SEED=TRUE docker-compose up -d --build --force-recreate database; make pause-30; make keycloak-sync;
+	TARGET_SPRINT=$(n) SEED=TRUE docker-compose up -d --build --force-recreate database;
```
devinleighsmith, line 257: cannot run keycloak sync automatically afterwards without starting the docker container, leaving that to the db-refresh command.

database/mssql/scripts/dbscripts/PSP_PIMS_S15_01/Alter Up/02_PSP_PIMS_S15_00_Alter.sql
```diff
@@ -484,18 +484,6 @@ GO
 IF @@ERROR <> 0 SET NOEXEC ON
 GO
 
--- Create foreign key constraint PIM_PIDSRT_PIM_PRPRTY_FK
```
devinleighsmith, line 487: Aqua is trying to add a foreign key constraint here to a code table that has no values. This will always fail as long as there is at least one PIMS_PROPERTY row with a DATA_SOURCE_TYPE_CODE.

Temporarily solved by moving this add statement until after the code table is populated.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -15,6 +15,27 @@ GO
...
+ from sys.tables t
+  join sys.default_constraints d on d.parent_object_id = t.object_id
+  join sys.columns c on c.object_id = t.object_id and c.column_id = d.parent_column_id
+ where t.name = @table_name
+  and t.schema_id = schema_id(@schema_name)
+  and c.name = @col_name
+
+--print @Command
+
+execute (@Command)
```
devinleighsmith, line 37: This is a lot of code for a relatively small issue, but basically we don't know the name of the default because it wasn't specified when it was originally created. As a result, we have to look up the name of the default programmatically and drop it.

Aqua seems to be assuming that generated Default names will be the same in every environment. This is not true.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -1220,7 +1241,7 @@ GO
...
-	SELECT [LEASE_ID], [LEASE_PAY_RVBL_TYPE_CODE], [LEASE_LICENSE_TYPE_CODE], [LEASE_CATEGORY_TYPE_CODE], [LEASE_PURPOSE_TYPE_CODE], [LEASE_PROGRAM_TYPE_CODE], [LEASE_INITIATOR_TYPE_CODE], [LEASE_RESPONSIBILITY_TYPE_CODE], [LEASE_PMT_FREQ_TYPE_CODE], [MOTI_NAME_ID], [L_FILE_NO], [TFA_FILE_NO], [PS_FILE_NO], SUBSTRING(LEASE_DESCRIPTION,1,2147483647), [IS_ORIG_EXPIRY_REQUIRED], [LEASE_PURPOSE_OTHER_DESC], SUBSTRING(LEASE_NOTES,1,2147483647), [INCLUDED_RENEWALS], [RENEWAL_COUNT], [RENEWAL_TERM_MONTHS], SUBSTRING(RETURN_NOTES,1,2147483647), [ORIG_START_DATE], [ORIG_EXPIRY_DATE], [LEASE_AMOUNT], [RESPONSIBILITY_EFFECTIVE_DATE], [INSPECTION_DATE], SUBSTRING(INSPECTION_NOTES,1,2147483647), [IS_SUBJECT_TO_RTA], [IS_COMM_BLDG], [IS_OTHER_IMPROVEMENT], [IS_EXPIRED], [HAS_PHYSICAL_FILE], [HAS_DIGITAL_FILE], [HAS_PHYSICIAL_LICENSE], [HAS_DIGITAL_LICENSE], [CONCURRENCY_CONTROL_NUMBER], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_GUID], [APP_CREATE_USER_DIRECTORY], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_GUID], [APP_LAST_UPDATE_USER_DIRECTORY], [DB_CREATE_TIMESTAMP], [DB_CREATE_USERID], [DB_LAST_UPDATE_TIMESTAMP], [DB_LAST_UPDATE_USERID]
```
devinleighsmith, line 1223: I have no idea why Aqua is doing this, but it was trying to insert the MOTI_NAME_ID as the LEASE_STATUS_TYPE_CODE. This makes no sense to me. I've hardcoded this to 'Active' for now.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -1311,16 +1332,6 @@ GO
 -- Rebuild table dbo.PIMS_LEASE_HIST
 PRINT N'Rebuild table dbo.PIMS_LEASE_HIST'
 GO
-ALTER TABLE [dbo].[PIMS_LEASE_HIST]
```
devinleighsmith, line 1314: Why drop these constraints when we are completing replacing the table below? Also, once again Aqua is trying to delete constraints based on generated ids that won't be consistent across database instances.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -1378,7 +1389,7 @@ GO
...
-	SELECT [_LEASE_HIST_ID], [EFFECTIVE_DATE_HIST], [END_DATE_HIST], [LEASE_ID], [LEASE_PAY_RVBL_TYPE_CODE], [LEASE_LICENSE_TYPE_CODE], [LEASE_CATEGORY_TYPE_CODE], [LEASE_PURPOSE_TYPE_CODE], [LEASE_PROGRAM_TYPE_CODE], [LEASE_INITIATOR_TYPE_CODE], [LEASE_RESPONSIBILITY_TYPE_CODE], [LEASE_PMT_FREQ_TYPE_CODE], SUBSTRING(LEASE_DESCRIPTION,1,20), [L_FILE_NO], [TFA_FILE_NO], [PS_FILE_NO], SUBSTRING(LEASE_NOTES,1,200), [LEASE_PURPOSE_OTHER_DESC], SUBSTRING(RETURN_NOTES,1,200), SUBSTRING(INSPECTION_NOTES,1,200), [MOTI_NAME_ID], [IS_ORIG_EXPIRY_REQUIRED], [INCLUDED_RENEWALS], [RENEWAL_COUNT], [ORIG_START_DATE], [ORIG_EXPIRY_DATE], [LEASE_AMOUNT], [RESPONSIBILITY_EFFECTIVE_DATE], [INSPECTION_DATE], [IS_SUBJECT_TO_RTA], [IS_COMM_BLDG], [IS_OTHER_IMPROVEMENT], [IS_EXPIRED], [HAS_PHYSICAL_FILE], [HAS_DIGITAL_FILE], [HAS_PHYSICIAL_LICENSE], [HAS_DIGITAL_LICENSE], [CONCURRENCY_CONTROL_NUMBER], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_GUID], [APP_CREATE_USER_DIRECTORY], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_GUID], [APP_LAST_UPDATE_USER_DIRECTORY], [DB_CREATE_TIMESTAMP], [DB_CREATE_USERID], [DB_LAST_UPDATE_TIMESTAMP], [DB_LAST_UPDATE_USERID]
```
devinleighsmith, line 1381: More weirdness, this time Aqua is trying to assign the value of SUBSTRING(LEASE_DESCRIPTION,1,20) to LEASE_STATUS_TYPE_CODE. I don't understand why.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -2042,54 +2053,6 @@ CREATE TABLE [dbo].[ADS_SSDATA_1637956464247]  (
 	PRIMARY KEY CLUSTERED([SECURITY_DEPOSIT_RETURN_ID])
  ON [PRIMARY])
 GO
-EXEC sp_addextendedproperty
```
devinleighsmith, line 2045: Alter fails as all of these already exist.

database/mssql/scripts/dbscripts/PSP_PIMS_S16_00/Alter Up/01_PSP_PIMS_S15_01_Alter_Up.sql
```diff
@@ -3621,18 +3584,6 @@ GO
 IF @@ERROR <> 0 SET NOEXEC ON
 GO
 
--- Create foreign key constraint dbo.PIM_LSSTYP_PIM_LEASE_FK
```
devinleighsmith, line 3624: Same issue as last time, Aqua should not be trying to add a foreign key constraint to a code table that hasn't been populated yet. Also fixed by adding this after the population script has run.

devinleighsmith:
@dfilteau here are the 15.1 and 16 alter up corrections. Note that Aqua's behaviour here seems extremely weird.

I also had a conversation with James re: these issues. In the future the dev team will not accept new schema change requests unless the Alter Up/Build/Alter Down scripts pass testing.

One final thought, If we can't get Aqua to generate scripts that are data aware, we may need to rethink our process. To create these fixes, I essentially worked through all the failures that I encountered one by one until I resolved all of the failures. However, this is not a good approach, as there may be other failures in the script that the data in my database didn't happen to trigger, but may trigger in dev, test, uat or event prod. Given how weird some of these errors are, I think it would be hard to predict and find them given how much code Aqua generates for an alter.

We may want to consider switching to an alternative tool or even writing alters by hand to avoid these kind of issues, but I'd like to hear your thoughts.

devinleighsmith:
> @dfilteau here are the 15.1 and 16 alter up corrections. Note that Aqua's behaviour here seems extremely weird.
> 
> I also had a conversation with James re: these issues. In the future the dev team will not accept new schema change requests unless the Alter Up/Build/Alter Down scripts pass testing.
> 
> One final thought, If we can't get Aqua to generate scripts that are data aware, we may need to rethink our process. To create these fixes, I essentially worked through all the failures that I encountered one by one until I resolved all of the failures. However, this is not a good approach, as there may be other failures in the script that the data in my database didn't happen to trigger, but may trigger in dev, test, uat or event prod. Given how weird some of these errors are, I think it would be hard to predict and find them given how much code Aqua generates for an alter.
> 
> We may want to consider switching to an alternative tool or even writing alters by hand to avoid these kind of issues, but I'd like to hear your thoughts.

devinleighsmith:
Also, I haven't updated the Alter Down scripts. I imagine they will probably fail as well. I think I'd prefer to prioritize my dev work at the moment. If I have time before end of sprint I will circle back and try and fix those as well.

asanchezr (approved):

PR #585, opened 2021-12-04
Contact view of individual - Psp-2278

FuriousLlama:
image

image


backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Provides a contact-oriented organization model.
+    /// </summary>
+    public class OrganizationModel
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - The primary key to identify the organization.
+        /// </summary>
+        public long Id { get; set; }
```
asanchezr, line 12: I don't see the corresponding **isDisabled** flag to organizations (they are in the DB). Shouldn't it be added to the model as well?

backend/api/Models/TypeModel.cs
```diff
@@ -1,7 +1,7 @@
-namespace Pims.Api.Areas.Lease.Models.Lease
+namespace Pims.Api.Models
```
asanchezr, line 1: 👍  I like that we are making shared/generic - not just for leases!

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
FuriousLlama: They should, however this pr is only for individual

backend/dal/Services/UserService.cs
```diff
@@ -81,7 +81,7 @@ public PimsUser Activate()
                 this.Logger.LogInformation($"User Activation: key:{key}, email:{email}, username:{username}, first:{givenName}, surname:{surname}");
 
                 var person = new PimsPerson() { Surname = surname, FirstName = givenName };
-                this.Context.PimsPeople.Add(person);
+                this.Context.PimsPerson.Add(person);
```
asanchezr, line 84: I know you are just renaming an existing service here but I'd question the need to add application "users"; e,g, an MOTI admin or gov employee to the PimsPerson table. Wasn't this table meant to contain just contacts associated with things that can be managed by the app (e.g. tenants of leases etc?). 

What value would it have to add an ADM as a contact for example? 

I know this PR might not be the best place to discuss but would love to hear your thoughts @FuriousLlama  and @devinleighsmith

backend/dal/Services/UserService.cs
```diff
@@ -81,7 +81,7 @@ public PimsUser Activate()
...
                 this.Context.CommitTransaction();
 
                 user = new PimsUser()
```
asanchezr, line 87: To support my previous comment - this line is correct (in my opinion). Every user that we activate should be added to the PimsUser table (note this is different from PimsPerson)

frontend/src/features/contacts/contact/detail/Container.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { ContactBreadcrumb } from '../..';
+import * as Styled from '../../styles';
+import PersonView from './Person';
+
+interface IContactViewContainerProps {
+  match?: any;
+}
+
+const ContactViewContainer: React.FunctionComponent<IContactViewContainerProps> = props => {
```
asanchezr, line 13: Best practice is to have filename match the name of the main exported component. I would expect to export Contact here. If you don't like having multiple imports for "Contact" then please rename the corresponding files?

Also, this is fine (but I get why it would be confusing)

- contact/details/Container (exports Container)
- contacts/create/Container (exports Container)

Somewhere else, say **AppRouter.tsx**:

```tsx

import Container as ViewContainer from "contact/details";
import Container as CreateContainer from "contacts/create";
import Container as UpdateContainer from "contacts/edit";

// Use as ViewContainer, CreateContainer and UpdateContainer...
```

frontend/src/features/contacts/contact/detail/Container.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { contact } = useContactDetail(props?.match?.params?.id);
+  console.log(props?.match?.params);
```
asanchezr, line 15: please remove `console.log()`

frontend/src/features/contacts/contact/detail/Person.tsx
```diff
@@ -0,0 +1,197 @@
...
+  phoneTypes[ContactMethodTypes.WorkPhone] = 'Fax';
+
+  const personPhoneNumbers: ContactInfoField[] = getContactInfo(person, phoneTypes);
+
+  const emailTypes: Dictionary<string> = {};
+  emailTypes[ContactMethodTypes.WorkEmail] = 'Work';
+  emailTypes[ContactMethodTypes.PerseEmail] = 'Personal';
+  const personEmails: ContactInfoField[] = getContactInfo(person, emailTypes);
+
+  const personAddresses: AddressField[] = person.addresses.reduce(
```
asanchezr, line 47: Just my opinion and `array.reduce()` is fine here but since you are not using the previous value to compute the next one - wouldn't it be simpler just to use `array.map()` ?

frontend/src/features/contacts/contact/detail/Person.tsx
```diff
@@ -0,0 +1,197 @@
...
+        });
+      }
+      return accumulator;
+    },
+    [],
+  );
+
+  // Sort according to the dictionary order
+  return filteredFields.sort((a, b) => {
+    return Object.values(validTypes).indexOf(a.label) - Object.values(validTypes).indexOf(b.label);
```
asanchezr, line 193: **nit**: given that `Object.values(validTypes)` is not going to change between invocations - would it be possible to pre-compute it so we don't have to generate it every time?

e.g.

```tsx
  const validValues = Object.values(validTypes);

  return filteredFields.sort((a, b) => {
    return validValues.indexOf(a.label) - validValues.indexOf(b.label)
  });


```

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
asanchezr: I know but you added the Org model so assumed the change would be included. Not a bug deal - we just need to add it back when orgs are implemented. Hopefully we don't forget

asanchezr (approved):
Great work! I'd like to see this merged as soon as we can to avoid duplication of work. 
I was building some frontend models that are already present in this pull request

PR #591, opened 2021-12-07
PSP_PIMS_S16_02 updates. Support address usage type changes.

devinleighsmith:
Mild update to support the fact that I updated S16_00 to add the PIMS_TENANT table. 
Note that 16_01 and 16_02 both only have alter up (not build) if you want to migrate to 16_02, you can use run 
**make db-deploy n=PSP_PIMS_S16_01
y
Alter Up**
and then
**make db-deploy n=PSP_PIMS_S16_02
y
Alter Up**


backend/entities/AddressType.cs
```diff
@@ -33,4 +33,13 @@ public PimsAddressUsageType(string id):this()
...
         #endregion
     }
+
+    public static class AddressUsageTypes
+    {
+        public const string Mailing = "MAILADDR";
+        public const string ProofOfInsurance = "INSURE";
+        public const string RentalPayment = "RENTAL";
+        public const string PropertyNotification = "PROPNOTIFY";
+        public const string Physical = "PHYSADDR";
```
FuriousLlama, line 43: This ones seem to be using the old codes

devinleighsmith: Corrected, ugh.

FuriousLlama (approved):

PR #599, opened 2021-12-08
Organization Contact Features/psp 2125

FuriousLlama:
image


backend/api/Areas/Contacts/Mapping/Contact/AddressMap.cs
```diff
@@ -26,6 +26,25 @@ public void Register(TypeAdapterConfig config)
...
+                .Map(dest => dest.StreetAddress2, src => src.Address.StreetAddress2)
+                .Map(dest => dest.StreetAddress3, src => src.Address.StreetAddress3)
+                .Map(dest => dest.Municipality, src => src.Address.MunicipalityName)
+                .Map(dest => dest.RegionId, src => src.Address.RegionCode)
+                .Map(dest => dest.Region, src => src.Address.RegionCodeNavigation.RegionName)
+                .Map(dest => dest.DistrictId, src => src.Address.DistrictCode)
+                .Map(dest => dest.District, src => src.Address.DistrictCodeNavigation.DistrictName)
+                .Map(dest => dest.ProvinceId, src => src.Address.ProvinceStateId)
+                .Map(dest => dest.Province, src => src.Address.ProvinceState.Description)
+                .Map(dest => dest.ProvinceCode, src => src.Address.ProvinceState.ProvinceStateCode)
```
devinleighsmith, line 43: This is slightly risky business, as if any of these properties are null the mapping will NPE out. Are all of these fields required right now? if so this is technically fine, but isn't defensive.

backend/api/Areas/Contacts/Mapping/Contact/OrganizationMap.cs
```diff
@@ -12,12 +10,14 @@ public void Register(TypeAdapterConfig config)
         {
             config.NewConfig<Entity.PimsOrganization, Model.OrganizationModel>()
                 .Map(dest => dest.Id, src => src.Id)
-                .Map(dest => dest.Name, src => src.OrganizationName)
-                .Map(dest => dest.Address, src => src.GetSingleAddress())
-                .Map(dest => dest.ContactName, src => src.GetFullName())
-                .Map(dest => dest.Landline, src => src.GetLandlinePhoneNumber())
-                .Map(dest => dest.Mobile, src => src.GetMobilePhoneNumber())
-                .Map(dest => dest.Email, src => src.GetEmail());
+                .Map(dest => dest.IsDisabled, src => src.IsDisabled)
```
devinleighsmith, line 13: Can you confirm that this doesn't break the view tenant page?

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
```diff
@@ -7,39 +9,49 @@ public class OrganizationModel
...
         public long Id { get; set; }
 
         /// <summary>
-        /// get/set - The organization's company name.
+        /// get/set - The organization's disabled status flag.
+        /// </summary>
+        public bool IsDisabled { get; set; }
+
+        /// <summary>
+        /// get/set - The organization's concatenated name.
```
devinleighsmith, line 22: concatenated name? what do you mean? isn't this a discrete field?

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
```diff
@@ -7,39 +9,49 @@ public class OrganizationModel
...
+
+        /// <summary>
+        /// get/set - The organization's incorporation number.
         /// </summary>
-        public string ContactName { get; set; }
+        public string IncorporationNumber { get; set; }
 
         /// <summary>
-        /// get/set - The organizations's email.
+        /// get/set - The organization's organizations.
```
devinleighsmith, line 37: Probably meant the organization's persons.

backend/api/Areas/Leases/Mapping/Lease/OrganizationMap.cs
```diff
@@ -14,7 +13,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.Name, src => src.OrganizationName)
                 .Map(dest => dest.Address, src => src.GetSingleAddress())
-                .Map(dest => dest.ContactName, src => src.GetFullName())
+                .Map(dest => dest.ContactName, src => src.GetFirstPersonFullName())
```
devinleighsmith, line 16: Like this, thanks for updating this.

backend/entities/Organization.cs
```diff
@@ -21,7 +21,7 @@ public partial class PimsOrganization : IDisableBaseAppEntity
         [NotMapped]
         public string Description { get => OrganizationAlias; set => OrganizationAlias = value; }
 
-        public ICollection<PimsPerson> GetPersons() => PimsPersonOrganizations?.Select(p=> p.Person).ToArray();
+        public ICollection<PimsPerson> GetPersons() => PimsPersonOrganizations?.Select(po => po.Person).Select(p => { p.PimsPersonOrganizations = null; return p; }).ToArray();
```
devinleighsmith, line 24: I don't get why the PimsPersonOrganizations is being set to null here.

devinleighsmith: maybe an empty collection instead? not sure why necessary though.

devinleighsmith: certainly seems weird to do this at the entity level. So what you are saying here is that whenever I get persons I should never be able to see the personorganizations? seems weird.

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+  label: string;
+}
+
+interface AddressField {
+  address: IContactAddress;
+  label: string;
+}
+
+const OrganizationView: React.FunctionComponent<OrganizationViewProps> = ({ organization }) => {
+  const phoneTypes: Dictionary<string> = {};
```
devinleighsmith, line 33: This is probably fine in this case, but be careful with this in stateful screens. All of this initialization will happen **every time** this component is re-rendered, which is an unnecessary use of resources. You can either memoize this work within a function component, or you can move this out of the component to prevent it from getting re-initialized unnecessarily.

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const emailTypes: Dictionary<string> = {};
+  emailTypes[ContactMethodTypes.WorkEmail] = 'Work';
+  emailTypes[ContactMethodTypes.PerseEmail] = 'Personal';
+  const personEmails: ContactInfoField[] = getContactInfo(organization, emailTypes);
+
+  let personAddresses: AddressField[];
+  if (organization.addresses === undefined) {
+    personAddresses = [];
+  } else {
+    personAddresses = organization.addresses.reduce(
```
devinleighsmith, line 50: nit: could be more concise using organization?.addresses.reduce(... ?? [] rather then needing an if/else block.

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  return (
+    <>
+      <FormSection key={'contact-org-' + organization.id + '-names'} className="mb-4">
+        <Styled.RowAligned>
+          <Col>
+            <Styled.H2>
+              <FaRegBuilding size={20} className="mr-2" />
+              <span data-testid="contact-organization-fullname">{organization.name}</span>
```
devinleighsmith, line 67: try to avoid data-testid where possible. react testing library's philosophy is:
**The more your tests resemble the way your software is used, the more confidence they can give you.**
data-testid's are not how your user navigates the application, so avoid where possible

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <Styled.RowAligned>
+          <Col className="ml-5">
+            <div>
+              <strong>Alias:</strong>
+            </div>
+            <div>
+              <span data-testid="contact-organization-alias">{organization.alias}</span>
+            </div>
+            <div>
+              <strong>Incorportation Number:</strong>
```
devinleighsmith, line 88: typo, should be Incorporation

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+        </Styled.RowAligned>
+      </FormSection>
+      <FormSection key={'contact-org-' + organization.id + '-contacts'} className="mb-4">
+        <Styled.H2Primary>Contact info</Styled.H2Primary>
+        <Styled.RowAligned className="pb-4">
+          <Col md={2}>
+            <strong>Email:</strong>
+          </Col>
+          <Col>
+            {personEmails.length === 0 && <span>N.A</span>}
```
devinleighsmith, line 105: is this correct? I don't remember this.

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+            {personEmails.map((field: ContactInfoField) => (
+              <Styled.RowAligned>
```
devinleighsmith, line 107: react will probably complain that you are missing a key here. (keys are required when using map to generate jsx from an array).

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <Styled.RowAligned>
+          <Col>
+            <Styled.H2>
+              <FaRegBuilding size={20} className="mr-2" />
+              <span data-testid="contact-organization-fullname">{organization.name}</span>
+            </Styled.H2>
+          </Col>
+          <Col md="auto" className="ml-auto">
+            <Styled.StatusIndicators className={organization.isDisabled ? 'inactive' : 'active'}>
+              <FaCircle size={7} className="mr-2" />
```
devinleighsmith, line 72: Nit: think the icon bobbi is using is different, but I wouldn't want you to waste time on this either.

frontend/src/features/contacts/contact/detail/Organization.tsx
devinleighsmith: same comment applies to other usages of map (I don't want to clog this review with duplicate comments).

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+      </FormSection>
+      <FormSection key={'contact-org-' + organization.id + '-address'} className="mb-4">
+        <Styled.H2Primary>Address</Styled.H2Primary>
+        {personAddresses.map((field: AddressField, index: number) => (
+          <Styled.RowAligned className="pb-3">
+            <Col md="4">
+              <strong>{field.label}:</strong>
+            </Col>
+            <Col data-testid="contact-organization-address">
+              <div>{field.address.streetAddress1} </div>
```
devinleighsmith, line 145: I'm still a bit considered about the potential for NPEs here, but field.address should always be non-null I guess?

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+              {field.address.streetAddress2 && <div>{field.address.streetAddress2} </div>}
+              {field.address.streetAddress3 && <div>{field.address.streetAddress3} </div>}
+              <div>{field.address.municipality + ' ' + field.address.provinceCode} </div>
+              <div>{field.address.postal} </div>
+              <div>{field.address.country}</div>
+              {index + 1 !== personAddresses.length && <hr></hr>}
```
devinleighsmith, line 151: nit: Id be tempted to wrap this in a const function for readability.

frontend/src/features/contacts/contact/detail/Organization.tsx
```diff
@@ -0,0 +1,215 @@
...
+            </Col>
+          </Styled.RowAligned>
+        ))}
+      </FormSection>
+      <FormSection key={'contact-org-' + organization.id + '-individual'} className="mb-4">
```
devinleighsmith, line 156: nit: I generally like breaking these into subcomponents, but that is a style thing, not strictly required.

frontend/src/features/contacts/contact/detail/Person.tsx
```diff
@@ -75,28 +82,30 @@ const PersonView: React.FunctionComponent<IPersonViewProps> = ({ person }) => {
...
           </Col>
           <Col md="auto">
-            {person.organizations.map((organization: IContactOrganization, index: number) => (
-              <>
-                <span data-testid={'contact-person-organization-' + index}>
-                  <Link to={'/organization/' + organization.id}>{organization.name}</Link>
-                </span>
-              </>
-            ))}
+            {person.organizations &&
```
devinleighsmith, line 96: is this a null check?

backend/api/Areas/Contacts/Mapping/Contact/AddressMap.cs
FuriousLlama: Ah good point. I am used to the compiler pointing to the possible nullability of the objects. I will add the checks.

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
FuriousLlama: good catch, it came from the person's name.

backend/api/Areas/Contacts/Models/Contact/OrganizationModel.cs
FuriousLlama: yep

backend/entities/Organization.cs
FuriousLlama: The problem is that there is a circular relationship between the organizations and persons at the model level. This is not a problem most of the time since we use limited models for each, but afaik these are the first "full-fledged" models, exposing the circular links between them.
```
PimsOrganization -> PimsPersonOrganizations -> PimsPersons-> PimsPersonOrganizations -> PimsOrganization....

Comment (FuriousLlama) [reply backend/entities/Organization.cs:24]: The section of code is removing the organization fields from the child persons so that the link is broken. It could be done elsewhere, like in the mapper. But since the purpose of these getters is to be used for the mappers it looked appropriate to happen here.

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 7 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:33]: That is a good point. I believe in this case is fine since if the organization changes the whole components needs to be re-calculated

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 8 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:50]: I think I tried this and the compiler didnt like it. I could give it a second go

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 9 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:67]: Interesting, I followed the example on the react site about testing, and most of the examples were using testing ids.
In regards to how is navigated, I would not try to simulate the user navigation or interaction with the page at the unit test level. In my experience front end unit-tests tent to become quite convoluted since the user navigates the page visually while the testing framework is doing it as markup. We should chat more about this to see what provides the best value both reading the tests and how maintainable they are

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 11 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:105]: The acceptance criteria do not specify what should happen if there is no data.

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 13 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:72]: Uhm, dont know which one Bobi is using

File: frontend/src/features/contacts/contact/detail/Organization.tsx (hunk 14 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Organization.tsx:145]: fair point, all street addresses could be null

File: frontend/src/features/contacts/contact/detail/Person.tsx (hunk 17 above)
Comment (FuriousLlama) [reply frontend/src/features/contacts/contact/detail/Person.tsx:96]: undefined check

File: frontend/src/features/contacts/contact/detail/Person.tsx [hunk 18]
```diff
@@ -142,27 +155,26 @@ const PersonView: React.FunctionComponent<IPersonViewProps> = ({ person }) => {
...
-                {field.address.municipality} {field.address.provinceCode}
-              </div>
-              <div>{field.address.postal}</div>
-              <div>{field.address.country}</div>
-              {index + 1 !== personAddresses.length && <hr></hr>}
+              <span data-testid="contact-person-address">
+                <div>{field.address.streetAddress1} </div>
+                {field.address.streetAddress2 && <div>{field.address.streetAddress2} </div>}
+                {field.address.streetAddress3 && <div>{field.address.streetAddress3} </div>}
+                <div>{field.address.municipality + ' ' + field.address.provinceCode} </div>
```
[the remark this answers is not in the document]
FuriousLlama: nope, it could be null

backend/api/Areas/Contacts/Mapping/Contact/OrganizationMap.cs
FuriousLlama: Yes, it will not break the tenant page. This model is exclusively used by the contacts page. Initially, it was used by the person section only so it was very minimal.

backend/entities/Organization.cs
devinleighsmith: Hmm... I don't actually think this is circular though, given that only related entities that you specify in .include and .includethen clauses are loaded by ef.

devinleighsmith: essentially the circular reference will only go as deep as whatever you specify in .include and .theninclude, so this will end up being redundant.

frontend/src/features/contacts/contact/detail/Organization.tsx
devinleighsmith: you can see it by clicking on it in uxpin (a download option appears). alternatively she will sometimes document in the uxpin comments section.

devinleighsmith (approved):

backend/entities/Organization.cs
FuriousLlama: it wasn't, it was causing a stack overflow on the mapper and breaking the json stringify if allowed the mapper to handle circular references. We can chat more about it later

PR #616, opened 2021-12-09
Cleaned unused usings on backend code

FuriousLlama:
Cleaned leftover usings that were causing a spike in issues on SonarQube
https://sonarqube-3cd915-tools.apps.silver.devops.gov.bc.ca/dashboard?id=483bc59d-f9ef-428c-866e-581d9ea69eab

Scaffold now uses data annotations
Data Annotations allow EF to use annotations on the entities for the table definitions instead of having to be defined in the context.
 https://docs.microsoft.com/en-us/ef/ef6/modeling/code-first/data-annotations

### DIFF
--- a/backend/dal.keycloak/Pims.Dal.Keycloak.csproj
+++ b/backend/dal.keycloak/Pims.Dal.Keycloak.csproj
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="Mapster.DependencyInjection" Version="1.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Identity" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\entities\Pims.Dal.Entities.csproj" />
--- a/backend/entities/ef/PimsAccessRequestOrganization.cs
+++ b/backend/entities/ef/PimsAccessRequestOrganization.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 
 #nullable disable
 
--- a/backend/entities/ef/PimsAccessRequestOrganization.cs
+++ b/backend/entities/ef/PimsAccessRequestOrganization.cs
@@ -1,30 +1,72 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+using Microsoft.EntityFrameworkCore;
 
 #nullable disable
 
 namespace Pims.Dal.Entities
 {
+    [Table("PIMS_ACCESS_REQUEST_ORGANIZATION")]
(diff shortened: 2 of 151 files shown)


backend/dal.keycloak/Pims.Dal.Keycloak.csproj
```diff
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="Mapster.DependencyInjection" Version="1.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Identity" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
```
asanchezr, line 13: We usually don't "downgrade" packages - any particular reason for this?

backend/entities/ef/PimsAccessRequestOrganization.cs
```diff
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
```
asanchezr, line 2: This sounds good to me but unless I'm missing something would be pointless - this file being auto-generated. 
Aren't all files under `entities/ef` regenerated by EF whenever the DB schema changes @devinleighsmith?

asanchezr (approved):

backend/entities/ef/PimsAccessRequestOrganization.cs
devinleighsmith: Yes, do not modify files in the ef* folder. If the generated files are generating warnings, you'll need to add a rule to ignore the ef folder.

FuriousLlama: why not?

backend/dal.keycloak/Pims.Dal.Keycloak.csproj
FuriousLlama: It was causing extra dependencies since the other Microsoft.Extensions.Logging libraries where on 5.0.0

backend/entities/ef/PimsAccessRequestOrganization.cs
asanchezr: This import in question is needed (I think) for **HashSet** which is used for collections of related entity (one-to-many, etc). Not sure that EF "scaffold" tool is so smart to only add the import when there is a collection in the generated entity. Most likely they hardcode the imports to cover all scenarios. 

In any case, we don't control what the tool generates so agree with Devin we need to add an exception to linting to ignore this folder

FuriousLlama: I re-ran it and it added the unused generic imports. 
We could:
- Add some rules to remove those folders
- or
- Run the formatter after the scaffold command is complete

backend/dal.keycloak/Pims.Dal.Keycloak.csproj
asanchezr: I see - good catch

backend/entities/ef/PimsAccessRequestOrganization.cs
asanchezr: I rather completely ignore an auto-generated folder than trying to "fix" it after each re-generation

devinleighsmith (approved):

PR #621, opened 2021-12-09
LISOS_PIMS IS15 updates for UAT

devinleighsmith:
Updates:
Handle address not being directly related to person/organization
The ETL_OPERATIONAL_SS_LOAD does not match the expected format, updated population script to handle
Handle defaults not working correctly by inserting values


asanchezr:
It's weird that an ETL pull request would fail the frontend unit tests... Maybe the a previous PR introduced it but the test has some running inconsistency?

image

devinleighsmith:
Instability possibly caused by tests running slower on github actions, unable to reproduce failure locally.

PR #633, opened 2021-12-13
Lisos etl fix

devinleighsmith:


devinleighsmith:
@FuriousLlama @asanchezr 

Sorry, I know these are hard to review.

This PR adds the following:
forcibly overwrite any existing data in the ETL_OPERATIONAL_SS_LOAD table, to ensure that the source data is consistent with what is expected by the ETL.

it also corrects an issue with some default sql where the database wasn't handling default values in the same manner as in IS14.

asanchezr (approved):

PR #640, opened 2021-12-15
psp-2564 psp-2557

devinleighsmith:
bugfixes for create lease/license.

looks like the scaffold was missing updates from 16.02, this caused some null behaviour errors from the changes to typecodes.


FuriousLlama:
Good catch. We should talk to dough about adding tables or some other marker to know what is the db versions and what update versions where applied.

FuriousLlama (approved):

devinleighsmith:
> Good catch. We should talk to dough about adding tables or some other marker to know what is the DB versions and what update versions were applied.

Does the DBVERSION in the PIMS_STATIC_VARIABLE table meet this requirement?

FuriousLlama:
> 

Sweet. I didn't know that table existed. Thanks!

It looks exactly like something we could expose alongside the API and front-end versions. 
An improvement would be a list of the versions that DB scripts applied instead of a single value. That way we would know if there is any gap in the Deltas as well as dates for the db execution for further debugging.

devinleighsmith:
> > 
> 
> Sweet. I didn't know that table existed. Thanks!
> 
> It looks exactly like something we could expose alongside the API and front-end versions. An improvement would be a list of the versions that DB scripts applied instead of a single value. That way we would know if there is any gap in the Deltas as well as dates for the db execution for further debugging.

agreed, that would be nice. Logged tech debt for Doug to handle that:
https://jira.th.gov.bc.ca/browse/PSP-2579

PR #641, opened 2021-12-15
PSP-2512: Refactor test data scripts

jdtoombs:
- ensure test data scripts are paired up with seed data
- make scripts more generic (remove PIMS_TST tag)
- ensure compatible with new column/ db structure 
- tested against PSP_PIMS_S16_00


testing/data-scripts/sprint-14/TENANT_TEST_DATA.sql
```diff
@@ -1,35 +1,144 @@
...
+        1,
+        'WORKMOBIL',
+        1,
+        '1111111111',
+        'Test Data',
+        'Test Data',
+        'Test Data',
+        'Test Data'
+    )
+INSERT INTO
```
devinleighsmith, line 37: not a big deal, but since the insert columns are the same, could add both rows in one insert instead of two for contact method, organization.

devinleighsmith (commented):
thank you for this.

devinleighsmith (approved):

testing/data-scripts/sprint-14/TENANT_TEST_DATA.sql
FuriousLlama: +1 , It would be more valuable to have more 'realistic' test data. Otherwise, we risk having our synthetic data sets not reflect actual application ussage.

devinleighsmith: @jdtoombs please log tech debt for the above, but don't do that for this PR. The purpose of this PR is just to get our test data working with our existing seed data, not to clean up the test data itself.

jdtoombs: Makes sense, will log!

FuriousLlama (approved):

PR #642, opened 2021-12-15
add claim/associate with roles for PIMS_ROLE_R to prevent it from bei…

devinleighsmith:
…ng removed during keycloak-sync.


FuriousLlama (approved):

database/mssql/scripts/dbscripts/PSP_PIMS_S17_01/Alter Up/74_DML_PIMS_CLAIM.sql
```diff
@@ -0,0 +1,2 @@
+INSERT [dbo].[PIMS_CLAIM] ([CLAIM_ID], [CLAIM_UID], [NAME], [KEYCLOAK_ROLE_ID], [DESCRIPTION], [IS_DISABLED], [CONCURRENCY_CONTROL_NUMBER], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_GUID], [APP_CREATE_USER_DIRECTORY], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_GUID], [APP_LAST_UPDATE_USER_DIRECTORY], [DB_CREATE_TIMESTAMP], [DB_CREATE_USERID], [DB_LAST_UPDATE_TIMESTAMP], [DB_LAST_UPDATE_USERID]) VALUES (21, N'a9b9075b-2ea5-43b6-91df-2cc075453428', N'ROLE_PIMS_R', N'a9b9075b-2ea5-43b6-91df-2cc075453428', N'Read access to Property-related map layers', 0, 3, CAST(N'2021-11-29T02:17:49.903' AS DateTime), N'SEED', NULL, N'SEED', CAST(N'2021-11-29T07:42:25.990' AS DateTime), N'service', N'1b93f614-91da-4b32-b36e-bd2c6ebd12e2', N'', CAST(N'2021-11-29T02:17:49.903' AS DateTime), N'SEED', CAST(N'2021-11-29T07:42:25.967' AS DateTime), N'SEED')
+GO
```
FuriousLlama, line 2: do we need Alter Down scripts that would revert these lines? same for the other script

devinleighsmith: For scripts like this that fix bugs, with existing data, I'm tempted to not have alter down scripts, because that would return the database to an invalid state. What do you think?

FuriousLlama: My recommendation would be to match the state of the schema with the deployment version that was run. If there is no Alter Down I think it could leave those out of sync in the case of a version rollback. Worst if there are consecutive rollbacks.
Otherwise, we risk not being able to reproduce the state of the DB unless certain operations are done in a particular order.

asanchezr (approved):

PR #646, opened 2021-12-16
PSP-2497: Fix IDIR not displaying

jdtoombs:
After some of the database first changes the business identifier field name underwent some changes. `businessIdentifier` is still being returned and the paged users was trying to map from a non existent (in this case) `businessIdentifierValue`.

I simply added the old value name to the` IUser` interface, but let me know if we should make a separate type and or rename more on the backend.

image

manageusersbug

### DIFF
--- a/frontend/src/features/admin/edit-user/EditUserPage.test.tsx
+++ b/frontend/src/features/admin/edit-user/EditUserPage.test.tsx
@@ -48,7 +48,7 @@ const selectedUser = {
   surname: 'User',
   position: '',
   rowVersion: 1,
-  businessIdentifierValue: 'admin',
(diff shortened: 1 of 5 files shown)


devinleighsmith:
@jdtoombs have you tested that this change is successfully able to save users back to the DB without losing any data? (ie. able to edit a user).

jdtoombs:
> have you tested that this change is successfully able to save users back to the DB without losing any data? (ie. able to edit a user).

@devinleighsmith - just added another quick change, similar problem when displaying idir on the edit page but now displays and saves as expected

devinleighsmith (approved):

PR #671, opened 2021-12-18
Add, Update & Delete - Lease and Licenses  |  psp 2117

FuriousLlama:
Added front end and backend to support this feature. 
Cleaned up code on the DB context that required tracking the entity before making updates.


backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
```diff
@@ -17,23 +17,12 @@ public static class EntityEntryExtensions
         /// <param name="entry"></param>
         public static void UndoAppCreatedEdits(this EntityEntry entry)
         {
-            if (entry.Entity is IDisableBaseAppEntity entity)
```
FuriousLlama, line 20: Fixes the context checking for the wrong entity type. BaseAppEntity contains the timestamps

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
```diff
@@ -17,23 +17,12 @@ public static class EntityEntryExtensions
...
-                    entity.AppCreateUserid = oCreatedBy;
-
-                var oCreatedByKey = (Guid?)entry.OriginalValues[nameof(entity.AppCreateUserGuid)];
-                if (!entity.AppCreateUserGuid.HasValue || !oCreatedByKey.Equals(entity.AppCreateUserGuid))
-                    entity.AppCreateUserGuid = oCreatedByKey;
-
-                var oCreatedByDirectory = (string)entry.OriginalValues[nameof(entity.AppCreateUserDirectory)];
-                if (String.IsNullOrWhiteSpace(entity.AppCreateUserDirectory) || !oCreatedByDirectory.Equals(entity.AppCreateUserDirectory))
-                    entity.AppCreateUserDirectory = oCreatedByDirectory;
+                entry.Property(nameof(IBaseAppEntity.AppCreateTimestamp)).IsModified = false;
```
FuriousLlama, line 22: For these time stamps to continue not being set the entity needed to be loaded first. Now is just not set to be tracked.

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
```diff
@@ -44,15 +33,10 @@ public static void UndoAppCreatedEdits(this EntityEntry entry)
...
             {
-                var oCreatedOn = (DateTime)entry.OriginalValues[nameof(entity.DbCreateTimestamp)];
-                if (!oCreatedOn.Equals(entity.DbCreateTimestamp))
-                    entity.DbCreateTimestamp = oCreatedOn;
-
-                var oCreatedBy = (string)entry.OriginalValues[nameof(entity.DbCreateUserid)];
-                if (String.IsNullOrWhiteSpace(entity.DbCreateUserid) || !oCreatedBy.Equals(entity.DbCreateUserid))
-                    entity.DbCreateUserid = oCreatedBy;
+                entry.Property(nameof(IBaseAppEntity.DbCreateTimestamp)).IsModified = false;
+                entry.Property(nameof(IBaseAppEntity.DbCreateUserid)).IsModified = false;
```
FuriousLlama, line 39: Same as above, fixed entity type and removed tracking of unnecessary entity change.

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
```diff
@@ -32,7 +33,9 @@ export const LeasePageForm: React.FunctionComponent<ILeasePageFormProps> = ({
           ...lease,
         }}
         enableReinitialize={true}
-        onSubmit={noop}
+        onSubmit={() => {
+          refreshLease();
+        }}
```
FuriousLlama, line 38: not sure if there is a cleaner way to refresh the parent container.

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
devinleighsmith: This is not correct, IDisableBaseAppEntity inherits from IBaseAppEntity.

frontend/src/features/leases/detail/LeasePages/insurance/edit/EditContainer.tsx
```diff
@@ -0,0 +1,202 @@
...
+                operation: UpdateOperation.ADD,
+              });
+            }
+
+            return accumulator;
+          },
+          [],
+        );
+
+        console.log('updated', updatedVals);
```
FuriousLlama, line 116: will remove these markers once the tests are good to go.

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
devinleighsmith: I don't think I understand, if the entity is not loaded we would not be modifying the audit fields on it. This code only operates on loaded entities that are either added, modified, or deleted. 

`public override int SaveChanges()
        {
            // get entries that are being Added or Updated
            var modifiedEntries = ChangeTracker.Entries()
                    .Where(x => x.State.IsIn(EntityState.Added, EntityState.Modified, EntityState.Deleted));

            // Default values are provided because depending on the claim source it may or may not have these values.
            var username = _httpContextAccessor.HttpContext.User.GetUsername() ?? "service";
            var key = _httpContextAccessor.HttpContext.User.GetUserKey();
            var directory = _httpContextAccessor.HttpContext.User.GetUserDirectory() ?? "";
            foreach (var entry in modifiedEntries)
            {
                entry.UpdateAppAuditProperties(username, key, directory);
            }

            return base.SaveChanges();
        }`

frontend/src/features/leases/hooks/useLeaseDetail.ts
```diff
@@ -34,14 +31,23 @@ export const useLeaseDetail = (leaseId?: number) => {
...
     } else {
       toast.error(
         'No valid lease id provided, go back to the lease and license list and select a valid lease.',
       );
     }
-  }, [getLease, leaseId, dispatch]);
+  }, [getLease, dispatch, leaseId, needsUpdate]);
+
+  const refresh = () => {
+    setRefreshIndex(r => r + 1);
```
FuriousLlama, line 45: looks a tad hacky. If there is a better way to force an update I would love to hear it!

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
FuriousLlama: Then what would happen if the entity does not have a disabled entry, has timestamps?

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
FuriousLlama: Ef allows performing updates on entities that have not been loaded into context. It is like doing a SQL update without having to do a select first. AFAIK the tracker keeps a tab of what entities are set to be added/modified/deleted and does not need to load said entities.

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
devinleighsmith: oh, sorry I thought this was on line 81.

backend/dal/Helpers/Extensions/EntityEntryExtensions.cs
devinleighsmith: definitely a good correction though, thanks for this.

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Updates a list of insurances for an insurance.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost()]
+        [HasPermission(Permissions.LeaseEdit)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Api.Models.BatchUpdate.Reply<InsuranceModel>), 200)]
+        [SwaggerOperation(Tags = new[] { "insurance" })]
+        public IActionResult BatchUpdate(int leaseId, Api.Models.BatchUpdate.Request<InsuranceModel> batchUpdate)
```
devinleighsmith, line 53: We can take this as it is for now, but I have some scenarios that I worry about that would be solved by checking the concurrency on the lease:

1) concurrent adds. If user 1 adds insurance A, and then user 2 adds insurance A' (equivalent to A) If the backend doesn't enforce the concurrency id of the lease the system will end up with two of the "same" leases.
2) concurrent delete/updates If user 1 deletes insurance A, and then user 2 updates insurance A, what will EFs update method do? update dynamically adds new and updates existing entities. Hopefully because you should be passing an id ef will throw an error instead of adding it?

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
```diff
@@ -0,0 +1,85 @@
...
+        public InsuranceController(IPimsService pimsService, IMapper mapper)
+        {
+            _pimsService = pimsService;
+            _mapper = mapper;
+        }
+        #endregion
+
+        #region Endpoints
+        /// <summary>
+        /// Updates a list of insurances for an insurance.
```
devinleighsmith, line 45: probably meant updates a list of insurances for a lease.

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
```diff
@@ -0,0 +1,85 @@
...
+                else if (modification.Operation == Constants.UpdateType.Delete)
+                {
+                    var insurance = _pimsService.Insurance.Delete(insuranceEntity, false);
+                    updatedEntities.Add(insurance);
+                }
+            }
+
+            _pimsService.Insurance.CommitTransaction();
+
+            var modelsss = _mapper.Map<IList<Models.InsuranceModel>>(updatedEntities);
```
devinleighsmith, line 79: sss?

backend/dal/Services/InsuranceService.cs
```diff
@@ -0,0 +1,95 @@
...
+        public InsuranceService(PimsContext dbContext, ClaimsPrincipal user, IPimsService service, ILogger<InsuranceService> logger, IMapper mapper) : base(dbContext, user, service, logger, mapper) { }
+        #endregion
+
+        #region Methods
+        public PimsInsurance Get(long id)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.LeaseView);
+            return this.Context.PimsInsurances
+                .Where(i => i.InsuranceId == id)
+                .FirstOrDefault() ?? throw new KeyNotFoundException();
```
devinleighsmith, line 35: nit, I think this can be shortened to FirstOrDefault(i => i.InsuranceId == id).

Could also consider making this non-tracking. That way if someone calls get, inadvertently updates the entity while doing other work, those changes wouldn't be saved.

backend/api/Models/BatchUpdate/Reply.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        /// <summary>
+        /// get/set - The updated payload.
+        /// </summary>
+        public IList<T> Payload { get; set; }
+
+        /// <summary>
+        /// get/set - Batch error messages if any.
+        /// </summary>
+        public IList<string> ErrorMessages { get; set; } = new List<string>();
```
devinleighsmith, line 23: These aren't used currently correct? what is the use case for this?

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
devinleighsmith: a common pattern is to use the return value from the add/update to update the lease being displayed.

devinleighsmith: saves a round trip.

frontend/src/features/leases/hooks/useLeaseDetail.ts
```diff
@@ -6,11 +6,8 @@ import { hideLoading, showLoading } from 'react-redux-loading-bar';
...
 import { logError } from 'store/slices/network/networkSlice';
 
-/**
- * hook that fetches the lease given the lease id.
- * @param leaseId
- */
-export const useLeaseDetail = (leaseId?: number) => {
+export function useLeaseDetail(leaseId: number) {
+  const [needsUpdate, setRefreshIndex] = useState<number>(0);
   const [lease, setLease] = useState<ILease>();
```
devinleighsmith, line 11: if you were going to update the lease, you would return this setLease to the leasecontainer

frontend/src/features/leases/detail/LeasePages/insurance/edit/EditContainer.tsx
```diff
@@ -0,0 +1,202 @@
...
+      initialValues={initialValues}
+      onSubmit={(values: IUpdateFormInsurance, formikHelpers) => {
+        const updatedVals = values.insurances.reduce(
+          (accumulator: IEntryModification<IInsurance>[], entry: FormInsurance) => {
+            const existingEntry = initialValues.insurances.find(i => i.id === entry.id && !i.isNew);
+            console.log(entry, existingEntry);
+
+            if (existingEntry) {
+              if (!entry.isShown) {
+                accumulator.push({
```
devinleighsmith, line 94: Was wondering how this works in the case of a local delete. Ie: if a user creates a new entry, and then removes it before saving it.

frontend/src/features/leases/detail/LeasePages/insurance/details/Insurance.tsx
```diff
@@ -0,0 +1,44 @@
...
+  insuranceTypes: ILookupCode[];
+}
+
+const InsuranceDetailsView: React.FunctionComponent<InsuranceDetailsViewProps> = ({
+  insuranceList,
+  insuranceTypes,
+}) => {
+  return (
+    <>
+      {insuranceList
```
devinleighsmith, line 19: worth noting that this sort will run whenever the user interacts with the form, ie. every time they type a character.

devinleighsmith (approved):

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
FuriousLlama: I was thinking about that, but I thought it would be better to get the entire tree that renders this component since we don't have each page acting independently. like /leases/65/insurances vs  /leases/65/pages=insurances

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
FuriousLlama: That's a fair point. In the case of insurances, I don't think that is a problem since afaik there can be only one type of insurance per lease, so duplicates would not be allowed.

backend/api/Models/BatchUpdate/Reply.cs
FuriousLlama: We should chat about this too. Id does not look like we have a way of passing errors from each layer of the application. Ideally, we would have db errors bubbling up to services that could tell us further information about the processing. This is the first step to making those.

frontend/src/features/leases/hooks/useLeaseDetail.ts
FuriousLlama: Not sure I follow. I tried this approach with the previous definition of useLeaseDetail but React complained about calling hooks in a callback.

frontend/src/features/leases/detail/LeasePages/insurance/edit/EditContainer.tsx
FuriousLlama: It works since the isShown is what defines if it was removed from the checklist.

frontend/src/features/leases/detail/LeasePages/insurance/details/Insurance.tsx
FuriousLlama: yikes. I think for now it works fine since those are only 5 categories

backend/api/Models/BatchUpdate/Reply.cs
devinleighsmith: Does ErrorHandlingMiddleware.cs partially meet this need or no?

frontend/src/features/leases/detail/LeasePages/insurance/InsuranceContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const lookupCodes = useLookupCodeHelpers();
+  const insuranceTypes = lookupCodes.getByType(INSURANCE_TYPES).sort((a, b) => {
+    return (a.displayOrder || 0) - (b.displayOrder || 0);
+  });
+
+  const [isEditing, setEditing] = useState(false);
+  return (
+    <>
+      {!isEditing && (
```
devinleighsmith, line 24: this needs to be controlled by the lease edit claim. So
and then hasClaim(Claims.LEASE_EDIT)

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
```diff
@@ -0,0 +1,85 @@
...
+            IList<Dal.Entities.PimsInsurance> updatedEntities = new List<Dal.Entities.PimsInsurance>();
+            foreach (Api.Models.BatchUpdate.EntryModification<InsuranceModel> modification in batchUpdate.Payload)
+            {
+                Dal.Entities.PimsInsurance insuranceEntity = _mapper.Map<Dal.Entities.PimsInsurance>(modification.Entry);
+                insuranceEntity.LeaseId = leaseId;
+                if (modification.Operation == Constants.UpdateType.Update)
+                {
+                    var insurance = _pimsService.Insurance.Update(insuranceEntity, false);
+                    updatedEntities.Add(insurance);
+                }
```
asanchezr, line 64: I don't know where we landed on a decision about this "batch" update pattern but in principle I don't like that someone external to the API (ie the frontend) is dictating what entities should be added vs updated vs deleted (via that operation type)

How is this better than having the backend decide what it needs to do? ie

```cs
public static void InsertUpdateOrDeleteGraph(BloggingContext context, Blog blog)
{
    var existingBlog = context.Blogs
        .Include(b => b.Posts)
        .FirstOrDefault(b => b.BlogId == blog.BlogId);

    if (existingBlog == null)
    {
        context.Add(blog);
    }
    else
    {
        context.Entry(existingBlog).CurrentValues.SetValues(blog);
        foreach (var post in blog.Posts)
        {
            var existingPost = existingBlog.Posts
                .FirstOrDefault(p => p.PostId == post.PostId);

            if (existingPost == null)
            {
                existingBlog.Posts.Add(post);
            }
            else
            {
                context.Entry(existingPost).CurrentValues.SetValues(post);
            }
        }

        foreach (var post in existingBlog.Posts)
        {
            if (!blog.Posts.Any(p => p.PostId == post.PostId))
            {
                context.Remove(post);
            }
        }
    }

    context.SaveChanges();
}
```

backend/api/Areas/Insurances/Mapping/InsuranceMap.cs
```diff
@@ -0,0 +1,32 @@
...
+                .Map(dest => dest.OtherInsuranceType, src => src.OtherInsuranceType)
+                .Map(dest => dest.CoverageDescription, src => src.CoverageDescription)
+                .Map(dest => dest.CoverageLimit, src => src.CoverageLimit)
+                .Map(dest => dest.ExpiryDate, src => src.ExpiryDate)
+                .Map(dest => dest.IsInsuranceInPlace, src => src.IsInsuranceInPlace)
+                .Inherits<Entity.IBaseEntity, Api.Models.BaseModel>();
+
+            config.NewConfig<Model.InsuranceModel, Entity.PimsInsurance>()
+                .Map(dest => dest.InsuranceId, src => src.Id)
+                .Map(dest => dest.InsuranceTypeCode, src => src.InsuranceType.Id)
```
asanchezr, line 23: There's potential for NPE here as we don't control the JSON sent to the API. It could have `src.InsuranceType` be undefined/null. Should probably check for null before trying to do `src.InsuranceType.Id` ?

backend/api/Models/BatchUpdate/Request.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Pims.Api.Models.BatchUpdate
+{
+    /// <summary>
+    /// Provides a model for batch update requests.
+    /// </summary>
+    public class Request<T>
+    {
+        /// <summary>
+        /// get/set - The insurance's Id
```
asanchezr, line 11: please update this comment - not related to insurance Id

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
FuriousLlama: That is an interesting option. I think we have a meeting to figure out what pattern we like better.

backend/api/Areas/Insurances/Mapping/InsuranceMap.cs
FuriousLlama: I think if the insurance does not have a type an exception should be valid since that is should never be null

frontend/src/components/common/form/RadioGroup.tsx
```diff
@@ -39,6 +39,8 @@ type OptionalAttributes = {
   toolTip?: string;
   /** id for tooltip */
   toolTipId?: string;
+  /** Flex direction for the group */
+  flexDirection?: string;
```
asanchezr, line 43: nit: a bit better typing for this would be:
```tsx
flexDirection?: 'column' | 'row'
```

That way the TS compiler would complain if you ever assign `flexDirection: 'cat'`

backend/api/Areas/Insurances/Controllers/InsuranceController.cs
devinleighsmith: I think for this PR we are allowing this pattern but we will need to agree on a pattern going forward. I will create a meeting for after the holidays to come up with a unified pattern.

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
devinleighsmith: Yes, actually my PR will be addressing that, and moving us to using a sub-router for the lease pages.

frontend/src/components/common/form/RadioGroup.tsx
```diff
@@ -72,6 +75,7 @@ export const RadioGroup = ({
     <StyledRadioGroup
       controlId={`input-${field}`}
       className={classNames(!!required ? 'required' : '', className)}
+      $isFlexColumn={flexDirection ? flexDirection === 'column' : true}
```
asanchezr, line 78: You really don't need this intermediate "$isFlexColumn" here. I think it could be dropped in favor of `flexDirection` (which you exposed in the component props..)

```tsx

export const RadioGroup = ({
  flexDirection = 'column' // so it sets it to column if not supplied
  ...rest
}) => {
 // ...
 <StyledRadioGroup
   flexDirection={flexDirection}
/>
}

export const StyledRadioGroup = styled(Form.Group)`
  &.form-group {
    display: flex;
    margin-bottom: 0;
    .radio-group {
      display: flex;
      flex-direction: column;
      row-gap: 0.4rem;
      flex-direction: ${props => props.flexDirection};  // we already assigned a default above
      row-gap: ${props => (props.flexDirection == 'column' ? '0.4rem' : 'unset')};
      column-gap: ${props => (props.flexDirection == 'column' ? 'unset' : '0.4rem')};
`;

File: frontend/src/components/common/form/styles.ts [hunk 21]
```diff
@@ -60,3 +60,11 @@ export const FormSection = styled.div`
   padding: 2.5rem;
   background-color: ${props => props.theme.css.filterBackgroundColor};
 `;
+
+export const FormSectionClear = styled.div`
```
asanchezr, line 64: nit: it took me a while to get why this was named "__Clear". If I understood correctly it wanted to indicate that this is a FormSection with no background. Would you consider a name closer to "FormSectionNoBg" ?

frontend/src/features/leases/detail/LeasePages/insurance/InsuranceContainer.tsx
asanchezr: To add to this - I thought we agreed to have separate "view" and "edit/create/delete" forms to avoid a single entry point like this container that either shows read-only form or edit form

Having separate "containers" for edit vs details would also allow to use proper claims on each one and each would be on their own FE route: (/leases/id/insurance vs leases/id/insurance/edit)

It would also solve the need for that "hacky" triggerRefresh - once you finish with the edit form you just redirect to the read route which would re-fetch the updated info from the server - or am I missing something?

frontend/src/components/common/form/styles.ts
FuriousLlama: That's right, just needed one without the background. I don't have a strong opinion either way.

frontend/src/features/leases/detail/LeasePages/insurance/InsuranceContainer.tsx
FuriousLlama: I like it. For now I rather leave it as is since that refactoring looks like is part of a bigger conversation about the sub-view organization.

backend/api/Models/BatchUpdate/Reply.cs
FuriousLlama: For logging, yes, but it does not look like is raising those errors cleanly to the front end. We could chat more about this later

frontend/src/components/common/form/RadioGroup.tsx
FuriousLlama: Yeah, I was a tad torn on this. I changed to compare against the string column.

PR #683, opened 2021-12-31
Psp-2382

devinleighsmith:
Contains the lease update functionality as well as add/update improvements functionality.

Also contains refactoring of lease components to use an internal router, also places global lease object into a context to facilitate access by child pages.


backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
```diff
@@ -9,30 +9,33 @@ public class PropertyMap : IRegister
...
                 .Map(dest => dest.PropertyId, src => src.Id)
                 .Map(dest => dest.Property, src => src)
-                .Map(dest => dest.LeaseArea, src => src.AreaUnit)
-                .Map(dest => dest.Property, src => src.AreaUnitType.GetTypeId());
+                .Map(dest => dest.LeaseArea, src => src.LandArea)
+                .Map(dest => dest.Property, src => src.AreaUnitType.GetTypeId())
+                .Map(dest => dest.ConcurrencyControlNumber, src => src.RowVersion);
 
             config.NewConfig<Model.PropertyModel, Entity.PimsProperty> ()
                 .Map(dest => dest.PropertyId, src => src.Id)
```
asanchezr, line 35: Not an expert in mapping but why are we assigning the **same** Id to two different EF entities here?

```cs
config.NewConfig<Model.PropertyModel,Entity.PimsPropertyLease>()
                .Map(dest => dest.PropertyId, src => src.Id)
```

```cs
config.NewConfig<Model.PropertyModel, Entity.PimsProperty> ()
                .Map(dest => dest.PropertyId, src => src.Id)
```

backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
```diff
@@ -9,30 +9,33 @@ public class PropertyMap : IRegister
...
                 .Map(dest => dest.Pid, src => src.PID == "" ? "-1" : src.PID)
```
asanchezr, line 36: why do we need to use "-1" here? is it because PID is not nullable in the database?

backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
devinleighsmith: This is unflattening a model.

backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
devinleighsmith: presumably? not part of this PR.

backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
asanchezr: ok you are right

backend/dal/Services/LookupService.cs
```diff
@@ -157,6 +157,11 @@ public IEnumerable<PimsInsuranceType> GetInsuranceTypes()
         {
             return this.Context.PimsInsuranceTypes.AsNoTracking().OrderBy(a => a.InsuranceTypeCode).ToArray();
         }
+
+        public IEnumerable<PimsPropertyImprovementType> GetPropertyImprovementTypes()
+        {
+            return this.Context.PimsPropertyImprovementTypes.AsNoTracking().OrderBy(a => a.PropertyImprovementTypeCode).ToArray();
```
asanchezr, line 163: Shouldn't we exclude those types that are disabled here?

frontend/src/components/common/LabelledIcon.tsx
```diff
@@ -32,5 +32,5 @@ const StyledLabelledIcon = styled.div`
 
 const StyledLabel = styled.label`
   font-size: 1.6rem;
-  $font-weight: 700;
+  font-family: 'BCSans=Bold';
```
asanchezr, line 35: I always wondered whether we had this font available in the assets folder but cannot find it anywhere. Where are we getting it from ?

asanchezr: NVM - I found it in package json

backend/dal/Services/LookupService.cs
devinleighsmith: property improvement table doesn't have a disabled column

frontend/src/components/common/LabelledIcon.tsx
devinleighsmith: there is a typo here however, will fix.

backend/dal/Services/LookupService.cs
devinleighsmith: yes, will fix.

asanchezr: this is for the improvement_type, right? the IS17 schema shows it has disabled...

image

asanchezr: And the entity has it (at least my copy)
image

devinleighsmith: actually, I take this back. The front end handles filtering disabled lookups via the lookup code helper, and I think there is value for the api to return disabled codes otherwise the FE has no way of interacting with them at all.

asanchezr: Fine but we will need to decide on a common strategy for all lookup codes. We were discussing to do things server side on my PR, now we move it to the frontend. Seems fine but need to settle on one way

asanchezr: Conceptually not sure why we would need the frontend to know about "disabled" lookup codes. To me disabled kinda equals "deleted from the app" but maybe I'm thinking about it wrong

frontend/src/features/leases/detail/LeasePages/LeasePageForm.tsx
```diff
@@ -25,39 +27,44 @@ export const LeasePageForm: React.FunctionComponent<ILeasePageFormProps> = ({
...
-            if (!leasePage.subRoute) {
+      {!edit ? (
+        <Formik<IFormLease>
+          initialValues={
+            leasePage.title !== leasePages.get(LeasePageNames.DETAILS)?.title
+              ? ({
+                  ...defaultFormLease,
+                  ...apiLeaseToFormLease(lease),
+                } as IFormLease)
+              : { ...defaultFormLease, ...apiLeaseToFormLease(lease) }
```
asanchezr, line 47: hmm, could you please comment on this logic a bit? It seems both code paths do the same here...

devinleighsmith: Oh yes, this was a side effect of deciding to simplify the lease page code by putting the lease in a context. I should be able to just have
{ ...defaultFormLease, ...apiLeaseToFormLease(lease) } 
now

backend/dal/Services/LookupService.cs
devinleighsmith: We already support the concept of disabled contacts, so for me it isn't a stretch that the same logic could apply to other values.

asanchezr (approved):

PR #703, opened 2022-01-08
PSP-2124 Create individual contact

asanchezr:
image


backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        #endregion
+
+        #region Endpoints
+
+        /// <summary>
+        /// Retrieves autocomplete predictions based on the supplied autocomplete request.
+        /// </summary>
+        /// <returns>An array of contacts matching the filter</returns>
+        [HttpGet("organizations")]
```
devinleighsmith, line 53: Hmm, maybe I need some context on this, but we already return the full list of organizations via the lookup controller. Wouldn't it have been significantly easier to just pull the full list of organizations from the context (or use the existing controller if you need an up to date list). and to the autocompletion logic locally?

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
```diff
@@ -0,0 +1,88 @@
...
+        [HasPermission(Permissions.ContactView)]
```
devinleighsmith, line 54: Do we want to limit this api to only be used with contacts?

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
asanchezr: The context is that the full list might cause the problems - we could have 1000s of organizations

asanchezr: My thought process for an "Autocomplete" service/controller is that we might want to have a central place to call these "auto-complete" functionalities. As the app grows, it wouldn't be feasible to get "all" contacts or all orgs or all leases. But we could have an auto-complete of those if needed

asanchezr: I was going to propose that we remove the list of organizations from the lookup controller in a future PR but happy to hear otherwise if we are guaranteed to only ever have a handful of orgs

backend/api/Models/Lookup/LookupModel.cs
```diff
@@ -11,6 +11,12 @@ public class LookupModel : BaseModel
         /// </summary>
         public string Id { get; set; }
 
+        /// <summary>
+        /// get/set - Optional parent of this lookup.
+        /// Allows to model parent/child relationships in lookup codes (ex Country -> Province)
+        /// </summary>
+        public string ParentId { get; set; }
```
devinleighsmith, line 18: yeah, this makes sense.

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
asanchezr: Don't have a strong opinion here - we could open it up. But given that is used in a single place (for now) I would leave it like that until the need arises

frontend/src/App.scss
```diff
@@ -126,27 +126,6 @@
     padding-right: 1.2rem;
   }
 
-  //by default highlight required fields in blue.
```
devinleighsmith, line 129: technically this is used by the manage users page, but on that page required field labels are also highlighted in red so I'm fine with this change.

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
asanchezr: If I remember correctly, the inclusion of orgs in the lookup controller comes from old PIMS where the list was set in advance and small. With contact manager letting users create organizations at will I don't think it would be advisable to keep that there. Thoughts?

frontend/src/components/common/Stack/Stack.tsx
```diff
@@ -0,0 +1,29 @@
+import styled from 'styled-components';
+
+interface IStackProps {
+  /** Defines the space between immediate children. */
+  gap?: string | number;
+  /** Defines the flex-direction style property. It is applied for all screen sizes. */
+  $direction?: 'row' | 'column';
```
devinleighsmith, line 7: $?

asanchezr: In styled-component speak if you prefix a property with "$" it will make sure to **not** passing it to the underlying DOM element. This is called a "transient prop". I had to use it here because **direction** is a valid DOM prop

https://styled-components.com/docs/api#transient-props

asanchezr: OR we could choose a different name for that prop that doesn't clash with an existing DOM prop

frontend/src/features/contacts/contact/create/Container.tsx
```diff
@@ -1,51 +1,51 @@
...
+import ContactTypeSelector from '../typeSelector/ContactTypeSelector';
+import CreatePersonForm from './Person/CreatePersonForm';
 
-interface MatchParams {
-  id?: string;
-}
+export interface IContactCreateContainerProps {}
 
-export interface IContactCreateContainerProps extends RouteComponentProps<MatchParams> {
-  type: ContactTypes;
```
devinleighsmith, line 14: is this type prop used?

frontend/src/App.scss
asanchezr: I see . That's why I was reluctant to remove global form styling. We need to file a tech-debt ticket to harmonize styling for all forms

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
asanchezr: My idea here was that Organizations ARE contacts so you should at least have read permission to be able to see an auto-complete of organizations or else we could be "leaking" information to a user without contact read access. But happy to hear other thoughts

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+  };
+
+  const onSubmit = async (
+    values: ICreatePersonForm,
+    { resetForm, setSubmitting }: FormikHelpers<ICreatePersonForm>,
+  ) => {
+    try {
+      await postPerson(personCreateFormToApiPerson(values));
+      setSubmitting(false);
+      resetForm({ values });
```
devinleighsmith, line 71: does this make sense? the rowversion will be off on the form if you do this. Also if the api enforces any logic on the request the FE won't see that.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+      toast.info('Contact added successfully');
+      history.push('/contact/list');
+    } catch (error) {
+      toast.error('Failed to add contact', { autoClose: 7000 });
+      history.push('/contact/list');
```
devinleighsmith, line 76: Is this correct? Normally I'd expect a form that I'm working on to stay in context if a save error occurs. That way I have a chance to try to resubmit/correct an error without losing all of my work.

Of course, do whatever the requirements say, but I personally don't like this behaviour.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: this is the create form not the update form so rowversion would always be undefined no? In essence I had to reset the form because the next line is a route change and if I didn't reset the form I would get the router guard saying that I had unsaved changes. Would like to hear how you solved this as I used the same router guard code you are using

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: I can remove the navigation. Requirements didn't specify what to do on error. What did you do for leases?

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+      <UnsavedChangesPrompt />
+
+      {/* Confirmation popup when Cancel button is clicked */}
+      <GenericModal
+        title="Unsaved Changes"
+        message="Confirm cancel adding this contact? Changes will not be saved."
+        cancelButtonText="No"
+        okButtonText="Confirm"
+        display={showConfirmation}
+        handleOk={() => {
```
devinleighsmith, line 150: strange, they want the route guard and a navigate away warning?

asanchezr: As per business requirements:

```
14. Cancel button* - on click:
If there is data in any fields, show modal popup: "Confirm cancel adding this contact? Changes will not be saved. [No] [Confirm]"
*No - remove modal, allow user to continue editing
*Confirm - do not save the record, and return the user to the Contact List screen.

If there is no data in any fields, do not save the record, and return the user to the Contact List screen.

16. If the form is partially filled, and the user clicks the other button, it would be advised to show the modal related to navigate away from the page "If the screen is closed before saving these changes, they will be lost. Do you want to continue without saving?"
```

asanchezr: In interpret the above as TWO separate popups. What do you think?

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+                    options={matchedOrgs}
+                    onSearch={handleTypeaheadSearch}
+                  />
+                </Col>
+              </Row>
+            </FormSection>
+
+            <FormSection>
+              <Styled.H2>Contact info</Styled.H2>
+              <Styled.SummaryText
```
devinleighsmith, line 200: seeing this I'm not in love with the idea of creating styled component wrappers like this, it seems to result in a weird mix of embedding pseudo-css in props that I don't like.

I think I'd rather we try to come up with some standardized form layouts (with Bobbi), and then codify those as css.

Of course those standardized layouts would be css so we could steal tweak them as needed, but that would give us a base at least.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: As I see a specific rules for the Cancel button. Requirement #16 speaks about when you click any other link in the app that causes it to navigate away (at least that is how I saw it)

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: agreed

asanchezr: We need base components to model basic form UI. We lack those so I attempted to create a few. This case is a component that either renders text in dark gray or red. I felt that creating two components for the color change was a waste but agree we need a set of form building blocks

frontend/src/features/contacts/contact/create/address/Address.tsx
```diff
@@ -0,0 +1,154 @@
...
+    (e: React.ChangeEvent<HTMLSelectElement>) => {
+      setSelectedCountryId(e.target.value);
+      setFieldValue(withNameSpace(namespace, 'provinceId'), '');
+      setFieldValue(withNameSpace(namespace, 'countryOther'), '');
+    },
+    [namespace, setFieldValue, setSelectedCountryId],
+  );
+
+  // this counter determines how many address lines we render in the form; e.g. street1, street2, etc
+  const { count, increment, decrement } = useCounter({ initial: 1, min: 1, max: 3 });
```
devinleighsmith, line 49: interesting, I take it that the need to limit the total number of elements is why this hook was created instead of using formik? something like:
values.addresses.length

it does seem strange to use external state to track the length of an array in a form, why not use the length of that array directly?

frontend/src/features/contacts/contact/create/address/useAddressHelpers.ts
```diff
@@ -0,0 +1,58 @@
...
+    { province: 'Province', provincePlaceholder: 'Select province', postal: 'Postal Code' },
+  ],
+  [CountryCodes.US, { province: 'State', provincePlaceholder: 'Select state', postal: 'Zip Code' }],
+  [CountryCodes.Other, { province: '', provincePlaceholder: '', postal: 'Postal Code' }],
+]);
+
+/**
+ * Hook that provides several helpers to the Address component.
+ */
+export default function useAddressHelpers() {
```
devinleighsmith, line 20: this makes sense, might as well group common logic like this to clean up the components that use this.

frontend/src/features/contacts/contact/create/address/Address.tsx
asanchezr: Nope, this was done to limit the street addresses fields to show. they are not an array, they have different names in the model and in the form:

- streetAddress1
- streetAddress2
- streetAddress3

The counter controls how many we show via the "add another" link

frontend/src/features/contacts/contact/create/contactInfo/ContactPhoneList.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { ContactPhone } from './ContactPhone';
+
+export interface IContactPhoneList {
+  field: string;
+  contactPhones: ICreateContactMethodForm[];
+}
+
+/**
+ * Formik Field array wrapper around email contacts.
```
devinleighsmith, line 14: nit: phone contacts

frontend/src/features/contacts/contact/create/address/Address.tsx
asanchezr: Tried to add a comment there but maybe is not clear

frontend/src/features/contacts/contact/create/contactInfo/ContactPhoneList.tsx
asanchezr: yeah - will fix. copy/paste

frontend/src/features/contacts/contact/create/address/useAddressHelpers.ts
asanchezr: yeah - I've been using "local hooks" to try to keep the component as lean as possible while moving logic off to the hook. I know this one might not be reused but keeps the component clean

frontend/src/features/contacts/hooks/useContactMethodHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+} from 'constants/contactMethodType';
+import useLookupCodeHelpers from 'hooks/useLookupCodeHelpers';
+import { useMemo } from 'react';
+import { ILookupCode } from 'store/slices/lookupCodes';
+
+/**
+ * Extending default ILookupCode interface to enforce the presence of `displayOrder` property
+ * This is okay here because both code tables have this field (not the case with other code tables)
+ */
+interface IHasDisplayOrder extends Omit<ILookupCode, 'displayOrder'> {
```
devinleighsmith, line 15: Is there a reason not to do this on the server side?

asanchezr: I guess I could have modified the server to always have displayOrder but I believe not all lookup codes have order so I went for the least disruptive way of modelling it

frontend/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "jsx": "react-jsx",
     "downlevelIteration": true,
     "noFallthroughCasesInSwitch": true,
-    "typeRoots": ["src/types", "node_modules/@types"]
+    "typeRoots": ["src/typings", "node_modules/@types"]
```
devinleighsmith, line 20: Hmm, what is this?

asanchezr: Prior to this PR we had TWO folder with custom typings: types and typings. Moved everything to the typings folder and fixed the json config

asanchezr: Or we didn't have the src/types folder - can't remember which

frontend/src/features/contacts/hooks/useContactMethodHelpers.ts
devinleighsmith: Sorry, I meant why not have the server return the list sorted by display order, then you would not need to return the display order at all.

frontend/tsconfig.json
asanchezr: We had both as per this screenshot

image

frontend/src/features/contacts/contact/create/address/Address.tsx
devinleighsmith: oh, ok. Wouldn't have thought of doing it that way, but makes sense now.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
devinleighsmith: I think this is correct given the above, but doesn't match the requirements I've been given on other forms. sigh.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
devinleighsmith: Just display the error, no navigation. Allows them to attempt to submit again.

frontend/src/features/contacts/hooks/useContactMethodHelpers.ts
asanchezr: I see. I didn't see any existing example where we do that. Also, another problem I noticed is that most lookup codes we do not check for **isDisabled** so there's potential to return old/stale lookup codes. I added the check on the ones I created.

I can sort by display order in the server in my next PR for the lookup codes I need. The others would need a separate ticket, no?

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
devinleighsmith: I have the prompt in context of formik, and so if the form is submitted successfully I have a flag that prevents the route guard from triggering.

frontend/src/components/common/Stack/Stack.tsx
devinleighsmith: Thanks, I understand now.

frontend/src/features/contacts/hooks/useContactMethodHelpers.ts
asanchezr: Checked and I already sort by display order in the server (for the ones I use in my PR)

```cs

  public IEnumerable<PimsCountry> GetCountries()
        {
            return this.Context.PimsCountries.AsNoTracking().OrderBy(a => a.DisplayOrder).ThenBy(a => a.CountryCode).ToArray();
        }

        /// <summary>
        /// Get all provinces sorted by DisplayOrder and Code
        /// </summary>
        public IEnumerable<PimsProvinceState> GetProvinces()
        {
            return this.Context.PimsProvinceStates.AsNoTracking().Where(c => c.IsDisabled != true).OrderBy(a => a.DisplayOrder).ThenBy(a => a.ProvinceStateCode).ToArray();
```

asanchezr: NOTE - PimsCountry doesn't have **isDisabled** (which it should I think) so I couldn't include that in the country query

frontend/src/features/contacts/contact/create/address/Address.tsx
asanchezr: well - Initially I though of having 3 vars controlled with `useState` but the counter seems more intuitive to others - maybe not I guess

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
devinleighsmith: I'm honestly not sure if this is premature optimization or wise future proofing at this point. Let's leave this in for now and talk to the business about this.

devinleighsmith (approved):

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: We will need to take it back to the business to define a common approach to Cancel & navigate away for all forms

frontend/src/features/contacts/hooks/useContactMethodHelpers.ts
devinleighsmith: yeah, don't need to fix it now, just wanted to make sure you are aware so you can save yourself some time/effort/loc on future stories.

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: okay, I can change in the next PR to check for that. What flag you use?

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: okay - will remove

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
asanchezr: Found the flag and made the change: `submitCount === 0`

backend/api/Areas/Autocomplete/Controllers/AutocompleteController.cs
devinleighsmith: Given that this is only used here this seems fine, if we get another consumer we can update this.

PR #715, opened 2022-01-11
PSP 2527 - Update oc builds to use unique tags

rafaelponcedeleon:
Requires that the following build configs are updated to respective tags:  latest-dev, latest-test, latest-uat
- pims-api.dev
- pims-api.test
- pims-api.master
- pims-app.dev
- pims-app.test
- pims-app.master


devinleighsmith:
Hey @rafaelponcedeleon. Shouldn't there be an update to the oc build configs so that they stop building to :latest as well? right now it looks like the pipeline is expecting the build configs to create a latest-env tag, but I don't see that tag being created anywhere in this PR.

openshift/4.0/pipelines/Jenkinsfile.cicd
```diff
@@ -219,14 +219,15 @@ pipeline {
...
           def img_frontend = "${APP_NAME}-app"
+          def img_tag = "latest-${DESTINATION}" // image tag on build config is based on environment, e.g. latest-dev, latest-test, latest-uat
 
           // tag release version (if found)
           if (RELEASE_VERSION != '') {
-            echo "Tagging images - ${img_backend}:latest, ${img_frontend}:latest..."
-            sh "oc tag ${img_backend}:latest ${img_backend}:${RELEASE_VERSION}"
-            sh "oc tag ${img_frontend}:latest ${img_frontend}:${RELEASE_VERSION}"
+            echo "Tagging images - ${img_backend}:latest-env, ${img_frontend}:latest-env"
+            sh "oc tag ${img_backend}:${img_tag} ${img_backend}:${RELEASE_VERSION}"
```
devinleighsmith, line 227: This retags :latest-${DESTINATION} to :${RELEASE_VERSION}
but I'm not seeing what is creating the :latest-${DESTINATION} 
tag in the first place.

database/mssql/scripts/dbscripts/db-upgrade.sh
```diff
@@ -11,6 +11,12 @@ eval $(grep -v '^#' ../../.env | xargs)
 
 echo "===== Begin DB Schema Upgrade ====="
 
+# 1. Find current DB version 
+#    select STATIC_VARIABLE_VALUE from PIMS_STATIC_VARIABLE where STATIC_VARIABLE_NAME = 'dbversion'
+# 2. Search for all SQL scripts under the 'Alter Up' folder and order them by folder name and then file name
+#    find ./PSP*/'Alter Up' -type f -iname "*.sql" | sort -n
+# 3. Execute all SQL scripts that have a higher version than the current DB version
+
```
asanchezr, line 19: This section is already present in lines 3-10 above - any reason to duplicate here?

openshift/4.0/pipelines/Jenkinsfile.cicd
rafaelponcedeleon: The tag **latest-${DESTINATION}** is created by the build config.
The tag is defined in the build.yaml template with the parameter OUTPUT_IMAGE_TAG. Are the parameter values stored somewhere? If not then there's no code to update. 

For reference, the api template is here: openshift\4.0\templates\api\build.yaml

asanchezr: I think the template provides default values. My guess is that Devin is requesting the default be changed in the **build.yaml** template

asanchezr: Also please make sure to modify the existing build configs right after this gets merged or else the pipeline will fail

asanchezr (approved):

PR #717, opened 2022-01-11
psp-2660 add simple retry to lazy load promise.

devinleighsmith:
(I will merge this after 2658)

### DIFF
--- a/backend/api/Pims.Api.csproj
+++ b/backend/api/Pims.Api.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <UserSecretsId>0ef6255f-9ea0-49ec-8c65-c172304b4926</UserSecretsId>
-    <Version>1.1.0-16.32</Version>
+    <Version>1.1.0-16.33</Version>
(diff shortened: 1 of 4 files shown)


devinleighsmith:
In our various environments, network performance blips on OCP would occasionally cause the dynamic import within the AppRouter to fail with a "chunk error". Given that these errors are fleeting, adding a simple retry should fix the majority of instances.

backend/api/Pims.Api.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <UserSecretsId>0ef6255f-9ea0-49ec-8c65-c172304b4926</UserSecretsId>
-    <Version>1.1.0-16.32</Version>
+    <Version>1.1.0-16.33</Version>
```
asanchezr, line 5: not a big deal - but shouldn't we be incrementing the hotfix version (not the build version) for every hotfix applied to master?

asanchezr (approved):

backend/api/Pims.Api.csproj
devinleighsmith: Right, will do on future PRs.

PR #724, opened 2022-01-12
Added logic to show modal and check for duplicates | PSP-2517

FuriousLlama:


jdtoombs (approved):

backend/dal/Services/PersonService.cs
```diff
@@ -66,17 +67,37 @@ public PimsPerson Get(long id)
...
+        public PimsPerson Add(PimsPerson person, bool userOverride)
         {
-            add.ThrowIfNull(nameof(add));
-            this.Context.PimsPeople.Add(add);
+            person.ThrowIfNull(nameof(person));
+            if (!userOverride)
+            {
+
+                List<PimsPerson> existingPersons = this.Context.PimsPeople.Where(
+                    p => EF.Functions.Collate(p.FirstName, "SQL_Latin1_General_CP1_CI_AS") == person.FirstName &&
```
asanchezr, line 80: suggestion: could we have the collection ID put in a constant/enum so we don't have to copy/paste every time (with potential of typos etc_?

backend/dal/Services/PersonService.cs
```diff
@@ -66,17 +67,37 @@ public PimsPerson Get(long id)
...
+                        EF.Functions.Collate(p.Surname, "SQL_Latin1_General_CP1_CI_AS") == person.Surname
+                ).Include(p => p.PimsContactMethods).ToList();
+
+                var isDuplicate = existingPersons.Any(
+                                p => p.PimsContactMethods.Any(
+                                    ec => person.PimsContactMethods.Any(
+                                        pc => pc.ContactMethodValue.ToLower() == ec.ContactMethodValue &&
```
asanchezr, line 87: shouldn't we **toLower()** both sides of the comparison here?

```cs
pc.ContactMethodValue.ToLower() == ec.ContactMethodValue.toLower() // <-- new
```

frontend/package-lock.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "frontend",
-  "version": "1.0.0-16.19",
+  "version": "1.0.0-18.1",
```
asanchezr, line 3: Please rebase your branch. This would cause conflicts if merged as-is. DEV is currently at IS19

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -62,27 +67,51 @@ export const CreatePersonForm: React.FunctionComponent<ICreatePersonFormProps> =
...
-      setSubmitting(false);
-      toast.info('Contact added successfully');
-      history.push('/contact/list');
-    } catch (error) {
-      toast.error('Failed to add contact', { autoClose: 7000 });
+      setDuplicateModal(false);
+      let newPerson = personCreateFormToApiPerson(formPerson);
+      const personResponse = await addPerson(newPerson, setDuplicateModal, allowDuplicate);
+
+      console.log(personResponse);
```
asanchezr, line 78: please remove console.log here

frontend/package-lock.json
FuriousLlama: I just checked dev and it has "1.0.0-16.19" as the version, but it should be 18.1 according to the packge.json file. Note that it is generated file from the package.json one

backend/dal/Services/PersonService.cs
FuriousLlama: sure can!

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -62,27 +67,51 @@ export const CreatePersonForm: React.FunctionComponent<ICreatePersonFormProps> =
...
+
+      if (!!personResponse?.id) {
+        history.push('/contact/list');
+      }
+    } finally {
+      formikRef?.current?.setSubmitting(false);
```
asanchezr, line 84: you don't really need to use formikRef here (but is okay). The **onSubmit** functions receives all formik helpers as parameters - see function signature above

frontend/src/features/contacts/hooks/useAddContact.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const addPerson = async (
+    person: ICreatePerson,
+    needsUserAction: (needsAction: boolean) => void,
+    userOverride: boolean = false,
+  ) => {
+    try {
+      dispatch(showLoading());
+      const response = await postPerson(person, userOverride);
+      toast.success('Lease/License saved');
```
asanchezr, line 25: This should probably be more like: "Contact/Person saved" ?

FuriousLlama: Good catch, will update it

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -62,27 +67,51 @@ export const CreatePersonForm: React.FunctionComponent<ICreatePersonFormProps> =
...
+        initialValues={defaultCreatePerson}
+        enableReinitialize
+        validate={onValidate}
+        onSubmit={onSubmit}
+        innerRef={formikRef}
+      />
+      <GenericModal
+        title="Duplicate Contact"
+        display={showDuplicateModal}
+        message="A contact matching this information already existts in the system."
```
asanchezr, line 108: nit: fix typo - `already exist[t]s`

asanchezr (approved):

frontend/package-lock.json
asanchezr: DEV is showing 19.3 for me. Maybe you have an old version cached in your browser?
image

asanchezr: Also **package.json** is showing the correct value in git
image

FuriousLlama: I think there is confusion, the comment is about package-lock.json. not package.json

asanchezr: Oh I see. Sorry, missed that. Disregard my previous comments. It seems the lock file gets updated only when someone does `npm install` locally. 

You are right - we don't care about the value in the lock file

PR #727, opened 2022-01-13
Updated phone list | PSP-2649

FuriousLlama:


frontend/src/features/contacts/contact/detail/Person.tsx
```diff
@@ -38,7 +38,7 @@ const PersonView: React.FunctionComponent<PersonViewProps> = ({ person }) => {
   const phoneTypes: Dictionary<string> = {};
   phoneTypes[ContactMethodTypes.WorkMobile] = 'Mobile';
   phoneTypes[ContactMethodTypes.WorkPhone] = 'Work';
-  phoneTypes[ContactMethodTypes.PersonalPhone] = 'Home';
+  phoneTypes[ContactMethodTypes.PersonalMobile] = 'Home';
```
asanchezr, line 41: I believe this would still result in bugs. The database supports **TWO** kinds of personal phones:

* Personal Phone
* Personal Mobile

 We would need to support both

image

asanchezr: The current enums on the frontend support all available phone types. I think the read form should display a type for each possible phone type (even fax)

FuriousLlama: the requirements specify the phone types to be displayed:
https://jira.th.gov.bc.ca/browse/PSP-2649?focusedCommentId=418270&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-418270

asanchezr: I see in the code we are already using the label "Mobile" for work mobile phones so we would have to come up with a new string for personal mobile phones. Maybe this is something to ask the business?

FuriousLlama: yep, please the link above.

asanchezr: hmm I see - well it seems weird to me to be able to create a phone with type X and then not show in the details. I'd reconfirm with them. Showing N/A doesn't seem great in my opinion but I agree you are coding as per their requirements

asanchezr (approved):

frontend/src/features/contacts/contact/detail/Person.tsx
devinleighsmith: Ie, that the create screen allows the user to add an option that will not display on the view screen based on his response to Manuel's question.

FuriousLlama: Yeah, that is a section that has been discussed multiple times since the confluence mapping was a tad confussing. That last link is the most straightforward definition of what the business needs.

PR #730, opened 2022-01-13
scaffold IS19. remove fields from FE/BE models/mappings.

devinleighsmith:
Note that this will temporarily leave the deposits and lease update forms in a "broken" state.
However, there is a story this sprint to fix the update lease form and a story to rewrite the deposits form next sprint.


devinleighsmith:
Also note that the ALTER DOWN script is broken currently. Doug is working on fixing it. In the interest of unblocking dev resources I merged Doug's PR, with the understanding that he would fix it before release.

FuriousLlama (approved):

PR #735, opened 2022-01-13
System Constants | psp-2178

FuriousLlama:


devinleighsmith:
thanks for this, looks good.

backend/api/Controllers/SystemConstantController.cs
```diff
@@ -0,0 +1,52 @@
...
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Pims.Api.Controllers
+{
+    /// <summary>
+    /// SystemConstantController class, provides endpoints to retrieve system information api.
+    /// </summary>
+    [ApiController]
+    [ApiVersion("1.0")]
+    [Route("v{version:apiVersion}/systemConstant")]
```
asanchezr, line 14: Not a big deal but ideally would prefer to stay away from mixed case URLs as they can be sometimes problematic for certain web servers. would you consider something like `/system-constants` instead?

asanchezr: Also URL should be plural

asanchezr (approved):

PR #738, opened 2022-01-14
rebuild image instead of container to ensure dbscripts folder up to d…

devinleighsmith:
…ate.


asanchezr (approved):
Thanks for fixing this

PR #740, opened 2022-01-14
rename services to repositories, add service scaffold for future serv…

devinleighsmith:
…ices.


backend/dal/PimsRepository.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Security.Claims;
+using Microsoft.Extensions.DependencyInjection;
+using Pims.Dal.Repositories;
+
+namespace Pims.Dal
+{
+    /// <summary>
+    /// PimsRepository class, provides a encapsulated way to references all the independent repositories.
+    /// </summary>
+    public class PimsRepository : IPimsRepository
```
FuriousLlama, line 11: Since we use dependency injection I am not sure this class provides much value other than wrapping up the service retrieval.

FuriousLlama:
Looking good! thanks for starting the change.

backend/dal/PimsRepository.cs
devinleighsmith: This is just a rename of the PimsService. If we want to refactor existing code, we can do that in another PR.

FuriousLlama (approved):

PR #747, opened 2022-01-15
Psp 2505

devinleighsmith:
image


devinleighsmith:
location of last payment received is wrong, will correct in psp-2020

backend/entities/LeasePayment.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsLeasePayment class, provides an entity for the datamodel to manage lease payments.
+    /// </summary>
+    public partial class PimsLeasePayment : IdentityBaseAppEntity<long>, IBaseAppEntity
+    {
+        #region Properties
+        [NotMapped]
+        public override long Id { get => this.LeasePaymentId; set => this.LeasePaymentId = value; }
```
FuriousLlama, line 11: this looks a tad hackish

frontend/src/features/leases/detail/LeasePages/payment/table/PaymentsForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const allPayments = orderBy(
+    (formikProps.values.terms ?? []).flatMap(p => p.payments),
+    'receivedDate',
+    'desc',
+  );
+  const lastPaymentDate = allPayments.length ? prettyFormatDate(allPayments[0]?.receivedDate) : '';
+
+  return (
+    <>
+      <>
```
FuriousLlama, line 36: Are two of these necessary?

frontend/src/features/leases/detail/LeasePages/payment/table/columns.tsx
```diff
@@ -0,0 +1,277 @@
...
+      original.startDate,
+      original.expiryDate,
+      original.paymentAmount,
+      gstDecimal ?? 0,
+    );
+    return expectedTerm !== undefined ? formatMoney(expectedTerm) : '-';
+  };
+
+function calculateExpectedGst(amount: number, gst: number) {
+  return amount * (gst / 100);
```
FuriousLlama, line 68: nit: gst/100 will always be the same value since gst is a constant.

frontend/src/store/slices/systemConstants/interfaces/ISystemConstant.ts
```diff
@@ -1,4 +1,4 @@
 export interface ISystemConstant {
-  name: number;
```
FuriousLlama, line 2: good catch

frontend/src/store/slices/systemConstants/useSystemConstants.ts
```diff
@@ -2,15 +2,28 @@ import * as actionTypes from 'constants/actionTypes';
...
+import { useAppDispatch, useAppSelector } from 'store/hooks';
 
 import { IGenericNetworkAction } from '../network/interfaces';
 import { logError, logRequest, logSuccess } from '../network/networkSlice';
 import { ISystemConstant, storeSystemConstants } from '.';
 
+export enum SystemConstants {
+  FYEND = 'FYEND',
+  FYSTART = 'FYSTART',
+  GST = 'GST',
```
FuriousLlama, line 14: Glad you added those ones. I think moving the enum to interfaces/ISystemConstant.ts  makes more sense, but its a nit.

frontend/src/features/leases/detail/LeasePages/payment/table/columns.tsx
```diff
@@ -0,0 +1,277 @@
...
+import { IFormLeaseTerm, ILeaseTerm } from 'interfaces/ILeaseTerm';
+import moment from 'moment';
+import { FaTrash } from 'react-icons/fa';
+import { MdEdit } from 'react-icons/md';
+import { CellProps } from 'react-table';
+import { ISystemConstant } from 'store/slices/systemConstants';
+import styled from 'styled-components';
+import { formatMoney, prettyFormatDate } from 'utils';
+
+function initialOrRenewalTerm({ row: { original, index } }: CellProps<ILeaseTerm, string>) {
```
FuriousLlama, line 20: The functions below seem to be doing the lift for the business calculations. It would be better if we have a class/object that would contain the transformations and methods for said transformation while the table columns would only care about how those values are rendered.

Something like: 
ILeaseTerm -> CalculatedLeaseTerm -> PaymentTableRow

FuriousLlama (approved):

backend/entities/LeasePayment.cs
devinleighsmith: alternatives?

frontend/src/features/leases/detail/LeasePages/payment/table/columns.tsx
devinleighsmith: Not sure what you mean, gst can't be hardcoded since it comes from the DB. I could memo this but I don't think the overhead is important.

frontend/src/features/leases/detail/LeasePages/payment/table/columns.tsx
devinleighsmith: I agree, would make sense to just update the IFormLeaseTerm to include the calculated rows and do the calculation in the initial conversion. Will try and get this in today, but likely will do it in psp-2020.

frontend/src/features/leases/detail/LeasePages/payment/table/PaymentsForm.tsx
devinleighsmith: will fix this in psp-2020

frontend/src/features/leases/detail/LeasePages/payment/table/columns.tsx
FuriousLlama: not hardcoded, but once calculated it won't change.

FuriousLlama: This would be more obvious if there is an object that contains the methods and members for the calculated information

backend/entities/LeasePayment.cs
FuriousLlama: Maybe I am missing something, is this only used by the UpdateChild method?

PR #755, opened 2022-01-18
Lease Region with code  - psp-2605

FuriousLlama:


backend/api/Areas/Leases/Mapping/Lease/RegionMap.cs
```diff
@@ -0,0 +1,16 @@
+using Mapster;
```
devinleighsmith, line 1: Hmm, does the existing type code mapping not work for regions? It looks like it should, in which case you can add/update the region partial class to extend : ITypeEntity<string>

frontend/.vscode/launch.json
```diff
@@ -26,6 +26,24 @@
         "CI": "true"
       },
       "disableOptimisticBPs": true
+    },
```
devinleighsmith, line 29: FYI I'm using jest-runner and Alejandro is using vscode-jest to assist in running tests. If you are interested one of us can give you a demo.

devinleighsmith (approved):

backend/api/Areas/Leases/Mapping/Lease/RegionMap.cs
FuriousLlama: Sadly not since PimsRegion attribute names are not standard like a typecode

frontend/.vscode/launch.json
FuriousLlama: that would a cool friday chat!

PR #760, opened 2022-01-18
Psp 2020

devinleighsmith:
CUD lease term functionality


backend/dal/Services/LeaseService.cs
```diff
@@ -1,294 +1,20 @@
...
-            bool newLeaseProperties = pimsPropertyLeases.Any(p => !existingLease.PimsPropertyLeases.Any(xp => xp.PropertyId == p.PropertyId));
-            existingLease.PimsPropertyLeases = pimsPropertyLeases;
-            var leaseWithAssociatedProperties = AssociatePropertyLeases(existingLease, userOverride, newLeaseProperties);
-
-            this.Context.UpdateChild<PimsLease, long, PimsPropertyLease>(l => l.PimsPropertyLeases, leaseId, leaseWithAssociatedProperties.PimsPropertyLeases.ToArray());
-            this.Context.CommitTransaction();
-
-            return Get(existingLease.LeaseId);
+            PimsLease lease = _pimsRepository.Lease.GetFlat(leaseId, true);
+            if(lease.ConcurrencyControlNumber != rowVersion) throw new DbUpdateConcurrencyException("Unable to save. Please refresh your page and try again");
```
FuriousLlama, line 17: It would result in a smaller query if instead of retrieving the entire lease, it would only give you the concurrency control number.

backend/dal/Services/LeaseService.cs
```diff
@@ -1,294 +1,20 @@
...
         }
-
-        /// <summary>
-        /// update the properties on the lease
-        /// </summary>
-        /// <param name="leaseId"></param>
-        /// <param name="pimsPropertyLeases"></param>
-        /// <returns></returns>
-        public PimsLease UpdatePropertyLeases(long leaseId, long rowVersion, ICollection<PimsPropertyLease> pimsPropertyLeases, bool userOverride = false)
+        public void CheckRowVersion(long leaseId, long rowVersion)
```
FuriousLlama, line 14: instead of throwing, I think this version should return a boolean. What to do about the values not matching sounds like a higher-level decision than this method

backend/tests/core/ServiceHelper.cs
```diff
@@ -115,7 +115,7 @@ public static class ServiceHelper
         /// <param name="helper"></param>
         /// <param name="args"></param>
         /// <returns></returns>
-        public static T CreateService<T>(this TestHelper helper, params object[] args) where T : IService
+        public static T CreateService<T>(this TestHelper helper, params object[] args) where T : IRepository
```
FuriousLlama, line 118: are these renames part of a different pr too?

FuriousLlama: Ideally, we would have the business/features changes on a different pr than the tech enhancements to keep changes logical and easier to comb though

FuriousLlama:
Are the changes here related to psp-2020? looking at both Prs and seeing similar changes.

backend/tests/core/ServiceHelper.cs
devinleighsmith: That is the case, When that PR is merged these renames will magically disappear, just couldn't wait for Alejandro's PR any longer.

devinleighsmith: Pr != commit

FuriousLlama: Ah I see. Well the PR shows the difference between the state of the branch to be merged and the set of commits on it.

backend/dal/Repositories/Interfaces/ILeaseService.cs
```diff
@@ -11,7 +11,7 @@ public interface ILeaseService : IRepository<PimsLease>
     {
         int Count();
         IEnumerable<PimsLease> Get(LeaseFilter filter);
-        PimsLease GetFlat(long id, bool asNoTracking = false);
+        long GetRowVersion(long id);
```
asanchezr, line 14: I see that we now have two folders: `dal/Repositories` and `dal/Services` - wouldn't it make sense to name the repo classes interface something like `WhateverRepo` instead of `WhateverService`. Seems a tad confusing

backend/api/Areas/Admin/Controllers/AccessRequestController.cs
```diff
@@ -33,7 +33,7 @@ public class AccessRequestController : Controller
         /// </summary>
         /// <param name="pimsService"></param>
         /// <param name="mapper"></param>
-        public AccessRequestController(IPimsService pimsService, IMapper mapper)
+        public AccessRequestController(IPimsRepository pimsService, IMapper mapper)
```
asanchezr, line 36: nit: I know this work is still in flux but would prefer if naming was consistent; i.e. `IPimsRepository pimsRepository`

backend/api/Areas/Leases/Controllers/LeaseTermController.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// Update the specified term on the passed lease.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost("{leaseId:long}/term")]
+        [HasPermission(Permissions.LeaseEdit)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(IEnumerable<Models.Lease.LeaseModel>), 200)]
+        [SwaggerOperation(Tags = new[] { "lease" })]
+        public IActionResult AddTerm(long leaseId, Models.Lease.TermModel termModel)
```
asanchezr, line 53: Not sure if this would be a big deal in C# / ASP but shouldn't we annotate the model as coming in the POST body via `[FromBody]` ?

backend/api/Areas/Leases/Controllers/LeaseTermController.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// Update the specified term on the passed lease.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPut("{leaseId:long}/term/{termId:long}")]
+        [HasPermission(Permissions.LeaseEdit)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(IEnumerable<Models.Lease.LeaseModel>), 200)]
+        [SwaggerOperation(Tags = new[] { "lease" })]
+        public IActionResult UpdateTerm(long leaseId, long termId, Models.Lease.TermModel termModel)
```
asanchezr, line 70: Same `[FromBody]` comment as above

backend/api/Areas/Admin/Controllers/AccessRequestController.cs
devinleighsmith: Yes, this was due to my decision to try and fit this into the sprint without it being planned. Ibudgeted some time for these changes and used up all of that budget without 100% completing everything I wanted, like renaming all of the existing services.

I made a call that the partial change was still worth keeping as it enables future services/repositories to be developed in the manner that we want. The downside is leftover naming like this.

devinleighsmith: Also, that change has already been merged, I'm in the process of updating a PR to only reflect the changes from psp-2020. Should be done shortly.

backend/api/Areas/Leases/Controllers/LeaseTermController.cs
devinleighsmith: good point, I'll add that in.

backend/dal/Repositories/Interfaces/ILeaseService.cs
devinleighsmith: agreed, we should rename all of the files in dal/repositories to whateverrepo.

devinleighsmith:
Okay, finally, PR looks as was originally intended now.

backend/dal/Helpers/Extensions/LeaseExtensions.cs
```diff
@@ -170,50 +170,6 @@ public static PimsLeaseTerm GetCurrentTerm(this Pims.Dal.Entities.PimsLease leas
             return GetCurrentTerm(lease)?.TermExpiryDate;
         }
 
-        /// <summary>
```
devinleighsmith, line 173: removed dead code.

backend/dal/Repositories/LeaseService.cs
```diff
@@ -181,7 +189,7 @@ private PimsLease AssociatePropertyLeases(PimsLease lease, bool userOverride = f
                 if (property?.PimsPropertyLeases.Any(p => p.LeaseId != lease.Id) == true && !userOverride && newLeaseProperties)
                 {
                     var genericOverrideErrorMsg = $"is attached to L-File # {property.PimsPropertyLeases.FirstOrDefault().Lease.LFileNo}";
-                    if (propertyLease?.Property?.Pin != null)
+                    if(propertyLease?.Property?.Pin != null)
```
devinleighsmith, line 192: some whitespace issues here that I need to clean up.

frontend/src/App.tsx
```diff
@@ -48,7 +47,6 @@ const App = () => {
         return (
           <>
             <AppRouter />
-            <OnLoadActions />
```
devinleighsmith, line 51: More dead code removed from PIMS.

frontend/src/components/common/form/FastCurrencyInput.tsx
```diff
@@ -18,7 +18,7 @@ export const defaultMaskOptions = {
   thousandsSeparatorSymbol: ',',
   allowDecimal: true,
   decimalSymbol: '.',
-  decimalLimit: 0, // how many digits allowed after the decimal
+  decimalLimit: 2, // how many digits allowed after the decimal
```
devinleighsmith, line 21: This is a global difference from legacy PIMS to PSP apparently.

frontend/src/features/leases/detail/LeasePages/payment/PaymentsContainer.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+describe('PaymentsContainer component', () => {
+  const setup = async (
+    renderOptions: RenderOptions &
+      Partial<IPaymentsContainerProps> & {
+        initialValues?: any;
+      } = {},
+  ) => {
+    // render component under test
+    const component = await renderAsync(
```
asanchezr, line 33: I saw in another file that we have `render` and `renderAsync` that both seem to do the same. renderAsync was added later. Could you please comment on the need for that? Just learning

asanchezr: Like the original react-test-library render does not return a promise so not sure how/why we would make it async

frontend/src/features/leases/detail/LeasePages/payment/PaymentsContainer.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    expect(component.asFragment()).toMatchSnapshot();
+  });
+
+  it('makes a post request when adding a new term', async () => {
+    const {
+      component: { getAllByText, getByText },
+    } = await setup({});
+    mockAxios.onPost().reply(200, { id: 1 });
+
+    const addButton = getAllByText('Add a Term')[0];
```
asanchezr, line 76: nit: are you finding multiple "Add a Term" buttons here? couldn't we use `getByText('...')` instead?

asanchezr (approved):

frontend/src/features/leases/detail/LeasePages/payment/PaymentsContainer.test.tsx
devinleighsmith: There is an add a term button and a add a term modal unfortunately.

frontend/src/features/leases/detail/LeasePages/payment/PaymentsContainer.test.tsx
devinleighsmith: renderasync automatically handles act related errors assuming no interactions are made to the page. So essentially it is a more concise way of rendering a component that has async logic. Without renderasync a snapshot test would need to have an async await on some part of the page that only renders after all promises have resolved.

PR #762, opened 2022-01-19
PSP-2294 Create Organization Contact

asanchezr:
image


backend/api/Areas/Organizations/Controllers/OrganizationController.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            // Business rule - support country free-form value if country code is "Other". Ignore field otherwise.
+            var otherCountry = _pimsService.Lookup.GetCountries().FirstOrDefault(x => x.Code == Dal.Entities.CountryCodes.Other);
+            foreach (var address in model?.Addresses)
+            {
+                if (otherCountry != null && address != null && address.CountryId != otherCountry.CountryId)
+                {
+                    address.CountryOther = null;
+                }
+            }
```
FuriousLlama, line 64: Is this necessary to make sure that the front end didnt send a badly formed model?

backend/api/Areas/Organizations/Models/Organization/OrganizationModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// get/set - The organization's alias name.
+        /// </summary>
+        public string Alias { get; set; }
+
+        /// <summary>
+        /// get/set - The organization's incorporation number.
+        /// </summary>
+        public string IncorporationNumber { get; set; }
+
+
```
FuriousLlama, line 45: nit: extra space

backend/api/Areas/Organizations/Models/Organization/OrganizationModel.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// get/set - The organization's disabled status flag.
+        /// </summary>
+        public bool IsDisabled { get; set; }
+
+        /// <summary>
+        /// get/set - The organization's name.
+        /// </summary>
+        [Required]
```
FuriousLlama, line 32: Does this change anything in the API model?

backend/dal/Services/UserOrganizationService.cs
```diff
@@ -0,0 +1,255 @@
...
+            this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin);
+
+            var query = this.Context.PimsOrganizations.AsNoTracking();
+
+            if (filter != null)
+            {
+                if (filter.Page < 1) filter.Page = 1;
+                if (filter.Quantity < 1) filter.Quantity = 1;
+                if (filter.Quantity > 50) filter.Quantity = 50;
+                if (filter.Sort == null) filter.Sort = System.Array.Empty<string>();
```
FuriousLlama, line 64: very nitty nit: inconsistent if style compared to the ones below as well as other ones not brackets

frontend/src/features/contacts/contact/create/CreateContactContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+  display: flex;
+  flex-direction: column;
+  text-align: left;
+  height: 100%;
+  width: 50%;
+  min-width: 30rem;
+  overflow: hidden;
+  padding: 1rem;
+  gap: 1.6rem;
+`;
```
FuriousLlama, line 46: Since it is already importing ```import * as Styled from 'features/contacts/styles';``` could this be moved there?

backend/api/Areas/Organizations/Models/Organization/OrganizationModel.cs
asanchezr: not really following, sorry. The only required field on this model is the org name. That is reflected here with a data annotation. The effect is that the API expects the json to have at least that property set.

This is standard MVC/webapi. If the controller receives json that doesn't pass validation it rejects the operation

backend/dal/Services/UserOrganizationService.cs
asanchezr: This is a "legacy" service that I renamed to make way for the "proper" OrganizationService. I would be inclined to leave as-is because not part of this PR (other than the name change)

frontend/src/features/contacts/contact/create/validation.ts
```diff
@@ -98,3 +97,12 @@ export const validationSchema = Yup.object().shape({
     }),
   }),
 });
+
+export const PersonValidationSchema = baseSchema.shape({
```
FuriousLlama, line 101: nit: PersonValidationSchema should be start with a capital letter since it does not define a class/interface

frontend/src/features/contacts/contact/create/CreateContactContainer.tsx
asanchezr: we could but would need a different naming. `contacts/styles` covers all common styling (for contacts list, create forms, update forms in the future etc). "ContactLayout" is too generic

FuriousLlama (approved):

backend/api/Areas/Organizations/Models/Organization/OrganizationModel.cs
FuriousLlama: right, but this is not running the mvc validations. We should remove it since it does not seem to be doing anything for us atm

frontend/src/features/contacts/contact/create/validation.ts
asanchezr: See other validation schemas in the code, they start with uppercase letter. I'm fine renaming but was following pre-existing patterns

backend/api/Areas/Organizations/Models/Organization/OrganizationModel.cs
asanchezr: could you please clarify why is not running validation. In my experience it does run by default (not by our code)

asanchezr: As per documentation [here](https://www.pragimtech.com/blog/blazor/rest-api-model-validation/) and [here](https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-6.0#automatic-http-400-responses) we are using model validation automatically in our project as all controllers are annotated with `[ApiController]` 

I didn't know this - I just noticed that our controllers were validating the input model and took advantage of that

frontend/src/features/contacts/contact/create/validation.ts
FuriousLlama: yeah, still looks like bad styling to me, but if that's the pattern for the code then sure.

PR #801, opened 2022-01-21
db schema fix.

devinleighsmith:
These scripts were added to 17.01 and then added again to IS19. 

17.01 has already been run in test and UAT, so these scripts need to be removed.


jdtoombs (approved):

devinleighsmith:
This will be backported immediately.

asanchezr:
@devinleighsmith where do you want to backport this to? The PR is targeting the **dev** branch. Or was it intended to target **test** instead?

PR #807, opened 2022-01-25
psp-2665 read only view of payment actuals.

devinleighsmith:
image


backend/dal/Repositories/LeaseTermRepository.cs
```diff
@@ -29,6 +29,7 @@ public void Delete(long leaseTermId)
 
         public PimsLeaseTerm Update(PimsLeaseTerm pimsLeaseTerm)
         {
+            pimsLeaseTerm.PimsLeasePayments = null; //Do not edit lease payments.
```
devinleighsmith, line 32: This is a conceptual problem with EF, but by default it wants to manipulate child relationships. Since we are trying to handle that explicitly, we need to come up with a strategy here.

frontend/package.json
```diff
@@ -31,7 +31,7 @@
     "react-draggable": "4.4.3",
     "react-error-boundary": "3.1.1",
     "react-fast-compare": "3.2.0",
-    "react-icons": "4.2.0",
+    "react-icons": "4.3.1",
```
devinleighsmith, line 34: missing some icons that are now available.

frontend/src/components/Table/Table.tsx
```diff
@@ -88,8 +99,8 @@ const getStyles = <T extends object>(
     props,
     {
       style: {
-        justifyContent: column?.align === 'right' ? 'flex-end' : 'flex-start',
-        textAlign: column?.align === 'right' ? 'right' : 'left',
+        justifyContent: getJustify(column.align),
```
devinleighsmith, line 102: this should respect the existing 'left by default' logic.

frontend/src/features/leases/detail/LeasePages/payment/styles.tsx
```diff
@@ -58,6 +59,61 @@ export const StyledTable = styled(Table)`
         }
       }
     }
+    > .tbody > .tr-wrapper > tr > .td:nth-of-type(2) {
+      margin-right: -5rem; /** TODO: prevent expander from pushing tds to the right */
```
devinleighsmith, line 63: This is a mess, but isn't trivial to fix given that the current logic just adds a td to the existing table with the expander icon in it. This pushes all of the tds in the table to the right. Will try to address this sprint or log tech debt for a better solution for future expandable tables.

backend/dal/Repositories/LeaseTermRepository.cs
FuriousLlama: Instead of setting it to null  you can mark the property not to be modified lilke:
```
pimsLeaseTerm.Property(nameof(<property you want>)).IsModified = false;
```

frontend/src/components/Table/Table.tsx
asanchezr: Potential NPE with **`column.align`** - please use **`column?.align`** instead

frontend/src/components/common/form/NotesModal.tsx
```diff
@@ -8,19 +8,24 @@ import { useState } from 'react';
 import { FaRegFileAlt } from 'react-icons/fa';
 import { withNameSpace } from 'utils/formUtils';
 
-export interface IAddLeaseTenantNoteProps {
+export interface INotesModalProps {
+  notesLabel: React.ReactNode;
+  title: string;
   nameSpace?: string;
+  onSave?: (values: any) => void;
```
asanchezr, line 15: I like making this a generic notes modal. One suggestion - could you please make the field name configurable via props? Right now is hardcoded to `notes` and it might not fit all forms

frontend/src/features/leases/detail/LeasePages/payment/table/PaymentsForm.tsx
```diff
@@ -2,16 +2,19 @@ import { Button } from 'components/common/form';
 import { TableProps } from 'components/Table/Table';
 import Claims from 'constants/claims';
 import { useFormikContext } from 'formik';
+import useDeepCompareMemo from 'hooks/useDeepCompareMemo';
 import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 import { IFormLease } from 'interfaces';
 import { IFormLeaseTerm } from 'interfaces/ILeaseTerm';
-import { orderBy } from 'lodash';
+import _, { noop, orderBy } from 'lodash';
```
asanchezr, line 9: importing only the methods we want from lodash helps with tree-shaking. Would you consider importing `find` directly here? It seems the only usage of `_` here...

frontend/src/interfaces/ILeasePayment.ts
```diff
@@ -12,3 +15,25 @@ export interface ILeasePayment {
...
+
+export const formLeasePaymentToApiPayment = (
+  formLeasePayment: IFormLeasePayment,
+  gstConstant?: number,
+): ILeasePayment => {
+  return {
+    ...formLeasePayment,
+    amountPreTax: stringToNull(formLeasePayment.amountPreTax),
+    amountGst: stringToNull(formLeasePayment.amountGst),
+    amountTotal: stringToNull(formLeasePayment.amountTotal),
```
asanchezr, line 37: quick question - we don't convert the amounts to number before we sent to the API? or is the conversion done prior to this?

asanchezr (approved):

asanchezr:
Approved but unit tests are failing - please fix

frontend/src/interfaces/ILeasePayment.ts
```diff
@@ -12,3 +15,25 @@ export interface ILeasePayment {
...
+    {
+      amountPreTax: NumberFieldValue;
+      amountGst: NumberFieldValue;
+      amountTotal: NumberFieldValue;
+    }
+  > {}
+
+export const formLeasePaymentToApiPayment = (
+  formLeasePayment: IFormLeasePayment,
+  gstConstant?: number,
```
FuriousLlama, line 31: this does not seem to be used

frontend/src/interfaces/ILeasePayment.ts
devinleighsmith: name of the conversion method is non-intuitive perhaps, but the method will:
if empty string, convert to null
if numeric, do not convert

frontend/src/interfaces/ILeasePayment.ts
devinleighsmith: getting ahead of myself it appears.

PR #821, opened 2022-01-30
PSP Sprint 20.00 Build

dfilteau:
PSP_PIMS S20.00 Sprint 20 2022-Jan-28
- Added table:
  - PIMS_STATIC_VARIABLE_HIST
- Altered table:
  - PIMS_STATIC_VARIABLE
- Added view:
  - PIMS_PERSON_CONTACT_VW
- Altered view:
  - PIMS_CONTACT_MANAGER_VW
- Modified seed script contents
  - 70_DML_PIMS_LEASE_PAYMENT_STATUS_TYPE.sql


database/mssql/scripts/dbscripts/PSP_PIMS_S19_00/Alter Down/01_PSP_PIMS_S19_00_Alter Down.sql
```diff
@@ -289,14 +289,6 @@ GO
 IF @@ERROR <> 0 SET NOEXEC ON
 GO
 
--- Drop index dbo.LSTERM_LEASE_TERM_STATUS_TYPE_CODE_IDX
```
devinleighsmith, line 292: @dfilteau surprised to see changes to the IS19 alter down/alter up in this PR.

The problem with doing that is that we have already shipped IS19 and executed the alter up / alter down in the dev and test environments. If we now to try to run these new alter ups/alter downs on test, I'm not sure what the result will be, since the database does not match the state that these updated scripts are expecting.

I'll wait to here back on why these changes are here, but if we do have changes that need to be made to IS19 we'll need a plan to get those into dev and test.

database/mssql/scripts/dbscripts/PSP_PIMS_S20_00/Alter Up/01_PSP_PIMS_S19_00_Alter_Up.sql
```diff
@@ -0,0 +1,282 @@
...
+	[APP_LAST_UPDATE_USER_GUID] uniqueidentifier NULL, 
+	[APP_LAST_UPDATE_USER_DIRECTORY] nvarchar(30) NOT NULL CONSTRAINT [STAVBL_APP_LAST_UPDATE_USER_DIRECTORY_DEF] DEFAULT (user_name())
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
+
+-- Create view dbo.PIMS_PROPERTY_BOUNDARY_VW
+PRINT N'Create view dbo.PIMS_PROPERTY_BOUNDARY_VW'
+GO
+CREATE VIEW [dbo].[PIMS_PROPERTY_BOUNDARY_VW] AS
```
devinleighsmith, line 53: Did this view change?

database/mssql/scripts/dbscripts/PSP_PIMS_S19_00/Alter Down/01_PSP_PIMS_S19_00_Alter Down.sql
devinleighsmith: @dfilteau I just simulated this on my local by setting my local db to the previous version of IS19. I wasn't able to run any combination of alter down/alter up scripts to successfully go from the previous version of IS19 to this version. As a result, these changes may need to be rolled into IS20.

devinleighsmith: or just strip out these changes for now, whatever is easier for you.

devinleighsmith (changes requested):
Can't accept this as is as it will not run on current IS19 databases.

devinleighsmith (approved):

PR #841, opened 2022-02-02
PSP-2819: Contact Method Validation Bug Fix

jdtoombs:
Tweaks so validation fired for contact methods in addition to other required fields.

**Previously:**
contactMethodBug

**Now:**
contactMethodFix

### DIFF
--- a/frontend/src/features/contacts/contact/create/Organization/CreateOrganizationForm.tsx
+++ b/frontend/src/features/contacts/contact/create/Organization/CreateOrganizationForm.tsx
@@ -53,21 +53,20 @@ export const CreateOrganizationForm: React.FunctionComponent = () => {
   );
 
   const formikRef = useRef<FormikProps<ICreateOrganizationForm>>(null);
-
   const onValidate = (values: ICreateOrganizationForm) => {
+    const errors = {} as any;
     try {
-      validateYupSchema(values, OrganizationValidationSchema, true, {
-        otherCountry: otherCountryId,
-      });
       // combine yup schema validation with custom rules
-      const errors = {} as any;
       if (!hasEmail(values) && !hasPhoneNumber(values) && !hasAddress(values)) {
         errors.needsContactMethod =
           'Contacts must have a minimum of one method of contact to be saved. (ex: email,phone or address)';
       }
+      validateYupSchema(values, OrganizationValidationSchema, true, {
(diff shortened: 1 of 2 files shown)


devinleighsmith:
@asanchezr please review.

frontend/src/features/contacts/contact/create/Organization/CreateOrganizationForm.tsx
```diff
@@ -53,21 +53,20 @@ export const CreateOrganizationForm: React.FunctionComponent = () => {
   );
 
   const formikRef = useRef<FormikProps<ICreateOrganizationForm>>(null);
-
   const onValidate = (values: ICreateOrganizationForm) => {
+    const errors = {} as any;
     try {
-      validateYupSchema(values, OrganizationValidationSchema, true, {
```
jdtoombs, line 59: This was previously throwing an error before the check for needsContactMethod was executed. Meaning only the validation schema could be executed and only once that was satisfied the custom validation would fire

frontend/src/features/contacts/contact/create/Person/CreatePersonForm.tsx
```diff
@@ -162,6 +163,17 @@ const CreatePersonComponent: React.FC<FormikProps<ICreatePersonForm>> = ({
     }
   };
 
+  const isContactMethodInvalid = useMemo(() => {
```
jdtoombs, line 166: Otherwise the custom validation will change the text to red before user attempts to save etc.

frontend/src/features/contacts/contact/create/Organization/CreateOrganizationForm.tsx
```diff
@@ -105,7 +104,6 @@ export const CreateOrganizationForm: React.FunctionComponent = () => {
       <Formik
         component={CreateOrganizationComponent}
         initialValues={defaultCreateOrganization}
-        enableReinitialize
```
asanchezr, line 108: question - why you took this away? most forms have it but wondering why was needed to remove here? did you find something odd?

jdtoombs: Oops I can add that back in this was while I was debugging! Thanks for catching

asanchezr (approved):
Thanks for fixing this! 💪

PR #849, opened 2022-02-02
PSP-2286 update individual contact

asanchezr:
image

Unit tests incoming on next commits or new PR


backend/api/Areas/Persons/Controllers/PersonController.cs
```diff
@@ -54,7 +73,7 @@ public PersonController(IPimsRepository pimsService, IMapper mapper)
         public IActionResult AddPerson([FromBody] Models.Person.PersonModel model, bool userOverride = false)
         {
             // Business rule - support country free-form value if country code is "Other". Ignore field otherwise.
-            var otherCountry = _pimsService.Lookup.GetCountries().FirstOrDefault(x => x.Code == Dal.Entities.CountryCodes.Other);
+            var otherCountry = _pimsRepository.Lookup.GetCountries().FirstOrDefault(x => x.Code == Dal.Entities.CountryCodes.Other);
```
devinleighsmith, line 76: Don't have to change this now - but going forward we don't want controllers to access repositories directly.

devinleighsmith: (The service would handle operations like this).

asanchezr: I know but notice this is for Add person (which I didn't want to refactor here). Agree, data access should be moved/proxied to services

backend/api/Areas/Persons/Models/Person/OrganizationLinkModel.cs
```diff
@@ -0,0 +1,22 @@
+
+namespace Pims.Api.Areas.Persons.Models.Person
+{
+    /// <summary>
+    /// Provides a contact-oriented organization model.
+    /// This model contains a sub-set of organization fields as it is only used to link a person to an organization.
+    /// </summary>
+    public class OrganizationLinkModel
```
FuriousLlama, line 8: This class looks like is formating the organization to be rendered already. Why not pass the organization instead and let the UI handle how it is rendered?

asanchezr: This is to match the results from the auto-complete for organizations. We could improve the auto-complete to return the whole organization but was a lot of work and didn't want to refactor the ADD form as well

frontend/src/features/contacts/list/ContactSearchResults/columns.tsx
```diff
@@ -104,16 +104,26 @@ const columns: ColumnWithProps<IContactSearchResult>[] = [
...
-          <MdContactMail size={22} />
-        </IconButton>
-      </StyledDiv>
-    ),
+    Cell: (props: CellProps<IContactSearchResult>) => {
+      const history = useHistory();
+
+      return (
+        <StyledDiv>
+          <IconButton
```
devinleighsmith, line 112: Do these need to be controlled by claims?

devinleighsmith (approved):

frontend/src/features/contacts/list/ContactSearchResults/columns.tsx
asanchezr: They weren't originally but can change. All I did was add an **onClick** handler to existing button as my story spoke about clicking on the button. I didn't go to the contact list story to check claims but could see that they are required here.

They should have been hidden since then but can change now

devinleighsmith:
@asanchezr I don't see any tests here, so please log tech debt to add unit tests. We are getting awfully close to our lower bound of 75%. We can't afford to hit that during end of sprint because our merges will get blocked and we'll have to drop features out of the sprint last minute.

asanchezr:
@devinleighsmith - I mentioned in PR description unit tests would be forthcoming. Will log tech-debt

FuriousLlama (approved):

backend/api/Areas/Persons/Models/Person/OrganizationLinkModel.cs
FuriousLlama: I see, we should standardize those too

devinleighsmith:
ah, sorry, didn't see that.

backend/api/Areas/Persons/Models/Person/OrganizationLinkModel.cs
asanchezr: we should

PR #885, opened 2022-02-04
Update gen env files.

devinleighsmith:
-Remove dead import .env
-Remove unused api postgres .env
-Add missing parameter for sql server make scripts.


devinleighsmith:
This should ensure that new resources setting up their envs for the first time have all required params for local database make commands.

asanchezr (approved):

PR #894, opened 2022-02-07
PSP-2844 Require proper claims to edit contacts

asanchezr:


frontend/src/features/contacts/list/ContactSearchResults/columns.tsx
```diff
@@ -108,27 +113,27 @@ const columns: ColumnWithProps<IContactSearchResult>[] = [
     maxWidth: 40,
     Cell: (props: CellProps<IContactSearchResult>) => {
       const history = useHistory();
-      const { hasClaim } = useKeycloakWrapper();
-
       return (
         <StyledDiv>
-          {hasClaim(Claims.CONTACT_EDIT) && (
+          <ProtectedComponent hideIfNotAuthorized claims={[Claims.CONTACT_EDIT]}>
```
FuriousLlama, line 118: The direct check here seems better than using a component than hides since is doing the same check, just one level below

asanchezr: I personally like the component better as it hides/encapsulates the boilerplate to check for claims. Maybe the component could be named better etc. Probably we should discuss on Friday's dev meeting

FuriousLlama: In this particular case, I don't think the logic needs encapsulation. For me is more obvious since the code that attempts to reduce the boilerplate code, is longer than the naive option.

asanchezr: Fine to change it back but also was following existing patterns in the code using the wrapper component. We should sync up this Friday on approach moving forward

asanchezr: Changed

devinleighsmith (approved):

PR #930, opened 2022-02-14
psp-2829

devinleighsmith:


backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
```diff
@@ -11,9 +12,9 @@ public class LeaseMap : IRegister
         private static void MapLease(Entity.PimsLease src, Model.LeaseModel dest)
         {
             dest.LFileNo = src.LFileNo;
-            dest.StartDate = src.OrigStartDate;
-            dest.EndDate = src.OrigExpiryDate;
-            dest.CurrentTermStartDate = src.GetCurrentTermStartDate();
+            dest.StartDate = src.OrigStartDate.FilterSqlMinDate();
+            dest.EndDate = src.OrigExpiryDate.FilterSqlMinDate();
```
FuriousLlama, line 16: if the sql date is min would that mean that it was not set? if that is allowed by design it should be set to null instead, otherwise, I think it would be a bug

devinleighsmith: no, this comes in from the ETL.

FuriousLlama (approved):

asanchezr (approved):

PR #940, opened 2022-02-15
PSP-2673: Manage users pagination

jdtoombs:
Updated the `pageIndex` to be state based in `ManageUsers.tsx`
paginationBug


frontend/src/components/Table/Table.tsx
```diff
@@ -370,8 +371,9 @@ const Table = <T extends IIdentifiedObject, TFilter extends object = {}>(
   }, [pageSizeProp, instance]);
 
   useEffect(() => {
+    props.setPageIndex && props.setPageIndex(pageIndex);
     onRequestData?.({ pageIndex: pageIndex, pageSize });
-  }, [onRequestData, pageIndex, pageSize]);
+  }, [onRequestData, pageIndex, pageSize, props]);
```
devinleighsmith, line 376: hmm, what field within props is fixing this issue?

devinleighsmith:
Looking for an overview on this one. Why do we need to track page state independently of the react table? Is the internal page tracker within react table not working properly? If so, do we need to make tracking external page state required? Does PIMS also have this issue?

jdtoombs:
> Looking for an overview on this one. Why do we need to track page state independently of the react table? Is the internal page tracker within react table not working properly? If so, do we need to make tracking external page state required? Does PIMS also have this issue?

@devinleighsmith reworked to match same process as `ContactSearchResults.tsx`

devinleighsmith (approved):

PR #941, opened 2022-02-15
PSP-2300 update organization

asanchezr:
image


backend/api/Areas/Organizations/Mapping/Organization/AddressMap.cs
```diff
@@ -24,8 +25,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.CountryId, src => src.Address.CountryId)
                 .Map(dest => dest.Postal, src => src.Address.PostalCode)
                 .Map(dest => dest.CountryOther, src => src.Address.OtherCountry)
-                .Map(dest => dest.AddressTypeId, src => src.AddressUsageTypeCodeNavigation)
-                .Inherits<Entity.IBaseAppEntity, Api.Models.BaseAppModel>();
```
FuriousLlama, line 28: should this line do the mapping to rowversion instead?

backend/api/Areas/Persons/Mapping/Person/PersonMap.cs
```diff
@@ -42,6 +41,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.PimsPersonAddresses, src => src.Addresses)
                 .Map(dest => dest.PimsContactMethods, src => src.ContactMethods)
                 .Map(dest => dest.PimsPersonOrganizations, src => src.Organization != null ? new[] { src } : null)
+                .Inherits<Api.Models.BaseAppModel, Entity.IBaseAppEntity>()
```
FuriousLlama, line 44: nice!

backend/api/Areas/Organizations/Mapping/Organization/AddressMap.cs
asanchezr: Sorry I don't understand the question. Could you be more specific please? 
Is it that the mapping is wrong?

asanchezr: Ah I understand the question now. If you look closely the RowVersion comes from the sub-entity -> `src.Address.ConcurrencyControlNumber`

This is due to flattening the model (which we won't do moving forward but this is old code). The generic mapping wants the RowVersion from the main entity (PimsOrgAddress)

PR #945, opened 2022-02-15
Lookup performance

devinleighsmith:
psp-2662


devinleighsmith:
due to the huge number of etl'd organizations, this had an unexpectedly large impact on lookup functionality. As a result, completely removing organizations from lookup api.

FuriousLlama (approved):

PR #1008, opened 2022-02-23
Mayan local dev - Features/psp 2607

FuriousLlama:
Docker container list
image

Mayan web fronty end. Port:7080


tools/mayan-edms/.mayan-env
```diff
@@ -0,0 +1,64 @@
+##################################################################
```
FuriousLlama, line 1: left this file mostly intact. We can change it more later and as it is is mostly examples of what could be set.

Makefile
```diff
@@ -305,5 +305,10 @@ env: ## Generate env files
 	@echo "$(P) Generate/Regenerate env files required for application (generated passwords only match if database .env file does not already exist)"
 	@./scripts/gen-env-files.sh;
 
-.PHONY: logs start destroy local setup restart refresh up down stop build rebuild clean client-test server-test pause-30 server-run db-clean db-drop db-seed db-refresh db-script db-scaffold npm-clean npm-refresh keycloak-sync convert backend-coverage frontend-coverage backend-test frontend-test env
+mayan-up: ## Calls the docker compose up for the mayan images
```
FuriousLlama, line 308: not 100% sure this would be useful, but add it for to match the rest of the dev tools

tools/mayan-edms/docker-compose.yml
```diff
@@ -0,0 +1,228 @@
+version: "3.9"
```
FuriousLlama, line 1: left most of the services that came from the default. I tried removing some of them but retrieving uploaded files would fail without redis, celery or index.

Also, I've let the services be on a separate file. I think it would make the existing docker file more complex for the rest of the services if merged into only one

devinleighsmith (approved):

Makefile
```diff
@@ -305,5 +305,10 @@ env: ## Generate env files
...
+	@echo "$(P) Create or start mayan-edms system"
+	@cd tools/mayan-edms
+	@docker-compose up -d
```
asanchezr, line 311: Not a big deal but this two lines should be made into one. Make executes each command/line in a new bash shell so the "cd" command does nothing for docker-compose. See other targets in this file like **frontend-test** for reference

asanchezr (approved):

Makefile
FuriousLlama: good catch, I've updated it

PR #1014, opened 2022-02-24
Psp 2474

devinleighsmith:
Navigation to the new slide out.
Remove unused draft marker code.
Remove unused property slice code.


frontend/src/utils/propertyUtils.ts
```diff
@@ -17,6 +17,26 @@ export const pidFormatter = (pid?: string) => {
   return '';
 };
 
+/**
+ * The pidParser is used to return a numeric pid value from a formatted pid.
+ * @param {string} pid This is the target PID to be parsed
+ */
+export const pidParser = (pid?: string): number | undefined => {
+  if (typeof pid === 'number') {
```
FuriousLlama, line 25: is it possible that the pid is a number? if so, the method definition should be ```pidParser = (pid?: string | number)```

frontend/src/utils/propertyUtils.ts
```diff
@@ -17,6 +17,26 @@ export const pidFormatter = (pid?: string) => {
...
+    return pid;
+  }
+  if (!!pid) {
```
FuriousLlama, line 28: nit: if (pid === undefined) looks more legible to me

FuriousLlama (approved):

frontend/src/utils/propertyUtils.ts
asanchezr: agree typing could be improved here

frontend/src/utils/propertyUtils.ts
asanchezr: There is a subtle difference with your change Manuel. === undefined would not catch null or empty string values
The code above encompasses all that (and even ensuring pid is not the number 0)

FuriousLlama: Right. My point is that the method accepts 3 types, String, Number, or Undefined (?). Lines 25-27 check that is a number, so only undefined is left. For that we only need now to check if it is undefined or empty, then we could do that. Using the !! as a catch-all suggests that the previous code might be missing something.
Reading down the code, if there is empty space it would be cached by the next set of tests when trying to match the regex.

PR #1017, opened 2022-02-25
PSP-2469 Link org mailing address to individual

asanchezr:
**NOTE** - working on unit tests. Will add them to this PR when ready.

image


backend/dal/Repositories/PersonRepository.cs
```diff
@@ -103,6 +103,14 @@ public PimsPerson Add(PimsPerson person, bool userOverride)
             }
 
             this.Context.PimsPeople.Add(person);
+
+            // Ensure we don't add existing address from linked organization again
+            var mailing = person.GetMailingAddress();
+            if (person.UseOrganizationAddress == true && mailing != null)
```
devinleighsmith, line 109: I see person.UseOrganizationAddress being used here to not create the linkage again. Am I correct in thinking that the actual logic to associate the organization address when this flag is true is on the frontend?

asanchezr: Right. When checkbox is checked, the frontend gets the organization details (incl address) and sets it as mailing address. This logic here is to avoid an error I saw on creating a person with existing mailing address. Note that Context.Add(person) marks the person (and all subentities within it) as Added so it was trying to re-insert an existing address. The DB complains due to triggers enforcing no duplicates so this logic basically ignores that address (and only that one)

asanchezr: The other use case where someone adds a net new mailing address (not the one linked from org) would not apply to this logic and would be inserted in the DB as expected

asanchezr: Also, to respond to your specific question about linkage here. The linkage IS created here. What is not re-created is the address row it points to:

```
person (add to db) -> person-address (add to db) -> address (existing already, do not add)
```

Hopefully that makes it clear

frontend/src/features/contacts/contact/create/components/address/Address.tsx
```diff
@@ -128,20 +158,29 @@ interface IProvinceOrCountryName {
...
 const ProvinceOrCountryName: React.FunctionComponent<IProvinceOrCountryName> = ({
   selectedCountry,
   provinces,
   formLabels,
   namespace,
+  disabled = false,
 }) => {
   if (selectedCountry === CountryCodes.Other) {
-    return <Input field={withNameSpace(namespace, 'countryOther')} label="Country name" />;
+    return (
```
devinleighsmith, line 172: confused why this is being added in this PR, was this missed in the original implementation?

frontend/src/features/contacts/contact/edit/Person/UpdatePersonForm.tsx
```diff
@@ -146,6 +163,34 @@ const UpdatePersonComponent: React.FC<FormikProps<IEditablePersonForm>> = ({
     }
   };
 
+  // update mailing address sub-form when "useOrganizationAddress" checkbox is toggled
+  useEffect(() => {
+    // toggle is on - set mailing address values to match organization address
+    if (useOrganizationAddress === true && organizationId) {
+      getOrganization(organizationId)
+        .then(({ data }) => getApiMailingAddress(data))
```
devinleighsmith, line 171: Is there a benefit to having multiple thens here?

devinleighsmith (approved):

frontend/src/features/contacts/contact/edit/Person/UpdatePersonForm.tsx
asanchezr: no benefit, personal preference to single lines but can be re-written as:

```tsx
  getOrganization(...)
    .then(({data)) => {
      const mailing = getApiMailingAddress(...)
      setFieldValue(..., mailing)
    })
   .catch(...)

```

devinleighsmith:
@asanchezr NOTE: this functionality was intended for release in Release 2.0, IS21. It will need to be included in TEST as well as DEV. Please notify the FT team when you have merged to TEST.

asanchezr:
Okay @devinleighsmith so I understand. You want a push to TST out of release cycle or you want me to do another PR to TST with this feature? not clear

asanchezr:
Or I can change the base branch of this PR to test but not sure on implications re: conflicts etc

PR #1018, opened 2022-02-25
Contact manager select for deposits | psp-2953

FuriousLlama:
New PR: https://github.com/bcgov/PSP/pull/1023

Added new common models
Updated contact manager view to support single selects
Wired up deposits with contact-manager
Created a new input type that allows opening the cm modal

image

image

image


backend/api/Models/Concepts/Address/AddressMap.cs
```diff
@@ -0,0 +1,36 @@
+using Mapster;
+using Entity = Pims.Dal.Entities;
+using Model = Pims.Api.Models.Concepts;
+
+namespace Pims.Api.Models.Concepts
```
FuriousLlama, line 5: Created a new folder under Models called Concepts (Couldn't think of anything better :) )
THey are isolated from all other models so that we can start refactoring the rest of the controllers. They pair with a models folder on the front end. Moving forward we should only create models there, and remove the other ones as we refactor

frontend/src/components/Table/Table.tsx
```diff
@@ -188,10 +201,15 @@ const IndeterminateCheckbox = React.forwardRef(
       if (isHeaderCheck) {
         setSelected([]);
       } else {
-        if (currentSelected.find(selected => selected.id === row.original.id)) {
-          remove(currentSelected, row.original);
-        } else {
+        if (isSingleSelect === true) {
```
FuriousLlama, line 204: If isSingleSelect is passed to the table, only one row can be selected at a time.

frontend/src/features/leases/detail/LeasePages/deposits/DepositsContainer.test.tsx
```diff
@@ -1,17 +1,13 @@
...
-} from 'interfaces';
+import { defaultFormLease, IFormLease } from 'interfaces';
 import { noop } from 'lodash';
+import { Api_SecurityDeposit, Api_SecurityDepositReturn } from 'models/api/SecurityDeposit';
 import { render, RenderOptions, RenderResult } from 'utils/test-utils';
 
 import DepositsContainer from './DepositsContainer';
 
-const mockDeposits: ILeaseSecurityDeposit[] = [
+const mockDeposits: Api_SecurityDeposit[] = [
```
FuriousLlama, line 10: Api_* models are isolated and to be matched with the new common models

devinleighsmith:
@FuriousLlama Note that this functionality is part of release 2.0, IS21. Please make sure this gets into test as well as dev, and notify the FT team as soon as it is available in TEST.

backend/api/Models/Concepts/Address/AddressMap.cs
devinleighsmith: Maybe we could talk about this during our meeting today, I don't mind the word, but mixing concept with _model seems weird.

backend/dal/Repositories/SecurityDepositReturnRepository.cs
```diff
@@ -44,62 +42,29 @@ public IEnumerable<PimsSecurityDepositReturn> GetByDepositId(long id)
...
+            return depositReturn;
         }
 
         public void Delete(long id)
         {
-            PimsSecurityDepositReturn term = this.Context.PimsSecurityDepositReturns.Find(id) ?? throw new KeyNotFoundException();
-            this.Context.Remove(term);
-        }
-
-        /*private PimsSecurityDepositReturn AssociateHolders(PimsSecurityDepositReturn securityDeposit)
```
devinleighsmith, line 66: eech, the commented out logic made it into the actual code base? ouch.

backend/dal/Services/SecurityDepositReturnService.cs
```diff
@@ -35,6 +35,12 @@ public PimsLease UpdateLeaseDepositReturn(long leaseId, long leaseRowVersion, Pi
         {
             _user.ThrowIfNotAuthorized(Permissions.LeaseEdit);
             ValidateServiceCall(leaseId, leaseRowVersion);
+            var currentHolder = _securityDepositReturnRepository.GetById(deposit.SecurityDepositReturnId).PimsSecurityDepositReturnHolder;
+            if (currentHolder != null)
```
devinleighsmith, line 39: Is this possible? or would a keyNotFoundException happen here?

backend/dal/Services/SecurityDepositReturnService.cs
```diff
@@ -35,6 +35,12 @@ public PimsLease UpdateLeaseDepositReturn(long leaseId, long leaseRowVersion, Pi
...
+            {
+                deposit.PimsSecurityDepositReturnHolder.SecurityDepositReturnHolderId = currentHolder.SecurityDepositReturnHolderId;
```
devinleighsmith, line 41: What is the purpose of this logic, is this to prevent someone from inadvertently updating the holder?

frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const ContactInput: React.FC<ContactInputProps> = ({
+  field,
+  setShowContactManager,
+  onClear,
+  label,
+  required,
+  displayErrorTooltips,
+}) => {
+  const { errors, touched, values } = useFormikContext<any>();
```
devinleighsmith, line 38: I think you can omit the any here for an equivalent (shorter) statement.

frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+    >
+      {!!label && <Form.Label>{label}</Form.Label>}
+
+      <TooltipWrapper toolTipId={`${field}-error-tooltip}`} toolTip={errorTooltip}>
+        <Row className="align-items-center">
+          <Col>
+            <StyledDiv className={!!touch && !!error ? 'is-invalid' : ''}>
+              {value?.summary}
+            </StyledDiv>
+            <Input field={field + '.id'} className="d-none"></Input>
```
devinleighsmith, line 57: wait, isn't this the numeric id from the database? Is there value to including this here?

frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+          </Col>
+          <Col xs="auto" className="pl-0 align-self-center">
+            <Button
+              icon={<MdClose size={20} />}
+              variant="secondary"
+              className="px-2"
```
devinleighsmith, line 63: Can we check up on what css this generates? We are using base 10 rem everwhere else, want to make sure this isn't using px or base 16 rem.

devinleighsmith: (I'm talking about the bootstrap shorthand).

backend/dal/Services/SecurityDepositReturnService.cs
FuriousLlama: This is needed since the front does not have the holder table id. In the front end, the contact is selected (contact view Id) wherein the backend that needs to be set to the existing holder id

frontend/src/components/common/form/ContactInput.tsx
FuriousLlama: This was necessary to get the tests running, Is quite hacky, but I couldn't think of a different way of triggering the yum validation when doind the jest test

frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -0,0 +1,96 @@
...
+            ></Button>
+          </Col>
+        </Row>
+      </TooltipWrapper>
+      {!displayErrorTooltips && <DisplayError field={field} errorPrompt />}
+    </Form.Group>
+  );
+};
+
+const StyledDiv = styled.div`
```
devinleighsmith, line 86: nit, normally I do:
`
{
  css for normal case
  {
    css for error/exception case
  }
}
`
because that seems clearer to me.

frontend/src/components/common/form/styles.ts
```diff
@@ -68,7 +68,6 @@ export const InlineCol = styled(Col)`
 export const InlineFastDatePicker = styled(FastDatePicker)`
   display: flex;
   align-items: baseline;
-  gap: 0.8rem;
```
devinleighsmith, line 71: This will probably affect a number of form, have you regression validated this?

frontend/src/components/common/form/ContactInput.tsx
FuriousLlama: The sizes are multiplied by a spacer. AFAIK that is 1rem by default.

frontend/src/components/common/form/styles.ts
FuriousLlama: I took a look at them, the problem was that this is the extra space that was showing up on the forms when they where stacked

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
```diff
@@ -29,14 +29,21 @@ export class DepositListEntry {
...
+    this.contactHolder = baseDeposit.contactHolder;
     this.depositReturnCount = baseDeposit.depositReturns.length;
   }
 }
 
 function renderHolder({ row: { original } }: CellProps<DepositListEntry, string>) {
-  return original.personDepositHolder?.fullName || '';
+  if (original.contactHolder) {
+    if (original.contactHolder.person) {
+      return original.contactHolder.person.firstName + ' ' + original.contactHolder.person.surname;
```
devinleighsmith, line 40: Do we have requirements on middle name here?

backend/api/Models/Concepts/Address/AddressMap.cs
asanchezr: Not a fan of that naming either. Let's discuss in our meeting. Don't see any problem in them just being "models" and live in the root model folder

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
devinleighsmith: also this will result in an extra space if the surname is null/empty/undefined.

backend/dal/Services/SecurityDepositReturnService.cs
FuriousLlama: If that happens then the update request would be bad since is using the deposit return id to get deposit returns

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
devinleighsmith: There is a utility in personUtils for contacts that we could maybe reuse here.

FuriousLlama: Given that we were talking about how to display names I am not sure what is the standard. I believe we will standardize the person name everywhere in a later story

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReturnedContainer/columns.tsx
```diff
@@ -41,13 +33,20 @@ export class ReturnListEntry {
...
-    this.organizationDepositHolder = baseDeposit.organizationDepositReturnHolder;
+    this.contactHolder = baseDeposit.contactHolder;
   }
 }
 
-function renderPerson({ row: { original } }: CellProps<ReturnListEntry, string>) {
-  return original.personDepositHolder?.fullName || '';
+function renderHolder({ row: { original } }: CellProps<ReturnListEntry, string>) {
+  if (original.contactHolder?.person !== undefined) {
+    var person = original.contactHolder?.person;
```
devinleighsmith, line 42: nit: can remove the ?. here.

devinleighsmith: could also store the dereference in a var to avoid dereferencing twice.

frontend/src/features/leases/detail/LeasePages/deposits/models/FormLeaseDeposit.ts
```diff
@@ -0,0 +1,57 @@
+import { fromContact, IContactSearchResult, toContact } from 'interfaces';
+import { Api_SecurityDeposit } from 'models/api/SecurityDeposit';
+import { NumberFieldValue } from 'typings/NumberFieldValue';
+
+export class FormLeaseDeposit {
```
devinleighsmith, line 5: I think we need to talk about this, I'm starting to think that having half of our forms using classes and half using interfaces isn't such a hot idea.

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
devinleighsmith: I'm not aware of such a story. Something like this isn't really appropriate for tech debt either since it affects requirements. Maybe check in with the BAs?

devinleighsmith:
Sorry about commenting on the wrong PR @FuriousLlama

frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
FuriousLlama: I double-checked and it states middle name too, I've added it

PR #1024, opened 2022-02-25
Reopening PR for contacts automation test cases

stairaku:


jdtoombs (approved):
@devinleighsmith I believe Sue cannot add tags as she is not own the codeowners. I may be mistaken though

PR #1029, opened 2022-02-28
Do not duplicate person contacts if a single person has multiple orga…

devinleighsmith:
…nizations.

### DIFF
--- a/database/mssql/scripts/dbscripts/PSP_PIMS_S21_00/Alter Up/01_PIMS_CONTACT_MGR_VIEW.sql
+++ b/database/mssql/scripts/dbscripts/PSP_PIMS_S21_00/Alter Up/01_PIMS_CONTACT_MGR_VIEW.sql
@@ -1,0 +1,60 @@
+-- Drop view dbo.PIMS_CONTACT_MGR_VW
+PRINT N'Drop view dbo.PIMS_CONTACT_MGR_VW'
+GO
+DROP VIEW [dbo].[PIMS_CONTACT_MGR_VW]
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
+
+-- Create view dbo.PIMS_CONTACT_MGR_VW
+PRINT N'Create view dbo.PIMS_CONTACT_MGR_VW'
+GO
+CREATE VIEW [dbo].[PIMS_CONTACT_MGR_VW] AS
+  SELECT CONCAT('P', PER.PERSON_ID)                                          AS ID
+       , PER.PERSON_ID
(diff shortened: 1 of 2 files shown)


devinleighsmith:
Note that this fixes code in TEST, so I'll need to run this in TEST and dev after merging.
Also @dfilteau please be aware of this change, as it fixes a functional regression in the contacts view.

database/mssql/scripts/dbscripts/PSP_PIMS_S21_00/Alter Up/01_PIMS_CONTACT_MGR_VIEW.sql
```diff
@@ -0,0 +1,60 @@
...
+       , ADR.MUNICIPALITY_NAME
+       , PRV.PROVINCE_STATE_CODE                                             AS PROVINCE_STATE
+       , COALESCE(PVW.WORK_EMAIL, PVW.PERSONAL_EMAIL)                        AS EMAIL_ADDRESS
+  FROM   PIMS_PERSON              PER                                                         LEFT JOIN
+         PIMS_PERSON_ADDRESS      PAD ON PAD.PERSON_ID                = PER.PERSON_ID
+                                     AND PAD.ADDRESS_USAGE_TYPE_CODE  = 'MAILING'             LEFT JOIN
+         PIMS_ADDRESS             ADR ON ADR.ADDRESS_ID               = PAD.ADDRESS_ID        LEFT JOIN
+         PIMS_PROVINCE_STATE      PRV ON PRV.PROVINCE_STATE_ID        = ADR.PROVINCE_STATE_ID LEFT JOIN
+         PIMS_PERSON_CONTACT_VW   PVW ON PVW.PERSON_ID                = PER.PERSON_ID         LEFT JOIN
+         (SELECT PERSON_ID, MAX(ORGANIZATION_ID) as ORGANIZATION_ID FROM PIMS_PERSON_ORGANIZATION GROUP BY PERSON_ID) as POR
```
devinleighsmith, line 33: This is the only change. we do not want to create multiple rows for a person that is in more then one organization. As a result, this join needs to only occur once. We don't have a sorting criteria here so I'm joining to the highest id.

asanchezr: We need to talk about the need for multiple organizations on a user. All my contact manager stories mention that a user should be associated with a single organization. Perhaps we can discuss this on our Friday meeting.

asanchezr (approved):

PR #1032, opened 2022-02-28
Made PID and PIN displaying optional

FuriousLlama:
image


devinleighsmith (approved):
Approved, although should this be formatting the pid as well?

PR #1041, opened 2022-03-01
Added a multi select to the lease search

FuriousLlama:
image


backend/api/Models/Concepts/Address/AddressMap.cs
```diff
@@ -0,0 +1,36 @@
+using Mapster;
+using Entity = Pims.Dal.Entities;
+using Model = Pims.Api.Models.Concepts;
+
+namespace Pims.Api.Models.Concepts
+{
+    public class AddressMap : IRegister
```
asanchezr, line 7: Doesn't seem like this file is related to this PR. Maybe you need to rebase @FuriousLlama ?

frontend/src/utils/personUtils.ts
```diff
@@ -4,7 +4,9 @@ export function formatFullName(person?: Partial<IEditablePerson>): string {
...
     return '';
   }
-  const nameParts = [person.firstName, person.middleNames, person.surname];
-  const fullName = nameParts.filter(n => n != null && n.trim() !== '').join(' ');
-  return fullName;
+  return formatNames([person.firstName, person.middleNames, person.surname]);
+}
+
+export function formatNames(nameParts: Array<string | undefined | null>): string {
+  return nameParts.filter(n => n !== null && n !== undefined && n.trim() !== '').join(' ');
```
asanchezr, line 11: Not sure I like this change. I rather have an interface that explicitly calls what name parts are we using (first, middle, last) than to rely on the position of the string in the incoming array. Thoughts?

asanchezr: NVM, I see you refactor the above method to call this one, should be fine

FuriousLlama: This one only concatenates the string, it provides the flexibility to change the order if needed be

backend/api/Models/Concepts/Address/AddressMap.cs
FuriousLlama: I rebased, its probably git not doing the diff correctly

frontend/src/features/leases/list/LeaseFilter/LeaseFilter.tsx
```diff
@@ -7,7 +7,8 @@ import { PropertyFilterOptions } from 'features/properties/filter';
 import { Formik } from 'formik';
 import useLookupCodeHelpers from 'hooks/useLookupCodeHelpers';
 import Multiselect from 'multiselect-react-dropdown';
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useState } from 'react';
```
asanchezr, line 11: nit: please put all react imports on same line

FuriousLlama: weird, I thought the lint command took care of this.

backend/api/Areas/Leases/Models/Search/LeaseFilterModel.cs
```diff
@@ -136,7 +136,7 @@ public override bool IsValid()
                 || !string.IsNullOrWhiteSpace(PinOrPid)
                 || !string.IsNullOrWhiteSpace(LFileNo)
                 || !string.IsNullOrWhiteSpace(Address)
-                || !string.IsNullOrWhiteSpace(LeaseStatusType)
+                || (LeaseStatusTypes.Count != 0)
```
devinleighsmith, line 139: nit: > is probably more correct, although negatives shouldn't be an issue here realistically.

frontend/src/features/leases/list/LeaseFilter/LeaseFilter.tsx
```diff
@@ -66,29 +66,57 @@ const idFilterPlaceholders = {
...
+  );
+
+  const [selectedStatus, setInitialSelectedStatus] = useState<MultiSelectOption[]>(
+    initialLeaseStatusList,
+  );
+
+  // Necessary since the lookup codes does not work properly with the multi-select on first render
+  useEffect(() => {
+    setInitialSelectedStatus([{ id: 'ACTIVE', text: 'Active' }]);
+  }, []);
```
devinleighsmith, line 114: likely the issue is that the lookup codes haven't been loaded yet, might make more sense to have this useeffect set the initial status until the lookup codes have loaded, or block somewhere above this to ensure lookup codes have loaded before executing this logic.

devinleighsmith: Or this, the above comment makes it seem like the reason why it doesn't work is unknown.

devinleighsmith:
looks like the multiselect doesn't display properly on smaller screens. css should be updated to accommodate.

image

devinleighsmith:
In the above case, clicking on the close X does not work.

devinleighsmith (approved):

PR #1049, opened 2022-03-03
PSP-2532 Update backend to .NET 6

asanchezr:
### Note
OpenShift build configs need to be updated prior to merging this PR or the build will fail.

### API container info
image


devinleighsmith (approved):

FuriousLlama:
Right on!

PR #1078, opened 2022-03-08
Common styles - PSP-3016

FuriousLlama:
Added common styles and a component that represents label and value pairs

### DIFF
--- a/frontend/src/features/mapSideBar/tabs/InventoryTabs.tsx
+++ b/frontend/src/features/mapSideBar/tabs/InventoryTabs.tsx
@@ -31,28 +31,39 @@ export const InventoryTabs: React.FunctionComponent<IInventoryTabsProps> = ({ Pr
         <StyledScrollable>
           <StyledFormSection>
             <StyledSectionHeader>Section Header</StyledSectionHeader>
-            <SectionField label="Single line" content={['A test description']} />
-            <SectionField
-              label="Many lines"
-              content={['Up', 'Down', 'Left', 'Right']}
-              isMultiLine
-            />
-            <SectionField label="Multiple inline" content={['Yes', 'No', "I dont't know"]} />
+            <SectionField label="Single line">A test description</SectionField>
+            <SectionField label="Many lines">
+              <div>Up</div> <div>Down</div>
+              <div>Left</div>
+              <div>Right</div>
+            </SectionField>
+            <SectionField label="Multiple inline">
+              <span>Up</span> <span>Down</span>
+              <span>Left</span>
+              <span>Right</span>
+            </SectionField>
           </StyledFormSection>
 
           <StyledFormSection>
             <StyledSectionHeader>Split section header</StyledSectionHeader>
             <Row>
               <Col>
-                <SectionField label="A single label" content={['A test description']} />
+                <SectionField label="Single line">A test description</SectionField>
               </Col>
               <Col>
-                <SectionField
-                  label="Many lines"
-                  content={['Up', 'Down', 'Left', 'Right']}
-                  isMultiLine
-                />
-                <SectionField label="Multiple inline" content={['Yes', 'No', "I dont't know"]} />
+                <Row>
+                  <Col>
+                    <SectionField label="Single line">A test description</SectionField>
+                  </Col>
+                  <Col>
+                    <SectionField label="Single line">A test description</SectionField>
+                    <SectionField label="Multiple inline">
+                      <span>Up</span> <span>Down</span>
+                      <span>Left</span>
+                      <span>Right</span>
+                    </SectionField>
+                  </Col>
+                </Row>
               </Col>
             </Row>
           </StyledFormSection>
--- a/frontend/src/features/mapSideBar/tabs/InventoryTabs.tsx
+++ b/frontend/src/features/mapSideBar/tabs/InventoryTabs.tsx
@@ -1,6 +1,9 @@
 import TabView from 'components/common/TabView';
 import * as React from 'react';
-import { Tab } from 'react-bootstrap';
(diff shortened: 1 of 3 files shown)


frontend/src/features/mapSideBar/tabs/InventoryTabs.tsx
```diff
@@ -10,17 +13,55 @@ export enum InventoryTabNames {
...
  */
 export const InventoryTabs: React.FunctionComponent<IInventoryTabsProps> = ({ PropertyForm }) => {
+  const longText =
+    'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam porttitor nisl at elit vestibulum vestibulum. Nullam eget consectetur felis, id porta eros. Proin at massa rutrum, molestie lorem a, congue lorem.';
   return (
     <TabView defaultActiveKey={InventoryTabNames.property}>
       <Tab eventKey={InventoryTabNames.title} title="Title"></Tab>
       <Tab eventKey={InventoryTabNames.value} title="Value"></Tab>
       <Tab eventKey={InventoryTabNames.property} title="Property Details"></Tab>
+      <Tab eventKey={InventoryTabNames.styles} title="STYLES">
```
devinleighsmith, line 30: I suppose we'll take this out sometime during the sprint? or left in by accident?

frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styled from 'styled-components';
+
+interface ISectionFieldProps {
+  label: string;
+  content: string[];
+  isMultiLine?: boolean;
+}
+
+export const SectionField: React.FunctionComponent<ISectionFieldProps> = props => {
+  console.log(props.content);
```
devinleighsmith, line 11: nit

frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -0,0 +1,37 @@
+import { Col, Row } from 'react-bootstrap';
+import styled from 'styled-components';
+
+interface ISectionFieldProps {
+  label: string;
+  content: string[];
```
devinleighsmith, line 6: could we make this a ReactNode? that would allow me to use this component.

devinleighsmith: or ReactNode[].

devinleighsmith: or ReactNode[]

devinleighsmith: Actually, it might be nice to have another version of this that just provides the Row/Col structure but allows you to pass through whatever you want via child props. I can add that myself though.

devinleighsmith: something like this: 
```
import { Col, Row } from 'react-bootstrap';

interface ISectionFieldWrapperProps {
  label: string;
}

export const SectionFieldWrapper: React.FunctionComponent<SectionFieldWrapper> = props => {
  return (
    <Row className="pb-2">
      <Col xs="4">
        <StyledFieldLabel>{props.label}:</StyledFieldLabel>
      </Col>
      <Col>{props.children}</Col>
    </Row>
  );
};

const StyledFieldLabel = styled.div`
  font-weight: bold;
`;
```

frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isMultiLine?: boolean;
+}
+
+export const SectionField: React.FunctionComponent<ISectionFieldProps> = props => {
+  return (
+    <Row className="pb-2">
+      <Col xs="4">
+        <StyledFieldLabel>{props.label}:</StyledFieldLabel>
```
devinleighsmith, line 14: could consider adding the : via css, that would allow easier consumers to override.

frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -0,0 +1,37 @@
...
+              {item}
+            </span>
+          ),
+        )}
+      </Col>
+    </Row>
+  );
+};
+
+const StyledFieldLabel = styled.div`
```
devinleighsmith, line 35: this should probably be a label not a div.

devinleighsmith: Also would be nice to be able to pass the 'for' as a prop so that jests *ByLabel methods work. Also this helps with accessibility.

devinleighsmith: another option would be to use a Form.Group with a controlId as that ensures the child Form.Label and Form.Control are linked appropriately.

frontend/src/features/mapSideBar/tabs/SectionField.tsx
asanchezr: I also prefer the version that just enforces layout and lets the consumer decide what child to render. That way we can use other things there than raw text

asanchezr (approved):

frontend/src/features/mapSideBar/tabs/SectionField.tsx
FuriousLlama: I like the label instead of div, the form group could be done later if needed

FuriousLlama: I opted to leave the div as is. If there is a need for inputs to match the label we could pass a isForm or something like that so it changes dynamicaly

PR #1085, opened 2022-03-10
Psp 2377

devinleighsmith:
Add ltsa tab, required logic to make request/load data.

### DIFF
--- a/frontend/src/features/mapSideBar/tabs/ltsa/LtsaLandSubForm.tsx
+++ b/frontend/src/features/mapSideBar/tabs/ltsa/LtsaLandSubForm.tsx
@@ -1,0 +1,53 @@
+import { Input } from 'components/common/form';
+import { FieldArray, getIn, useFormikContext } from 'formik';
+import { DescriptionOfLand, LtsaOrders } from 'interfaces/ltsaModels';
+import * as React from 'react';
+import { withNameSpace } from 'utils/formUtils';
+
+import { SectionFieldWrapper } from '../SectionFieldWrapper';
+import { StyledSectionHeader } from '../SectionStyles';
+import LtsaLegalNotationsSubForm from './LtsaLegalNotationsSubForm';
+
+export interface ILtsaLandSubFormProps {
+  nameSpace?: string;
+}
+
+export const LtsaLandSubForm: React.FunctionComponent<ILtsaLandSubFormProps> = ({ nameSpace }) => {
+  const { values } = useFormikContext<LtsaOrders>();
+  const lands = getIn(values, withNameSpace(nameSpace, 'descriptionsOfLand')) ?? [];
+  return (
+    <>
+      <StyledSectionHeader>Land</StyledSectionHeader>
+      {lands.length === 0 ? (
+        'this title has no land'
+      ) : (
+        <>
+          <FieldArray
+            name={withNameSpace(nameSpace, 'descriptionsOfLand')}
+            render={({ name }) => (
+              <React.Fragment key={`land-row-${name}`}>
+                {lands.map((land: DescriptionOfLand, index: number) => {
+                  const innerNameSpace = withNameSpace(nameSpace, `descriptionsOfLand.${index}`);
+                  return (
+                    <>
+                      <SectionFieldWrapper label="PID">
+                        <Input field={`${withNameSpace(innerNameSpace, 'parcelIdentifier')}`} />
+                      </SectionFieldWrapper>
+                      <SectionFieldWrapper label="Legal Description">
+                        <p>{getIn(land, 'fullLegalDescription')}</p>
+                      </SectionFieldWrapper>
+                      {index < lands.length - 1 && <hr></hr>}
+                    </>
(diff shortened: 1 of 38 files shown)


FuriousLlama (approved):

frontend/src/features/mapSideBar/tabs/ltsa/LtsaTransferSubForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { FieldArray, getIn, useFormikContext } from 'formik';
+import { LtsaOrders, TitleTransferDisposition } from 'interfaces/ltsaModels';
+import * as React from 'react';
+import { withNameSpace } from 'utils/formUtils';
+
+import { SectionFieldWrapper } from '../SectionFieldWrapper';
+import { StyledSectionHeader } from '../SectionStyles';
+
+export interface ILtsaTransferSubFormProps {
+  nameSpace?: string;
```
FuriousLlama, line 11: can this be undefined?

frontend/src/features/mapSideBar/tabs/ltsa/LtsaLegalNotationsSubForm.tsx
```diff
@@ -0,0 +1,64 @@
+import { InlineInput } from 'components/common/form/styles';
+import { FieldArray, getIn, useFormikContext } from 'formik';
+import { DescriptionOfLand, LtsaOrders } from 'interfaces/ltsaModels';
+import * as React from 'react';
+import { Col, Row } from 'react-bootstrap';
+import { withNameSpace } from 'utils/formUtils';
+
+export interface ILtsaLegalNotationsSubFormProps {
+  nameSpace?: string;
```
FuriousLlama, line 9: This field does not seem to be taking undefined values. Is there a reason for leaving the type like that?

frontend/src/features/mapSideBar/tabs/ltsa/LtsaLandSubForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+      {lands.length === 0 ? (
+        'this title has no land'
+      ) : (
+        <>
+          <FieldArray
+            name={withNameSpace(nameSpace, 'descriptionsOfLand')}
+            render={({ name }) => (
+              <React.Fragment key={`land-row-${name}`}>
+                {lands.map((land: DescriptionOfLand, index: number) => {
+                  const innerNameSpace = withNameSpace(nameSpace, `descriptionsOfLand.${index}`);
```
FuriousLlama, line 30: would this cause React to complain about the key not being present?

frontend/src/features/mapSideBar/tabs/ltsa/LtsaTransferSubForm.tsx
devinleighsmith: yes

frontend/src/features/mapSideBar/tabs/ltsa/LtsaLegalNotationsSubForm.tsx
devinleighsmith: These are composable forms. If used in a different context they may require a different, or possibly undefined namespace.

frontend/src/features/mapSideBar/tabs/ltsa/LtsaLandSubForm.tsx
devinleighsmith: Ill double check

frontend/src/features/mapSideBar/tabs/ltsa/LtsaLegalNotationsSubForm.tsx
FuriousLlama: But are they meant to be common? given the file location and contents, my guess is that they specific for that. But I could be wrong. Same as the other places where the namespace is passed as possibly undefined

devinleighsmith: Even if they are not common, the order/hierarchy that they are used on this page could change.

devinleighsmith: Think of it this way, a subform has no way of knowing what the parent formik context looks like. As a result, the object that this form operates on could be:
1) at the root level
2) at some child level
3) part of an array
4) etc.

Passing in the an optional namespace reflects the above uncertainty.

FuriousLlama: I see your point, adding that gives some flexibility about how the component could be used in the future. What I am saying is that I have doubts about its future usage, I've seen components that are not reused at all. I prefer the "code for the current intent, not future one" if there is no clear idea of how that would be used in the future. Anyhow, this is a minor thing.

asanchezr (approved):

PR #1110, opened 2022-03-16
Psp 3042

devinleighsmith:
correct issue with mismatched navigation properties causing: 
NotSupportedException: Collection was of a fixed size. at System.SZArrayHelper.Remove[T](T value)


FuriousLlama (approved):

FuriousLlama:
Do we have a techdebt story to separate the different subsections sections vs loading the entire lease tree again?

PR #1111, opened 2022-03-17
psp-3042 correct issue with lease max page size.

devinleighsmith:
Handle issue with extremely large leases.


backend/dal/Repositories/LeaseRepository.cs
```diff
@@ -68,7 +68,7 @@ public PimsLease Get(long id)
         {
             this.User.ThrowIfNotAuthorized(Permissions.LeaseView);
 
-            PimsLease lease = this.Context.PimsLeases
+            PimsLease lease = this.Context.PimsLeases.AsSplitQuery()
```
FuriousLlama, line 71: now this is interesting. Good find!

FuriousLlama (approved):

PR #1128, opened 2022-03-21
Katalon scripts for linking address, and new marker popout

jdtoombs:
- Updated scripts to accommodate for new slide out, updates to be made when v2 comes through
- Script for linking organization mailing address to an individual


jdtoombs:
@devinleighsmith 
Yup this has been ready, just need a review! Updates will come in the next PR. Will update branch right now

PR #1143, opened 2022-03-26
PSP-2376 Property Information slideout

asanchezr:
UNIT TESTS to come

image


frontend/package.json
```diff
@@ -18,7 +19,7 @@
     "lodash": "4.17.21",
     "moment": "2.29.1",
     "moment-timezone": "0.5.33",
-    "multiselect-react-dropdown": "2.0.7",
+    "multiselect-react-dropdown": "2.0.21",
```
asanchezr, line 22: The updated version supports displaying multi-selects in read-only/disabled mode

frontend/package.json
```diff
@@ -9,6 +9,7 @@
     "axios": "0.23.0",
     "bootstrap": "4.6.0",
     "classnames": "2.3.1",
+    "convert": "4.4.2",
```
asanchezr, line 12: extra dependency to convert between various units of measure - e.g. meters -> acres etc

frontend/src/components/common/form/Text.tsx
```diff
@@ -0,0 +1,23 @@
...
+  // if no field is supplied, then we render the children as-is
+  const { values } = useFormikContext();
+  const value = field ? getIn(values, field) : children;
+
+  return <StyledSpan className={className}>{value}</StyledSpan>;
+};
+
+const StyledSpan = styled.span`
+  color: ${props => props.theme.css.formControlTextColor};
+`;
```
asanchezr, line 23: New component to support read-only forms. Avoids having to re-skin inputs etc without borders.

frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
```diff
@@ -9,6 +9,15 @@ export const ELECTORAL_LAYER_URL =
...
 
+export const MOTI_REGION_LAYER_URL =
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_ADMIN_BOUNDARIES.TADM_MOT_REGIONAL_BNDRY_POLY/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_ADMIN_BOUNDARIES.TADM_MOT_REGIONAL_BNDRY_POLY';
+export const HWY_DISTRICT_LAYER_URL =
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_ADMIN_BOUNDARIES.TADM_MOT_DISTRICT_BNDRY_POLY/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_ADMIN_BOUNDARIES.TADM_MOT_DISTRICT_BNDRY_POLY';
+export const ALR_LAYER_URL =
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_LEGAL_ADMIN_BOUNDARIES.OATS_ALR_BOUNDARY_LINES_SVW/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_LEGAL_ADMIN_BOUNDARIES.OATS_ALR_BOUNDARY_LINES_SVW';
+export const INDIAN_RESERVES_LAYER_URL =
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_ADMIN_BOUNDARIES.ADM_INDIAN_RESERVES_BANDS_SP/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_ADMIN_BOUNDARIES.ADM_INDIAN_RESERVES_BANDS_SP';
+
```
asanchezr, line 20: Additional DataBC map layers that we query in the new property information slideout

frontend/src/components/common/form/Text.tsx
asanchezr: Disabled only apply to input fields. This component renders text as a span

backend/api/Areas/Property/Models/Property/PropertyModel.cs
```diff
@@ -1,12 +1,14 @@
...
+using Pims.Api.Models;
+using Pims.Api.Models.Concepts;
 
 namespace Pims.Api.Areas.Property.Models.Property
 {
     /// <summary>
     /// PropertyModel class, provides a model to represent the property whether Land or Building.
     /// </summary>
-    public class PropertyModel
+    public class PropertyModel : BaseAppModel
```
FuriousLlama, line 11: nit: this one adds the app changes stamps. I dont think we used those. Insead just implementing the interface 
 BaseModel would suffice

asanchezr: I can change but see no harm in exposing that.

backend/entities/PropertyRoadType.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// get/set - Primary key to identify adjacent land type.
+        /// </summary>
+        [NotMapped]
+        public string Id { get => PropertyRoadTypeCode; set => PropertyRoadTypeCode = value; }
+        #endregion
+
+        #region Constructors
+        /// <summary>
+        /// Create a new instance of a PimsPropertyAdjacentLandType class.
```
devinleighsmith, line 19: copy pasta

backend/entities/VolumetricType.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsVolumetricType class, provides an entity for the datamodel to manage a list of volumetric types.
+    /// </summary>
+    public partial class PimsVolumetricType : ITypeEntity<string>
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - Primary key to identify property tenure type.
```
devinleighsmith, line 11: nit

frontend/src/features/mapSideBar/tabs/propertyDetails/PropertyDetailsTabView.helpers.tsx
```diff
@@ -0,0 +1,210 @@
...
+    opacity: 1,
+    background: '#F2F2F2',
+    borderRadius: '4px',
+    color: 'black',
+    fontSize: '16px',
+    marginRight: '1em',
+  },
+};
+
+export const defaultPropertyInfo: Api_Property = {
```
devinleighsmith, line 118: Mild preference for leaving fields blank rather then filling them in with fake values during loading.

frontend/src/features/mapSideBar/tabs/propertyDetails/components/LandMeasurementTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+const columns: ColumnWithProps<any>[] = [
+  {
+    accessor: 'value',
+    align: 'right',
+    Cell: ({ value }: CellProps<any, number>) => formatNumber(value, 0, 2),
+  },
+  { accessor: 'unit', align: 'left' },
+];
+
+const StyledTable = styled(Table)`
```
devinleighsmith, line 33: is this something that should be made common?

frontend/src/features/mapSideBar/tabs/propertyDetails/usePropertyDetails.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+      // query BC Geographic Warehouse layers
+      const motiRegion = await findByLocation(motiRegionService, location, 'GEOMETRY');
+      const highwaysDistrict = await findByLocation(highwaysDistrictService, location, 'GEOMETRY');
+      const electoralDistrict = await findByLocation(electoralService, location);
+      const alr = await alrService.findOneWhereContains(location, 'GEOMETRY');
+      const firstNations = await findByLocation(firstNationsService, location, 'GEOMETRY');
+
+      if (isMounted()) {
+        setPropertyDetails(prevState => {
```
devinleighsmith, line 65: it looks like this is intended to run AFTER the property info has loaded, but I'm not seeing any mechanism to enforce this. If our API returns after the calls to the various layers, won't that response overwrite these responses?

frontend/src/features/mapSideBar/tabs/propertyDetails/usePropertyDetails.ts
```diff
@@ -0,0 +1,105 @@
...
+    fn();
+  }, [getPropertyWithPid, isMounted, pid]);
+
+  useEffect(() => {
+    async function fn() {
+      if (location === undefined) {
+        return;
+      }
+      // query BC Geographic Warehouse layers
+      const motiRegion = await findByLocation(motiRegionService, location, 'GEOMETRY');
```
devinleighsmith, line 58: are there any requirements on error handling for these?

frontend/src/hooks/useProperties.ts
```diff
@@ -91,6 +94,47 @@ export const useProperties = () => {
...
+   * Make an AJAX request to fetch the specified 'property' from inventory.
+   * @param params PID of the property
+   */
+  const fetchPropertyWithPid = useCallback(
+    async (pid: string): Promise<Api_Property> => {
+      dispatch(logRequest(actionTypes.GET_PARCEL_DETAIL));
+      dispatch(showLoading());
+      // Due to spatial information being stored in BC Albers in the database, we need to make TWO requests here:
+      //   1. to the REST API to fetch property field attributes (e.g. address, etc)
+      //   2. to GeoServer to fetch latitude/longitude in expected web mercator projection (EPSG:4326)
```
devinleighsmith, line 107: I'm no expert, but can't leaflet handle projections as well? that way we can save a request.

devinleighsmith: ie something like: L.Projection.Mercator.project(latlng);

devinleighsmith (approved):

frontend/src/features/mapSideBar/tabs/propertyDetails/PropertyDetailsTabView.helpers.tsx
FuriousLlama: Agree, especially since the values are all undefinable

frontend/src/models/api/Property.ts
```diff
@@ -0,0 +1,42 @@
...
+  // multi-selects
+  anomalies?: Api_TypeCode<string>[];
+  tenure: Api_TypeCode<string>[];
+  roadType?: Api_TypeCode<string>[];
+  adjacentLand?: Api_TypeCode<string>[];
+
+  dataSource?: Api_TypeCode<string>;
+  dataSourceEffectiveDate?: string;
+
+  address: Api_Address;
```
FuriousLlama, line 41: address  could be undefined

frontend/src/models/api/Property.ts
```diff
@@ -0,0 +1,42 @@
+import { Api_Address } from './Address';
+import { Api_ConcurrentVersion } from './ConcurrentVersion';
+import Api_TypeCode from './TypeCode';
+
+export interface Api_Property extends Api_ConcurrentVersion {
```
FuriousLlama, line 5: I could not find a new PropertyModel on the Concepts namespace. Is this mapping to the old one or one of the feature-specific ones? If so, I believe it would be better if we have a property model on the front end that is specific to that one until we get the general one on the backend

frontend/src/models/api/Property.ts
```diff
@@ -0,0 +1,42 @@
...
+  id: number;
```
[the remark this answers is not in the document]
FuriousLlama: Similar to other values below. Since we don't promise non-nullability on the backend API, this should not expect it.

frontend/src/features/mapSideBar/tabs/propertyDetails/components/LandMeasurementTable.tsx
asanchezr: could/should be but this is tailored to the land measurements data so didn't see need for a generic table. Can be refactored at a later date if needed

frontend/src/features/mapSideBar/tabs/propertyDetails/usePropertyDetails.ts
asanchezr: the underlying custom axios that queries map layers handles errors. The story does not call for specific error action (that I can see)

frontend/src/features/mapSideBar/tabs/propertyDetails/usePropertyDetails.ts
asanchezr: The logic might not be clear upfront but this what happens that guarantees the order (at least in my mind, happy to hear otherwise):

1. location is a memo that only gets re-calculated when the property info comes back from the API with proper lat, long values. So until API results comes back it will be `undefined`
2. The second `useEffect` aborts early if location is undefined so it won't call the layers until a location is available (because that is needed to query the layers). So effectively if the API takes longer to return then the second effect will return early and re-run whenever the API response is set via setState on propertyDetails

Hopefully that logic was somewhat ok to follow. If deemed to complicated a comment like above might help other devs figure out that code

frontend/src/hooks/useProperties.ts
asanchezr: I'd have to check. I was under the impression that leaflet wouldn't handle the specific projection in our DB (BC Albers) as it is not one of the popular ones. There indeed are client-side projection libraries (turf-js) but we haven't included that in our project (yet). We might look at refactoring this code (and other places where we do the same) via a local client-side re-projection library. 

It would be outside of this PR though

asanchezr: sorry - meant https://github.com/proj4js/proj4js
turf-js is more for shape manipulation and measuring client side

frontend/src/models/api/Property.ts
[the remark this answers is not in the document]
asanchezr: For properties we currently do not support CREATE, only READ and UPDATE as they are brought into the system via ETL. So that's why I left the id as non-null as it is guaranteed (for now). 

I will turn it into nullable for consistency but note that we don't plan on creating properties for now so it will always come back set to a value

frontend/src/models/api/Property.ts
asanchezr: PropertyModel was already existing on it's own Area and other controllers/logic depend on it so I felt that refactoring it to a common "concept" was not part of this PR. Can be addresses later I think

frontend/src/models/api/Property.ts
asanchezr: image

ADDRESS foreign key is required (not null) on the database so there won't be an instance of a property without address

asanchezr: Unless DB schema changes, then yes we can modify the frontend to match (if/when it happens). I think it makes sense as all properties imported into the system need a valid address

frontend/src/models/api/Property.ts
FuriousLlama: Yeah I agree. This model, however, is a new one that should be common (Api_Property). If we are using the old model on the backend it should map to the old model on the front end.

frontend/src/models/api/Property.ts
FuriousLlama: From the DB point of view yes. But from the API point of view everything is nullable

frontend/src/components/maps/leaflet/LayerPopup/hooks/useLayerQuery.tsx
```diff
@@ -12,7 +12,10 @@ export interface IUserLayerQuery {
    * function to find GeoJSON shape containing a point (x, y)
    * @param latlng = {lat, lng}
    */
-  findOneWhereContains: (latlng: LatLngLiteral) => Promise<FeatureCollection>;
+  findOneWhereContains: (
```
FuriousLlama, line 15: nit the docs dont match the function params

frontend/src/features/mapSideBar/tabs/propertyDetails/usePropertyDetails.ts
```diff
@@ -0,0 +1,116 @@
...
+} from 'components/maps/leaflet/LayerPopup';
+import { useProperties } from 'hooks';
+import useIsMounted from 'hooks/useIsMounted';
+import { LatLngLiteral } from 'leaflet';
+import { useEffect, useMemo, useState } from 'react';
+import { pidFormatter } from 'utils';
+
+import { IPropertyDetailsForm, toFormValues } from './PropertyDetailsTabView.helpers';
+
+export function usePropertyDetails(pid?: string) {
```
FuriousLlama, line 18: I was hoping this method would be general enough to just return a Property from the api. Would it be possible to make this file return that instead? I can refactor that if necessary

asanchezr: I'm merging right now. You can refactor if needed

asanchezr: and it does return a property from the api - not sure I follow

asanchezr: if you just need an API_Property instance then use `const { getPropertyWithPid } = useProperties();`

PR #1144, opened 2022-03-29
Psp 3040

devinleighsmith:
Correct display of currency input (was displaying as numeric only)

### DIFF
--- a/frontend/src/features/leases/detail/LeasePages/insurance/edit/EditContainer.test.tsx
+++ b/frontend/src/features/leases/detail/LeasePages/insurance/edit/EditContainer.test.tsx
@@ -111,9 +111,10 @@ describe('Lease Insurance', () => {
     const forms = result.getAllByTestId('insurance-form');
     expect(forms.length).toBe(2);
 
-    const formLimits = result.getAllByTestId('insurance-form-limit');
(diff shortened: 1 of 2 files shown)


devinleighsmith:
confirmed that this is the only instance of this issue in the application at this time.

frontend/src/features/leases/detail/LeasePages/insurance/edit/InsuranceForm.tsx
```diff
@@ -59,10 +59,9 @@ const InsuranceForm: React.FunctionComponent<IInsuranceRowProps> = ({ nameSpace
           <div>
             <strong>Limit ($)</strong>
           </div>
-          <InlineInput
-            type="number"
+          <FastCurrencyInput
+            formikProps={formikProps}
             field={withNameSpace(nameSpace, 'coverageLimit')}
-            data-testid="insurance-form-limit"
```
FuriousLlama, line 65: Without this line the the test will fail
``` const formLimits = result.getAllByTestId('insurance-form-limit');```

devinleighsmith: test failure corrected.
Can't use testid as it is not supported by the masked input.

FuriousLlama (approved):

PR #1153, opened 2022-03-29
Psp 2907

devinleighsmith:
Add the ability to select multiple properties by clicking on the map.


frontend/src/components/common/PopupWindow.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /** provide the size of the modal, default width is 50.0rem */
+  modalSize?: ModalSize;
+  className?: string;
+}
+
+/**
+ * Generic Component used to display modal popups to the user.
+ * @param props customize the component with custom text, and an operation to take when the component is closed.
+ */
+const PopupWindow = (props: BsModalProps) => {
```
FuriousLlama, line 80: this looks very similar to the generic modal implementation. Can we make the Modal more extensible or get similar functionality (like checking the handling on cancel/ok)? I am concerned that we might need to mirror fixes on both files since they look so similar

devinleighsmith: thats fair, I'll see if I can make this happen via a prop instead.

FuriousLlama (approved):

PR #1156, opened 2022-03-30
Property information header - Features/psp 2375

FuriousLlama:
image


frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -238,8 +238,27 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+              provinceId: propertyAddress?.province?.id || 0,
+              streetAddress1: propertyAddress?.streetAddress1 || '',
+              streetAddress2: propertyAddress?.streetAddress2,
+              streetAddress3: propertyAddress?.streetAddress3,
+              municipality: propertyAddress?.municipality || '',
+              postal: propertyAddress?.postal || '',
+            },
+            areaUnit: propertyData.areaUnit?.id,
+            landArea: propertyData.landArea || 0,
+            landLegalDescription: propertyData.landLegalDescription || '',
```
FuriousLlama, line 261: The property information coming back from the backend does not match the models we use on the front end (IProperty). @asanchezr @devinleighsmith am not too familiar with the leaflet code; can you take a look to see if this mapping is sufficient?

frontend/src/features/mapSideBar/tabs/ltsa/LtsaChargeOwnerSubForm.tsx
```diff
@@ -9,7 +9,7 @@ import * as React from 'react';
 import { Col, Row } from 'react-bootstrap';
 import { withNameSpace } from 'utils/formUtils';
 
-import { SectionFieldWrapper } from '../SectionFieldWrapper';
+import { SectionField } from '../SectionField';
```
FuriousLlama, line 12: These two where pretty much the same

frontend/src/interfaces/IPropertyApiModel.ts
```diff
@@ -1,42 +1,51 @@
...
+import { Api_ConcurrentVersion } from 'models/api/ConcurrentVersion';
+import Api_TypeCode from 'models/api/TypeCode';
 
-export interface Api_Property extends Api_ConcurrentVersion {
+import { ILease } from './ILease';
+
+/**
+ * Pairs with the current implementation of Pims.Api.Areas.Property.Models.Property.PropertyModel
+ */
+export interface IPropertyApiModel extends Api_ConcurrentVersion {
```
FuriousLlama, line 10: Aligned the model with the model that is coming back from the backend.

frontend/src/components/maps/leaflet/PointClusterer.tsx
asanchezr: I think this code should be using API_Property but you could double check what is returned from the `PropertyController.GetProperty(int id)`

asanchezr: IProperty was the old one from CITZ pims, I believe the one I created in the frontend `API_Property` matches the backend `PropertyModel` under the Property area

FuriousLlama: The reason why I believe it does not belong with the shared API_* models is that the model coming from the backend is not common and does not match our current standards yet (I think there are some fields that are not matching the DB). Once we get a model on the same level as the new Address/person/org we can match it with an API_Property. That distinction helps with the refactor.

frontend/src/hooks/useProperties.ts
```diff
@@ -55,51 +50,12 @@ export const useProperties = () => {
...
-   * @param params unique id of the property
-   * @param position optional override for the lat/lng of the returned property.
-   */
-  const fetchProperty = useCallback(
-    async (id: number, position?: [number, number]): Promise<IProperty> => {
-      dispatch(logRequest(actionTypes.GET_PARCEL_DETAIL));
-      dispatch(showLoading());
-      // Due to spatial information being stored in BC Albers in the database, we need to make TWO requests here:
-      //   1. to the REST API to fetch property field attributes (e.g. address, etc)
-      //   2. to GeoServer to fetch latitude/longitude in expected web mercator projection (EPSG:4326)
```
asanchezr, line 69: Did you check that this code is not user anywhere please? I recall putting this comment here recently to fix a bug in PSP...

frontend/src/interfaces/IPropertyApiModel.ts
asanchezr: I like your refactoring and personally have no preference on what we call the frontend models but didn't we agree to start naming things API_* ? happy to hear your thoughts

asanchezr (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
asanchezr: Fair points. We do need to do some cleaning on the various models in the backend. I agree we should keep this separate from API_* until refactor happens .

frontend/src/interfaces/IPropertyApiModel.ts
asanchezr: NVM, your comment on another file answered why we are not using API_Property here

frontend/src/hooks/useProperties.ts
FuriousLlama: Yeah, removing this function did not cascade into any of the front end breaking

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: Yes, we agreed to do a progressive refactor, so we shouldn't be expecting anything other then person and possibly some lease models to match our new standards. We are explicitly scheduling that work as tech debt, so refactoring property will wait until we create/assign that story.

PR #1173, opened 2022-04-01
IS24 Up/Down Alters Edited

dfilteau:
Changes made to provide idempotent behaviour


devinleighsmith:
@dfilteau getting the following error:

Msg 2111, Level 16, State 1, Server 0eca209b131d, Procedure PIMS_STAVBL_I_S_U_TR, Line 1
Cannot create trigger 'PIMS_STAVBL_I_S_U_TR' on table 'PIMS_STATIC_VARIABLE' because an INSTEAD OF UPDATE trigger already exists on this object.
======= SCRIPT PSP_PIMS_S24_00/Alter Up/01_PSP_PIMS_S23_01_Alter_Up.sql RETURNS AN ERROR. =========
make: *** [Makefile:265: db-deploy] Error 1

dfilteau:
@devinleighsmith, fix applied.

.gitignore
```diff
@@ -14,4 +14,86 @@ auth/keycloak/config/hsperfdata_jboss/
 auth/keycloak/config/*.log
 trufflehog_report.*
 datastore.xml
-**/.vscode/
\ No newline at end of file
+**/.vscode/
+*.bak
+database/mssql/scripts/dbscripts/PSP_PIMS_S23_01/._Alter Down
```
devinleighsmith, line 19: @dfilteau can you explain what is going on here? also, If you need to ignore files I'd suggest using wildcards instead of creating one row per file.

database/mssql/scripts/dbscripts/PSP_PIMS_S24_00/Alter Up/25_PIMS_CONTACT_MGR_VW_Fix.SQL
```diff
@@ -0,0 +1,90 @@
+/* Drop and recreate the PIMS_CONTACT_MGR_VW view to include the subselect in
```
devinleighsmith, line 1: This file seems the same in the alter up and the alter down, may not be necessary?

database/mssql/scripts/dbscripts/PSP_PIMS_S24_00/Build/04_PSP_PIMS_PMBC_Tables_Build.sql
```diff
@@ -0,0 +1,294 @@
...
+               FROM   sys.schemas 
+               WHERE  name = N'pmbc')
+  EXEC('CREATE SCHEMA [pmbc]');
+GO
+IF @@ERROR <> 0 SET NOEXEC ON
+GO
+
+
+-- Drop table pmbc.PMBC_PARCEL_POLYGON_FABRIC
+PRINT N'Drop table pmbc.PMBC_PARCEL_POLYGON_FABRIC'
```
devinleighsmith, line 29: @dfilteau can we change this to not drop the pmbc table, just only create it if it doesn't exist? This file is the reason that the pmbc table was empty in all of our environments.

.gitignore
devinleighsmith: Pr looks good besides this.

devinleighsmith (changes requested):
would like the gitignore cleaned up first.

devinleighsmith (approved):

PR #1178, opened 2022-04-01
IS24 scaffold

devinleighsmith:


backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
```diff
@@ -22,8 +23,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.DataSourceEffectiveDate, src => src.PropertyDataSourceEffectiveDate)
                 .Map(dest => dest.ClassificationId, src => src.PropertyClassificationTypeCode)
                 .Map(dest => dest.Classification, src => src.PropertyClassificationTypeCodeNavigation.Description)
-                .Map(dest => dest.TenureId, src => src.PropertyTenureTypeCode)
-                .Map(dest => dest.Tenure, src => src.PropertyTenureTypeCodeNavigation.Description)
+                .Map(dest => dest.TenureId, src => src.PimsPropPropTenureTypes.FirstOrDefault())
```
devinleighsmith, line 26: @asanchezr please confirm, did not want to change the format of models.

asanchezr: I did not work on this area (property search) so not sure what to do here. concatenating all values into one string would seem reasonable. This is what would get displayed in the property list view - right?

asanchezr (approved):

backend/api/Areas/Property/Mapping/Search/PropertyMap.cs
FuriousLlama: We should work on the standardization of the Property model before we move on and tie our front end more

FuriousLlama (approved):

PR #1196, opened 2022-04-04
PSP-3188 Fix property information slide-out

asanchezr:
Map layers are now properly queried after API returns property info object

### DIFF
--- a/frontend/src/features/mapSideBar/tabs/ltsa/LtsaTabView.tsx
+++ b/frontend/src/features/mapSideBar/tabs/ltsa/LtsaTabView.tsx
@@ -1,4 +1,5 @@
 import { Input, TextArea } from 'components/common/form';
+import LoadingBackdrop from 'components/maps/leaflet/LoadingBackdrop/LoadingBackdrop';
 import { Form, Formik, getIn } from 'formik';
 import { LtsaOrders, OrderParent, ParcelInfo, TaxAuthority } from 'interfaces/ltsaModels';
 import { noop } from 'lodash';
@@ -19,8 +20,11 @@ export interface ILtsaTabViewProps {
 
 export const LtsaTabView: React.FunctionComponent<ILtsaTabViewProps> = ({ ltsaData }) => {
   const titleNameSpace = 'titleOrders.0.orderedProduct.fieldedData';
+  const isLoading = ltsaData === undefined;
+
   return (
     <StyledScrollable>
+      <LoadingBackdrop show={isLoading} parentScreen={true} />
       <Formik initialValues={ltsaData ?? defaultLtsaData} onSubmit={noop} enableReinitialize={true}>
         <StyledForm>
           <StyledFormSection>
--- a/frontend/src/features/mapSideBar/MapSideBarContainer.test.tsx
+++ b/frontend/src/features/mapSideBar/MapSideBarContainer.test.tsx
@@ -19,14 +19,6 @@ describe('MapSideBarContainer component', () => {
     };
   };
 
-  it('renders a spinner when the ltsa data is loading', () => {
-    const {
-      component: { getByTestId },
-    } = setup();
-
-    const spinner = getByTestId('filter-backdrop-loading');
-    expect(spinner).toBeVisible();
-  });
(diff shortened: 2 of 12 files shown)


frontend/src/features/mapSideBar/tabs/ltsa/LtsaTabView.tsx
```diff
@@ -19,8 +20,11 @@ export interface ILtsaTabViewProps {
 
 export const LtsaTabView: React.FunctionComponent<ILtsaTabViewProps> = ({ ltsaData }) => {
   const titleNameSpace = 'titleOrders.0.orderedProduct.fieldedData';
+  const isLoading = ltsaData === undefined;
+
   return (
     <StyledScrollable>
+      <LoadingBackdrop show={isLoading} parentScreen={true} />
```
FuriousLlama, line 27: why split the loading indicator to each section instead of the entire pane?

FuriousLlama: Note that the header also would need the indicator too.

asanchezr: What I noticed was that if the LTSA info failed to load but the property information loaded ok the spinner would remain. I recall Devin put the loading spinner on his tab originally. Perhaps having multiple spinners might be a bit better than a single one?

asanchezr: We can discuss in chat instead of here if preferred. In my opinion when a page loads data from multiple sources (where some sources can fail while others return data) I'd rather confine the spinner to the area that uses each data source instead of one big spinner that requires all sources to succeed. What do you think about this @FuriousLlama and @devinleighsmith ? Open to ideas

asanchezr: In any case, this fixes a bug. We can deal with the spinner issue in a another PR (in my opinion). We might need to chat about spinner usage in general when we have different data sources in the same page/form. Thoughts?

FuriousLlama: Either looks fine to me. But if we are going to have spinners for each component, then the header needs one too.

asanchezr: ok, will add to header and we can discuss general strategy moving forward this Friday

FuriousLlama: On the same line of thought; Checking for undefined as the indicator for Loading data is an anti-pattern since the data could be undefined if no results come back or if there is an error. Instead, the loading indicator should be the composition of all sub-components being loaded/fetched. I believe that the IsLoading state can be added/used when using Axios to mark the state of the retrieval and use that instead.

asanchezr: Agree with you on that but I feel that perhaps this is part of a larger conversation? I'd prefer this PR to focus on fixing the bug instead of implementing "loading" behaviour. we use the spinner on other places in the app and they check for undefined. I rather discuss a spinner/loading strategy separately. Would you agree to that?

asanchezr: For example, as part of that general discussion I like hooks like https://github.com/simoneb/axios-hooks that return the loading state AND the error state back to consumers so they know whether to render a spinner or an error message

asanchezr: Something like:

```tsx
import { useAxios } from 'use-axios-client';
export default () => {
  const { data, error, loading } = useAxios({
    url: 'https://example/api',
  });
  return (
    <>
      {loading && <div>Loading...</div>}
      {error && <div>{error.message}</div>}
      {data && <div>{data}</div>}
    </>
  );
};
```

FuriousLlama (approved):

PR #1200, opened 2022-04-05
Scaffold for s25

FuriousLlama:


devinleighsmith:
@FuriousLlama noticed that the test data for s25 doesn't have geospatial attributes. Can you replace this file from S25:
05_DML_LEASE_PROPERTY.sql
with the file of the same name from S24 please?

devinleighsmith (approved):

PR #1205, opened 2022-04-07
Create research | Features/psp 2874

FuriousLlama:


backend/.editorconfig
```diff
@@ -84,6 +84,30 @@ dotnet_diagnostic.S4136.severity = none
 # S1128 Remove this unnecessary 'using'
 dotnet_diagnostic.S1128.severity = warning
 
+# StyleCop
+# SA1600 Elements should be documented
+dotnet_diagnostic.SA1600.severity = none
```
FuriousLlama, line 89: Removes warnings that we are not following/dont care atm

backend/dal/Repositories/PropertyService.cs
```diff
@@ -100,26 +100,27 @@ public PimsProperty Get(int id)
                 .Include(p => p.PropertyDataSourceTypeCodeNavigation)
                 .Include(p => p.PropertyClassificationTypeCodeNavigation)
                 .Include(p => p.PimsPropPropTenureTypes)
+                    .ThenInclude(t => t.PropertyTenureTypeCodeNavigation)
```
FuriousLlama, line 103: missing navigation include. Was causing an exception if there was a tenure on the property

backend/api/Models/Concepts/ResearchFile/PropertyModel.cs
```diff
@@ -0,0 +1,13 @@
+namespace Pims.Api.Models.Concepts
+{
+    // TODO: should this be refactored?
+    public class PropertyModel : BaseModel
```
devinleighsmith, line 4: yeah, I'd assume that this would be a general property concept and not one specific to research (otherwise we are back to creating multiple models for every entity again). I'm doing this in my PR that is coming soon.

FuriousLlama: That is what I thought too. For this PR I decided to leave most of the property model empty and only have what we need. In the future we should add to this model and move away from the other one.

backend/dal/Helpers/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -57,11 +58,12 @@ public static IServiceCollection AddPimsDalServices(this IServiceCollection serv
             services.AddScoped<ILeaseService, LeaseService>();
             services.AddScoped<ILeaseReportsService, LeaseReportsService>();
             services.AddScoped<ILeaseTermService, LeaseTermService>();
-            services.AddScoped<Services.ILeasePaymentService, Services.LeasePaymentService>();
```
devinleighsmith, line 60: yes, I noticed this too.

frontend/package.json
```diff
@@ -1,9 +1,11 @@
 {
   "name": "frontend",
-  "version": "1.1.0-25.10",
+  "version": "1.1.0-25.11",
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "1.0.1",
+    "@mdi/js": "^6.6.95",
```
devinleighsmith, line 7: Hmm, we want to be careful about this one I think. For one react-icons (should) contain the material design icons
https://react-icons.github.io/react-icons/icons?name=md
Icon libraries can be a bit on the big side so I wouldn't want us to bring this in unless:
1. the library is small **or** 
2. only adds icons that we are actively using to the build.
3. **and** the icons being added aren't part of react-icons.

frontend/src/components/layout/SideNavBar/SideNavbar.test.tsx
```diff
@@ -17,7 +16,7 @@ const renderComponent = (props?: IRenderProps) =>
     <TestCommonWrapper
       history={history}
       roles={props?.roles ?? [Roles.REAL_ESTATE_MANAGER]}
-      claims={[Claims.LEASE_VIEW]}
+      claims={[Claims.LEASE_VIEW, Claims.RESEARCH_VIEW]}
```
devinleighsmith, line 19: hmm, is this necessary? does the user have the ability to view research files on the sidenavbar?

frontend/src/features/mapSideBar/MapSideBarContainer.tsx
```diff
@@ -19,14 +19,32 @@ import { PropertyDetailsTabView } from './tabs/propertyDetails/PropertyDetailsTa
  * container responsible for logic related to map sidebar display. Synchronizes the state of the parcel detail forms with the corresponding query parameters (push/pull).
  */
 export const MotiInventoryContainer: React.FunctionComponent = () => {
+  const isMounted = useIsMounted();
```
devinleighsmith, line 22: Hey @FuriousLlama it seems like your PR may be pulling in unrelated commits? I could be wrong of course but I don't think this commit is part of the add research file logic? Actually didn't Alejandro originally commit this?

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { useAddResearch } from '../hooks/useAddResearch';
+import { AddResearchFileYupSchema } from './AddResearchFileYupSchema';
+import { ResearchForm } from './models';
+import ResearchProperties from './ResearchProperties';
+
+const AddResearchForm: React.FunctionComponent = () => {
+  const initialForm = new ResearchForm();
+  const history = useHistory();
+
+  const { addResearchFile } = useAddResearch();
```
devinleighsmith, line 21: I'm guessing tests are coming with the add properties logic?

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const submitForm = async (researchFile: Api_ResearchFile) => {
+    const response = await addResearchFile(researchFile);
+    if (!!response?.name) {
+      history.push(`/research/${response?.name}`);
+    }
+  };
+
+  const onCancel = () => {
+    history.push(`/`);
```
devinleighsmith, line 31: I think this takes you back to the mapview, is that what is being asked for here?

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  formikProps.setSubmitting(true);
+                  formikProps.submitForm();
+                }}
+              >
+                Save
+              </Button>
+            </Col>
+          </Row>
+
+          <Prompt
```
devinleighsmith, line 83: Do you think we should create a component to wrap this or is this terse enough as is?

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const onCancel = () => {
+    history.push(`/`);
+  };
+  return (
+    <Formik<ResearchForm>
+      initialValues={initialForm}
+      onSubmit={async (values: ResearchForm, formikHelpers) => {
+        const researchFile: Api_ResearchFile = values.toApi();
+        formikHelpers.setSubmitting(false);
```
devinleighsmith, line 38: This should possibly be after the submitForm call, and you may want to await the result of that call.

frontend/src/features/research/add/ResearchProperties.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { PropertyForm } from './models';
+
+export interface IResearchPropertiesProps {
+  properties: PropertyForm[];
+  namespace: string;
+  onRemove: (id: string) => void;
+}
+
+const ResearchProperties: React.FunctionComponent<IResearchPropertiesProps> = props => {
```
devinleighsmith, line 15: Sorry, I think we had some overlap here.

frontend/src/features/research/add/models.ts
```diff
@@ -0,0 +1,40 @@
...
+  public pid: string;
+  public description?: string;
+
+  constructor(id: string, pid: string) {
+    this.id = id;
+    this.pid = pid;
+  }
+
+  public toApi(): Api_Property {
+    // TODO: description
```
devinleighsmith, line 37: a description for the pid?

frontend/src/features/research/context/LeaseContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { noop } from 'lodash';
+import * as React from 'react';
+import { useState } from 'react';
+
+export interface ILeaseState {
+  lease?: ILease;
+  setLease: (lease: ILease) => void;
+}
+
+export const LeaseStateContext = React.createContext<ILeaseState>({
```
devinleighsmith, line 11: Hmm, I'm a bit lost here, why does the research logic need the lease context?

frontend/src/features/research/hooks/useAddResearch.ts
```diff
@@ -0,0 +1,46 @@
+import axios, { AxiosError } from 'axios';
```
devinleighsmith, line 1: This seems odd, github isn't applying the .ts markup properly here.

frontend/src/features/research/hooks/useUpdateLease.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  <Provider store={store}>{children}</Provider>
+);
+
+const setup = (values?: any) => {
+  const { result } = renderHook(useUpdateLease, { wrapper: getWrapper(getStore(values)) });
+  return result.current;
+};
+const defaultLeaseWithId = { ...defaultLease, id: 1 };
+
+describe('useUpdateLease functions', () => {
```
devinleighsmith, line 46: I feel like something inadvertent happened here maybe? this is another lease specific file that is now in the features/research area. Maybe I'm not understanding something.

devinleighsmith (changes requested):
requesting changes re: the lease files that were moved to features/research

frontend/package.json
FuriousLlama: The reason I added this one is because that is the icon on the mockups. I think there is a version of it as an svg. But maybe we want to stick with some icon packs.

frontend/src/components/layout/SideNavBar/SideNavbar.test.tsx
FuriousLlama: I think it does. The visibility of the link on the side bar is controlled by the claims

frontend/src/features/mapSideBar/MapSideBarContainer.tsx
FuriousLlama: Thanks for bringing it up. I think is an issue with git not refreshing the diff if a rebase is done

frontend/src/features/research/add/AddResearchForm.tsx
FuriousLlama: Yes, at this point this does not add properties

frontend/src/features/research/add/AddResearchForm.tsx
FuriousLlama: Yep:
```
On Cancel:

- A confirmation message appears to prompt the user that data will not be saved if they cancel now.
- No properties are highlighted .
- User returns to full screen map (research panel closes)

```

frontend/src/features/research/add/AddResearchForm.tsx
FuriousLlama: Good to know. I think I copied this from another onSubmit

frontend/src/features/research/add/AddResearchForm.tsx
FuriousLlama: I think is good enough now. the params are flexible enough and is not much boilerplate

frontend/src/features/research/add/models.ts
FuriousLlama: Right, since this PR does not have anything with the properties I will leave this as is. Will get fixed in the next PR

frontend/src/features/research/context/LeaseContext.tsx
FuriousLlama: Not sure what is going on here. My guess is a bad merge :(

frontend/src/features/research/hooks/useUpdateLease.test.tsx
FuriousLlama: yeah, this should not be here. I'll clean it up

frontend/package.json
devinleighsmith: yeah, like I said, all of the md icons should be part of react-icons, can you confirm if it is/isn't?

devinleighsmith:
@FuriousLlama in order for new permissions to be applied consistently, we need an update to the db to add the roles/claims to the relevant tables. This is because we use keycloak sync to set the role/claims/role_claims in keycloak in an automated way. Here is an example of that:

https://github.com/bcgov/PSP/commit/fbff0d630895a39e5240c0df6aae5a86058775b0#diff-f3b5c4742b9b76768fb96b8a235d2e29eaade1ecaf51750b1428737176c2e884

we may want to discuss how we do these permissions updates in the future, maybe in a separate story so we get the requirements documented correctly?

in this case we want the following:
the admin role should have all of the new research claims
the real estate analyst should have all of the new claims except for delete
the real estate manager should only have the view claim.

I have time do to the above if you need.

frontend/package.json
FuriousLlama: This is the one I found that matches the mockup:
https://github.com/bcgov/PSP/pull/1205

FuriousLlama:
Closing this PR since the next one has the addressed changes with included refactoring.
Followed up by.
https://github.com/bcgov/PSP/pull/1205

PR #1219, opened 2022-04-11
Features/psp 3125 With properties

FuriousLlama:
image

Follow up of https://github.com/bcgov/PSP/pull/1205.

Includes changes to the map sidebar layout


backend/dal/Services/ResearchFileService.cs
```diff
@@ -0,0 +1,61 @@
...
+                {
+                    _logger.LogDebug("Adding new property with pid:{prop}", pid);
+                    researchProperty.Property.PropertyTypeCode = "UNKNOWN";
+                    researchProperty.Property.PropertyClassificationTypeCode = "CORESTRAT";
+                    researchProperty.Property.PropertyDataSourceTypeCode = "PAIMS";
+                    researchProperty.Property.PropertyStatusTypeCode = "UNSURVYED";
+                    researchProperty.Property.SurplusDeclarationTypeCode = "UNKNOWN";
+                    researchProperty.Property.RegionCode = 1; // TODO: this reallly needs to come from the app
+                    researchProperty.Property.DistrictCode = 1; // TODO: this reallly needs to come from the app
+                    researchProperty.Property.AddressId = 1; // TODO: this would be nullable
```
FuriousLlama, line 51: This needs some extra work. Are these coming from the front end or can we make some assumptions?

backend/dal/Services/ResearchFileService.cs
```diff
@@ -0,0 +1,61 @@
...
+            foreach (var researchProperty in researchFile.PimsPropertyResearchFiles)
+            {
+                //var currentProperty = researchProperty.Property;
+                var pid = researchProperty.Property.Pid.Value;
+                try
+                {
+                    var foundProperty = _propertyRepository.GetByPid(pid);
+                    researchProperty.Property = foundProperty;
+                }
+                catch (KeyNotFoundException e)
```
FuriousLlama, line 41: IMO the result of a query not finding anything should return an empty object, null, or a wrapped result with a payload. Throwing an exception indicates that it should ALLWAYS find something, that is not the case here.

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
```diff
@@ -1,115 +1,108 @@
...
-  min-width: 93rem;
-  margin-left: -93rem;
-  max-width: 93rem;
-  padding: 2.4rem 3.6rem;
-  display: grid;
-  grid: 4.2rem 5.8rem 1fr / 1fr;
-  grid-template-areas:
-    'title'
-    'header'
-    'content';
```
FuriousLlama, line 106: This was necessary since the child layout was being pushed out of view if there was anything on the 'header'. Without it, the extra calculations are not necessary since the view is not absolute anymore.

frontend/src/features/properties/map/MapView.scss
```diff
@@ -1,11 +1,6 @@
 @import '~assets/scss/variables';
 @import '~assets/scss/colors.scss';
 
-.sidebar .map {
-  margin-left: 93rem;
-  width: calc(100% - 93rem);
-}
-
```
FuriousLlama, line 8: The sidebar map is no longer absolute so this is not necessary.

frontend/src/features/properties/map/MapView.scss
```diff
@@ -14,16 +9,6 @@
...
-.mapview {
-  &.draft-cursor,
-  &.draft-cursor .leaflet-grab,
-  &.draft-cursor .leaflet-interactive {
-    cursor: url('data:image/svg+xml,%3Csvg version="1.2" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" overflow="visible" preserveAspectRatio="none" viewBox="0 0 24 24" width="97" height="97"%3E%3Cg%3E%3Cpath xmlnsDefault="http://www.w3.org/2000/svg" id="map-marker" d="M13.88,11.22c-1.04,1.04-2.72,1.04-3.76,0.01c0,0-0.01-0.01-0.01-0.01C9.06,10.18,9.06,8.5,10.1,7.46 c0,0,0.01-0.01,0.01-0.01c1.04-1.04,2.72-1.04,3.76-0.01c0,0,0.01,0.01,0.01,0.01C14.86,8.48,14.8,10.14,13.88,11.22z M17.33,9.34 c0.02-1.42-0.54-2.78-1.56-3.77C14.78,4.55,13.42,3.98,12,4c-1.42-0.02-2.78,0.55-3.77,1.57C7.21,6.56,6.65,7.92,6.67,9.33 C6.64,9.97,6.75,10.61,7,11.2l3.8,8.06c0.1,0.23,0.27,0.41,0.48,0.54c0.43,0.26,0.98,0.26,1.41,0c0.21-0.13,0.38-0.32,0.49-0.54 L17,11.2c0.25-0.59,0.36-1.22,0.34-1.86l0,0H17.33z" style="fill: rgb(66, 139, 202);" vectorEffect="non-scaling-stroke"/%3E%3C/g%3E%3C/svg%3E')
-        48.5 81,
-      pointer;
-  }
-}
-
```
FuriousLlama, line 26: moved the cursor styles to the individual component

backend/dal/Services/ResearchFileService.cs
devinleighsmith: the address being nullable is coming in 25.01, what should be done with the region and district is defined here:

https://jira.th.gov.bc.ca/browse/PSP-3125?focusedCommentId=434144&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-434144

backend/dal/Services/ResearchFileService.cs
devinleighsmith: Sure, this is a nice side effect of having a wrapping service layer, we can hide internal db errors and wrap them with appropriate business logic.

frontend/src/features/properties/selector/MapSelectorModalContainer.tsx
```diff
@@ -1,77 +0,0 @@
-import { Button } from 'components/common/form';
```
devinleighsmith, line 1: I don't see a replacement for this, is there still a way to pop up a window with the selector embedded into it? this may be required for future stories.

frontend/src/features/properties/selector/PropertySelectorFormView.tsx
```diff
@@ -1,72 +1,51 @@
...
 }
 
 const PropertySelectorFormView: React.FunctionComponent<IPropertySelectorFormViewProps> = ({
-  onClickDraftMarker,
-  onClickAway,
-  formikRef,
-  properties,
+  onSelectedProperty,
 }) => {
+  const { setCursor } = React.useContext(SelectedPropertyContext);
```
devinleighsmith, line 20: Ideally containers interact with contexts and not form components. That allows for easier segregation of display components and logic components.

frontend/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "1.0.1",
+    "@mdi/react": "^1.5.0",
```
devinleighsmith, line 7: same comment as earlier, please confirm that this is necessary. We do not want to increase our bundle size unless absolutely necessary.

frontend/src/features/mapSideBar/MotiInventoryContainer.test.tsx
```diff
@@ -3,27 +3,41 @@ import MockAdapter from 'axios-mock-adapter';
...
+    );
 
     return {
       component,
     };
   };
 
   it('requests ltsa data by pid', async () => {
-    history.push('mapview?pid=9212434&searchBy=pinOrPid&sidebar=true');
+    //history.push('mapview?pid=9212434&searchBy=pinOrPid&sidebar=true');
```
devinleighsmith, line 33: can remove comment.

frontend/src/components/maps/leaflet/Map.test.tsx
```diff
@@ -28,6 +28,8 @@ const mockParcels = [
 jest.mock('hooks/useApi');
 jest.mock('hooks/pims-api');
 
+const onPropertyMarkerClick = jest.fn();
```
devinleighsmith, line 31: are the tests not done for this? it seems odd to add a mocked function but never verify it, reset it, etc.

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -1,27 +1,32 @@
...
+
 /**
  * container responsible for logic related to map sidebar display. Synchronizes the state of the parcel detail forms with the corresponding query parameters (push/pull).
  */
-export const MotiInventoryContainer: React.FunctionComponent = () => {
+export const MotiInventoryContainer: React.FunctionComponent<IMotiInventoryContainerProps> = props => {
   const isMounted = useIsMounted();
-  const formikRef = React.useRef<FormikProps<FormikValues>>();
-  const { showSideBar, setShowSideBar, pid } = useMapSideBarQueryParams(formikRef);
+  //const { showSideBar, setShowSideBar, pid } = useMapSideBarQueryParams();
```
devinleighsmith, line 29: comment.

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
devinleighsmith: This has led to a regression in how the sidebar is animated, the forms now squish/unsquish as the sidebar is closed.

frontend/src/components/maps/leaflet/Map.test.tsx
FuriousLlama: not at the moment. Added it to fulfill the props for the component

frontend/src/features/properties/selector/PropertySelectorFormView.tsx
FuriousLlama: that's an interesting point. We should talk about it more to standardize the expected behaviour of each part

frontend/src/features/properties/selector/MapSelectorModalContainer.tsx
FuriousLlama: There is not one at the moment. The refactor render this one unusable, and since is not being used atm I did not refactor this one.

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
```diff
@@ -1,115 +1,108 @@
...
+  height: 100%;
+  width: 100%;
 `;
 
-const LotIcon = styled(LotSvg)`
-  width: 3rem;
-  height: 3rem;
-  align-self: flex-end;
-`;
+const Header = styled.div``;
```
devinleighsmith, line 81: Is there a reason for these empty styled components?

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
```diff
@@ -1,115 +1,108 @@
...
           </>
         )}
       </VisibilitySensor>
     </StyledMapSideBarLayout>
   );
 };
 
-const Content = styled.div`
-  grid-area: content;
+const StyledBody = styled.div`
```
devinleighsmith, line 67: It seems odd to use rows and cols everywhere else but a custom div here, what was the rationale?

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
```diff
@@ -1,115 +1,108 @@
...
-  padding: 2.4rem 3.6rem;
-  display: grid;
-  grid: 4.2rem 5.8rem 1fr / 1fr;
-  grid-template-areas:
-    'title'
-    'header'
-    'content';
-  transition: 1s;
+
+  min-width: ${props => (props.show ? `93rem` : `0px`)};
```
devinleighsmith, line 96: nit: mixing and matching 0px vs 0rem

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
devinleighsmith: Also the screen is cut off:

image

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { Button, Col, Row } from 'react-bootstrap';
+import { Prompt, useHistory } from 'react-router-dom';
+import styled from 'styled-components';
+
+import { useAddResearch } from '../hooks/useAddResearch';
+import { AddResearchFileYupSchema } from './AddResearchFileYupSchema';
+import { ResearchForm } from './models';
+import ResearchProperties from './ResearchProperties';
+
+const AddResearchForm: React.FunctionComponent = () => {
```
devinleighsmith, line 14: getting a 404 from here: /research/ogs-internal/ows

devinleighsmith: also: 
image

frontend/src/models/api/Property.ts
```diff
@@ -0,0 +1,7 @@
+import { Api_ConcurrentVersion } from './ConcurrentVersion';
+
+export interface Api_Property extends Api_ConcurrentVersion {
+  id?: string;
```
devinleighsmith, line 4: Looks like you are trimming out some content from 2907 with this.
2907 supports:
pid, pin, plan number, latitude, and longitude.
it will eventually support address and legal description when we get bc assessment and the full attributed parcel map working.

frontend/src/features/mapSideBar/MotiInventoryHeader.tsx
```diff
@@ -6,12 +6,12 @@ import styled from 'styled-components';
...
 import { HeaderField } from './tabs/HeaderField';
 
-interface IMapSlideBarHeaderProps {
+interface IMotiInventoryHeaderProps {
   ltsaData?: LtsaOrders;
   property?: IPropertyApiModel;
 }
 
-export const MapSlideBarHeader: React.FunctionComponent<IMapSlideBarHeaderProps> = props => {
+export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderProps> = props => {
```
devinleighsmith, line 14: odd, didn't this get updated to display an independent spinner?

frontend/src/features/mapSideBar/MotiInventoryHeader.tsx
```diff
@@ -57,7 +57,6 @@ export const MapSlideBarHeader: React.FunctionComponent<IMapSlideBarHeaderProps>
 };
 
 const HeaderWrapper = styled.div`
-  position: relative;
```
devinleighsmith, line 60: Ah, I see, another display regression then, likely caused by this. without an absolute/relative parent the spinner doesn't know what to draw on top of and so is not drawing on top of the header correctly anymore.

frontend/src/features/mapSideBar/hooks/useMapSideBarQueryParams.ts
```diff
@@ -20,7 +20,7 @@ export enum MapSidebarContextType {
 }
 
 /** control the state of the side bar via query params. */
-export const useMapSideBarQueryParams = (formikRef?: any): IMapSideBar => {
```
devinleighsmith, line 23: thanks for these

frontend/src/features/properties/selector/MapSelectorContainer.test.tsx
```diff
@@ -1,85 +0,0 @@
-import { PropertyPopUpContext } from 'components/maps/providers/PropertyPopUpProvider';
```
devinleighsmith, line 1: This is a real shame, I like a lot of the changes here but losing out on test coverage hurts, and doesn't feel like a good use of either of our time. I think we should get together and discuss what happened here so we can avoid something like this from happening in the future.

frontend/src/features/properties/selector/MapSelectorModalContainer.tsx
devinleighsmith: same comment then, lets meet offline so I can understand why the refactoring was necessary.

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const onCancel = () => {
+    history.push(`/`);
+  };
+  return (
+    <Formik<ResearchForm>
+      initialValues={initialForm}
+      onSubmit={async (values: ResearchForm, formikHelpers) => {
+        const researchFile: Api_ResearchFile = values.toApi();
+        formikHelpers.setSubmitting(false);
```
devinleighsmith, line 35: same comment as earlier about order of operations here.

frontend/src/features/research/add/models.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.pid = property.pid;
+    this.pin = property.pin;
+    this.latitude = property.latitude;
+    this.longitude = property.longitude;
+    this.planNumber = property.planNumber;
+  }
+
+  public toApi(): Api_Property {
+    // TODO: description
+    return { pid: Number(this.pid), pin: Number(this.pin) };
```
devinleighsmith, line 46: I believe you'll need to pass the plan number, lat, and lng as well, otherwise the property will be missing those values when it is created.

devinleighsmith (changes requested):
being able to display existing markers in addition to new markers is definitely a requirement, that will need to be fixed before this can be merged.

devinleighsmith:
nit: the save/cancel buttons on this form are supposed to be sticky.

frontend/src/features/research/add/AddResearchForm.tsx
FuriousLlama: Found the issue, it turns out the baseAxios was taking the current url as a base if nothing was passed as baseUrl parameter

devinleighsmith:
I see what you mean now that you moved the list of selected properties out of the tab. You'll need to get that approved by Bobbi otherwise the story will fail FT. @FuriousLlama

backend/api/Areas/Research/Controllers/ResearchFileController.cs
```diff
@@ -0,0 +1,65 @@
...
+using Pims.Api.Models.Concepts;
+using Pims.Api.Policies;
+using Pims.Dal.Security;
+using Pims.Dal.Services;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Pims.Api.Areas.ResearchFile.Controllers
+{
+    /// <summary>
+    /// ResearchFileController class, provides endpoints for interacting with leases.
```
devinleighsmith, line 14: nit: leases

backend/api/Models/Concepts/ResearchFile/ResearchFilePropertyMap.cs
```diff
@@ -0,0 +1,30 @@
...
+using Entity = Pims.Dal.Entities;
+
+namespace Pims.Api.Models.Concepts
+{
+    public class ResearchFilePropertyMap : IRegister
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config.NewConfig<Entity.PimsPropertyResearchFile, ResearchFilePropertyModel>()
+                .PreserveReference(true)
```
devinleighsmith, line 11: Is this needed to prevent a circular reference from research -> propertyresearch -> research?

FuriousLlama: Yes

backend/dal/Security/Permissions.cs
```diff
@@ -70,6 +70,18 @@ public enum Permissions
         LeaseEdit = 21,
 
         [Display(GroupName = "lease", Name = "lease-delete", Description = "Can delete existing leases.")]
-        LeaseDelete = 22
+        LeaseDelete = 22,
```
devinleighsmith, line 73: hmm, thats odd. This doesn't seem to match up with the DB, let me check something.

devinleighsmith: not relevant for this pr anyways I guess, will fix in mine re: permissions.

backend/dal/Services/ResearchFileService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            _user = user;
+            _logger = logger;
+            _researchFileRepository = researchFileRepository;
+            _propertyRepository = propertyRepository;
+        }
+
+        public PimsResearchFile Add(PimsResearchFile researchFile)
+        {
+            _logger.LogInformation("Adding research file...");
```
devinleighsmith, line 28: we should discuss our use of logging, we aren't doing this enough and it would be nice to come up with some patterns.

frontend/package-lock.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "frontend",
-  "version": "1.1.0-24.12",
```
devinleighsmith, line 3: I see a package lock change but no package change, intended?

frontend/src/assets/images/pins/icon-draft.svg
```diff
@@ -1 +1,7 @@
-<svg version="1.2" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink" overflow="visible" preserveAspectRatio="none" viewBox="0 0 24 24" width="97" height="97"><g><path xmlnsDefault="http://www.w3.org/2000/svg" id="map-marker" d="M13.88,11.22c-1.04,1.04-2.72,1.04-3.76,0.01c0,0-0.01-0.01-0.01-0.01C9.06,10.18,9.06,8.5,10.1,7.46  c0,0,0.01-0.01,0.01-0.01c1.04-1.04,2.72-1.04,3.76-0.01c0,0,0.01,0.01,0.01,0.01C14.86,8.48,14.8,10.14,13.88,11.22z M17.33,9.34  c0.02-1.42-0.54-2.78-1.56-3.77C14.78,4.55,13.42,3.98,12,4c-1.42-0.02-2.78,0.55-3.77,1.57C7.21,6.56,6.65,7.92,6.67,9.33  C6.64,9.97,6.75,10.61,7,11.2l3.8,8.06c0.1,0.23,0.27,0.41,0.48,0.54c0.43,0.26,0.98,0.26,1.41,0c0.21-0.13,0.38-0.32,0.49-0.54  L17,11.2c0.25-0.59,0.36-1.22,0.34-1.86l0,0H17.33z" style="fill: rgb(66, 139, 202);" vectorEffect="non-scaling-stroke"/></g></svg>
\ No newline at end of file
+<svg version="1.2"
```
devinleighsmith, line 1: what was the reason for this change?

frontend/package-lock.json
FuriousLlama: yep, now it matches the version on package.json

frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
FuriousLlama: Leaving them here since I am not 100% we will need to revisit this as we work on the rest of the sidebar components

frontend/src/assets/images/pins/icon-draft.svg
FuriousLlama: react was complaining it could not parse the  xmlns="http://www.w3.org/2000/svg" part.

frontend/src/features/research/add/AddResearchForm.tsx
devinleighsmith: The issue with the icon is still a problem @FuriousLlama

frontend/src/features/research/add/AddResearchForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+            </Col>
+            <Col xs="auto">
+              <Button
+                disabled={formikProps.isSubmitting}
+                onClick={async () => {
+                  formikProps.setSubmitting(true);
+                  formikProps.submitForm();
+                }}
+              >
+                Save
```
devinleighsmith, line 66: Technically these buttons are supposed to be sticky (in the mockup)

frontend/src/features/mapSideBar/MotiInventoryHeader.tsx
devinleighsmith: Now there is just one spinner that takes up the whole screen, I would consider that a regression as well.

frontend/src/features/research/add/AddResearchView.tsx
```diff
@@ -0,0 +1,70 @@
...
+  lat: 52.81604319154934,
+  lng: -124.67285156250001,
+};
+
+interface ResearchMapViewProps {
+  showParcelBoundaries?: boolean;
+  onMarkerPopupClosed?: (obj: IProperty) => void;
+}
+
+const ResearchMapView: React.FC<ResearchMapViewProps> = props => {
```
devinleighsmith, line 26: The only problem I see with this is that the two map instances don't share state, so if we ever need to navigate between map instances the state will be reset. Not a problem at the moment but could be in the future.

devinleighsmith (approved):

devinleighsmith:
just got some conflicts and then can merge.

PR #1220, opened 2022-04-12
Psp 2877

devinleighsmith:
Research List View


devinleighsmith:
note that this will need to be merged with Manuel's changes when available.

PR #1226, opened 2022-04-12
PSP-3086: Contact modal default to 5 items

jdtoombs:
Quick fix to pass initial page size prop to `useSearch`  to match AC of 5 items shown by default on the modals
contactitemsize


devinleighsmith:
some test failures here.

devinleighsmith (approved):

PR #1230, opened 2022-04-12
update claims/roles

devinleighsmith:
-add research claims associate with appropriate roles
-ensure that existing roles/claims are correct


database/mssql/scripts/dbscripts/PSP_PIMS_S25_01/Alter Down/88_DML_PIMS_ROLES_CLAIMS.sql
```diff
@@ -0,0 +1,87 @@
...
+. . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . . .
+Author        Date         Comment
+------------  -----------  -----------------------------------------------------
+Devin Smith  2021-Nov-18  Initial version
+----------------------------------------------------------------------------- */
+
+DECLARE @admin BIGINT;
+DECLARE @rem BIGINT;
+DECLARE @rea BIGINT;
+SELECT @admin = ROLE_ID FROM PIMS_ROLE where NAME = 'System Administrator';
```
devinleighsmith, line 12: @dfilteau it seems like the order of roles was changed when a new row was inserted at some point, This was leading to a number of roles not having the correct claims associated to it. Hopefully by getting the correct ID explicitly here that will not happen again in the future.

My question is, do I need to do this with the PIMS_CLAIM table as well? currently this script is depending on all of the claims ids to be static. If they were ever re-ordered this script would assign incorrect claims to roles.

devinleighsmith:
@dfilteau please be aware of these changes when you are creating your next schema migration (and do not overwrite them).

dfilteau:
Doug Filteau
Little Mountain Systems Inc.


m: 1.306.201.8531
e:  ***@***.***> ***@***.***
w:  <https://hilltopz.com> https://hilltopz.com
 
“In our profession, precision and perfection are not a dispensable luxury, but a simple necessity.”

                • Nicklaus Wirth

database/mssql/scripts/dbscripts/PSP_PIMS_S25_01/Alter Down/88_DML_PIMS_ROLES_CLAIMS.sql
dfilteau: Hi @devinleighsmith, I made no changes to the order of the roles so I'm not sure where this change came from.  If you could provide the roles in the correct order, I'd appreciate it.

devinleighsmith: This PR addresses the issue.

The issue appears to have been introduced when the value UNDETERMINED was introduced that value was introduced before the pre-existing roles of Real Estate Manager and Real Estate Analyst

image

asanchezr (approved):

PR #1231, opened 2022-04-12
PSP-2919 Property Details - NON-INVENTORY 

asanchezr:
This PR will hide the property information tab in the map slide-out for non-inventory properties. Basically when the API returns an error for a property from the map it means it's not on our inventory. In that case, hide the tab and make LTSA the default selected tab.

Also refactored the way we launch the sidebar to only require a PID instead of a full property object. For non-inventory we have no property object so wouldn't make sense there.

### DIFF
--- a/frontend/src/features/mapSideBar/MapSideBarContainer.tsx
+++ b/frontend/src/features/mapSideBar/MapSideBarContainer.tsx
@@ -46,7 +46,7 @@ export const MotiInventoryContainer: React.FunctionComponent = () => {
         // We get an error because PID is not on our database
         if (axios.isAxiosError(e)) {
           const axiosError = e as AxiosError<IApiError>;
-          if (axiosError?.response?.status === 400) {
+          if (axiosError?.response?.status === 404) {
             setShowPropertyInfoTab(false);
           }
         }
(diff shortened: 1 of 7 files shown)


frontend/src/features/mapSideBar/MapSideBarContainer.tsx
```diff
@@ -25,15 +27,28 @@ export const MotiInventoryContainer: React.FunctionComponent = () => {
...
+            setApiProperty(propInfo);
+            setShowPropertyInfoTab(true);
+          }
+        }
+      } catch (e) {
+        // PSP-2919 Hide the property info tab for non-inventory properties
+        // We get an error because PID is not on our database
+        if (axios.isAxiosError(e)) {
+          const axiosError = e as AxiosError<IApiError>;
+          if (axiosError?.response?.status === 400) {
```
devinleighsmith, line 49: darn, does the api really return a 400 and not a 404 in this case?

devinleighsmith: eech, yep, it does. Should we return a 404 from here instead? 
            else if (ex is KeyNotFoundException)
            {
                code = HttpStatusCode.BadRequest;
                message = "Item does not exist.";

                _logger.LogDebug(ex, "Middleware caught unhandled exception.");
            }

devinleighsmith (approved):

frontend/src/features/mapSideBar/MapSideBarContainer.tsx
asanchezr: Sadly it does. I was hoping for a 404 but we get 400. Might want to file a tech debt ticket to see if we can change it across the app for that keynotfound exception

asanchezr: I can fix it there. Good eye

PR #1234, opened 2022-04-13
Psp 2877 reuse

devinleighsmith:
Research list view


asanchezr:
PR failing with a weird error. Did you include all relevant code @devinleighsmith ?

image

backend/api/Models/Concepts/Property/PropertyModel.cs
```diff
@@ -1,42 +1,183 @@
...
+
+        /// <summary>
+        /// get/set - The property zoning potential.
+        /// </summary>
+        public string ZoningPotential { get; set; }
+
+        /// <summary>
+        /// get/set - The property notes.
+        /// </summary>
+
```
asanchezr, line 168: nit:  comments should be colocated with field

backend/dal/Helpers/Extensions/ResearchExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        /// <summary>
+        /// Generate a new R File in format R-XXX-YYY using the research file id. Add the research file id and rfileNo to the passed research file.
+        /// </summary>
+        /// <param name="context"></param>
+        public static PimsResearchFile GenerateRFileNo(this PimsContext context, PimsResearchFile researchFile)
+        {
+            Int64 researchFileId = GetNextResearchFileSequenceValue(context);
+            researchFile.ResearchFileId = researchFileId;
+            researchFile.RfileNumber = $"L-{researchFile.ResearchFileId.ToString().PadLeft(6, '0').Insert(3, "-")}";
```
asanchezr, line 136: As per comments this should generate R-XXX-YY identifiers. However I see it being generated with "L-XXX-YY"  - perhaps a typo?

backend/dal/Services/Interfaces/IResearchService.cs
```diff
@@ -0,0 +1,10 @@
+using Pims.Dal.Entities;
+using Pims.Dal.Entities.Models;
+
+namespace Pims.Dal.Services
+{
+    public interface IResearchService
+    {
+        Paged<PimsResearchFile> GetPage(ResearchFilter filter);
+    }
+}
```
asanchezr, line 10: Do we have a use for this interface - given that we already have a **IResearchFileService** above? seems like a duplicate

devinleighsmith: will remove

backend/dal/Helpers/Extensions/ResearchExtensions.cs
devinleighsmith: this was refactored to be part of the research repository, I can remove it.

backend/api/Mapping/TypeMap.cs
```diff
@@ -14,6 +14,12 @@ public void Register(TypeAdapterConfig config)
                 .Map("IsDisabled", "IsDisabled")
                 .Map("DisplayOrder", "DisplayOrder");
 
+            config.ForType(typeof(Entity.ICodeEntity<short>), typeof(Model.TypeModel<short>))
+                .Map("Id", "Id")
+                .Map("Description", "Description")
+                .Map("IsDisabled", "IsDisabled")
+                .Map("DisplayOrder", "DisplayOrder");
+
```
asanchezr, line 22: I like this generic mapping but ICodeEntity doesn't have Id on the interface (instead has Code). 
I see that of the two code entities below (Region and District) one has Id and the other hasn't.
What do you think of adding an "Id" property to the ICodeEntity interface to make it consistent across all models? 
Otherwise this mapping would fail for some code entities I believe? 
Thoughts?

backend/tests/unit/api/Controllers/Research/ResearchControllerTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// <summary>
+        /// Make a successful request that passes the filter in the query string.
+        /// </summary>
+        [Theory]
+        [MemberData(nameof(ResearchQueryFilters))]
+        public void GetResearchFiles_Query_Success(Uri uri)
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var controller = helper.CreateController<SearchController>(Permissions.PropertyView, uri);
```
asanchezr, line 86: Not sure how this test passes as the controller requires LeaseView permission instead?

backend/api/Mapping/TypeMap.cs
devinleighsmith: This was created to handle some exceptions to the normal structure of typed entities, I think Doug created these early in the project before he had created an established pattern.

So for these ones I think mapping from the Id to the code makes sense (as the code is the only identifier on these tables).

backend/tests/unit/api/Controllers/Research/ResearchControllerTest.cs
devinleighsmith: thanks for the reminder, need to update these now that I have merged with Manuel's code.

frontend/src/AppRouter.tsx
```diff
@@ -208,6 +208,15 @@ const AppRouter: React.FC = () => {
           claim={[Claims.CONTACT_VIEW]}
           title={getTitle('View Contact')}
         />
+        <AppRoute
+          protected
+          exact
+          path="/research/list"
+          component={ResearchListView}
+          layout={AuthLayout}
+          claim={Claims.PROPERTY_VIEW}
```
asanchezr, line 217: The backend controller requires LeaseView for the endpoints but the frontend is requiring PropertyView? I don't know story well enough but seems odd.

frontend/src/components/Table/DateCell.tsx
```diff
@@ -0,0 +1,6 @@
+import { ILeaseTerm } from 'interfaces';
+import { CellProps } from 'react-table';
+import { prettyFormatDate } from 'utils';
+
+export const DateCell = ({ cell: { value } }: CellProps<ILeaseTerm, string>) =>
```
asanchezr, line 5: nit: Generic component should expect to be `CellProps<Date, string>` instead of ILeaseTerm

frontend/src/components/common/form/InputGroup.tsx
```diff
@@ -94,22 +94,13 @@ export const InputGroup: React.FC<InputGroupProps> = ({
         </ColPrepend>
       )}
       <ColContent>
-        {fast ? (
-          <FastInput
-            formikProps={formikProps}
```
asanchezr, line 99: Seems like the only usage of formikProps have been deleted from InputGroup implementation. Maybe would be helpful to remove from interface as well? That way no need to pass ` formikProps={null as any}` ?

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -147,21 +143,6 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
-        mapInstance.setView(
-          [selected?.latitude as number, selected?.longitude as number],
-          Math.max(MAX_ZOOM, mapInstance.getZoom()),
-        );
-      }
-    } else {
-      setCurrentSelected(null);
-    }
-  }, [selected, setCurrentSelected]);
-
```
asanchezr, line 164: thanks for cleaning this up!

frontend/src/features/research/list/ResearchFilter/ResearchFilter.tsx
```diff
@@ -0,0 +1,189 @@
...
+  updatedOnEndDate: '',
+  updatedOnStartDate: '',
+  rfileNumber: '',
+  researchSearchBy: 'name',
+  createOrUpdateRange: 'updatedOnStartDate',
+  createOrUpdateBy: 'appLastUpdateUserid',
+};
+
+/**
+ * Filter bar for leases and license.
```
asanchezr, line 41: nit: copy /paste

frontend/src/features/research/list/ResearchListView.test.tsx
```diff
@@ -0,0 +1,553 @@
...
+  });
+
+  it('searches by R-file number', async () => {
+    setupMockSearch([]);
+    const { container, searchButton } = setup();
+    fillInput(container, 'researchSearchBy', 'rFileNumber', 'select');
+    fillInput(container, 'rFileNumber', '101');
+    await act(async () => userEvent.click(searchButton));
+
+    expect(getResearchFiles).toHaveBeenCalledWith(expect.objectContaining({}));
```
asanchezr, line 224: I think this test will always pass? shouldn't we include actual fields being submitted here?

devinleighsmith: yes, went back and forth on these a few times, will correct.

frontend/src/features/research/list/ResearchListView.tsx
```diff
@@ -1,8 +1,102 @@
...
+      <Styled.Scrollable>
+        <Styled.PageHeader>Research Files</Styled.PageHeader>
+        <Styled.PageToolbar>
+          <Row>
+            <Col>
+              <ResearchFilter filter={filter} setFilter={changeFilter} />
+            </Col>
+          </Row>
+        </Styled.PageToolbar>
+        {hasClaim(Claims.LEASE_ADD) && (
```
asanchezr, line 73: just double checking this is the right claim?

asanchezr (approved):
Amazing work!

PR #1282, opened 2022-04-21
Mayan prototype with doc type | Features/psp 3232

FuriousLlama:


devinleighsmith:
If there are any openshift changes required, please update the relevant file in the openshift folder in git. Also, lets remember to update this page:
https://confluence.th.gov.bc.ca/display/PIMS/2.1
in the future with any manual steps required for release.

backend/api/Repositories/MayanDocumentRepository.cs
```diff
@@ -50,6 +50,87 @@ public class MayanDocumentRepository : IDocumentRepository
...
+                        retVal.Status = ExternalResultStatus.Error;
+                        retVal.Message = "Forbidden";
+                        break;
+                    default:
+                        retVal.Status = ExternalResultStatus.Error;
+                        retVal.Message = $"Unable to contact endpoint {endpoint}. Http status {response.StatusCode}";
+                        break;
+                }
+            }
+            catch (Exception e)
```
devinleighsmith, line 123: Is catching Exception considered an antipattern the way it is in Java?

devinleighsmith (approved):

backend/api/Repositories/MayanDocumentRepository.cs
FuriousLlama: It would be if you catch and then re-throw without adding anything. On this case is adding information to the reply wrapper

PR #1289, opened 2022-04-22
psp-3254 allow other unit types besides hectares to be selected.

devinleighsmith:


backend/api/Areas/Leases/Mapping/Lease/PropertyMap.cs
```diff
@@ -15,8 +14,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.PID, src => src.Property.Pid)
                 .Map(dest => dest.PIN, src => src.Property.Pin)
                 .Map(dest => dest.Name, src => src.Property.Name)
-                .Map(dest => dest.AreaUnitId, src => src.Property.PropertyAreaUnitTypeCode)
```
devinleighsmith, line 18: Really what needs to happen here is that the lease models need to be updated to the concept format. That would be a pretty major change however as that would affect all lease pages and subpages. 

For the time being I've just corrected the issue in-place.

asanchezr (approved):

PR #1294, opened 2022-04-25
Update research file | Features/psp 2876

FuriousLlama:
image


backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -46,8 +65,23 @@ public PimsResearchFile Add(PimsResearchFile researchFile)
...
+        /// </summary>
+        /// <param name="researchFile"></param>
+        /// <returns></returns>
+        public PimsResearchFile Update(PimsResearchFile researchFile)
+        {
+            researchFile.ThrowIfNull(nameof(researchFile));
+
+            researchFile.RfileNumber = GenerateRFileNumber();
+
+            this.Context.PimsResearchFiles.Update(researchFile);
```
devinleighsmith, line 78: I'm always worried about doing this. I believe this updates the db with the passed research file, but also updates all child relationships. For example, if this research file entity contained a number of changes to PimsProperty entities that would change those as well. Probably fine but we need to be careful about possible inadvertent updates.

backend/entities/RequestSourceType.cs
```diff
@@ -0,0 +1,28 @@
+using System.ComponentModel.DataAnnotations.Schema;
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// ResearchFileStatusType class, provides an entity for the datamodel to manage research status types.
```
devinleighsmith, line 5: nit

backend/entities/RequestSourceType.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public partial class PimsRequestSourceType : ITypeEntity<string>
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - Primary key to identify research file status type.
```
devinleighsmith, line 11: nit - name throughout file

frontend/src/features/mapSideBar/hooks/useMapSideBarQueryParams.tsx
```diff
@@ -74,6 +84,18 @@ export const useMapSideBarQueryParams = (map?: L.Map): IMapSideBar => {
         setSidebarComponent(<AddResearchContainer onClose={handleClose} />);
         setShowSideBar(true);
         break;
+      case MapViewState.RESEARCH_EDIT:
+        setSidebarComponent(
+          <UpdateResearchContainer researchFileId={Number(parts[3])} onClose={handleClose} />,
```
devinleighsmith, line 89: nit: store parts[3] in a variable to avoid dereferencing multiple times.

frontend/src/features/properties/map/research/add/AddResearchContainer.tsx
```diff
@@ -25,6 +25,7 @@ export const AddResearchContainer: React.FunctionComponent<IAddResearchContainer
   const saveResearchFile = async (researchFile: Api_ResearchFile) => {
     const response = await addResearchFile(researchFile);
     if (!!response?.name) {
+      formikRef.current?.resetForm();
```
devinleighsmith, line 28: I'm not so sure about this, Do we really want the behaviour of an error to be wiping the form? what if the error is correctible on the frontend? what if the error is due to a network timeout and just requires the form to be resubmitted?

frontend/src/features/properties/map/research/hooks/useGetResearch.ts
```diff
@@ -0,0 +1,49 @@
...
+        toast.success('Research File retrieved');
+        return response?.data;
+      } catch (e) {
+        if (axios.isAxiosError(e)) {
+          const axiosError = e as AxiosError<IApiError>;
+
+          if (axiosError?.response?.status === 400) {
+            toast.error(axiosError?.response.data.error);
+          } else {
+            toast.error('Save error. Check responses and try again.');
```
devinleighsmith, line 31: we aren't saving anything with this action.

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
```diff
@@ -0,0 +1,115 @@
...
+      showCloseButton
+      onClose={handleCancel}
+    >
+      <Formik<UpdateResearchFormModel>
+        enableReinitialize
+        innerRef={formikRef}
+        initialValues={initialForm}
+        onSubmit={async (values: UpdateResearchFormModel, formikHelpers) => {
+          const researchFile: Api_ResearchFile = values.toApi();
+          console.log(researchFile);
```
devinleighsmith, line 84: can remove this.

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
```diff
@@ -0,0 +1,115 @@
...
+          saveResearchFile(researchFile);
```
devinleighsmith, line 85: I'm guessing that this form is supposed to redirect to the view screen on success. Not seeing that here.

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const [researchFile, setResearchFile] = useState<Api_ResearchFile | undefined>(undefined);
+  const [initialForm, setForm] = useState<UpdateResearchFormModel | undefined>(undefined);
+
+  useEffect(() => {
+    async function fetchResearchFile() {
+      var retrieved = await retrieveResearchFile(props.researchFileId);
+      setResearchFile(retrieved);
+      if (retrieved !== undefined) {
+        setForm(UpdateResearchFormModel.fromApi(retrieved));
```
devinleighsmith, line 35: Not seeing loading logic on this screen, is there a reason for that?

frontend/src/features/properties/map/research/update/UpdateResearchForm.tsx
```diff
@@ -0,0 +1,200 @@
...
+  id: string;
+  text: string;
+}
+
+export interface IUpdateResearchFormProps {
+  formikProps: FormikProps<UpdateResearchFormModel>;
+}
+
+const UpdateResearchForm: React.FunctionComponent<IUpdateResearchFormProps> = props => {
+  //const { values, setFieldValue } = props.formikProps;
```
devinleighsmith, line 30: remove

frontend/src/features/properties/map/research/update/UpdateResearchForm.tsx
```diff
@@ -0,0 +1,200 @@
...
+      purposeType.researchPurposeTypeCode = x.id;
+      purposeType.researchPurposeTypeDescription = x.text;
+      return purposeType;
+    });
+    props.formikProps.setFieldValue('researchFilePurposes', mapped);
+  }
+
+  function handleRequesterSelected() {
+    var selectedContact = selectedContacts[0];
+    console.log(selectedContact);
```
devinleighsmith, line 73: remove

frontend/src/features/properties/map/research/update/UpdateResearchHeader.tsx
```diff
@@ -0,0 +1,67 @@
...
+import * as React from 'react';
+import { Col, Row } from 'react-bootstrap';
+import { prettyFormatDate } from 'utils';
+
+export interface IUpdateResearchHeaderProps {
+  researchFile?: Api_ResearchFile;
+}
+
+const UpdateResearchHeader: React.FunctionComponent<IUpdateResearchHeaderProps> = props => {
+  const lefttColumnWidth = '6';
```
devinleighsmith, line 12: nit, spelling.

frontend/src/features/properties/map/research/update/UpdateResearchHeader.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const leftColumnLabel = '4';
```
devinleighsmith, line 13: nit leftColumnLabelWidth

devinleighsmith: personally, I think I'd rather these variables are removed and the values just inlined, otherwise I can't see the value and the usage in the same area.

frontend/src/features/properties/map/research/update/models.ts
```diff
@@ -0,0 +1,120 @@
...
+      rfileNumber: this.rfileNumber,
+      statusTypeCode: { id: this.statusTypeCode },
+      //researchProperties: this.researchProperties,
+      requestDate: this.requestDate,
+      requestDescription: this.requestDescription,
+      requestSourceDescription: this.requestSourceDescription,
+      researchResult: this.researchResult,
+      researchCompletionDate: this.researchCompletionDate,
+      isExpropriation:
+        this.isExpropriation === undefined
```
devinleighsmith, line 107: this seems like a good candidate for the use of a utility function. we may already have one for this.

frontend/src/features/properties/map/research/update/models.ts
```diff
@@ -0,0 +1,120 @@
...
+    model.rfileNumber = base.rfileNumber;
+    model.statusTypeCode = base.statusTypeCode?.id;
+    model.statusTypeDescription = base.statusTypeCode?.description;
+    //model.researchProperties = base.researchProperties;
+    model.requestDate = base.requestDate;
+    model.requestDescription = base.requestDescription;
+    model.requestSourceDescription = base.requestSourceDescription;
+    model.researchResult = base.researchResult;
+    model.researchCompletionDate = base.researchCompletionDate;
+    model.isExpropriation = base.isExpropriation ? 'true' : 'false';
```
devinleighsmith, line 71: surprised by this, are we displaying these as strings?

frontend/src/features/properties/map/research/update/models.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export class UpdateResearchFormModel {
+  public id?: number;
+  public name?: string;
+  public roadName?: string;
+  public roadAlias?: string;
+  public rfileNumber?: string;
+  public statusTypeCode?: string;
+  public statusTypeDescription?: string;
+  //public researchProperties?: Api_ResearchFileProperty[];
```
devinleighsmith, line 42: Right, we have a follow up story for this I believe.

devinleighsmith (approved):
Are tests coming later or do you need to log a follow up story?

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
```diff
@@ -0,0 +1,115 @@
...
+          isSubmitting={formikRef.current?.isSubmitting}
+          onSave={handleSave}
+          onCancel={handleCancel}
+        />
+      }
+      header={<UpdateResearchHeader researchFile={researchFile} />}
+      showCloseButton
+      onClose={handleCancel}
+    >
+      <Formik<UpdateResearchFormModel>
```
devinleighsmith, line 78: I wonder if there is value putting in validation schema even when there are no required fields? just for basic stuff like the types yup.string, yup.number, yup.date etc.

frontend/src/features/properties/map/research/update/UpdateResearchHeader.tsx
FuriousLlama: This just makes sure that all columns are the same instead of having to manually update each attribute

frontend/src/features/properties/map/research/update/models.ts
FuriousLlama: This was necessary since the current select component requires a string for its value.

FuriousLlama:
Will address all comments on https://github.com/bcgov/PSP/pull/1295

frontend/src/features/mapSideBar/hooks/useMapSideBarQueryParams.tsx
```diff
@@ -54,6 +58,12 @@ export const useMapSideBarQueryParams = (map?: L.Map): IMapSideBar => {
     } else if (parts[2] === 'research') {
```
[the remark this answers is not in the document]
FuriousLlama: sure we can!

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
FuriousLlama: Yep, thats on the view screen since this pr did not have that

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
FuriousLlama: I moved the loading logic to a simple return of the component.

frontend/src/features/properties/map/research/update/UpdateResearchContainer.tsx
FuriousLlama: not sure, the problem I saw was that if it was specified the submit wo[uld not work but no error was shown

PR #1297, opened 2022-04-26
psp-3280 enforce order of address fields on contact view screens.

devinleighsmith:


frontend/src/features/contacts/contact/detail/utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    return 1;
+  }
+
+  if (a2.addressType.id === 'RESIDNT' && a1.addressType.id !== 'MAILING') {
+    return 1;
+  }
+  return -1;
+};
+
+export const fakeAddresses = [
```
FuriousLlama, line 18: If this is for testing only, lets move it to a testing file instead of just util

frontend/src/features/contacts/contact/detail/utils.ts
```diff
@@ -0,0 +1,89 @@
+import { IContactAddress } from 'interfaces/IContact';
+
+export const sortAddresses = (a1: IContactAddress, a2: IContactAddress) => {
+  if (a2.addressType.id === a1.addressType.id) {
+    return 0;
+  }
+
+  if (a2.addressType.id === 'MAILING') {
```
asanchezr, line 8: nit: I prefer we use enums/constants here. I believe we have that in the frontend already

asanchezr (approved):

frontend/src/features/contacts/contact/detail/utils.ts
FuriousLlama, line 8: this looks rather hard to maintain, how about having the order of each type hardcoded on a separate array and order it by that.

FuriousLlama (approved):

PR #1298, opened 2022-04-26
correct issue with pims_property sequence not being set by paims etl.

devinleighsmith:
pid/pin is not required when adding a property
set the sequence to be equal to the latest value in the PIMS_PROPERTY table.


frontend/src/features/properties/map/research/add/AddResearchFileYupSchema.ts
```diff
@@ -2,13 +2,5 @@ import * as Yup from 'yup';
...
-  properties: Yup.array().of(
-    Yup.object().shape({
-      pid: Yup.string().when('pin', {
-        is: (pin: string) => !!pin,
-        then: Yup.string().nullable(),
-        otherwise: Yup.string().required('valid PID or PIN required'),
-      }),
-    }),
-  ),
+  properties: Yup.array(),
```
asanchezr, line 5: Is this related to this PR? Doesn't look like...

asanchezr (approved):

database/mssql/scripts/dbscripts/PSP_PIMS_S25_01/Alter Up/89_DML_PIMS_PROPERTY_ID_SEQ.sql
```diff
@@ -0,0 +1,6 @@
+DECLARE @PIMS_PROPERTY_ID VARCHAR(MAX);
+DECLARE @sql VARCHAR(MAX);
+SET @PIMS_PROPERTY_ID = (SELECT max(PROPERTY_ID) from PIMS_PROPERTY) + 1;
+SET @sql = 'ALTER SEQUENCE PIMS_PROPERTY_ID_SEQ RESTART WITH ' + @PIMS_PROPERTY_ID
+EXEC (@sql)
+GO
```
FuriousLlama, line 6: Is this necessary? Should the etl migration not do this instead? The problem is that since this is only executed on the s25 alter up, if another set of properties are added after this wont work

frontend/src/features/properties/map/research/add/AddResearchFileYupSchema.ts
devinleighsmith: it is, psp-3287 has two issues:
1) research files containing properties without pids or pins can't be saved.
2) research files with any properties throw an error when saving

database/mssql/scripts/dbscripts/PSP_PIMS_S25_01/Alter Up/89_DML_PIMS_PROPERTY_ID_SEQ.sql
devinleighsmith: This will allow IS 25.1 in UAT to go ahead. Future ETL work will need to add this script as well

PR #1311, opened 2022-04-27
PSP-3053: Read - Property Details (v1) - Title (LTSA) - OWNERSHIP INF…

Maheshbabu28:
…ORMATION


devinleighsmith:
noticed that the collapse function is missing from this section. I believe that this section is supposed to be collapseable.

frontend/src/features/mapSideBar/tabs/ltsa/LtsaOwnershipInformationTitleOwnerForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  nameSpace?: string;
+}
+
+export const LtsaOwnershipInformationTitleOwnerForm: React.FunctionComponent<ILtsaOwnershipInformationTitleOwnerFormProps> = ({
+  nameSpace,
+}) => {
+  const { values } = useFormikContext<LtsaOrders>();
+  const titleOwners = getIn(values, withNameSpace(nameSpace, 'titleOwners')) ?? [];
+
+  const getJoinedFieldValues = (input1?: string, input2?: string): string => {
```
devinleighsmith, line 21: could this be summarized as return [input1, input2].join(', ') ? this seems a bit verbose.

frontend/src/interfaces/ltsaModels.ts
```diff
@@ -2041,13 +2041,13 @@ export interface TitleOwner {
    * @type {string}
    * @memberof TitleOwner
    */
-  surnameOrCorpName1: string;
```
devinleighsmith, line 2044: this is surprising, these are generated so I wouldn't expect any of these to be wrong. Are you sure about this?

frontend/src/features/mapSideBar/tabs/ltsa/LtsaOwnershipInformationTitleOwnerForm.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {
+        ...renderOptions,
+      },
+    );
+
+    return {
+      component,
+    };
+  };
+  it('renders ownership information title owner form', () => {
```
devinleighsmith, line 32: I think this is fine for now, in the future it is nice to have at least one test to sanity check some actual data on the form. Ie, if you pass in an owner, does the owner name display correctly, that sort of thing. The problem with snapshots is that it is easy to ignore failures since they are so verbose.

devinleighsmith (approved):

frontend/src/features/mapSideBar/tabs/ltsa/LtsaOwnershipInformationTitleOwnerForm.test.tsx
Maheshbabu28: Sure, I will take care of this in future stories when I work

PR #1312, opened 2022-04-27
PSP-3167 fix button css

asanchezr:


frontend/src/components/common/__snapshots__/ErrorModal.test.tsx.snap
```diff
@@ -4,6 +4,12 @@ exports[`Error modal tests... renders correctly 1`] = `
 <div
   aria-hidden="true"
 >
+  <div
```
devinleighsmith, line 7: hm... this seems unexpected.

devinleighsmith (approved):

devinleighsmith:
FYI going to change your task into a story so that the tester's retest buttons.

frontend/src/components/common/__snapshots__/ErrorModal.test.tsx.snap
asanchezr: This is because now ErrorModal is rendered inside `<TestCommonWrapper>` - it wasn't before. We need the test wrapper because our new button needs the ThemeContext for its styling to work

frontend/src/components/common/buttons/Button.tsx
```diff
@@ -0,0 +1,235 @@
...
+          as="span"
+          style={{ marginLeft: '.8rem', padding: '.8rem' }}
+        />
+      )}
+    </StyledButton>
+  );
+});
+
+const StyledButton = styled(BootstrapButton)`
+  &.btn {
```
FuriousLlama, line 62: Instead of having the css change based on the variant, what do you think about returning different styles that inherit from bootstrap? Just asking to see if would make the maintenance of the buttons easier

FuriousLlama (approved):

PR #1314, opened 2022-04-27
prevent NPE when property address is null.

devinleighsmith:
based on the is26 db migration, ensure that frontend handles nullable address properly.


FuriousLlama:
Was the backend model and front end model changed to be undefined too? that would help us find other places where the address was expected to be non-null

asanchezr (approved):

PR #1332, opened 2022-04-28
psp-3301 enable quality gate on PR (not just merge).

devinleighsmith:


devinleighsmith:
Quality gate running in all environments but only on PR merge where it may be missed. Changing to run always, will need to check how that affects overall action run times.

FuriousLlama (approved):

asanchezr (approved):

asanchezr:
Approved but I think the problem is that PRs don't get the secrets needed to run sonarqube quality gate (because they come from forks)

asanchezr:
It still fails because the SONAR_PROJECT key is a secret and will not work with forks. Not sure 100% that project key should be a secret but was told yes initially so kept it that way

PR #1349, opened 2022-05-02
PSP-3296: Fix UI to show 0 at the 1/100th place

Maheshbabu28:


frontend/src/components/common/form/FastCurrencyInput.tsx
```diff
@@ -124,7 +124,16 @@ const CurrencyInput = ({
...
+            let value: string | number = '';
+            if (cleanValue) {
+              const cleanValueArr = cleanValue.split('.');
+              if (cleanValueArr.length > 1 && cleanValueArr[1].length >= 2) {
+                value = parseFloat(cleanValue).toFixed(2);
+              } else {
+                value = parseFloat(cleanValue);
+              }
+            }
+            setFieldValue(field, value);
```
asanchezr, line 136: This function will now set the field value to either a string - e.g. '1000.00' or a number - e.g. 1000 depending on what you type (1000.00 will get a string, 1000 will get a number)

I think this should always return a number or a string because consumers of this component would not expect to receive a string in some cases. What are your thoughts @devinleighsmith ?

asanchezr: To make it a bit clear. What I'm saying is that branch that does `parseFloat().toFixed(2)` returns a string while the else branch with just `parseFloat()` returns a number. I'm fine if we make both strings but should be consistent.

Maheshbabu28: Thanks for your feedback, I also thought we shall format the value to two decimal places always to be more consistent like below, please let me know your thoughts so that I can push it
image

asanchezr (approved):

PR #1380, opened 2022-05-11
Update Research property | Features/psp 2941

FuriousLlama:
image


backend/dal/Services/ResearchFileService.cs
```diff
@@ -85,5 +84,37 @@ public Paged<PimsResearchFile> GetPage(ResearchFilter filter)
...
+        {
+            _logger.LogInformation("Updating property research file...");
+            _user.ThrowIfNotAuthorized(Permissions.ResearchFileEdit);
+            ValidateVersion(researchFileId, researchFileVersion);
+            return _researchFileRepository.UpdateProperty(researchFileId, propertyResearchFile);
+        }
+
+        private void PopulateResearchFile(PimsProperty property)
+        {
+            property.PropertyClassificationTypeCode = "CORESTRAT"; // Todo: should be "UNKNOWN"
```
devinleighsmith, line 98: same comment as last time for visibility, are we tracking on these TODOs? Doug should be made aware of this one for example.

frontend/src/features/properties/map/research/common/ResearchMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const ResearchMenu: React.FunctionComponent<IResearchMenuProps> = props => {
+  const handleClick = (index: number) => {
+    props.onChange(index);
+  };
+  return (
+    <StyledMenuWrapper>
+      {props.items.map((label: string, index: number) => (
+        <StyledRow
```
devinleighsmith, line 18: I don't see a key here so you may be getting a console warning from react about not having a unique key per mapped row.

frontend/src/features/properties/map/research/detail/PropertyResearchView.tsx
```diff
@@ -0,0 +1,123 @@
...
+import * as React from 'react';
+import { useEffect, useState } from 'react';
+import { pidFormatter } from 'utils';
+
+export interface IPropertyResearchViewProps {
+  researchFileProperty: Api_ResearchFileProperty;
+  setEditMode: (isEditing: boolean) => void;
+}
+
+const PropertyResearchView: React.FunctionComponent<IPropertyResearchViewProps> = props => {
```
devinleighsmith, line 27: This seems more like a container then a view to me, lots of logic and almost no display

backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -79,6 +84,67 @@ public PimsResearchFile Update(PimsResearchFile researchFile)
...
 
+        /// <summary>
+        /// Update the properties on the research file.
+        /// </summary>
+        /// <param name="researchFileId"></param>
+        /// <param name="researchFileProperty"></param>
+        /// <returns></returns>
+        public PimsResearchFile UpdateProperty(long researchFileId, PimsPropertyResearchFile researchFileProperty)
+        {
+            var currentTypes = Context.PimsPropertyResearchFiles
```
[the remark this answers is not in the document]
FuriousLlama: I spent quite some time trying to make it work but got quite a lot of issues. IMO having the logic like this is more explicit anyway.

frontend/src/features/properties/map/research/detail/PropertyResearchView.tsx
devinleighsmith: alternatively we could revisit this:

Naming
Top level react components accessible via the main AppRouter should be postfixed with Page.
A Page is composed of one or more Containers, that provide logic for child Views.
A View is ideally a stateless via only react component. All logic should be provided by a Container.
A View is comprised of multiple components, which do not have an enforced naming pattern.
Note that there may be multiple Containers in a Page, and that a Container may contain other nested Containers.
Note that some components will not implement the previous pattern, but all new code should.

(from https://confluence.th.gov.bc.ca/display/PIMS/Best+Practices)

devinleighsmith (approved):

backend/dal/Services/ResearchFileService.cs
FuriousLlama: I've updated it since the type exists in the db

backend/dal/Services/ResearchFileService.cs
```diff
@@ -85,5 +84,37 @@ public Paged<PimsResearchFile> GetPage(ResearchFilter filter)
...
+            property.PropertyDataSourceEffectiveDate = System.DateTime.Now;
+            property.PropertyDataSourceTypeCode = "PAIMS"; // Todo: should be "PMBC"
```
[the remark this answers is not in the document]
FuriousLlama: same as above

frontend/src/features/properties/map/research/detail/PropertyResearchView.tsx
FuriousLlama: Fair point, I think I just named view since it was at the same level of the other files

PR #1381, opened 2022-05-11
Changes to create the UserNameTooltip component

Maheshbabu28:


frontend/src/components/common/UserNameTooltip.tsx
```diff
@@ -0,0 +1,48 @@
...
+/** Generic user info tooltip component that displays user name on hover */
+export const UserNameTooltip: React.FunctionComponent<IUserNameTooltipProps> = ({
+  userGuid,
+  userName,
+}) => {
+  const { getUserInfo } = useApiUsers();
+  const [userNameInfo, setUserNameInfo] = React.useState<string>('');
+  React.useEffect(() => {
+    if (userGuid) {
+      getUserInfo(userGuid).then(({ data }) => {
```
devinleighsmith, line 21: this is fine for this sprint, but in future sprints lets start using the apiWrapper. That handles errors, as well as not updating state when the component is unmounted.

devinleighsmith: For this sprint at least you'll want to add an ismounted check in addition to the check for data.

frontend/src/features/properties/map/research/common/ResearchHeader.tsx
```diff
@@ -26,7 +27,11 @@ const ResearchHeader: React.FunctionComponent<IResearchHeaderProps> = props => {
           </Col>
           <Col className="text-right">
             Created: <strong>{prettyFormatDate(researchFile?.appCreateTimestamp)}</strong> by
-            <span> {researchFile?.appCreateUserid} </span>
+            <UserNameTooltip
+              key={'appCreateBy'}
```
devinleighsmith, line 31: if this is a string you can just use "" instead of {''}

devinleighsmith: are you using the key within the usernametooltip?

devinleighsmith (approved):

frontend/src/components/common/UserNameTooltip.tsx
Maheshbabu28: For now I have used ismounted and pushed the changes

frontend/src/features/properties/map/research/common/ResearchHeader.tsx
Maheshbabu28: For now I have removed the `key` as I thought its not needed for now.

PR #1420, opened 2022-05-16
Add missing line so that Property gets mapped. fixes psp-3449

devinleighsmith:
@FuriousLlama please review. I'm assuming that this was not intentional but want to make sure that this change wont break anything else.

### DIFF
--- a/backend/api/Models/Concepts/ResearchFile/ResearchFilePropertyMap.cs
+++ b/backend/api/Models/Concepts/ResearchFile/ResearchFilePropertyMap.cs
@@ -24,6 +24,7 @@ namespace Pims.Api.Models.Concepts
 
             config.NewConfig<ResearchFilePropertyModel, Entity.PimsPropertyResearchFile>()
                 .Map(dest => dest.PropertyResearchFileId, src => src.Id)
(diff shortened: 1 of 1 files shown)


devinleighsmith:
@FuriousLlama for context, this prevents any research files with at least one property from being saved.

backend/api/Models/Concepts/ResearchFile/ResearchFilePropertyMap.cs
```diff
@@ -24,6 +24,7 @@ public void Register(TypeAdapterConfig config)
 
             config.NewConfig<ResearchFilePropertyModel, Entity.PimsPropertyResearchFile>()
                 .Map(dest => dest.PropertyResearchFileId, src => src.Id)
+                .Map(dest => dest.Property, src => src.Property)
```
devinleighsmith, line 27: @FuriousLlama this is the line I'm not sure about.

FuriousLlama: That looks very much necessary. Not sure why it was not there

FuriousLlama (approved):

PR #1429, opened 2022-05-18
PSP-3011 Edit property info

asanchezr:
Screenshot 2022-05-17 175109


backend/dal/Repositories/Interfaces/ILookupService.cs
```diff
@@ -40,6 +41,10 @@ public interface ILookupService : IRepository
         IEnumerable<PimsRequestSourceType> GeRequestSourceTypes();
         IEnumerable<PimsResearchPurposeType> GetResearchPurposeTypes();
         IEnumerable<PimsPropResearchPurposeType> GetPropertyResearchPurposeTypes();
+        IEnumerable<PimsPropertyAnomalyType> GetPropertyAnomalyTypes();
+         IEnumerable<PimsPropertyRoadType> GetPropertyRoadTypes();
```
devinleighsmith, line 45: nit: weird formatting.

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsContainer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    formikHelpers.setSubmitting(false);
+  };
+
+  const onCancel = useCallback(() => {
+    if (!!initialForm?.pid) {
+      history.push(`/mapview/property/${initialForm.pid}`);
+    }
+  }, [history, initialForm?.pid]);
+
+  if (initialForm === undefined) {
```
devinleighsmith, line 73: Could we update this to use the loading booleans from the wrapped api requests instead?

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsForm.tsx
```diff
@@ -0,0 +1,267 @@
...
+  );
+  const roadTypeOptions = getByType(API.PROPERTY_ROAD_TYPES).map(x =>
+    PropertyRoadFormModel.fromLookup(x),
+  );
+  const adjacentLandOptions = getByType(API.PROPERTY_ADJACENT_LAND_TYPES).map(x =>
+    PropertyAdjacentLandFormModel.fromLookup(x),
+  );
+
+  // multi-selects
+  const tenureStatus = getIn(values, 'tenures') as PropertyTenureFormModel[];
```
devinleighsmith, line 56: Why use getIn instead of just values.tenures ? values.tenures seems better because it is more concise and doesn't rely on type coercion.

asanchezr: no hard preference there, can change. was kinda using formik helpers but agree on type coercion

frontend/src/features/mapSideBar/tabs/propertyDetails/update/components/LandMeasurementTable.tsx
```diff
@@ -0,0 +1,197 @@
...
+        };
+      });
+    }
+  }, [focus, ha]);
+
+  useEffect(() => {
+    if (focus === AreaUnitTypes.Acres) {
+      setState(prevState => {
+        const value = prevState[AreaUnitTypes.Acres];
+        const newSqMeters = convertArea(value, AreaUnitTypes.Acres, AreaUnitTypes.SquareMeters);
```
devinleighsmith, line 99: nit: extract out 99-107 into a common function to reduce duplicated code.

frontend/src/features/mapSideBar/tabs/propertyDetails/update/components/LandMeasurementTable.tsx
```diff
@@ -0,0 +1,197 @@
...
+              </div>
+              <div role="cell" title="" className="td left">
+                hectares
+              </div>
+            </div>
+          </div>
+          <div className="tr-wrapper">
+            <div role="row" className="tr">
+              <div role="cell" title="" className="td right">
+                <input
```
devinleighsmith, line 181: Just want to put a reminder here that you can override the contents of a cell with a custom component (such as an input).
`{
    Header: 'Summary',
    accessor: 'summary',
    align: 'left',
    clickable: true,
    sortable: true,
    width: 80,
    maxWidth: 120,
    Cell: (props: CellProps<IContactSearchResult>) => {
      if (hasClaim(Claims.CONTACT_VIEW)) {
        return <Link to={`/contact/${props.row.original.id}`}>{props.row.original.summary}</Link>;
      }
      return props.row.original.summary;
    },
  },`

asanchezr: Not sure if you recall that I mentioned a re-rendering issue with that approach. I tried using a table and putting the input like you show above but the whole table was re-rendering on each keypress so the input would loose focus (as it is a new rendered component) and that didn't work.

So I had to "re-create" the look and feel of a table manually. It works because `Table.scss` exists and the styles are global. If/when Table gets converted to styled components we would loose those styles but I felt that I didn't need to duplicate the css for now

frontend/src/hooks/useQueryMapLayersByLocation.ts
```diff
@@ -0,0 +1,84 @@
...
+  const queryAllCallback = useCallback(
+    async function(location: LatLngLiteral | null): Promise<IMapLayerResults> {
+      // Query BC Geographic Warehouse layers - ONLY if lat, long have been provided!
+      if (location === null) {
+        return initialState;
+      }
+
+      try {
+        // We are using spatial reference = 3005 (BC Albers) here because that's how the backend is returning spatial location
+        const alr = await alrService.findOneWhereContains(location, 'GEOMETRY', 3005);
```
devinleighsmith, line 50: I think something like this would help with performance
```
/** 
 * These will be run in series, because we call
 * a function and immediately wait for each result, 
 * so this will finish in 1s.
 */
async function series() {
  return {
    result1: await wait(500, 'seriesTask1'),
    result2: await wait(500, 'seriesTask2'),
  }
}

/** 
 * While here we call the functions first,
 * then wait for the result later, so 
 * this will finish in 500ms.
 */
async function parallel() {
  const task1 = wait(500, 'parallelTask1');
  const task2 = wait(500, 'parallelTask2');

  return {
    result1: await task1,
    result2: await task2,
  }
}
```

frontend/src/hooks/useQueryMapLayersByLocation.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+async function findByLocation(
+  service: IUserLayerQuery,
+  latlng: LatLngLiteral,
+  geometryName: string = 'SHAPE',
+  spatialReferenceId: number = 4326,
+) {
+  const response = await service.findOneWhereContains(latlng, geometryName, spatialReferenceId);
+  const featureProps = response?.features?.[0]?.properties;
```
devinleighsmith, line 82: I think this will throw an array out of bounds if you click outside of a layer.

frontend/src/models/api/AuditFields.ts
```diff
@@ -0,0 +1,6 @@
+export interface Api_AuditFields {
```
devinleighsmith, line 1: This is nice.

frontend/src/hooks/useQueryMapLayersByLocation.ts
asanchezr: you are right - will add array length check

frontend/src/hooks/useQueryMapLayersByLocation.ts
asanchezr: will change

devinleighsmith (approved):

asanchezr:
Will address PR comments in my next PR

PR #1437, opened 2022-05-19
feat(PSP-3379):Update Research File - Header Behaviour

Maheshbabu28:


devinleighsmith (approved):

devinleighsmith:
I checked this out locally, looks good. Seems like this was easier then we anticipated.

PR #1438, opened 2022-05-19
limit sonarqube scan, failure notification to runs within bcgov repo …

devinleighsmith:
…context.

### DIFF
--- a/.github/workflows/api-dotnetcore.yml
+++ b/.github/workflows/api-dotnetcore.yml
@@ -101,7 +101,7 @@ jobs:
       - name: SonarScanner for .NET 6 with pull request decoration support
         id: scan
         uses: highbyte/sonarscan-dotnet@v2.1.2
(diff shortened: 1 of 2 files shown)


FuriousLlama (approved):

.github/workflows/api-dotnetcore.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - name: SonarScanner for .NET 6 with pull request decoration support
         id: scan
         uses: highbyte/sonarscan-dotnet@v2.1.2
-        if: github.event_name == 'push'
+        if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
```
asanchezr, line 104: question: shouldn't this be AND (&&) instead of OR?
Seems to be that it would still run when I push to my origin...

PR #1445, opened 2022-05-20
psp-3053 fix contact list view count and add mocked contact repositor…

devinleighsmith:
…y tests.


backend/tests/unit/mockdal/Pims.Dal.Mock.Test.csproj
```diff
@@ -0,0 +1,38 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
devinleighsmith, line 1: New test library - Moq.EntityFrameworkCore required to allow testing of keyless entities.
Is not compatible with in memory test database.
Created a new testing library to isolate tests that will use the above library.

FuriousLlama (approved):

PR #1446, opened 2022-05-20
PSP-3383 Add Region and District to properties in research file

asanchezr:
image


backend/api/Mapping/TypeMap.cs
```diff
@@ -20,6 +20,12 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.IsDisabled, src => src.IsDisabled)
                 .Map(dest => dest.DisplayOrder, src => src.DisplayOrder);
 
+            config.NewConfig<Model.TypeModel<short>, Entity.ICodeEntity<short>>()
+                .Map(dest => dest.Code, src => src.Id)
+                .Map(dest => dest.Description, src => src.Description)
+                .Map(dest => dest.IsDisabled, src => src.IsDisabled)
+                .Map(dest => dest.DisplayOrder, src => src.DisplayOrder);
+
```
asanchezr, line 28: Missing mapping

backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -48,6 +48,10 @@ public PimsResearchFile GetById(long id)
                 .Include(r => r.RequestorOrganizationNavigation)
                 .Include(r => r.PimsPropertyResearchFiles)
                     .ThenInclude(rp => rp.Property)
+                    .ThenInclude(p => p.RegionCodeNavigation)
+                .Include(r => r.PimsPropertyResearchFiles)
+                    .ThenInclude(rp => rp.Property)
+                    .ThenInclude(p => p.DistrictCodeNavigation)
```
asanchezr, line 54: Need to expose region and district when fetching research files

backend/dal/Services/ResearchFileService.cs
```diff
@@ -101,11 +101,8 @@ private void PopulateResearchFile(PimsProperty property)
 
             property.PropertyTypeCode = "UNKNOWN";
 
-            property.PropertyStatusTypeCode = "UNSURVYED"; // Todo: should be 'UNKNOWN';
+            property.PropertyStatusTypeCode = "UNKNOWN";
```
asanchezr, line 104: Now (IS28) we have UNKNOWN as a valid value in the DB

frontend/src/components/maps/hooks/useActiveFeatureLayer.test.tsx
```diff
@@ -15,16 +15,18 @@ import useActiveFeatureLayer from './useActiveFeatureLayer';
...
 });
-(useLayerQuery as jest.Mock).mockReturnValue({
-  findOneWhereContains: findOneWhereContains,
-});
+
+const useLayerQueryMock = {
+  findOneWhereContains: jest.fn(),
+  findMetadataByLocation: jest.fn(),
+};
+(useLayerQuery as jest.Mock).mockReturnValue(useLayerQueryMock);
```
asanchezr, line 29: New function needs to be mocked

frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
```diff
@@ -55,7 +61,14 @@ const useActiveFeatureLayer = ({
...
 
-    const parcel = await parcelsService.findOneWhereContains(latLng);
+    // call these APIs in parallel - notice there is no "await"
+    const task1 = parcelsService.findOneWhereContains(latLng);
+    const task2 = regionService.findMetadataByLocation(latLng, 'GEOMETRY');
+    const task3 = districtService.findMetadataByLocation(latLng, 'GEOMETRY');
+
+    const parcel = await task1;
+    const region = await task2;
+    const district = await task3;
```
asanchezr, line 71: The comment is quite explicit - we prefer to make these requests in parallel, rather than awaiting each task in sequential order

frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
```diff
@@ -82,7 +95,7 @@ const useActiveFeatureLayer = ({
       feature = {
         geometry: { coordinates: [latLng.lng, latLng.lat], type: 'Point' },
         type: 'Feature',
-        properties: [],
+        properties: {},
```
asanchezr, line 98: `Feature.properties` is an object (dictionary), not an array

frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
```diff
@@ -102,6 +115,12 @@ const useActiveFeatureLayer = ({
         ...feature.properties,
         IS_SELECTED: isSelecting,
         CLICK_LAT_LNG: latLng,
+        REGION_NUMBER: isNumber(region.REGION_NUMBER) ? region.REGION_NUMBER : RegionCodes.Unknown,
+        REGION_NAME: region.REGION_NAME ?? 'Cannot determine',
+        DISTRICT_NUMBER: isNumber(district.DISTRICT_NUMBER)
+          ? district.DISTRICT_NUMBER
+          : DistrictCodes.Unknown,
+        DISTRICT_NAME: district.DISTRICT_NAME ?? 'Cannot determine',
```
asanchezr, line 123: Acceptance criteria call for:
* If no match is found map it to Type Unknown - "Can not determine."

frontend/src/constants/districtCodes.ts
```diff
@@ -0,0 +1,3 @@
+export enum DistrictCodes {
+  Unknown = 12,
+}
```
asanchezr, line 3: Value here matches the DB

frontend/src/constants/regionCodes.ts
```diff
@@ -0,0 +1,3 @@
+export enum RegionCodes {
+  Unknown = 4,
+}
```
asanchezr, line 3: Value here matches the DB

frontend/src/features/properties/selector/PropertySelectorSubForm.tsx
```diff
@@ -33,6 +37,12 @@ export const PropertySelectorSubForm: React.FunctionComponent<IPropertySelectorS
         <SectionField label="PID">{pid}</SectionField>
         <SectionField label="Plan #">{planNumber}</SectionField>
         <SectionField label="Address">{address}</SectionField>
+        <SectionField label="Region">
+          {[region, regionName].filter(Boolean).join(' - ')}
+        </SectionField>
+        <SectionField label="District">
+          {[district, districtName].filter(Boolean).join(' - ')}
+        </SectionField>
```
asanchezr, line 45: Took the liberty to include the region and district in the property picker sub-form as I think it is quite useful information. Happy to take it out or modify if UX deems it not useful. See screenshot for new look

backend/api/Mapping/TypeMap.cs
FuriousLlama: the catch with this is that we need to make sure that when updating the type model should not be created most of the time

frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
```diff
@@ -55,7 +61,14 @@ const useActiveFeatureLayer = ({
     let title = 'Municipality Information';
     let feature: Feature | undefined = undefined;
 
-    const parcel = await parcelsService.findOneWhereContains(latLng);
+    // call these APIs in parallel - notice there is no "await"
+    const task1 = parcelsService.findOneWhereContains(latLng);
+    const task2 = regionService.findMetadataByLocation(latLng, 'GEOMETRY');
+    const task3 = districtService.findMetadataByLocation(latLng, 'GEOMETRY');
```
FuriousLlama, line 67: nit: could be retrieve parcelTask, retrieveRegionTask ...

asanchezr: agree - but felt that they are close together so task1..taskn was okay

frontend/src/components/maps/leaflet/LayerPopup/hooks/useLayerQuery.tsx
```diff
@@ -126,13 +140,29 @@ export const useLayerQuery = (url: string): IUserLayerQuery => {
...
 
+  const findMetadataByLocation = useCallback(
+    async (
+      latlng: LatLngLiteral,
+      geometryName: string = 'SHAPE',
+      spatialReferenceId: number = 4326,
+    ): Promise<Record<string, any>> => {
+      const collection = await findOneWhereContains(latlng, geometryName, spatialReferenceId);
+      if (collection?.features?.length > 0) {
+        return collection.features[0].properties || {};
```
FuriousLlama, line 151: would it not be better to just return undefined instead of an empty object?

FuriousLlama (approved):

frontend/src/components/maps/leaflet/LayerPopup/hooks/useLayerQuery.tsx
asanchezr: In my opinion an empty object here is okay as it is saying "there is no metadata". That way when you ask for obj.SOME_FIELD you don't have to check for undefined first. 

Again personal preference. The method signature is that it returns a `Record<string,any>` not `Record | undefined`

backend/api/Mapping/TypeMap.cs
asanchezr: I will remove it then to avoid bugs - we can re-add if needed

PR #1452, opened 2022-05-24
Bugs/psp 3511

FuriousLlama:


frontend/src/features/properties/map/research/update/property/UpdatePropertyForm.tsx
```diff
@@ -63,7 +63,7 @@ const UpdatePropertyForm: React.FunctionComponent<IUpdatePropertyFormProps> = pr
       <StyledFormSection>
         <StyledSectionHeader>Property of Interest</StyledSectionHeader>
         <SectionField label="Descriptive name">
-          <Input field="propertyName" />
+          <Input field="propertyName" pattern={/^.{1,499}$/} />
```
devinleighsmith, line 66: This works, although I think I prefer using yup for this in the future.

devinleighsmith (approved):

devinleighsmith:
This is a good point however, we should make a point of enforcing these character limits in our yup schemas across the board. It may make more sense to do this holistically. I will log tech debt for this.

PR #1454, opened 2022-05-24
feat(PSP-3230): Changes to enable sorting on the Lease Result table

Maheshbabu28:
- Added a custom method in LeaseExtensions.cs to map the model sort fields to entity soft fields
 - Enabled sorting in the LeaseSearchResults.tsx for required columns and updated the UTs to pass according to change

### DIFF
--- a/backend/dal/Helpers/Extensions/LeaseExtensions.cs
+++ b/backend/dal/Helpers/Extensions/LeaseExtensions.cs
@@ -107,6 +107,10 @@ namespace Pims.Dal.Helpers.Extensions
 
             if (filter.Sort?.Any() == true)
             {
+                MapSortField("ExpiryDate", "OrigExpiryDate", filter.Sort);
+                MapSortField("StatusType", "LeaseStatusTypeCodeNavigation", filter.Sort);
+                MapSortField("ProgramName", "LeaseProgramTypeCodeNavigation.Description", filter.Sort);
+
                 query = query.OrderByProperty(filter.Sort);
             }
             else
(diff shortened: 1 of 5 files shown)


FuriousLlama (approved):

backend/dal/Helpers/Extensions/LeaseExtensions.cs
```diff
@@ -140,6 +144,25 @@ private static IQueryable<Entities.PimsLease> GenerateCommonLeaseQuery(this IQue
...
+        /// <param name="sourceField">Sort field name from model</param>
+        /// <param name="targetField">Sort field name from entity</param>
+        /// <param name="sortDef">Find and replaces the soft field in this array</param>
+        private static void MapSortField(string sourceField, string targetField, string[] sortDef)
+        {
+            var sortField = sortDef.FirstOrDefault(x => x.Contains(sourceField));
+            if (sortField != null)
+            {
+                var sortFieldIndex = sortDef.ToList().IndexOf(sortField);
+                if (sortFieldIndex > -1)
```
devinleighsmith, line 159: I think I'm in favour of throwing an error here, as this not having a matching sort field is likely a development-time error which we would want to fail fast on.

devinleighsmith (approved):

backend/dal/Helpers/Extensions/LeaseExtensions.cs
FuriousLlama: That makes sense. I am thinking we could improve the sorting + filters to avoid having possible runtime errors

PR #1484, opened 2022-06-03
PSP-3011 unit tests

asanchezr:


backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -136,8 +136,7 @@ public PimsProperty Get(long id)
                     .ThenInclude(l => l.Lease)
                     .ThenInclude(l => l.PimsLeaseTenants)
                     .ThenInclude(l => l.Organization)
-                .Where(p => p.PropertyId == id)
-                .FirstOrDefault() ?? throw new KeyNotFoundException();
+                .FirstOrDefault(p => p.PropertyId == id) ?? throw new KeyNotFoundException();
```
asanchezr, line 139: stylecop suggestion

backend/entities/Property.cs
```diff
@@ -60,6 +60,8 @@ public PimsProperty(int pid, PimsPropertyType type, PimsPropertyClassificationTy
             this.PropertyDataSourceTypeCodeNavigation = dataSource ?? throw new ArgumentNullException(nameof(dataSource));
             this.PropertyDataSourceTypeCode = dataSource.Id;
             this.PropertyDataSourceEffectiveDate = dataSourceEffectiveDate;
+            this.PropertyStatusTypeCodeNavigation = status ?? throw new ArgumentNullException(nameof(status));
+            this.PropertyStatusTypeCode = status.Id;
```
asanchezr, line 64: Required relationship was missing on unit tests

backend/api/Areas/Property/Controllers/PropertyController.cs
```diff
@@ -112,15 +112,15 @@ public IActionResult GetConceptPropertyWithPid(string pid)
...
+        /// Update the specified property, and attached properties.
         /// </summary>
         /// <returns></returns>
         [HttpPut("concept/{pid}")]
         [HasPermission(Permissions.PropertyEdit)]
         [Produces("application/json")]
         [ProducesResponseType(typeof(Pims.Api.Models.Concepts.PropertyModel), 200)]
         [SwaggerOperation(Tags = new[] { "property" })]
-        public IActionResult UpdatePerson([FromBody] Pims.Api.Models.Concepts.PropertyModel propertyModel)
+        public IActionResult UpdateConceptProperty([FromBody] Pims.Api.Models.Concepts.PropertyModel propertyModel)
```
FuriousLlama, line 123: Why the change of name?

backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
```diff
@@ -0,0 +1,209 @@
...
+        public void GetPage_Properties(PropertyFilter filter, int expectedCount)
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission(Permissions.PropertyView);
+
+            using var init = helper.InitializeDatabase(user);
+
+            init.CreateProperty(2);
+            init.CreateProperty(3, pin: 111);
```
FuriousLlama, line 86: I think creating a property and passing os more readable than just calling CreateProperty

backend/api/Areas/Property/Controllers/PropertyController.cs
asanchezr: It was a copy/paste mistake from previous PR - this is the property controller, not the person controller

backend/tests/unit/dal/Services/PropertyServiceTest.cs
```diff
@@ -0,0 +1,173 @@
...
+        #region Tests
+        #region GetById
+        [Fact]
+        public void GetById_Success()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission(Permissions.PropertyView);
+
+            var property = EntityHelper.CreateProperty(1);
```
FuriousLlama, line 31: it does not look like the property is being used. Is it necessary to be created?

asanchezr: Look at the next line - it is adding it to the in-memory DB

```cs
helper.CreatePimsContext(user, true).AddAndSaveChanges(property);
```

backend/tests/unit/dal/Services/PropertyServiceTest.cs
```diff
@@ -0,0 +1,173 @@
...
+            helper.CreatePimsContext(user, true).AddAndSaveChanges(property);
```
FuriousLlama, line 32: if we are testing the service, is the context necessary?

backend/tests/unit/dal/Services/PropertyServiceTest.cs
asanchezr: We need to add it to be able to find it by ID - the whole purpose of this test

backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
```diff
@@ -0,0 +1,209 @@
...
+            // Assert
+            Assert.Throws<KeyNotFoundException>(() => repository.Update(property));
+        }
+
+        [Fact]
+        public void Update_Property_ThrowIfNull()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission(Permissions.PropertyView);
```
FuriousLlama, line 196: is the user relevant for repository tests?

backend/tests/unit/dal/Services/PropertyServiceTest.cs
asanchezr: I guess we might not need it but I followed other existing tests that do that. Wouldn't want to go a different way and find that it doesn't work. In any case, do you see harm in that?

backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
asanchezr: This is an existing test - it appears as new because I renamed this file from PropertyServiceTest -> PropertyRepositoryTest

I'd lean to not changing it for now as it has been there before - unless there is something fundamentally wrong with it

backend/tests/unit/dal/Services/PropertyServiceTest.cs
FuriousLlama: Is just that any code that does not add value should be removed if possible.

backend/tests/unit/dal/Services/PropertyServiceTest.cs
FuriousLlama: Looks like the mock on line 36 is not checking the ID passed. Line 42 is not checking that the property passed is being returned (In this case wont be returned given the repository mocked function is not using the created context). So from that I gather that the property is not being used for the purpose of the test

FuriousLlama (approved):
It would be a good idea to revisit our testing practices

backend/tests/unit/dal/Services/PropertyServiceTest.cs
asanchezr: ok - will remove inserting the property

PR #1494, opened 2022-06-07
Updated property update to have selects for region and districtg

FuriousLlama:
image

### DIFF
--- a/backend/api/Areas/Property/Models/Property/PropertyModel.cs
+++ b/backend/api/Areas/Property/Models/Property/PropertyModel.cs
@@ -113,6 +113,16 @@ namespace Pims.Api.Areas.Property.Models.Property
         public TypeModel<string> AreaUnit { get; set; }
 
         /// <summary>
+        /// get/set - District type.
+        /// </summary>
+        public TypeModel<short> DistrictType { get; set; }
+
+        /// <summary>
+        /// get/set - Region type.
+        /// </summary>
+        public TypeModel<short> RegionType { get; set; }
+
+        /// <summary>
         /// get/set - The land area of the parcel.
         /// </summary>
         public float LandArea { get; set; }
--- a/frontend/src/features/mapSideBar/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
+++ b/frontend/src/features/mapSideBar/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
@@ -72,13 +72,16 @@ const FormComponent: React.FC<FormikProps<IPropertyDetailsForm>> = ({ values }) 
     <StyledReadOnlyForm>
       <Section header="Property attributes">
         <SectionField label="MOTI region">
-          <Text field="motiRegion.REGION_NAME" />
+          <Text field="regionType.description" />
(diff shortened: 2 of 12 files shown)


stairaku (approved):

backend/api/Areas/Property/Models/Property/PropertyModel.cs
```diff
@@ -112,6 +112,16 @@ public class PropertyModel : BaseModel
...
+        /// <summary>
+        /// get/set - District type.
+        /// </summary>
+        public TypeModel<short> DistrictType { get; set; }
+
+        /// <summary>
+        /// get/set - Region type.
+        /// </summary>
+        public TypeModel<short> RegionType { get; set; }
+
```
asanchezr, line 124: nit: would prefer if these are named District, Region to align better with the "concept" Property. Ideally in the future we would want to refactor this specific model into the concept one

frontend/src/features/mapSideBar/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
```diff
@@ -72,13 +72,16 @@ const FormComponent: React.FC<FormikProps<IPropertyDetailsForm>> = ({ values })
...
         <SectionField label="MOTI region">
-          <Text field="motiRegion.REGION_NAME" />
+          <Text field="regionType.description" />
         </SectionField>
         <SectionField label="Highways district">
           <InlineContainer>
-            <Text field="highwaysDistrict.DISTRICT_NUMBER" />
-            {'-'}
-            <Text field="highwaysDistrict.DISTRICT_NAME" />
+            {values.districtType?.description !== 'Cannot determine' && (
```
asanchezr, line 79: nit: would you consider using a constant for 'Cannot determine' ? might help re-use in other places

frontend/src/interfaces/IPropertyApiModel.ts
```diff
@@ -36,6 +36,9 @@ export interface IPropertyApiModel extends Api_ConcurrentVersion {
   propertyType?: Api_TypeCode<string>;
   status?: Api_TypeCode<string>;
 
+  regionType?: Api_TypeCode<number>;
+  districtType?: Api_TypeCode<number>;
+
```
asanchezr, line 41: as per previous comment would love to see this named just region, district to align with concept API_Property

asanchezr (approved):

frontend/src/features/mapSideBar/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
FuriousLlama: I thought about but given it only is needed on this page decided to leave it as is. We could have a enum for regions in the future

frontend/src/interfaces/IPropertyApiModel.ts
FuriousLlama: the problem is that these are not part of the concept, but the old model. We should get rid of it completly

PR #1499, opened 2022-06-07
PSP-3231 Convert leaflet lat/long from 4326 to 3005

asanchezr:


backend/api/Models/Concepts/Property/GeometryMap.cs
```diff
@@ -12,7 +12,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Coordinate, src => src.Coordinate);
 
             config.NewConfig<GeometryModel, Geometry>()
-                .ConstructUsing(src => GeometryHelper.CreatePoint(src.Coordinate.X, src.Coordinate.Y));
+                .ConstructUsing(src => GeometryHelper.CreatePoint(src.Coordinate.X, src.Coordinate.Y, 4326));
```
FuriousLlama, line 15: nit: would be better to use a constants instead of a number

asanchezr: good idea - done

backend/dal/Helpers/GeometryHelper.cs
```diff
@@ -1,20 +1,33 @@
 namespace Pims.Dal.Helpers
 {
     /// <summary>
-    /// GeometryHelper static class, provides methods to help with goemetric shapes.
+    /// GeometryHelper static class, provides methods to help with geometric shapes.
```
asanchezr, line 4: fixing typo

backend/dal/Services/CoordinateTransformService.cs
```diff
@@ -0,0 +1,87 @@
...
+            var enumeration = DefaultProjections();
+            foreach (var sridWkt in enumeration)
+            {
+                var cs = CreateCoordinateSystem(sridWkt.Value);
+                if (cs != null)
+                {
+                    _projections.Add(sridWkt.Key, cs);
+                }
+            }
+        }
```
FuriousLlama, line 48: This is probably fine, but it looks like commented out code on git

asanchezr: hmm - looks weird it should not be commented out.

asanchezr: image

backend/dal/Services/CoordinateTransformService.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <summary>
+        /// Creates an instance of this class
+        /// </summary
+        public CoordinateTransformService()
+        {
+            var enumeration = DefaultProjections();
+            foreach (var sridWkt in enumeration)
```
FuriousLlama, line 40: is sridWkt a known abreviation?

backend/dal/Services/CoordinateTransformService.cs
asanchezr: Looks/works good on VS Code - perhaps github colorization gets confused by the documentation?

backend/dal/Services/CoordinateTransformService.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                var cs = CreateCoordinateSystem(sridWkt.Value);
```
FuriousLlama, line 42: nit cs=> coordinateSytstem

backend/dal/Services/CoordinateTransformService.cs
asanchezr: yes, srid = spatial reference id (e.g. 3005)
wkt = well known text (e.g. the string that defines the projection parameters)

backend/dal/Services/PropertyService.cs
```diff
@@ -12,33 +14,59 @@ public class PropertyService : IPropertyService
...
             _logger.LogInformation("Getting property with pid {pid}", pid);
             _user.ThrowIfNotAuthorized(Permissions.PropertyView);
-            return _propertyRepository.GetByPid(pid);
+
+            // return property spatial location in lat/long (4326)
+            var property = _propertyRepository.GetByPid(pid);
+            if (property?.Location != null)
+            {
+                var newCoords = _coordinateService.TransformCoordinates(SpatialReference.BC_ALBERS, SpatialReference.WGS_84, property.Location.Coordinate);
+                property.Location = GeometryHelper.CreatePoint(newCoords, SpatialReference.WGS_84);
```
FuriousLlama, line 52: since this is overriding the entity information, what do you think about using a new variable instead? Not to be done now but something to think about since the location does not represent what is in the repository anymore.

asanchezr: I've been thinking about where's the best place to do this conversion. We can talk more this Friday and decide on approach.

PR #1583, opened 2022-06-17
PSP-3648 bug fix - When adding a pin based on Plan# search, 400 error is thrown

asanchezr:
Root cause was not sending MOTI region and district for properties added via text search - e.g. by PID, PIN, Address


frontend/src/features/properties/selector/search/LayerFilter.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { mockGeocoderOptions } from 'hooks/pims-api/useApiGeocoder';
+import { mockGeocoderOptions } from 'mocks';
```
asanchezr, line 1: Moved mock from hook to dedicated **mocks** folder for easier discoverability

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.test.tsx
```diff
@@ -39,87 +44,149 @@ describe('PropertySelectorSearchContainer component', () => {
...
+          'https://openmaps.gov.bc.ca/geo/pub/WHSE_ADMIN_BOUNDARIES.TADM_MOT_DISTRICT_BNDRY_POLY/wfs',
+        ),
+      )
+      .reply(200, mockDistrictLayerResponse)
+      .onGet(new RegExp('tools/geocoder/parcels/pids/*'))
+      .reply(200, mockGeocoderPidsResponse)
+      .onGet(new RegExp('tools/geocoder/addresses'))
+      .reply(200, mockGeocoderOptions);
+  });
+
```
asanchezr, line 76: This mocks requests to various services consumed by the form being tests. A bit of an improvement over a generic `mockAxios.onGet().reply(...)` that applies to every web request

frontend/src/hooks/pims-api/useApiGeocoder.ts
```diff
@@ -25,31 +25,3 @@ export const useApiGeocoder = () => {
 };
 
 export type IPaginateResearch = IPaginateRequest<IResearchFilter>;
-
-export const mockGeocoderPidsResponse: IGeocoderPidsResponse = {
-  pids: ['9025ea39-da19-4655-85e6-e34c00e765f8'],
-  siteId: '312312',
-};
-
-export const mockGeocoderOptions: IGeocoderResponse[] = [
```
FuriousLlama, line 34: +1 Thank you for removing the test code from the production code!

PR #1612, opened 2022-06-20
Psp 3656 unit tests for keycloak, access request, user management.

devinleighsmith:


backend/api/appsettings.json
```diff
@@ -58,7 +58,7 @@
       "Logout": "/protocol/openid-connect/logout",
       "Register": "/protocol/openid-connect/registrations",
       "Token": "/protocol/openid-connect/token",
-      "TokenIntrospect": "/protocol/openid-connect/token/introspect",
+      "TokenIntrospect": "/protocol/openid-connect/token/introspect/fake",
```
asanchezr, line 61: double checking that this URL is correct?

backend/dal/Repositories/Interfaces/IAccessRequestService.cs
```diff
@@ -4,9 +4,9 @@
 namespace Pims.Dal.Repositories
 {
     /// <summary>
-    /// IAccessRequestService interface, provides functions to interact with access requests within the datasource.
+    /// IAccessRequestRepository interface, provides functions to interact with access requests within the datasource.
```
asanchezr, line 7: nit: ideally the file name should be renamed as well to match

backend/tests/unit/dal/Repositories/LeaseReportsServiceTest.cs
```diff
@@ -25,7 +26,7 @@ namespace Pims.Dal.Test.Services
     public class LeaseReportsServiceTest : IDisposable
```
asanchezr, line 26: nit: ideally file name could be changed to repository

asanchezr (approved):

backend/api/appsettings.json
devinleighsmith: this is unused, but will change it back to match the endpoint.

PR #1621, opened 2022-06-20
feature(PSP-3663): Enforce character limits on FE - User Management

Maheshbabu28:


devinleighsmith:
Something is off with this PR.
1) This PR is for user management field limits but includes a change to research files.
2) This PR does not contain any changes to the edit user yup schema, only to the access request yup schema.

devinleighsmith (changes requested):
see previous comment

Maheshbabu28:
1 - frontend/src/utils/YupSchema.ts Contains changes for Update User and Access User Request both
2 - Removed extra file that was added by mistake

devinleighsmith (approved):

PR #1645, opened 2022-06-23
version bumps requested by dependabot.

devinleighsmith:


frontend/package.json
```diff
@@ -22,7 +22,7 @@
     "moment": "2.29.2",
     "moment-timezone": "0.5.33",
     "multiselect-react-dropdown": "2.0.21",
-    "npm": "8.1.0",
+    "npm": "8.11.0",
```
asanchezr, line 25: @devinleighsmith  as I mentioned in the original dependabot PR this change will break our build. We need to stay at 8.5.5

This change is problematic because our frontend does not build with npm 8.11 due to some changes in peer dependency resolution. The latest we can use is 8.5.5

asanchezr: I suggest you bump it to 8.5.5 - anything beyond that will break our frontend build

asanchezr (approved):

PR #1653, opened 2022-06-24
psp-3247 add acquisition files link from map. Unit tests.

devinleighsmith:


frontend/src/components/maps/leaflet/LayerPopup/components/LayerPopupFlyout.tsx
```diff
@@ -1,25 +1,40 @@
...
+import { Claims } from 'constants/claims';
+import { useKeycloakWrapper } from 'hooks/useKeycloakWrapper';
 import noop from 'lodash/noop';
 import React from 'react';
 import { ListGroup } from 'react-bootstrap';
 
 export interface ILayerPopupFlyoutProps {
   onViewPropertyInfo: () => void;
   onCreateResearchFile?: () => void;
+  onCreateAcquisitionFile: () => void;
```
asanchezr, line 11: nit: not sure if this should be optional like onCreateResearchFile()...

asanchezr (approved):

PR #1685, opened 2022-07-01
Sync document and metada types with mayan | Features/psp 3621

FuriousLlama:

PR #1693, opened 2022-07-05
Mayan and database syncronization | psp-3803 & psp-3621 

FuriousLlama:
Added backend repository and controller methods to update Mayan pims database entries. 
Response from the sync operations not results in a batch result with information pertaining multiple possible operations

Note: Changes for the synchronization between Mayan and source JSON can be seen separately here https://github.com/bcgov/PSP/pull/1685


backend/api/Controllers/DocumentController.cs
```diff
@@ -40,12 +49,12 @@ public DocumentController(IDocumentService documentService)
         /// </summary>
         [HttpGet("types")]
         [HasPermission(Permissions.PropertyAdd)]
```
asanchezr, line 51: question - is this the right permission? PropertyAdd ?

backend/api/Controllers/DocumentController.cs
```diff
@@ -54,38 +63,77 @@ public IActionResult GetDocumentTypes()
...
+        {
+            var result = _documentSyncService.SyncMayanDocumentTypes(model);
+            return new JsonResult(result);
+        }
+
+        /// <summary>
+        /// Uploads the passed document.
+        /// </summary>
+        [HttpPatch("sync/mayan/metadatatype")]
+        //[HasPermission(Permissions.PropertyAdd)]
```
asanchezr, line 117: nit: putting a TODO here might help come back to this later when we have the appropriate permissions set up

backend/api/Repositories/Mayan/MayanBaseRepository.cs
```diff
@@ -0,0 +1,241 @@
...
+                    case HttpStatusCode.NoContent:
+                        result.Status = ExternalResultStatus.Success;
+                        result.Message = "No content was returned from the call";
+                        break;
+                    case HttpStatusCode.Forbidden:
+                        result.Status = ExternalResultStatus.Error;
+                        result.Message = "Request was forbidden";
+                        break;
+                    case HttpStatusCode.BadRequest:
+                        result.Status = ExternalResultStatus.Error;
```
asanchezr, line 163: question - why not use ProcessResponse<T>() here like the other methods do?

asanchezr (approved):

backend/api/Controllers/DocumentController.cs
FuriousLlama: thats a good idea

backend/api/Controllers/DocumentController.cs
FuriousLlama: I don't think so, we don't have a document permission at the moment

backend/api/Repositories/Mayan/MayanBaseRepository.cs
FuriousLlama: The delete response needed different logic from the retrieval (GET) or update/create (post, patch) since it does not return a payload

PR #1702, opened 2022-07-06
PSP-3693 Add Notes

asanchezr:
**IMPORTANT -** This will have heavy merge conflicts with Mahesh's PR as I created basically the same files as him with somewhat different code. 

Also, unit tests to come but wanted to start the review of this PR as soon as possible.

screenshot


backend/dal/Models/Note/EntityNoteModel.cs
```diff
@@ -0,0 +1,29 @@
+namespace Pims.Dal.Entities.Models
+{
+    /// <summary>
+    /// EntityNoteModel class, provides a model to represent notes associated to entities.
+    /// </summary>
+    public class EntityNoteModel
```
asanchezr, line 6: This is a generic model to attach notes to any entity in the database

backend/dal/Models/Note/NoteModel.cs
```diff
@@ -0,0 +1,29 @@
+
+namespace Pims.Dal.Entities.Models
+{
+    /// <summary>
+    /// NoteModel class, provides a model to represent notes associated to entities.
+    /// </summary>
```
asanchezr, line 6: This is a model for a note withing PIMS_NOTE table

backend/dal/Pims.Dal.csproj
```diff
@@ -36,6 +36,8 @@
     <PackageReference Include="ProjNet" Version="2.0.0" />
     <PackageReference Include="System.Data.SqlClient" Version="4.8.2" />
     <PackageReference Include="System.Text.RegularExpressions" Version="4.3.1" />
+    <PackageReference Include="Macross.Json.Extensions" Version="2.2.0" />
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
```
asanchezr, line 40: Needed to add these to be able to use `JsonConverter(typeof(JsonStringEnumMemberConverter))` in the enum - basically supporting string enums in JSON

backend/dal/Repositories/EntityNoteRepository.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        /// <typeparam name="T">The entity type - e.g. PimsActivityInstanceNote.</typeparam>
+        /// <param name="entity">The entity to add to the datastore.</param>
+        /// <returns>The created association.</returns>
+        public T Add<T>(T entity) where T : class
+        {
+            entity.ThrowIfNull(nameof(entity));
+            this.Context.Add<T>(entity);
+            return entity;
+        }
```
asanchezr, line 42: This allows to add notes associated to an entity in a generic way. 

```cs
  // now
  repo.Add<PimsActivityInstanceNote>(value)
  
  // in the future...
  repo.Add<PimsFileNote>(value)
  
```

backend/dal/Repositories/EntityNoteRepository.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Retrieves all notes associated with a parent entity
+        /// </summary>
+        /// <typeparam name="T">The entity type - e.g. PimsActivityInstanceNote.</typeparam>
+        /// <param name="predicate">The predicate to filter all notes - e.g. where parentId == parent.Id</param>
+        /// <returns>The entity-notes associations</returns>
+        public IEnumerable<T> GetAll<T>(Func<T, bool> predicate) where T : class
```
asanchezr, line 50: Currently not in use in this PR but thought it would be nice to have a generic way to "get all notes from this entity type" - e.g. get all "activity notes"

would be something like:

```cs
// in service class
public IEnumerable<PimsActivityInstanceNote> GetAllActivityNotes(long activityId) 
{
  return repo.GetAll<PimsActivityInstanceNote>(x => x.ActivityInstanceId == activityId);
}

// same can be applied to GetAllFileNotes() etc...
```

backend/dal/Security/Permissions.cs
```diff
@@ -82,6 +82,18 @@ public enum Permissions
...
+        NoteView = 27,
+
+        [Display(GroupName = "note", Name = "note-add", Description = "Can add new notes.")]
+        NoteAdd = 28,
+
+        [Display(GroupName = "note", Name = "note-edit", Description = "Can update existing notes.")]
+        NoteEdit = 29,
+
+        [Display(GroupName = "note", Name = "note-delete", Description = "Can delete existing notes.")]
+        NoteDelete = 30,
```
asanchezr, line 97: I'm not sure how we add new claims/permissions to our system @devinleighsmith? I suspect there is more to it than just this enum?

backend/dal/Repositories/EntityNoteRepository.cs
FuriousLlama: We might need to limit the classes the generic type can be in the future

backend/dal/Repositories/EntityNoteRepository.cs
FuriousLlama: As of now it looks like these classes can get any type of entity, not only notes

FuriousLlama (approved):

backend/dal/Repositories/EntityNoteRepository.cs
asanchezr: good idea - not sure how to implement that though

PR #1731, opened 2022-07-11
psp-3832 psp-3833 remove researchfile-add claim from restricted user.…

devinleighsmith:
… Restore role/claims removed from previous db migration.


database/mssql/scripts/dbscripts/PSP_PIMS_S32_00/Build/74_DML_PIMS_CLAIM.sql
```diff
@@ -61,4 +61,17 @@ GO
 INSERT [dbo].[PIMS_CLAIM] ([CLAIM_ID], [CLAIM_UID], [NAME], [KEYCLOAK_ROLE_ID], [DESCRIPTION], [IS_DISABLED], [CONCURRENCY_CONTROL_NUMBER], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_GUID], [APP_CREATE_USER_DIRECTORY], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_GUID], [APP_LAST_UPDATE_USER_DIRECTORY], [DB_CREATE_TIMESTAMP], [DB_CREATE_USERID], [DB_LAST_UPDATE_TIMESTAMP], [DB_LAST_UPDATE_USERID]) VALUES (24, N'38028c27-0a87-4875-a5d4-40879d13cfd6', N'researchfile-edit', N'38028c27-0a87-4875-a5d4-40879d13cfd6', N'Ability to edit existing research files.', 0, 3, CAST(N'2022-04-12T02:17:49.903' AS DateTime), N'SEED', NULL, N'SEED', CAST(N'2022-04-12T07:42:25.890' AS DateTime), N'service', N'1b93f614-91da-4b32-b36e-bd2c6ebd12e2', N'', CAST(N'2022-04-12T02:17:49.903' AS DateTime), N'SEED', CAST(N'2022-04-12T07:42:25.873' AS DateTime), N'SEED')
```
devinleighsmith, line 61: @dfilteau note that some changes made in 31_01 seem to have been overwritten in 32_00. I've added those changes back, but wanted to let you know so that the same thing doesn't happen in 33_00

auth/keycloak/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # >docker build -t keycloak .
 # >docker run --env-file=../../.env -d keycloak
 # https://hub.docker.com/r/jboss/keycloak/
-FROM jboss/keycloak:11.0.0
+FROM jboss/keycloak:16.1.1
```
asanchezr, line 4: I'm fine with this but shouldn't this warrant it's own PR in case we need to revert?

frontend/src/features/properties/map/research/ResearchContainer.tsx
```diff
@@ -167,12 +170,12 @@ export const ResearchContainer: React.FunctionComponent<IResearchContainerProps>
         <ResearchFileLayout
           leftComponent={
             <>
-              {selectedMenuIndex === 0 && (
+              {selectedMenuIndex === 0 && hasClaim(Claims.RESEARCH_ADD) ? (
```
asanchezr, line 173: quick question - as per PSP-3833

> EXPECTED RESULTS: The user with functional (restricted) role should only be able to create research files with already mapped properties or create research files without linked properties.

With this if you have the research add then you can add/edit properties right? Do we have a specific claim for adding properties to a research file upon creation / edit?

asanchezr (approved):

auth/keycloak/Dockerfile
devinleighsmith: It could, I was considering this low risk since this is a PR to dev that affects local dev only.

frontend/src/features/properties/map/research/ResearchContainer.tsx
devinleighsmith: No, we don't.

devinleighsmith: The idea with the claims is to provide something that is functionally equivalent. In this case, the ability to add a research file allows a user to create new properties of interest by attaching those properties to a research file. Similiarly, a user would be creating new POIs when adding new properties to an existing research file. Either way (adding new or updating existing) new POIs are created.

PR #1744, opened 2022-07-13
Psp 3821 - add support for local clamav scanning

devinleighsmith:


backend/api/Services/DocumentService.cs
```diff
@@ -40,8 +43,9 @@ public ExternalResult<FileDownload> DownloadFile(int documentId, int fileId)
             return task.Result;
         }
 
-        public ExternalResult<DocumentDetail> UploadDocument(int documentType, IFormFile fileRaw)
+        public async Task<ExternalResult<DocumentDetail>> UploadDocument(int documentType, IFormFile fileRaw)
         {
+            await this.avService.Scan(fileRaw);
```
devinleighsmith, line 48: Upload documents now makes a call to the new av service.

backend/api/av.Local.json
```diff
@@ -0,0 +1,5 @@
+{
+  "Av": {
+    "DisableScan": "true"
```
devinleighsmith, line 3: In general most devs will not ever have a use for the av scan. For local dev the only reason you would want this is if you are working on the av code itself, which won't happen often. As a result the default for local is off.

backend/api/av.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Av": {
+    "DisableScan": "false",
```
devinleighsmith, line 3: opposite for any other env, we want to be scanning by default in all of our envs.

backend/clamav/IAvService.cs
```diff
@@ -0,0 +1,11 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Pims.Av
+{
+    public interface IAvService
+    {
+        public Task Scan(IFormFile file);
```
devinleighsmith, line 8: In theory this could support another av implementation besides clam, but currently the only service that implements this is clam.

backend/tests/unit/dal/Services/DocumentServiceTest.cs
```diff
@@ -0,0 +1,51 @@
...
+using Pims.Dal.Services;
+using Xunit;
+
+namespace Pims.Dal.Test.Services
+{
+    [Trait("category", "unit")]
+    [Trait("category", "dal")]
+    [Trait("group", "document")]
+    [ExcludeFromCodeCoverage]
+    public class DocumentServiceTest
```
devinleighsmith, line 21: None of the document service code has tests because it is WIP proof of concept, but added one test just to ensure that the av service is called during upload.

docker-compose.yml
```diff
@@ -90,6 +90,22 @@ services:
     networks:
       - psp
 
+  ####################### Clam AV #######################
+  # example av test file here: https://serverfault.com/questions/484082/how-should-i-test-clam-anti-virus
+  clamav:
```
devinleighsmith, line 95: integrate clamav with existing docker-compose.

backend/clamav/ClamAvService.cs
```diff
@@ -0,0 +1,97 @@
...
+                return;
+            }
+            this._logger.LogInformation("ClamAV scan start");
+            var scanResult = await this.Client.SendAndScanFileAsync(fileData);
+
+            switch (scanResult.Result)
+            {
+                case ClamScanResults.Unknown:
+                    this._logger.LogError("ClamAV scan failed", scanResult);
+                    throw new AvException("Virus scan failed with unknown response. If you continue to see this error, please contact your site administrator.");
```
devinleighsmith, line 82: handle all failure cases with exceptions, these will be reported normally by the frontend using our existing mechanisms.

docker-compose.yml
```diff
@@ -90,6 +90,22 @@ services:
     networks:
       - psp
 
+  ####################### Clam AV #######################
+  # example av test file here: https://serverfault.com/questions/484082/how-should-i-test-clam-anti-virus
```
devinleighsmith, line 94: My computer keeps removing this test file so I assume others would have the same problem. Link to instructions on how to create that file instead.

backend/api/av.Local.json
asanchezr: nit: would prefer if config option is named in the positive - ie "EnableScan" vs "DisableScan"
not a big deal though

backend/api/Services/DocumentService.cs
```diff
@@ -40,8 +43,9 @@ public ExternalResult<FileDownload> DownloadFile(int documentId, int fileId)
...
             Task<ExternalResult<DocumentDetail>> task = documentRepository.UploadDocumentAsync(documentType, fileRaw);
             task.Wait();
```
asanchezr, line 50: another nit: would prefer we use await across the whole method

backend/clamav/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// Add the AddClamAvService to the dependency injection service collection.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="section"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddClamAvService(this IServiceCollection services, IConfigurationSection section)
+        {
+
+
```
asanchezr, line 22: nit: whitespace

backend/tests/unit/dal/Services/DocumentServiceTest.cs
```diff
@@ -0,0 +1,51 @@
...
+            var documentRepository = helper.GetService<Mock<IEdmsDocumentRepository>>();
+            documentRepository.Setup(x => x.UploadDocumentAsync(It.IsAny<int>(), It.IsAny<IFormFile>()));
+            var avService = helper.GetService<Mock<IAvService>>();
+            avService.Setup(x => x.Scan(It.IsAny<IFormFile>()));
+
+            // Act
+            var updatedLease = service.UploadDocument(1, helper.GetFormFile(""));
+
+            // Assert
+            documentRepository.Verify(x => x.UploadDocumentAsync(It.IsAny<int>(), It.IsAny<IFormFile>()), Times.Once);
```
asanchezr, line 45: question - shouldn't we verify that the AV service is called here as well?

asanchezr (approved):

PR #1747, opened 2022-07-14
fix debugging in jest-runner

devinleighsmith:


asanchezr (approved):
I don't have the extension installed but I trust you verified this on you local?

PR #1748, opened 2022-07-14
update properties screens to use id instead of pid where appropriate.

devinleighsmith:
not all properties in inventory have pids, therefore when we are retrieving properties from our inventory we should use the id if possible.

### DIFF
--- a/backend/api/Areas/Property/Controllers/PropertyController.cs
+++ b/backend/api/Areas/Property/Controllers/PropertyController.cs
@@ -86,7 +86,7 @@ namespace Pims.Api.Areas.Property.Controllers
         [Produces("application/json")]
         [ProducesResponseType(typeof(PropertyAssociationModel), 200)]
         [SwaggerOperation(Tags = new[] { "property" })]
-        public IActionResult GetPropertyAssociationsWithPid(long id)
+        public IActionResult GetPropertyAssociationsWithId(long id)
         {
             var property = _pimsRepository.Property.GetAssociations(id);
 
--- a/backend/api/Areas/Property/Controllers/PropertyController.cs
+++ b/backend/api/Areas/Property/Controllers/PropertyController.cs
@@ -78,17 +78,17 @@ namespace Pims.Api.Areas.Property.Controllers
         }
 
         /// <summary>
-        /// Get the property associations for the specified unique 'pid'.
+        /// Get the property associations for the specified unique 'id'.
         /// </summary>
         /// <returns></returns>
-        [HttpGet("{pid}/associations")]
+        [HttpGet("{id}/associations")]
         [HasPermission(Permissions.PropertyView)]
         [Produces("application/json")]
         [ProducesResponseType(typeof(PropertyAssociationModel), 200)]
         [SwaggerOperation(Tags = new[] { "property" })]
-        public IActionResult GetPropertyAssociationsWithPid(string pid)
+        public IActionResult GetPropertyAssociationsWithPid(long id)
         {
-            var property = _pimsRepository.Property.GetAssociations(pid);
+            var property = _pimsRepository.Property.GetAssociations(id);
 
             return new JsonResult(_mapper.Map<PropertyAssociationModel>(property));
         }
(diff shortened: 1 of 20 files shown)


devinleighsmith:
@FuriousLlama can you review these changes? they affect a large amount of code that you worked on, want to make sure I don't break anything.

backend/api/Areas/Property/Controllers/PropertyController.cs
```diff
@@ -78,43 +78,43 @@ public IActionResult GetProperty(int id)
...
         /// </summary>
         /// <returns></returns>
-        [HttpGet("{pid}/associations")]
+        [HttpGet("{id}/associations")]
         [HasPermission(Permissions.PropertyView)]
         [Produces("application/json")]
         [ProducesResponseType(typeof(PropertyAssociationModel), 200)]
         [SwaggerOperation(Tags = new[] { "property" })]
-        public IActionResult GetPropertyAssociationsWithPid(string pid)
+        public IActionResult GetPropertyAssociationsWithPid(long id)
```
FuriousLlama, line 89: "GetPropertyAssociationsWithPid" to Id

frontend/src/features/mapSideBar/hooks/useMapSideBarQueryParams.tsx
```diff
@@ -98,7 +104,12 @@ export const useMapSideBarQueryParams = (map?: L.Map): IMapSideBar => {
         break;
       case MapViewState.PROPERTY_INFORMATION:
         setSidebarComponent(
-          <MotiInventoryContainer onClose={handleClose} pid={propertyId} onZoom={onZoom} />,
+          <MotiInventoryContainer
+            onClose={handleClose}
+            id={!!propertyId ? +propertyId : undefined}
```
FuriousLlama, line 109: this looks very odd. Can this logic be done when setting the propertyId variable instead?

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsContainer.tsx
```diff
@@ -13,7 +13,7 @@ import { UpdatePropertyDetailsForm } from './UpdatePropertyDetailsForm';
 import { UpdatePropertyDetailsYupSchema } from './validation';
 
 export interface IUpdatePropertyDetailsContainerProps {
-  pid: string;
+  id?: number;
```
FuriousLlama, line 16: if we reached this container, should the id not be known? The id should not be undefined at this point

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -34,16 +35,16 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
...
-    const parsedPid = pidParser(property.pid);
-    history.push(`/mapview/property/${parsedPid}`);
+    history.push(`/mapview/property/${property.id}?pid=${property.pid}`);
   };
 
   const onPropertyViewClicked = (pid?: string | null) => {
     if (pid !== undefined && pid !== null) {
       const parsedPid = pidParser(pid);
-      history.push(`/mapview/property/${parsedPid}`);
+      history.push(`/mapview/non-inventory-property/${parsedPid}`);
```
FuriousLlama, line 44: In this case the id of the property would not be the pid. Thoughts on using "?pid=" on both?

like:
/mapview/property?id=${propertyId}&pid={propertyPid}
/mapview/property?pid=${propertyPid}

frontend/src/features/mapSideBar/hooks/useMapSideBarQueryParams.tsx
devinleighsmith: yes, that works

frontend/src/features/properties/map/MapView.tsx
devinleighsmith: Hmm, I thought this would be ok since a non-inventory property would never have an id so the main identifier would be a pid, but the above also works.

devinleighsmith:
I think the best identifier we have is whether or not we have an id, as our system is the only place that populates an id.

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsForm.tsx
```diff
@@ -74,24 +74,25 @@ export const UpdatePropertyDetailsForm: React.FunctionComponent<IUpdatePropertyD
...
     if (!isVolumetricParcel) {
-      formikProps.setFieldValue('volumetricMeasurement', 0);
-      formikProps.setFieldValue('volumetricUnitTypeCode', undefined);
-      formikProps.setFieldValue('volumetricParcelTypeCode', undefined);
+      setFieldValue('volumetricMeasurement', 0);
+      setFieldValue('volumetricUnitTypeCode', undefined);
+      setFieldValue('volumetricParcelTypeCode', undefined);
     }
-  }, [isVolumetricParcel, formikProps, formikProps.setFieldValue]);
+  }, [isVolumetricParcel, setFieldValue]);
```
devinleighsmith, line 85: FYI @FuriousLlama this code was causing infinite render loops, this may have been the root cause of some of your test issues? if any tests got removed/skipped as a result of that this should allow them to be enabled again.

FuriousLlama (approved):

FuriousLlama:
> I think the best identifier we have is whether or not we have an id, as our system is the only place that populates an id.

The problem with that is that we would be implying the source, but given the models are not immutable it could have been just set to undefine or not being populated

devinleighsmith:
@FuriousLlama there is only one source for inventory properties, our database. For property information in general there are many sources:

- parcelmap layers
- bc assessment
- ltsa
- other layers

devinleighsmith:
So I guess what would make the most sense to me is to have the unified property object that we discussed, with an inventory property sub-object. We will then only populate that object with an inventory property from our db.

asanchezr (approved):

PR #1749, opened 2022-07-15
Property Edit Update | Features/psp 3524

FuriousLlama:
image


frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -34,163 +29,180 @@ export interface IMotiInventoryContainerProps {
...
+          setComposedProperty(property => ({ ...property, apiProperty: propInfo }));
+        }
+      }
+    } catch (e) {
+      // PSP-2919 Hide the property info tab for non-inventory properties
+      // We get an error because PID is not on our database
+      if (axios.isAxiosError(e)) {
+        const axiosError = e as AxiosError<IApiError>;
+        if (axiosError?.response?.status === 404) {
+          setComposedProperty(property => ({ ...property, apiProperty: undefined }));
```
devinleighsmith, line 88: nice, this looks similar to what we talked about today.

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -34,163 +29,180 @@ export interface IMotiInventoryContainerProps {
...
-      setLtsaDataRequestedOn(new Date());
-      setLtsaData(undefined);
+      setComposedProperty(property => ({
+        ...property,
+        ltsaDataRequestedOn: new Date(),
+        ltsaData: undefined,
+      }));
+
       if (!!props.pid) {
         const ltsaData = await getLtsaData(pidFormatter(props.pid));
```
devinleighsmith, line 119: my pr updates a lot of these to use the id instead of the pid, but I don't mind if you go first and I resolve afterwords.

FuriousLlama: I rather that. This one gave me a lot of headache when updating the tests

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsForm.test.tsx
```diff
@@ -193,7 +192,7 @@ describe('UpdatePropertyDetailsForm component', () => {
   ) => {
     const utils = render(
       <Formik onSubmit={onSubmit} initialValues={renderOptions.initialValues}>
-        {formikProps => <UpdatePropertyDetailsForm {...formikProps} onCancel={onCancel} />}
+        {formikProps => <UpdatePropertyDetailsForm formikProps={formikProps} />}
```
devinleighsmith, line 195: If the tests are passing this is probably fine, but this passes the same object instance, so if multiple tests are using that instance and a test modifies the object this _may_ be a problem.

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsContainer.tsx
```diff
@@ -1,28 +1,33 @@
...
 export const UpdatePropertyDetailsContainer: React.FC<IUpdatePropertyDetailsContainerProps> = props => {
   const isMounted = useIsMounted();
-  const history = useHistory();
   const { retrieveProperty, retrievePropertyLoading } = useGetProperty();
   const { updateProperty } = useUpdateProperty();
   const { queryAll } = useQueryMapLayersByLocation();
 
+  const formikRef = useRef<FormikProps<UpdatePropertyDetailsFormModel>>(null);
+
+  props.setFormikRef(formikRef);
```
devinleighsmith, line 29: another option is to just pass the formikref from the parent that needs access to formik, and then that ref will get set by formik.

frontend/src/features/properties/map/research/update/properties/UpdateProperties.tsx
```diff
@@ -73,7 +73,7 @@ export const UpdateProperties: React.FunctionComponent<IUpdatePropertiesProps> =
         title={'Property selection'}
         icon={undefined}
         footer={
-          <ResearchFooter
+          <SidebarFooter
```
devinleighsmith, line 76: Yes, this name makes more sense given the usage.

frontend/src/features/properties/map/propertyInformation/PropertyContainer.tsx
```diff
@@ -0,0 +1,96 @@
...
+  tabViews.push({
+    content: <></>,
+    key: InventoryTabNames.value,
+    name: 'Value',
+  });
+
+  var defaultTab = InventoryTabNames.title;
+
+  // TODO: this should have a loading flag
+  const propertyViewForm = usePropertyDetails(composedProperty.apiProperty);
```
devinleighsmith, line 51: Does this todo make sense? it looks like the PropertyDetailsTabView is being sent a loading flag that corresponds to this object. Maybe I don't understand.

frontend/src/features/properties/map/propertyInformation/ComposedProperty.ts
```diff
@@ -0,0 +1,14 @@
+import { IPropertyApiModel } from 'interfaces/IPropertyApiModel';
+import { LtsaOrders } from 'interfaces/ltsaModels';
+import { Api_PropertyAssociations } from 'models/api/Property';
+
+export default interface ComposedProperty {
```
devinleighsmith, line 5: this is a good start.

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -34,163 +29,180 @@ export interface IMotiInventoryContainerProps {
...
-  }
-
-  const onEditPropertyInformation = useCallback(
-    () => history.push(`/mapview/property/${props.pid}/edit`),
-    [history, props.pid],
-  );
+  };
+
+  const handleCancelConfirm = () => {
+    if (formikRef !== undefined) {
```
devinleighsmith, line 157: nit: formikRef?.current !== undefined

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -34,163 +29,180 @@ export interface IMotiInventoryContainerProps {
...
-      />
+      <>
+        <PropertyViewSelector
+          composedProperty={composedProperty}
+          isEditMode={isEditing}
+          setEditMode={setIsEditing}
+          setFormikRef={setFormikRef}
+          onSuccess={onSuccess}
+        />
+        <GenericModal
```
devinleighsmith, line 189: I wonder if we should register some of these common ones somewhere at the app level, and then create a context to control their display. That way we wouldn't have to keep registering the modals in various containers/forms throughout the app.

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
```diff
@@ -34,163 +29,180 @@ export interface IMotiInventoryContainerProps {
...
-    async function fetchResearchFile() {
+    fetchPimsProperty();
+  }, [fetchPimsProperty]);
+
+  useEffect(() => {
+    const getAssociations = async () => {
       if (props.pid !== undefined) {
         const response = await getPropertyAssociations(props.pid);
-        setPropertyAssociations(response);
+        if (response !== undefined) {
```
devinleighsmith, line 102: if the response was undefined would we want to maintain the old response or clear it?

devinleighsmith (approved):

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsForm.test.tsx
FuriousLlama: That's interesting. Will keep an eye in the future for tests failures

frontend/src/features/mapSideBar/tabs/propertyDetails/update/UpdatePropertyDetailsContainer.tsx
FuriousLlama: We could do that. I think is up to how we define the responsibilities to handle the formik logic and how the communication should happen for complex hierarchies like these

frontend/src/features/properties/map/propertyInformation/PropertyContainer.tsx
FuriousLlama: Yes. Unfortunately, this causes the page to stutter when loading information. The loading screen goes away once apiProperty is retrieved, but updated again once usePropertyDetails completes since that one retrieves extra information. IMO this should follow the same pattern of retrieving the information alongside a isLoading

frontend/src/features/mapSideBar/MotiInventoryContainer.tsx
FuriousLlama: Thats a good idea... The only problem is that if is not put in the correct place it might be hidden by other modals. Might need to tinker with it to make sure the Z prop is set

PR #1796, opened 2022-07-26
Added document list actions hooks and base detail modal information| Features/psp 3799

FuriousLlama:
image

image


frontend/src/features/admin/access-request/AccessRequestForm.tsx
```diff
@@ -56,28 +56,28 @@ export const AccessRequestForm: React.FunctionComponent<IAccessRequestFormProps>
           />
         </SectionField>
 
-        <SectionField label="First name" wideScreen>
+        <SectionField label="First name" labelWidth="2">
```
FuriousLlama, line 59: changes not directly related to the documents screen. Changed for consistency and flexibility of the SectionField

frontend/src/features/admin/edit-user/EditUserForm.tsx
```diff
@@ -46,7 +46,7 @@ const EditUserForm: React.FunctionComponent<IEditUserFormProps> = ({
     >
       {formikProps => (
         <Form className="userInfo">
-          <SectionField label="IDIR/BCeID" wideScreen>
+          <SectionField label="IDIR/BCeID" labelWidth="2">
```
FuriousLlama, line 49: changes not directly related to the documents screen. Changed for consistency and flexibility of the SectionField

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -38,8 +39,8 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
         documentItems = documentItems.filter(document => {
           return (
             (!filters.documentTypeId ||
-              document.documentTypeId === Number(filters.documentTypeId)) &&
-            (!filters.status || document.statusTypeCode?.id === filters.status)
+              document?.documentType?.id === Number(filters.documentTypeId)) &&
+            (!filters.status || document?.statusTypeCode?.id === filters.status)
```
FuriousLlama, line 43: The type was not flattened on the backend, but it was on the front end. This fixes that

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -59,6 +60,17 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
...
 
+  const [isDetailsVisible, setIsDetailsVisible] = useState(false);
+  const [selectedDocument, setSelectedDocument] = useState<Api_Document | undefined>(undefined);
+
+  const handleViewDetails = (values: Api_Document) => {
+    setIsDetailsVisible(true);
+    setSelectedDocument(values);
+  };
+  const handleDelete = (values: Api_Document) => {
+    // Todo: need to have the parent entity in context
```
FuriousLlama, line 71: At this point there is no activity/file to make the delete functionality work

frontend/src/features/mapSideBar/tabs/Section.tsx
```diff
@@ -23,31 +19,47 @@ export const Section: React.FC<SectionProps> = ({
...
-              <ArrowDropUpIcon
-                onClick={() => {
-                  setIsCollapsed(!isCollapsed);
-                }}
-              />
-            )}
-          </Col>
-        </Row>
-      </StyledSectionHeader>
+      {header && (
```
FuriousLlama, line 22: Allows the section to be without a header if necessary. Allows the use of this component instead of only the style

frontend/src/features/properties/selector/search/PropertySearchSelectorFormView.tsx
```diff
@@ -33,7 +33,7 @@ export const PropertySearchSelectorFormView: React.FunctionComponent<IPropertySe
 }) => {
   return (
     <>
-      <StyledFormSection>
+      <Section header={undefined}>
         <Styled.H3>Search for a property</Styled.H3>
```
FuriousLlama, line 37: We should figure out if we want different types of section headers and standardize those so we can re-use the component instead of just using the style

frontend/src/features/mapSideBar/tabs/SectionStyles.tsx
```diff
@@ -3,20 +3,6 @@ import { Col } from 'react-bootstrap';
 import { MdArrowDropDown, MdArrowDropUp } from 'react-icons/md';
 import styled from 'styled-components';
 
-export const StyledFormSection = styled.div`
-  margin: 1.5rem;
-  padding: 1.5rem;
-  background-color: white;
-  text-align: left;
```
FuriousLlama, line 10: The removal of these styles caused a lot of snapshots to be updated.

frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -5,17 +5,23 @@ import styled from 'styled-components';
...
   return (
     <Row className={props.className ?? 'pb-2'}>
-      <Col xs={props.wideScreen ? 2 : 4}>
+      <Col xs={props.labelWidth ?? '4'} className="pr-0 text-left">
         <StyledFieldLabel>{props.label}:</StyledFieldLabel>
       </Col>
-      <StyledCol className={clsx({ required: props.required })}>{props.children}</StyledCol>
+      <StyledCol
+        xs={props.contentWidth ?? true}
+        className={clsx({ required: props.required }) + ' text-left'}
```
FuriousLlama, line 21: This will make the text inside section fields to align left instead of center by default. I checked against dev and saw no big differences. But it would be worthy to keep an eye in case the alignment causes issues.

frontend/src/features/properties/selector/MapSelectorContainer.tsx
```diff
@@ -71,7 +71,7 @@ export const MapSelectorContainer: React.FunctionComponent<IMapSelectorContainer
           Add to selection
         </Button>
       ) : null}
-      <StyledFormSection>
+      <Section header={undefined}>
```
FuriousLlama, line 74: left as a marker so that the H3 below can be removed if we decided to have sections with h2 or h3 headings

backend/api/Controllers/DocumentController.cs
```diff
@@ -59,7 +59,7 @@ public DocumentController(IDocumentService documentService, IDocumentSyncService
         [SwaggerOperation(Tags = new[] { "documents" })]
         public IActionResult GetDocumentTypes()
         {
-            var result = _documentService.GetDocumentTypes();
+            var result = _documentService.GetStorageDocumentTypes();
```
devinleighsmith, line 62: at some point we'll need claims for document crud. I don't think we called that out specifically with any of these stories but if it doesn't get done with these stories we may need a follow-up

backend/api/Controllers/DocumentController.cs
```diff
@@ -73,20 +73,33 @@ public IActionResult GetDocumentTypes()
...
 
         /// <summary>
         /// Downloads the file for the correspoding file and document id.
         /// </summary>
         [HttpGet("{documentId}/files/{fileId}/download")]
-        [HasPermission(Permissions.PropertyAdd)]
+        //[HasPermission(Permissions.PropertyAdd)]
+        [ProducesResponseType(typeof(ExternalResult<FileDownload>), 200)]
+        [SwaggerOperation(Tags = new[] { "documents" })]
+        public async Task<IActionResult> DownloadFile(int documentId, int fileId)
```
devinleighsmith, line 87: is int correct here? I just now that a lot of our ids are longs in other controllers/services.

backend/api/Controllers/DocumentController.cs
```diff
@@ -73,20 +73,33 @@ public IActionResult GetDocumentTypes()
         [SwaggerOperation(Tags = new[] { "documents" })]
         public IActionResult GetDocumentList()
         {
-            var result = _documentService.GetDocumentList();
+            var result = _documentService.GetStorageDocumentList();
             return new JsonResult(result);
         }
 
         /// <summary>
         /// Downloads the file for the correspoding file and document id.
```
devinleighsmith, line 81: I realize this was from a previous sprint, but these controllers also were WIP at that time.
nit: correspoding -> corresponding

backend/api/Models/Mayan/Document/DocumentMetadata.cs
```diff
@@ -0,0 +1,41 @@
+using System.Text.Json.Serialization;
+using Pims.Api.Models.Mayan.Metadata;
+
+namespace Pims.Api.Models.Mayan.Document
+{
+    /// <summary>
+    /// Represents the document metada. Note, this does not contain the stored document.
```
devinleighsmith, line 7: nit: metadata

frontend/src/features/mapSideBar/tabs/Section.tsx
asanchezr: Also I think this component should be moved to common folder. We are using it in most forms

frontend/src/features/mapSideBar/tabs/SectionStyles.tsx
asanchezr: did you check that all affected forms look ok?

asanchezr: there's a lot of margin & padding removed...

backend/api/Controllers/DocumentController.cs
FuriousLlama: That is true, I will fix that on the next PR going over the types for the mayan interface

frontend/src/features/mapSideBar/tabs/SectionField.tsx
asanchezr: nit: **clsx** already let's you merge styles so we should avoid string concatenation

```tsx
clsx({ required: props.required, 'text-left': true})
```

backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -124,6 +124,35 @@ public async Task<ExternalResult<QueryResult<DocumentDetail>>> GetDocumentsListA
...
+        {
+            _logger.LogDebug("Retrieving document metadata...");
+
+            string authenticationToken = await _authRepository.GetTokenAsync();
+
+            var queryParams = GenerateQueryParams(ordering, page, pageSize);
+
+            string endpointString = $"{_config.BaseUri}/documents/{documentId}/metadata/";
+            Uri endpoint = new(QueryHelpers.AddQueryString(endpointString, queryParams));
+            var response = await GetAsync<QueryResult<DocumentMetadata>>(endpoint, authenticationToken).ConfigureAwait(true);
```
devinleighsmith, line 150: Is there a reason we are using configureAwait true here and not in the delete or create functions in this file?

backend/api/Controllers/DocumentController.cs
asanchezr: agree with Devin - ids should be typed as **long**

devinleighsmith: Hmm, when I first read the fileId I think I assumed it was the Id of a pims file, but it appears to by the id of the file in mayan. Not sure if that is worth a rename or not, up to you.

backend/api/Repositories/Mayan/MayanDocumentRepository.cs
FuriousLlama: No particular one. Once we add the warning back to the config we should see all the ones where configureAwait should have been set

backend/api/Controllers/DocumentController.cs
FuriousLlama: I believe we should rename it so that is clear that the id is from mayan or pims

backend/api/Controllers/DocumentController.cs
FuriousLlama: I agree. I prefer would prefer a story to go over the claims

frontend/src/features/mapSideBar/tabs/Section.tsx
FuriousLlama: Agreed. Can you log a tech debt to do this exclusively?

frontend/src/features/mapSideBar/tabs/SectionStyles.tsx
FuriousLlama: I did check some of them and saw no difference. But having more eyes would be better

backend/api/Controllers/DocumentController.cs
FuriousLlama: Will update it in the a different pr for the next part of the document features

frontend/src/features/documents/DownloadDocumentButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+    } else {
+      const data = await provider.downloadDocumentFileLatest(mayanDocumentId);
+      if (data) {
+        showFIle(data.payload);
+      }
+    }
+  }
+
+  const showFIle = (file: FileDownload) => {
```
asanchezr, line 30: nit: showFIle -> showFile

frontend/src/features/documents/DownloadDocumentButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export interface IDownloadDocumentButtonProps {
+  mayanDocumentId: number;
+  mayanFileId?: number;
+}
+
+const DownloadDocumentButton: React.FunctionComponent<IDownloadDocumentButtonProps> = props => {
+  let provider = useDocumentProvider();
+
+  async function DownloadFile(mayanDocumentId: number, mayanFileId?: number) {
```
asanchezr, line 16: nit: `DownloadFile(...)` -> `downloadFile(...)`

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+            ...document,
+            mayanMetadata: results,
+          }));
+        }
+      }
+    };
+
+    fetch();
+  }, [props.pimsDocument.mayanDocumentId, retrieveDocumentMetadata]);
+  return <DocumentDetailView metadata={document} isLoading={retrieveDocumentMetadataLoading} />;
```
asanchezr, line 34: We need to come up with a naming standard for forms. We have been using *View and *Form - would be nice to pick one. Can discuss this Friday

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const DocumentDetailContainer: React.FunctionComponent<IDocumentDetailContainerProps> = props => {
+  const [document, setDocument] = useState<ComposedDocument>({ pimsDocument: props.pimsDocument });
+
+  const { retrieveDocumentMetadata, retrieveDocumentMetadataLoading } = useDocumentProvider();
+  useEffect(() => {
+    const fetch = async () => {
+      if (props.pimsDocument.mayanDocumentId !== undefined) {
+        const axiosResponse = await retrieveDocumentMetadata(props.pimsDocument.mayanDocumentId);
+        console.log(axiosResponse?.status, ExternalResultStatus.Success);
```
asanchezr, line 21: console.log

frontend/src/features/documents/documentDetail/hooks/useDocumentProvider.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  return {
+    retrieveDocumentMetadata,
+    retrieveDocumentMetadataLoading,
+    downloadDocumentFile,
+    downloadDocumentFileLoading,
+    downloadDocumentFileLatest,
+    downloadDocumentFileLatestLoading,
+    deleteActivityDocument,
+    deleteActivityDocumentLoading,
```
asanchezr, line 128: nit: we should also return the error state for consumers of this hook to check for errors (if they want to). We should discuss on Friday how to achieve that (if we want)

I've been returning the whole object from "useApiRequestWrapper" in my hooks.  For example:

```tsx
const retrieveDocumentMetadataApi =  useApiRequestWrapper(...)


// export
return {
  retrieveDocumentMetadata: retrieveDocumentMetadataApi

}

/// in a consumer of this hook

const { retrieveDocumentMetadata } = useDocumentProvider()
...
retrieveDocumentMetadata.execute(...)

retrieveDocumentMetadata.loading etc
```

frontend/src/features/documents/list/DocumentListView.test.tsx
```diff
@@ -13,6 +14,17 @@ const storeState = {
...
+(useKeycloak as jest.Mock).mockReturnValue({
+  keycloak: {
+    userInfo: {
+      organizations: [1],
+      roles: [],
+    },
+    subject: 'test',
+  },
+});
+
```
asanchezr, line 27: nit: we don't need to do this anymore. The **render** function has this built-in if you pass the option useMockAuthentication: true

```tsx

const component = render(<Blah />, { ...renderOptions, store: storeState, useMockAuthentication: true});

```

asanchezr: FYI - you DO need to keep `jest.mock('@react-keycloak/web');`

frontend/src/features/admin/edit-user/EditUserForm.tsx
```diff
@@ -101,7 +101,7 @@ const EditUserForm: React.FunctionComponent<IEditUserFormProps> = ({
             </TooltipWrapper>
           </SectionField>
 
-          <SectionField label="MoTI Region(s)" required wideScreen>
+          <SectionField label="MoTI Region(s)" required labelWidth="2">
```
devinleighsmith, line 104: this is fine, but please make sure to leave a note for Sue to make sure she runs regression on these screens this sprint.

frontend/src/features/documents/list/DocumentListView.test.tsx
```diff
@@ -38,6 +50,13 @@ describe('Document List View', () => {
     mockAxios.reset();
     mockAxios.onGet(`documents/document-types`).reply(200, mockDocumentTypesResponse());
   });
+  afterEach(() => {
+    mockAxios.reset();
```
asanchezr, line 54: Do we need to reset mockAxios before AND after each call? I'd think only one time would suffice...

FuriousLlama: not sure, but given we are seeing some issues with concurrency I figured it would not hurt to be add the extra reset

frontend/src/features/documents/list/DocumentListView.tsx
asanchezr: How about the ** entityId: number** that is passed in as props? I believe that would be the parent entity

frontend/src/features/documents/DownloadDocumentButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+      const data = await provider.downloadDocumentFileLatest(mayanDocumentId);
+      if (data) {
+        showFIle(data.payload);
+      }
+    }
+  }
+
+  const showFIle = (file: FileDownload) => {
+    const aElement = document.createElement('a');
```
devinleighsmith, line 31: @FuriousLlama have you tested this in FF? If I remember correctly you may need to attach the a to the dom for this to work in FF.

frontend/src/features/documents/list/DocumentListView.tsx
asanchezr: The way we did it with notes is create a "placeholder" activity on our local databases and have the Test page use that ID to create/update/delete notes from it. See example in TestNotes.tsx page - I believe a similar pattern could be followed here while we wait for Activity forms to be implemented...

frontend/src/features/documents/DownloadDocumentButton.tsx
FuriousLlama: I dont have FF atm. But will give it a try once it installs

frontend/src/features/documents/list/DocumentResults/DocumentResults.test.tsx
```diff
@@ -8,12 +10,31 @@ const setSort = jest.fn();
...
+(useKeycloak as jest.Mock).mockReturnValue({
+  keycloak: {
+    userInfo: {
+      organizations: [1],
+      roles: [],
+    },
+    subject: 'test',
+  },
+});
+
```
asanchezr, line 22: see prior comment re: useMockAuthentication

frontend/src/features/documents/DownloadDocumentButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { FaDownload } from 'react-icons/fa';
+
+import { useDocumentProvider } from './documentDetail/hooks/useDocumentProvider';
+
+export interface IDownloadDocumentButtonProps {
+  mayanDocumentId: number;
+  mayanFileId?: number;
+}
+
+const DownloadDocumentButton: React.FunctionComponent<IDownloadDocumentButtonProps> = props => {
```
devinleighsmith, line 13: In my mind this component is doing a lot for a button, it seems like a mix of a display component and a container.

Up to you, but I'd be tempted to just pass in functions to handle the click event and then have a parent container/component handle the click. you should be able to pass any loading/disabled state logic via props.

frontend/src/features/documents/DownloadDocumentButton.tsx
FuriousLlama: @devinleighsmith  The download worked on FF

frontend/src/features/documents/list/DocumentResults/DocumentResults.tsx
```diff
@@ -1,57 +1,132 @@
...
+  return function({ row: { original, index } }: CellProps<Api_Document, string>) {
+    const { hasClaim } = useKeycloakWrapper();
+    return (
+      <StyledIcons className="no-gutters">
+        {hasClaim(Claims.PROPERTY_ADD) && original.mayanDocumentId !== undefined && (
+          <Col>
+            <DownloadDocumentButton mayanDocumentId={original.mayanDocumentId} />
+          </Col>
+        )}
+        {hasClaim(Claims.PROPERTY_VIEW) && (
```
asanchezr, line 89: This needs at least a TODO so we don't forget to use the correct claim (Document_View)

frontend/src/features/documents/list/DocumentResults/DocumentResults.tsx
```diff
@@ -1,57 +1,132 @@
...
+          <Col>
+            <Button
+              title="document view details"
+              icon={<FaEye size={24} id={`document-view-${index}`} title="document view details" />}
+              onClick={() => original.id && onViewDetails(original)}
+            ></Button>
+          </Col>
+        )}
+        {hasClaim(Claims.PROPERTY_VIEW) && (
```
asanchezr, line 98: This needs at least a TODO so we don't forget to use the correct claim (Document_Delete)

frontend/src/features/documents/DownloadDocumentButton.tsx
FuriousLlama: I thought so too. The problem is that we need the same functionality multiple times, including the visual aspect. I believe it encapsulated as a widget works better. Could be renamed given is not only a button

frontend/src/features/documents/documentDetail/ComposedDocument.ts
```diff
@@ -0,0 +1,7 @@
+import { Api_Document } from 'models/api/Document';
+import { Mayan_DocumentMetadata } from 'models/api/DocumentManagement';
+
+export interface ComposedDocument {
```
devinleighsmith, line 4: Hmm, not sure if we need this idea of composition in this case. it makes sense for properties since their are literally multiple different data sources at play, but here can't our backend hide that some data is in mayan and some is in our db? does our frontend need to know about this implementation detail?

frontend/src/features/documents/list/DocumentResults/DocumentResults.tsx
asanchezr: I read your prior comments that you don't want to add the claims to the database + API with this story but if we don't do it here, we will need to be very careful of not forgetting to update to proper claims later

frontend/src/features/documents/documentDetail/ComposedDocument.ts
FuriousLlama: We could abstract it, but given we are not changing the model on the other concepts I believe is better if leave that as is. We could just called StoredDocumentModel or something like that to differentiate it from the pims one

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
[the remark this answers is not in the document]
FuriousLlama: This should be gone. If you refresh the gh page is the change still there?

frontend/src/features/documents/list/DocumentResults/DocumentResults.tsx
```diff
@@ -1,57 +1,132 @@
...
+  return [
+    {
+      Header: 'Document type',
+      accessor: 'documentType',
+      align: 'left',
+      sortable: true,
+      minWidth: 30,
+      maxWidth: 40,
+      Cell: renderDocumentType,
+    },
```
asanchezr, line 32: nit: would be nice to move this to a separate **colums.tsx** file to unclutter this file a bit

frontend/src/features/mapSideBar/tabs/SectionField.tsx
devinleighsmith: Yes, Sue should probably regression test leases, especially since this change would be in the 2.2 release.

frontend/src/features/properties/selector/search/PropertySearchSelectorFormView.tsx
```diff
@@ -33,7 +33,7 @@ export const PropertySearchSelectorFormView: React.FunctionComponent<IPropertySe
 }) => {
   return (
     <>
-      <StyledFormSection>
+      <Section header={undefined}>
```
devinleighsmith, line 36: nit: just use <Section>

frontend/src/models/api/ExternalResult.ts
```diff
@@ -5,6 +5,6 @@ export interface ExternalResult<T> {
 }
 
 export enum ExternalResultStatus {
-  Success,
-  Error,
+  Success = 'Success',
```
devinleighsmith, line 8: this is functionally equivalent is it not?

devinleighsmith:
@FuriousLlama something worth mentioning, when a lot of the documents backend was implemented we were ok with little to no test coverage since that code was POC. If we are moving that code into general usage, we will want test coverage.

backend/api/Controllers/DocumentController.cs
devinleighsmith: would you mind logging something so that we don't forget?

devinleighsmith (approved):

frontend/src/features/notes/list/NoteResults/NoteResults.test.tsx
```diff
@@ -14,7 +14,13 @@ jest.mock('@react-keycloak/web');
...
   const { results, ...rest } = renderOptions;
   const utils = render(
-    <NoteResults sort={{}} results={results ?? []} setSort={setSort} onDelete={noop} />,
+    <NoteResults
+      sort={{}}
+      results={results ?? []}
+      setSort={setSort}
+      onDelete={noop}
+      onShowDetails={noop}
+    />,
```
asanchezr, line 23: I had already fixed this in my PR but thanks for catching it

frontend/src/features/properties/selector/search/PropertySearchSelectorFormView.tsx
asanchezr: Agreed

frontend/src/pages/TestFileManagement.tsx
```diff
@@ -183,6 +175,32 @@ export const TestFileManagement: React.FunctionComponent = () => {
...
+        </Col>
+      </Row>
+      <Row className="no-gutters border">
+        <Col>
+          <DocumentDetailContainer pimsDocument={{ id: 1, mayanDocumentId: 13 }} />
+        </Col>
+      </Row>
+      {
+        //<DocumentDetailModal pimsDocument={{ id: 1, mayanDocumentId: 13 }} />
+      }
```
asanchezr, line 196: nit: I know this is a test page but perhaps remove commented code? thx

asanchezr (approved):

PR #1812, opened 2022-07-28
Delete document action | Features/psp 3799

FuriousLlama:
image

### DIFF
--- a/backend/api/Areas/Documents/DocumentController.cs
+++ b/backend/api/Areas/Documents/DocumentController.cs
@@ -10,6 +10,7 @@ using Pims.Api.Models.Concepts;
 using Pims.Api.Models.Mayan;
 using Pims.Api.Models.Mayan.Document;
 using Pims.Api.Services;
+using Pims.Dal.Entities;
 using Swashbuckle.AspNetCore.Annotations;
 using Concepts = Pims.Api.Models.Concepts;
 
@@ -72,7 +73,7 @@ namespace Pims.Api.Controllers
         //[HasPermission(Permissions.NoteView)]
         [ProducesResponseType(typeof(IList<DocumentRelationshipModel>), 200)]
         [SwaggerOperation(Tags = new[] { "document" })]
-        public IActionResult GetActivityDocuments(DocumentRelationType relationshipType, long parentId)
+        public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipType, long parentId)
         {
             switch (relationshipType)
             {
@@ -80,6 +81,30 @@ namespace Pims.Api.Controllers
                     var documents = _documentService.GetActivityDocuments(parentId);
                     var mappedDocuments = _mapper.Map<List<DocumentRelationshipModel>>(documents);
                     return new JsonResult(mappedDocuments);
+                default:
+                    throw new BadRequestException("Relationship type not valid.");
+            }
+        }
+
+        /// <summary>
+        /// Gets a collection of documents for the specified type and owner id.
+        /// </summary>
+        /// <param name="type">Used to identify document type.</param>
+        /// <param name="entityId">Used to identify document's parent entity.</param>
+        /// <returns></returns>
+        [HttpDelete("{relationshipType}")]
+        [Produces("application/json")]
+        //[HasPermission(Permissions.NoteView)]
+        [ProducesResponseType(typeof(bool), 200)]
+        [SwaggerOperation(Tags = new[] { "document" })]
+        public async Task<IActionResult> DeleteDocumentRelationship(DocumentRelationType relationshipType, [FromBody] DocumentRelationshipModel model)
+        {
+            switch (relationshipType)
+            {
+                case DocumentRelationType.Activities:
+                    var activityRelationship = _mapper.Map<PimsActivityInstanceDocument>(model);
+                    var result = await _documentService.DeleteActivityDocumentAsync(activityRelationship);
+                    return new JsonResult(result);
                 default:
                     throw new BadRequestException("Relationship type not valid.");
             }
--- a/backend/api/Areas/Documents/DocumentController.cs
+++ b/backend/api/Areas/Documents/DocumentController.cs
@@ -87,7 +87,7 @@ namespace Pims.Api.Controllers
         }
 
         /// <summary>
-        /// Gets a collection of documents for the specified type and owner id.
+        /// Deletes the specific document relationship for the given type.
         /// </summary>
         /// <param name="type">Used to identify document type.</param>
         /// <param name="entityId">Used to identify document's parent entity.</param>
--- a/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
+++ b/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
@@ -217,6 +217,7 @@ namespace Pims.Api.Repositories.Mayan
         public async Task<ExternalResult<string>> DeleteDocument(long documentId)
         {
             _logger.LogDebug("Deleting document...");
+            _logger.LogTrace("Document id {documentId}", documentId);
 
             string authenticationToken = await _authRepository.GetTokenAsync();
 
--- a/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
+++ b/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
@@ -214,6 +214,20 @@ namespace Pims.Api.Repositories.Mayan
             return retVal;
         }
 
+        public async Task<ExternalResult<string>> DeleteDocument(long documentId)
+        {
+            _logger.LogDebug("Deleting document...");
+
+            string authenticationToken = await _authRepository.GetTokenAsync();
+
+            Uri endpoint = new($"{this._config.BaseUri}/documents/{documentId}/");
+
+            var response = await DeleteAsync(endpoint, authenticationToken);
+
+            _logger.LogDebug($"Finished deleting document");
+            return response;
+        }
+
         public async Task<ExternalResult<DocumentDetail>> UploadDocumentAsync(int documentType, IFormFile file)
         {
             _logger.LogDebug("Uploading document...");
--- a/backend/api/Services/DocumentService.cs
+++ b/backend/api/Services/DocumentService.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -16,15 +17,23 @@ namespace Pims.Api.Services
     /// </summary>
     public class DocumentService : IDocumentService
     {
+        private readonly IDocumentRepository documentRepository;
         private readonly IDocumentActivityRepository documentActivityRespostory;
         private readonly IEdmsDocumentRepository documentStorageRepository;
         private readonly IDocumentTypeRepository documentTypeRepository;
         private readonly IAvService avService;
 
-        public DocumentService(IDocumentActivityRepository documentActivityRespostory, IEdmsDocumentRepository documentRepository, IDocumentTypeRepository documentTypeRepository, IAvService avService)
+        public DocumentService(
+            IDocumentRepository documentRepository,
+            IDocumentActivityRepository documentActivityRespostory,
+            IEdmsDocumentRepository documentStorageRepository,
+            IDocumentTypeRepository documentTypeRepository,
+            IAvService avService)
         {
+            this.documentRepository = documentRepository;
             this.documentActivityRespostory = documentActivityRespostory;
-            this.documentStorageRepository = documentRepository;
+            this.documentActivityRespostory = documentActivityRespostory;
+            this.documentStorageRepository = documentStorageRepository;
             this.documentTypeRepository = documentTypeRepository;
             this.avService = avService;
         }
@@ -37,6 +46,44 @@ namespace Pims.Api.Services
         public IList<PimsActivityInstanceDocument> GetActivityDocuments(long activityId)
         {
             return documentActivityRespostory.GetAll(activityId);
+        }
+
+        public async Task<bool> DeleteActivityDocumentAsync(PimsActivityInstanceDocument activityDocument)
+        {
+            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespostory.GetAll(activityDocument.ActivityInstanceId);
+            if (existingActivityDocuments.Count == 1)
+            {
+                return await DeleteDocumentAsync(activityDocument.Document);
+            }
+            else
+            {
+                documentActivityRespostory.Delete(activityDocument);
+                documentActivityRespostory.CommitTransaction();
+                return true;
+            }
+        }
+
+        public async Task<bool> DeleteDocumentAsync(PimsDocument document)
+        {
+            int count = documentRepository.GetTotalRelationCount(document.DocumentId);
+            if (count > 1)
+            {
+                throw new InvalidOperationException("Documents can only be removed if there is one or less relationships");
+            }
+            else
+            {
+                ExternalResult<string> result = await documentStorageRepository.DeleteDocument(document.MayanId);
+                if (result.Status == ExternalResultStatus.Success)
+                {
+                    documentRepository.Delete(document);
+                    documentRepository.CommitTransaction();
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+            }
         }
 
         public async Task<ExternalResult<QueryResult<DocumentType>>> GetStorageDocumentTypes(string ordering = "", int? page = null, int? pageSize = null)
--- a/backend/api/Services/DocumentService.cs
+++ b/backend/api/Services/DocumentService.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Net;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Pims.Api.Models;
@@ -53,6 +54,7 @@ namespace Pims.Api.Services
             IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespostory.GetAll(activityDocument.ActivityInstanceId);
             if (existingActivityDocuments.Count == 1)
             {
+                documentActivityRespostory.Delete(activityDocument);
                 return await DeleteDocumentAsync(activityDocument.Document);
             }
             else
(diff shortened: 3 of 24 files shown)


backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -72,7 +73,7 @@ public IActionResult GetDocumentTypes()
         //[HasPermission(Permissions.NoteView)]
         [ProducesResponseType(typeof(IList<DocumentRelationshipModel>), 200)]
         [SwaggerOperation(Tags = new[] { "document" })]
-        public IActionResult GetActivityDocuments(DocumentRelationType relationshipType, long parentId)
+        public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipType, long parentId)
```
asanchezr, line 76: We are still missing relevant permissions here? Did you file a JIRA to address that?

backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -85,6 +86,30 @@ public IActionResult GetActivityDocuments(DocumentRelationType relationshipType,
             }
         }
 
+        /// <summary>
+        /// Gets a collection of documents for the specified type and owner id.
```
asanchezr, line 90: copy / pasta

backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -214,6 +214,20 @@ public async Task<ExternalResult<FileDownload>> DownloadFileAsync(long documentI
             return retVal;
         }
 
+        public async Task<ExternalResult<string>> DeleteDocument(long documentId)
+        {
+            _logger.LogDebug("Deleting document...");
```
asanchezr, line 219: nit: would be nice to log the **documentId** here for tracing/debugging

backend/api/Services/DocumentService.cs
```diff
@@ -16,15 +18,23 @@ namespace Pims.Api.Services
     /// </summary>
     public class DocumentService : IDocumentService
     {
+        private readonly IDocumentRepository documentRepository;
         private readonly IDocumentActivityRepository documentActivityRespostory;
```
devinleighsmith, line 22: nit: documentActivityRespostory-> documentActivityRepository

backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -214,6 +214,20 @@ public async Task<ExternalResult<FileDownload>> DownloadFileAsync(long documentI
...
+
+            string authenticationToken = await _authRepository.GetTokenAsync();
+
+            Uri endpoint = new($"{this._config.BaseUri}/documents/{documentId}/");
+
+            var response = await DeleteAsync(endpoint, authenticationToken);
+
+            _logger.LogDebug($"Finished deleting document");
```
asanchezr, line 227: same comment re: documentId

backend/api/Services/DocumentService.cs
```diff
@@ -39,6 +49,45 @@ public IList<PimsActivityInstanceDocument> GetActivityDocuments(long activityId)
             return documentActivityRespostory.GetAll(activityId);
         }
 
+        public async Task<bool> DeleteActivityDocumentAsync(PimsActivityInstanceDocument activityDocument)
+        {
+            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespostory.GetAll(activityDocument.ActivityInstanceId);
```
asanchezr, line 54: nit: fix typo `documentActivityRespostory` -> `documentActivityRespos(i)tory`

devinleighsmith, line 54: Hmm, I don't think this is correct. I thought that the logic for deleting is:
1) If this document is used by **any other activity** then don't delete it, just remove the reference between this activity and the document.
2) otherwise remove the relationship and also delete the document in mayan.

Whereas this logic appears to be saying: if this is the only document associated with this activity, then delete the document and remove the reference. Otherwise only remove the reference

devinleighsmith (changes requested):

backend/dal/Repositories/DocumentActivityRepository.cs
```diff
@@ -41,25 +41,42 @@ public IList<PimsActivityInstanceDocument> GetAll(long activityId)
...
+        /// <param name="activityDocument"></param>
         /// <returns></returns>
-        public PimsActivityInstanceDocument Add(PimsActivityInstanceDocument documentActivity)
+        public PimsActivityInstanceDocument Add(PimsActivityInstanceDocument activityDocument)
         {
-            if (documentActivity == null)
+            if (activityDocument == null)
             {
-                throw new ArgumentNullException(nameof(documentActivity), "documentActivity cannot be null.");
+                throw new ArgumentNullException(nameof(activityDocument), "documentActivity cannot be null.");
```
asanchezr, line 57: nit: "activityDocument cannot be null."

backend/api/Areas/Documents/DocumentController.cs
FuriousLlama: I dont think there is a jira for documents permissiosn just yet

backend/dal/Repositories/DocumentActivityRepository.cs
```diff
@@ -41,25 +41,42 @@ public IList<PimsActivityInstanceDocument> GetAll(long activityId)
...
+        /// Deletes the passed document activity to the database.
+        /// </summary>
+        /// <param name="activityDocument"></param>
+        /// <returns></returns>
+        public bool Delete(PimsActivityInstanceDocument activityDocument)
+        {
+            if (activityDocument == null)
+            {
+                throw new ArgumentNullException(nameof(activityDocument), "activityDocument cannot be null.");
+            }
```
asanchezr, line 74: nit: I believe we have an extension method to reduce the amount of boilerplate code like this

```cs
using Pims.Core.Extensions;
...

activityDocument.ThrowIfNull(nameof(activityDocument));
```

backend/api/Services/DocumentService.cs
```diff
@@ -39,6 +49,45 @@ public IList<PimsActivityInstanceDocument> GetActivityDocuments(long activityId)
...
+            if (existingActivityDocuments.Count == 1)
```
devinleighsmith, line 55: Hmm, I don't think I understand this, we only try and delete the document if it is the only document associated with this activity?

devinleighsmith (changes requested):

backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        #region Methods
+
+        public int GetTotalRelationCount(long documentId)
+        {
+            var document = this.Context.PimsDocuments.AsNoTracking()
+                .Include(d => d.PimsActivityInstanceDocuments)
+                .Where(d => d.DocumentId == documentId)
+                .AsNoTracking()
+                .FirstOrDefault();
```
asanchezr, line 41: nit: I think this query doesn't need two "AsNoTracking()" - isn't it?
Also no need for Where

```cs
var document = this.Context.PimsDocuments.AsNoTracking()
                .Include(d => d.PimsActivityInstanceDocuments)
                .FirstOrDefault(d => d.DocumentId == documentId);

```

FuriousLlama: It was needed since we dont want to make modifications to any of the document information retrieved by the query

backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// Deletes the passed document from the database.
+        /// </summary>
+        /// <param name="document"></param>
+        /// <returns></returns>
+        public bool Delete(PimsDocument document)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document), "document cannot be null.");
+            }
```
asanchezr, line 57: same comment re: boilerplate

backend/api/Services/DocumentService.cs
FuriousLlama: If there is only one document activity relationship, then remove it alongside the document and the stored document, otherwise only remove this relationship

frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -16,24 +16,38 @@ const DocumentListContainer: React.FunctionComponent<IDocumentListContainerProps
...
-      if (documents !== undefined) {
-        setDocumentResults(documents);
-      }
-    };
+    retrieveDocuments();
+  }, [retrieveDocuments]);
 
-    fetch();
-  }, [props.parentId, props.relationshipType, retrieveDocumentRelationship]);
+  const deleteHandle = (documentRelationship: Api_DocumentRelationship): Promise<boolean> => {
```
devinleighsmith, line 33: handle?

frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -16,24 +16,38 @@ const DocumentListContainer: React.FunctionComponent<IDocumentListContainerProps
...
+    return new Promise<boolean>((resolve, reject) => {
```
devinleighsmith, line 34: Any reason why we are using a promise and not await here?

frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -16,24 +16,38 @@ const DocumentListContainer: React.FunctionComponent<IDocumentListContainerProps
...
+      deleteDocumentRelationship(props.relationshipType, documentRelationship).then(result => {
+        if (result) {
+          retrieveDocuments();
```
devinleighsmith, line 37: do we need to check if the user is still on the page (hasn't navigated away) here? (ismounted)

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -65,12 +70,36 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
...
+
+  const onDeleteConfirm = async () => {
+    if (selectedDocument) {
+      const documentRelationship = documentResults.find(
+        x => x.document?.id === selectedDocument.id,
+      );
+
+      if (documentRelationship !== undefined) {
+        props.deleteHandle(documentRelationship).then(result => {
+          console.log(result);
```
devinleighsmith, line 95: console.log also using promise instead of await.

devinleighsmith: also may need ismounted here.

frontend/src/hooks/pims-api/useApiDocuments.ts
```diff
@@ -27,8 +27,8 @@ export const useApiDocuments = () => {
 
       deleteDocumentRelationshipApiCall: (
```
asanchezr, line 28: I realize this is from previous PR but why break the naming convention here? 
We haven't used "*ApiCall" before - any reason why this cannot be simply named **deleteDocumentRelationship** ?

frontend/src/features/documents/hooks/useDocumentRelationshipProvider.ts
```diff
@@ -46,16 +46,24 @@ export const useDocumentRelationshipProvider = () => {
...
     ),
     requestName: 'deleteDocumentRelationship',
-    onSuccess: useCallback(() => toast.success('Delete document relationship'), []),
+    onSuccess: useCallback((response?: boolean) => {
+      if (response !== undefined && response) {
+        toast.success('Delete document relationship');
+      } else {
+        toast.error('Delete document relationship error. Check responses and try again.');
+      }
+    }, []),
```
asanchezr, line 66: question to learn - I'm a but confused by this... so the API could return 200 and still be an error?

backend/api/Services/DocumentService.cs
devinleighsmith: So if there are two different documents on an activity and we remove one, we will not delete that document even if that is the only instance of that document in the system? Won't that create an orphaned document in Mayan?

frontend/src/features/documents/list/DocumentListContainer.tsx
FuriousLlama: No hard reason, I can change to follow the standard

frontend/src/features/documents/hooks/useDocumentRelationshipProvider.ts
FuriousLlama: Yes, that would mean that the call to the external API was successful (httpstatus 200 /202) but the action was not what we wanted (i.e. not found).

frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -16,24 +16,38 @@ const DocumentListContainer: React.FunctionComponent<IDocumentListContainerProps
...
+        }
+        resolve(!!result);
+      });
+    });
+  };
```
asanchezr, line 42: nit: this is a matter of developer preference but would love to see this in async/await format as it lends to better code refactoring/maintenance etc

not a big deal of course

frontend/src/hooks/pims-api/useApiDocuments.ts
FuriousLlama: Mostly because it confuses the function that is being returned from the hook/provider vs the one that is used by the api call. I am open to suggestions about it but we should come to a stable pattern to avoid confusion in the future

frontend/src/features/documents/list/DocumentListContainer.tsx
FuriousLlama: handler?

frontend/src/features/documents/list/DocumentListView.test.tsx
```diff
@@ -34,6 +36,7 @@ describe('Document List View', () => {
         parentId={0}
         relationshipType={DocumentRelationshipType.FILES}
         documentResults={mockDocumentsResponse()}
+        deleteHandle={deleteMock}
```
asanchezr, line 39: please use `renderProps.deleteMock` instead to support overriding it with passed props (as shown in tests below)

asanchezr: Also, don't see any relevant unit test to cover the delete document scenario

frontend/src/features/documents/list/DocumentListView.test.tsx
```diff
@@ -70,8 +73,8 @@ describe('Document List View', () => {
       isLoading: false,
       parentId: 0,
       relationshipType: DocumentRelationshipType.FILES,
-
       documentResults: mockDocumentsResponse(),
+      deleteHandle: deleteMock,
```
asanchezr, line 77: GH doesn't let me comment on non-modified line but this test 'should have the Documents type in the component' and the next one 'should NOT have the Documents status in the component' seem to be doing the same? kinda confused by the test duplication...

frontend/src/features/documents/list/DocumentListView.tsx
asanchezr, line 95: console.log

backend/api/Services/DocumentService.cs
devinleighsmith: Sorry @FuriousLlama either I'm not understanding this still or this is incorrect, but I want to resolve this before I approve.

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -84,13 +113,36 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
...
+            </div>
+            <br />
+            <strong>Do you wish to continue deleting this document?</strong>
+          </>
+        }
+        handleOk={onDeleteConfirm}
+        handleCancel={() => setShowDeleteConfirmModal(false)}
+        okButtonText="Continue"
+        cancelButtonText="Cancel"
+        show
```
asanchezr, line 145: question - do we need to pass this **show** property when the modal display is controlled by he **display** props?

backend/api/Services/DocumentService.cs
FuriousLlama: I see what you are saying. This was selected based on the activity side of the relationship. The changed one selects based on the document.

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -17,6 +18,7 @@ export interface IDocumentListViewProps {
   documentResults: Api_DocumentRelationship[];
   hideFilters?: boolean;
   defaultFilters?: IDocumentFilter;
+  deleteHandle: (relationship: Api_DocumentRelationship) => Promise<boolean>;
```
asanchezr, line 21: nit: perhaps could name this "onDelete" to align with FE naming conventions? not a big deal

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -65,12 +70,36 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
...
     setIsDetailsVisible(true);
-    setSelectedDocument(values);
+    setSelectedDocument(document);
+  };
+
+  const handleViewDetailsClose = () => {
+    setIsDetailsVisible(false);
+    setSelectedDocument(undefined);
+  };
+  const handleDeleteClicked = (document: Api_Document) => {
```
asanchezr, line 82: nit: I think we use present tense when naming click handlers; ie "handleDeleteClick"

asanchezr (approved):

frontend/src/features/documents/list/DocumentListView.test.tsx
FuriousLlama: good catch, these tests were written for the list view and do not seem to provide value. I will remove one.

frontend/src/features/documents/list/DocumentListView.tsx
FuriousLlama: I dont think so.

backend/api/Areas/Documents/DocumentController.cs
asanchezr: would be nice if you file one as you are the developer for this feature - I worry we forget about permissions. Usually permissions are set on the story that handles update/read/create like this one (and the one before). 

If it wasn't called out specifically then we could end up not having permissions in place for this feature

asanchezr: more of a process thing - perhaps we can discuss process re: when to add claims to DB on Friday

backend/dal/Repositories/DocumentRepository.cs
asanchezr: then how about ?

```cs
this.Context.PimsDocuments
  .Include(d => d.PimsActivityInstanceDocuments)
  .AsNoTracking()
...
```

frontend/src/features/documents/list/DocumentListView.test.tsx
FuriousLlama: There are almost none at this point. The reason is that is changing so quickly that I don't think it would be good to add tests until it gets into a more stable point.

devinleighsmith (approved):

PR #1818, opened 2022-07-30
PSP-3250 Create Acquisition File

asanchezr:
create-acq-file


backend/api/Controllers/LookupController.cs
```diff
@@ -153,6 +156,9 @@ public IActionResult GetAll()
             codes.AddRange(propertyVolumetricTypes);
             codes.AddRange(pphStatusType);
             codes.AddRange(documentStatusTypes);
+            codes.AddRange(acquisitionFileStatusTypes);
+            codes.AddRange(acquisitionPhysFileStatusTypes);
+            codes.AddRange(acquisitionTypes);
```
asanchezr, line 161: new code tables

backend/api/Services/AcquisitionFileService.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            acquisitionFile.ThrowIfNull(nameof(acquisitionFile));
+
+            _logger.LogInformation("Updating acquisition file with id {id}", acquisitionFile.Id);
+            _user.ThrowIfNotAuthorized(Permissions.AcquisitionFileEdit);
+
+            ValidateVersion(acquisitionFile.Id, acquisitionFile.ConcurrencyControlNumber);
+
+            // TODO: Implementation pending
+            throw new System.NotImplementedException();
```
asanchezr, line 61: This is a placeholder implementation for now that throws an exception - but includes boilerplate for whenever this story gets implemented

backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        /// <summary>
+        /// Updates the specified acquisition file.
+        /// </summary>
+        /// <param name="acquisitionFile"></param>
+        /// <returns></returns>
+        public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile)
+        {
+            acquisitionFile.ThrowIfNull(nameof(acquisitionFile));
```
asanchezr, line 64: boilerplate for update - pending implementation

backend/dal/Security/Permissions.cs
```diff
@@ -84,20 +84,40 @@ public enum Permissions
...
+        DocumentView = 35,
+
+        [Display(GroupName = "document", Name = "document-add", Description = "Can add new documents.")]
+        DocumentAdd = 36,
+
+        [Display(GroupName = "document", Name = "document-edit", Description = "Can update existing  documents.")]
+        DocumentEdit = 37,
+
+        [Display(GroupName = "document", Name = "document-delete", Description = "Can delete existing documents.")]
+        DocumentDelete = 38,
```
asanchezr, line 121: I added the rest of documents permissions here because only DocumentView got added on a prior PR.

Note this is not the "real" claims that need to be added to the DB - this is just an enum matching the claims name

frontend/src/components/layout/SideNavBar/AcquisitionTray.tsx
```diff
@@ -20,7 +20,7 @@ export const AcquisitionTray = ({ onLinkClick }: ISideTrayPageProps) => {
         </Link>
       )}
       {hasClaim(Claims.ACQUISITION_ADD) && (
-        <Link onClick={onLinkClick} to="/acquisition/new">
+        <Link onClick={onLinkClick} to="/mapview/acquisition/new">
```
asanchezr, line 23: Fix route

frontend/src/features/admin/access-request/__snapshots__/AccessRequestContainer.test.tsx.snap
```diff
@@ -352,15 +352,27 @@ exports[`AccessRequestContainer component makes a request based on the access re
             </option>
             <option
               class="option"
-              value="5"
+              value="6"
             >
-              Real Estate Analyst
+              Finance
```
asanchezr, line 357: Lot of updates to snapshots because I updated the mock code table lookups with up to date values

frontend/src/features/mapSideBar/tabs/Section.tsx
```diff
@@ -18,7 +18,7 @@ export const Section: React.FC<SectionProps> = ({
 }) => {
   const [isCollapsed, setIsCollapsed] = useState<boolean>(!initiallyExpanded && true);
   return (
-    <StyledFormSection>
+    <StyledFormSection className="form-section">
```
asanchezr, line 21: added this "marker" class to facilitate style overrides - For example this component defines a lot of padding and margin. Perhaps that kind of spacing should be left out to the parent containers - maybe discuss on Friday

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React, { useCallback } from 'react';
+
+interface IAcquisitionContainerProps {
+  acquisitionFileId: number;
+  onClose: () => void;
+}
+
+export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainerProps> = props => {
+  const { onClose } = props;
+  const close = useCallback(() => onClose && onClose(), [onClose]);
```
asanchezr, line 12: Placeholder container for Acquisition Details (to be implemented)

frontend/src/features/properties/map/acquisition/add/AddAcquisitionFileYupSchema.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-template-curly-in-string */
```
asanchezr, line 1: need to disable this lint rule to be able to parameterize yup validation string below. 

Perhaps we should discuss adding a project-wide exception for yup files

frontend/src/features/properties/map/acquisition/add/AddAcquisitionFileYupSchema.ts
```diff
@@ -0,0 +1,10 @@
...
+import * as Yup from 'yup';
+
+export const AddAcquisitionFileYupSchema = Yup.object().shape({
+  name: Yup.string()
+    .required('Acquisition file name is required')
+    .max(500, 'Acquisition file name must be at most ${max} characters'),
```
asanchezr, line 7: this is the same as saying "blah must be at most 500 chars" but with advantage that if we ever change the limit we don't have to update the message as well. Seems useful

frontend/src/features/mapSideBar/tabs/Section.tsx
FuriousLlama: Yeah, I agree that the parent is the one that should figure out if the child needs spacing or not since it might be composed of multiple children,

FuriousLlama (approved):

PR #1820, opened 2022-08-02
Upload Document + Document details adjustments | Features/psp 2942

FuriousLlama:
image

image

image


backend/dal/Repositories/AccessRequestRepository.cs
```diff
@@ -29,10 +29,9 @@ public class AccessRequestRepository : BaseRepository<PimsAccessRequest>, IAcces
         /// </summary>
         /// <param name="dbContext"></param>
         /// <param name="accessRequest"></param>
-        /// <param name="service"></param>
         /// <param name="logger"></param>
         /// <returns></returns>
-        public AccessRequestRepository(PimsContext dbContext, System.Security.Claims.ClaimsPrincipal user, ClaimsPrincipal accessRequest, IPimsRepository service, ILogger<AccessRequestRepository> logger, IMapper mapper) : base(dbContext, user, service, logger, mapper) { }
+        public AccessRequestRepository(PimsContext dbContext, System.Security.Claims.ClaimsPrincipal user, ClaimsPrincipal accessRequest, ILogger<AccessRequestRepository> logger) : base(dbContext, user, logger) { }
```
FuriousLlama, line 34: These where just general cleanup of the base classes givenrepositories dont need access to the general service or repository provider

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+                <Select field="documentStatusCode" options={documentStatusTypes} />
+              </SectionField>
+
+              <StyledH3>Details</StyledH3>
+              <StyledScrollable>
+                {props.mayanMetadata?.length === 0 && (
+                  <StyledNoData>No additional data</StyledNoData>
+                )}
+                {props.mayanMetadata?.map(value => (
+                  <SectionField
```
FuriousLlama, line 131: Section left here as placeholder once upload metadata functionality is added

backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -87,11 +87,38 @@ public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipT
...
         /// <summary>
-        /// Deletes the specific document relationship for the given type.
+        /// Uploads a document for the given relationship.
         /// </summary>
         /// <param name="type">Used to identify document type.</param>
         /// <param name="entityId">Used to identify document's parent entity.</param>
         /// <returns></returns>
+        [HttpPost("upload/{relationshipType}/{parentId:long}")]
+        [Produces("application/json")]
+        //[HasPermission(Permissions.PropertyAdd)]
```
asanchezr, line 97: we really need to enforce permissions here - there has been a couple PRs but not actual permissions in place

backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -87,11 +87,38 @@ public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipT
...
+            )
+        {
+            switch (relationshipType)
+            {
+                case DocumentRelationType.Activities:
+                    var response = await _documentService.UploadActivityDocumentAsync(parentId, uploadRequest);
+                    return new JsonResult(response);
+                default:
+                    throw new BadRequestException("Relationship type not valid.");
+            }
```
asanchezr, line 113: question - I believe we kinda agreed to do this kind of logic in the service to try to keep controllers as lean as possible

backend/api/Repositories/Mayan/IEdmsDocumentRepository.cs
```diff
@@ -29,7 +29,7 @@ public interface IEdmsDocumentRepository
 
         Task<ExternalResult<string>> DeleteDocument(long documentId);
 
-        Task<ExternalResult<DocumentDetail>> UploadDocumentAsync(int documentType, IFormFile file);
+        Task<ExternalResult<DocumentDetail>> UploadDocumentAsync(long documentType, IFormFile file);
```
asanchezr, line 32: Thanks for fixing this!

frontend/package.json
```diff
@@ -141,6 +141,7 @@
     "start": "react-app-rewired start",
     "build": "cross-env CI=true GENERATE_SOURCEMAP=false INLINE_RUNTIME_CHUNK=false react-app-rewired build",
     "test": "tsc && react-app-rewired test --ci --watchAll=false",
+    "test-u": "tsc && react-app-rewired test --ci --watchAll=false -u",
```
asanchezr, line 144: I'm okay with this but we already have an "update-snapshots" npm task - perhaps update that one instead?

frontend/src/features/documents/documentDetail/DocumentDetailView.tsx
```diff
@@ -16,17 +16,13 @@ interface IDocumentDetailsViewProps {
...
+  const documentFileName = props.document.pimsDocument?.fileName;
+  const mayanDocumentId = props.document.pimsDocument?.mayanDocumentId || -1;
 
+  let mayanFileId = undefined;
   if (props.document.mayanMetadata !== undefined && props.document.mayanMetadata?.length > 0) {
     const document = props.document.mayanMetadata[0].document;
-    documentTypeLabel = document.document_type.label;
-    documentFileName = document.label;
-    mayanDocumentId = document.id;
     mayanFileId = document.file_latest.id;
```
asanchezr, line 26: There's still a chance for NPE if `document.mayanMetadata[0] == null` 

perhaps consider:
```tsx
const document = props.document.mayanMetadata[0]?.document;
```

frontend/src/features/documents/documentUpload/ComposedDocument.ts
```diff
@@ -0,0 +1,7 @@
+import { Api_Document } from 'models/api/Document';
+import { Api_Storage_DocumentMetadata } from 'models/api/DocumentStorage';
+
+export interface ComposedDocument {
+  mayanMetadata?: Api_Storage_DocumentMetadata[];
```
asanchezr, line 5: nit: I like how we moved from `Mayan_DocumentMetadata` to `Api_Storage_DocumentMetadata` to avoid leaking mayan   details. Would you consider renaming this property name to match - ie "storageMetadata" ?

frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    const fetch = async () => {
+      const axiosResponse = await getDocumentTypes();
+      if (axiosResponse && isMounted()) {
+        setDocumentTypes(axiosResponse.data);
+      }
+    };
+
+    fetch();
+  }, [getDocumentTypes, isMounted, retrieveDocumentMetadata]);
```
asanchezr, line 36: question re: useEffect dependency array - do we need to check for "retrieveDocumentMetadata" here? Doesn't seem to be in use inside the effect

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+interface DocumentUploadForm {
+  documentTypeId: string;
+  documentStatusCode: string;
+}
+
+interface IDocumentUploadViewProps {
+  documentTypes: Api_DocumentType[];
+  isLoading: boolean;
+  mayanMetadata: Api_Storage_DocumentTypeMetadataType[];
```
asanchezr, line 24: nit: please consider renaming this prop to "storageMetadata"

asanchezr: Also, it is usually helpful (in tests etc) to be able to import the interface type - please consider exporting the interfaces as well here

backend/api/Areas/Documents/DocumentController.cs
FuriousLlama: Well we chated on it, but did not conclude on where the logic should be. I think we have a meeting later this week to go about that as well as the routes

frontend/package.json
FuriousLlama: Not sure abut the update-snapshots since it looks like it does some extra stuff. I will happily merge them if the other one is not used/useful

frontend/src/features/documents/documentUpload/ComposedDocument.ts
FuriousLlama: For sure. That is something that could be done on the next pr that does something for documents given it would be another set of changes

frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
FuriousLlama: good catch. Is not needed so i've removed it

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+    return { label: x.documentType || '', value: x.id?.toString() || '' };
+  });
+
+  const { getOptionsByType } = useLookupCodeHelpers();
+  const documentStatusTypes = getOptionsByType(API.DOCUMENT_STATUS_TYPES);
+
+  const initialFormState: DocumentUploadForm = {
+    documentTypeId: documentTypes.length > 0 ? documentTypes[0].value.toString() || '' : '',
+    documentStatusCode:
+      documentStatusTypes.length > 0 ? documentStatusTypes[0].value.toString() || '' : '',
```
asanchezr, line 46: I'm not sure what comes back in these arrays but feel like we could be defensive and check for nulls - ie `documentTypes[0]?.value?.toString() || ''`

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+  };
+
+  const handleFileInput = (changeEvent: ChangeEvent<HTMLInputElement>) => {
+    // handle validations
+    if (changeEvent.target !== null) {
+      var target = changeEvent.target;
+      if (target.files !== null) {
```
asanchezr, line 53: nit: there's a chance that the files array is empty (and thus not null) but `files[0]` would return undefined

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+              var request: Api_UploadRequest = {
+                documentStatusCode: values.documentStatusCode,
+                documentType: selectedDocumentType,
+                file: selectedFile,
+              };
+
+              await props.onUploadDocument(request);
+              setSubmitting(false);
+            } else {
+              console.error('Selected document type is not valid');
```
asanchezr, line 81: nit: I believe we don't use `console.error` for these - perhaps `toast.error(...)` ?

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }}
+      >
+        {formikProps => (
+          <>
+            <div className="pb-4">
+              Choose the document type and select "Browse" to choose the file to upload from your
+              computer or network to PIMS.
+            </div>
+            <SectionField label={'Document type'} labelWidth="3" className="pb-2">
```
asanchezr, line 92: nit: if label is just a string better to just `<SectionField label="Blah" ...>`

frontend/src/features/documents/list/DocumentListView.tsx
```diff
@@ -100,10 +115,29 @@ export const DocumentListView: React.FunctionComponent<IDocumentListViewProps> =
...
   return (
-    <Styled.ListPage>
-      <Styled.Scrollable vertical={true}>
-        <Styled.PageHeader>Documents</Styled.PageHeader>
+    <div>
+      <Section
+        header={
+          <Row>
+            <Col xs="auto">Documents</Col>
+            {hasClaim(Claims.RESEARCH_ADD) && (
```
asanchezr, line 129: is this the right claim?

frontend/src/features/documents/documentUpload/DocumentUploadView.tsx
FuriousLlama: This would be more for us. I've not seen this actually happening in live code, but since the type could be undefined is here.

frontend/src/features/documents/list/DocumentListView.tsx
FuriousLlama: I dont think so. It suffers from the same claims issues other document sections has

frontend/src/models/api/Document.ts
```diff
@@ -15,6 +17,7 @@ export interface Api_Document extends Api_ConcurrentVersion, Api_AuditFields {
 export interface Api_DocumentType extends Api_ConcurrentVersion, Api_AuditFields {
   id?: number;
   documentType?: string;
+  mayanId?: number;
```
asanchezr, line 20: nit: would prefer to see this named "storageId"

FuriousLlama: it could. The mayan* could be abstracted to be document storage

asanchezr (approved):

frontend/package.json
asanchezr: I don't believe we are using the other for anything else so I'm okay replacing it with your version

asanchezr: `--updateSnapshot == -u`

PR #1833, opened 2022-08-03
fix overwritten build scripts.

devinleighsmith:
The build scripts were overwritten from is32 -> is29


devinleighsmith:
Hey @dfilteau not sure how this happened but some build scripts in IS33 got reverted back to the state they were in from IS29. I've set up a meeting to discuss this tmrw.

asanchezr (approved):

PR #1861, opened 2022-08-04
HOTFIX psp-4084 control deposit, terms, payments via LEASE_EDIT permission.

devinleighsmith:


frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "2.1.1-32.26",
+  "version": "2.1.1-32.27",
```
asanchezr, line 3: hotfix # should be `2.1.2-32.27`

asanchezr (approved):

PR #1864, opened 2022-08-05
Fix issue with github action logic, should run either on a bcgov pr o…

devinleighsmith:
…r on push (not both which is not possible)

### DIFF
--- a/.github/workflows/api-dotnetcore.yml
+++ b/.github/workflows/api-dotnetcore.yml
@@ -101,7 +101,7 @@ jobs:
       - name: SonarScanner for .NET 6 with pull request decoration support
         id: scan
         uses: highbyte/sonarscan-dotnet@v2.1.2
-        if: ${{ github.event_name == 'push' && github.event.pull_request.head.repo.full_name == github.repository }}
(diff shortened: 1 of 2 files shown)


.github/workflows/api-dotnetcore.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - name: SonarScanner for .NET 6 with pull request decoration support
         id: scan
         uses: highbyte/sonarscan-dotnet@v2.1.2
-        if: ${{ github.event_name == 'push' && github.event.pull_request.head.repo.full_name == github.repository }}
+        if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == 'bcgov/PSP' }}
```
devinleighsmith, line 104: We want this to run if
1) there is a push within the repo
2) there is a pr from a branch in the repo to another branch in the repo (ie. not a fork).

FuriousLlama (approved):

asanchezr (approved):

FuriousLlama (changes requested):
It does not look like it worked

FuriousLlama (approved):

PR #1867, opened 2022-08-05
HOTFIX: Fix keycloak sync

devinleighsmith:
keycloak sync was not successfully adding roles to newly created users. This was due to a route change a model change.

### DIFF
--- a/tools/keycloak/sync/Properties/launchSettings.json
+++ b/tools/keycloak/sync/Properties/launchSettings.json
@@ -1,8 +1,7 @@
 {
   "profiles": {
     "Pims.Tools.Keycloak.Sync": {
-      "commandName": "Project",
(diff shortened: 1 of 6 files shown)


tools/keycloak/sync/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Pims.Tools.Keycloak.Sync": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "workingDirectory": "C:\\git\\PSP\\tools\\keycloak\\sync"
```
asanchezr, line 5: is `workingDirectory` required? this will not match the local for each developer

asanchezr (approved):

PR #1871, opened 2022-08-08
Updated initial gen files and makefile

FuriousLlama:


frontend/src/setupProxy.js
```diff
@@ -4,7 +4,7 @@ module.exports = function(app) {
   app.use(
     '/api',
     createProxyMiddleware({
-      target: process.env.API_URL || 'http://backend:8080/',
+      target: process.env.API_URL || 'http://localhost:5000/',
```
FuriousLlama, line 7: This was most likely going nowhere. Thoughts @asanchezr @devinleighsmith?

scripts/gen-env-files.sh
```diff
@@ -95,7 +95,7 @@ if test -f "./frontend/.env"; then
 else
 echo \
 "NODE_ENV=development
-API_URL=http://backend:8080/
+API_URL=http://localhost:5000/
```
FuriousLlama, line 98: @asanchezr @devinleighsmith do you know if the gen configuration is used by non development? otherwise this should be fine

frontend/README.md
```diff
@@ -8,15 +8,15 @@ Create a .env file and populate it with this.
 
 ```conf
 NODE_ENV=development
-API_URL=http://backend:8080/
+API_URL=http://localhost:5000/
```
FuriousLlama, line 11: not 100% sure if these configurations are used for non-development. If they are they can be reverted

Makefile
```diff
@@ -313,7 +313,7 @@ env: ## Generate env files
 
 mayan-up: ## Calls the docker compose up for the mayan images
 	@echo "$(P) Create or start mayan-edms system"
-	@cd tools/mayan-edms; docker-compose up -d
+	@cd tools/mayan-edms; docker-compose --profile all up -d
```
FuriousLlama, line 316: this was causing issues when deploying mayan from scratch

Makefile
```diff
@@ -274,7 +274,7 @@ db-upgrade: ## Upgrade an existing database to the TARGET_VERSION (if passed) or
 
 db-scaffold: ## Requires local install of sqlcmd
 	@echo "$(P) regenerate ef core entities from database"
-	@cd backend/dal; eval $(grep -v '^#' .env | xargs) dotnet ef dbcontext scaffold Name=PIMS Microsoft.EntityFrameworkCore.SqlServer -o ../entities/ef --schema dbo --context PimsContext --context-namespace Pims.Dal --context-dir . --startup-project ../api --no-onconfiguring --namespace Pims.Dal.Entities --data-annotations -v -f
+	@cd backend/entities; eval $(grep -v '^#' .env | xargs) dotnet ef dbcontext scaffold Name=PIMS Microsoft.EntityFrameworkCore.SqlServer -o ../entities/ef --schema dbo --context PimsBaseContext --context-namespace Pims.Dal --context-dir . --no-onconfiguring --namespace Pims.Dal.Entities --data-annotations -v -f
```
FuriousLlama, line 277: forgot to update the makefile the last time I made changes to the scaffold process. This should match it. Note that because it cannot find PIMS on my machine it fails. Does that need to be setup separately

PryancaJSharma (approved):

frontend/src/setupProxy.js
asanchezr: this was going to the backend API docker container when we used to run the API in docker locally. We don't anymore so I'm okay with this change. You should have API_URL defined in your env anyways

scripts/gen-env-files.sh
asanchezr: same comment as above

.vscode/launch.json
```diff
@@ -21,7 +21,7 @@
       "request": "launch",
       "console": "internalConsole",
       "preLaunchTask": "build",
-      "program": "${workspaceRoot}/tools/converters/excel/bin/Debug/net5.0/Pims.Tools.Converters.ExcelConverter.dll",
+      "program": "${workspaceRoot}/tools/converters/excel/bin/Debug/net6.0/Pims.Tools.Converters.ExcelConverter.dll",
```
asanchezr, line 24: This whole section can be safely removed - we don't have the excel converter in the solution anymore

.vscode/launch.json
```diff
@@ -31,16 +31,16 @@
       "request": "launch",
       "console": "internalConsole",
       "preLaunchTask": "build",
-      "program": "${workspaceRoot}/tools/import/bin/Debug/net5.0/Pims.Tools.Import.dll",
```
asanchezr, line 34: This whole section can be safely removed - we don't have the data import tool in the solution anymore

Makefile
asanchezr: Will let @devinleighsmith answer this one - not sure about that error

frontend/README.md
asanchezr: As said before - these are here since the initial dev setup that had the API running in docker containers (instead of locally in VS Code). Since then we have moved on from that. Unless @devinleighsmith has a different opinion, I'm okay with this change

asanchezr (approved):
Approved but want to hear comments on that **db-scaffold** error. Make commands should run as-is without errors...

.vscode/launch.json
FuriousLlama: Removed !

PR #1875, opened 2022-08-10
BUGFIX: PSP-4036 Edit notes: pop-up has an "x" button at the top right that doesn't work

asanchezr:


frontend/src/features/notes/detail/NoteDetailsFormModal.tsx
```diff
@@ -79,6 +79,7 @@ export const NoteDetailsFormModal: React.FC<INoteDetailsFormModalProps> = props
       message={loading ? spinner : body}
       okButtonText="Close"
       handleOk={() => onCloseClick && onCloseClick()}
+      handleCancel={() => onCloseClick && onCloseClick()}
```
FuriousLlama, line 82: these could be avoided if the setDisplay param is passed. Also, given the function have the same parameters it could be passed directly 
```
handleOk={onCloseClick}
```

FuriousLlama (approved):

PryancaJSharma (approved):

PR #1879, opened 2022-08-12
PSP-3253 List View - Acquisition Files

asanchezr:
image


frontend/src/features/acquisition/list/AcquisitionFilter/AcquisitionFilter.tsx
```diff
@@ -0,0 +1,137 @@
...
+  .idir-input-group {
+    .input-group-prepend select {
+      width: 16rem;
+    }
+    input {
+      width: 18rem;
+      max-width: 100%;
+    }
+  }
+`;
```
FuriousLlama, line 133: this is something we could standardize along the other filters

FuriousLlama (approved):
Looks good. It might need to be re-factored to match the schema of s34 models

PR #1885, opened 2022-08-13
psp-4065 prevent requests to LTSA when no valid pid is associated.

devinleighsmith:


frontend/src/features/mapSideBar/tabs/ltsa/LtsaTabView.tsx
```diff
@@ -22,22 +22,35 @@ export interface ILtsaTabViewProps {
...
+      {!pid ? (
+        <FormSection>
+          <b>
+            This property does not have a valid PID.
+            <br />
+            <br /> Only properties that are associated to a valid PID can display corresponding data
+            from LTSA.
+          </b>
+        </FormSection>
+      ) : !loading && !ltsaData ? (
```
FuriousLlama, line 49: Instead of having both on the same markup it could be in two different returns.
```
if(notValidLtsa)
{
   return <>Sorry, no ltsdataa</>
}
else
{
   return <>...Stuff that goes with ltsa</>
}
```

FuriousLlama (approved):

PR #1892, opened 2022-08-15
Standarized the display of map properties

FuriousLlama:


frontend/src/features/properties/selector/utils.ts
```diff
@@ -1,15 +1,56 @@
...
+  label: NameSourceType;
+  value: string;
+}
+
+export const getPropertyName = (property: IMapProperty): PropertyName => {
+  if (property.pid !== undefined && property.pid?.length > 0 && property.pid !== '0') {
+    return { label: NameSourceType.PID, value: pidFormatter(property.pid.toString()) };
+  } else if (property.pin !== undefined && property.pin?.length > 0 && property.pin !== '0') {
+    return { label: NameSourceType.PIN, value: property.pin.toString() };
+  } else if (property.planNumber !== undefined) {
```
devinleighsmith, line 25: assuming this is a string (I think it is), shouldn't this also check for non-zero length? also worth mentioning that !!property.pid checks for non-undefined, non-null, and also non-empty.

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.tsx
```diff
@@ -189,7 +188,6 @@ export const featuresToIdentifiedMapProperty = (
           longitude: boundedCenter[0],
           legalDescription: feature?.properties?.LEGAL_DESCRIPTION,
         };
-        property.id = getPropertyIdentifier(property);
```
devinleighsmith, line 192: This seems like it could break something. Can you confirm that duplicate detection still works when attempting to add multiple of the same property?

frontend/src/features/properties/selector/search/PropertySearchSelectorFormView.test.tsx
```diff
@@ -187,54 +189,35 @@ describe('PropertySearchSelectorFormView component', () => {
...
   describe('selecting results', () => {
     it('search results can be selected', async () => {
-      const { findByTestId } = setup({ searchResults: mockPropertyLayerSearchResponse.features });
-      const checkbox = await findByTestId(
-        'selectrow-WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW.fid-674bf6f8_180d8c9b18e_7c12',
-      );
+      const { findByTestId, container } = setup({
+        searchResults: toMapProperty(mockPropertyLayerSearchResponse.features),
+      });
+      console.log(container.outerHTML);
```
devinleighsmith, line 195: can we remove this to avoid cluttering up the logs?

frontend/src/features/properties/selector/components/MapClickMonitor.tsx
```diff
@@ -40,7 +40,7 @@ export const mapFeatureToProperty = (
     .getBounds()
     .getCenter();
   return {
-    pid: selectedFeature?.properties?.PID_NUMBER ?? undefined,
+    pid: selectedFeature?.properties?.PID ?? undefined,
```
devinleighsmith, line 43: what is the rationale for this change?

frontend/src/features/properties/map/research/add/SelectedPropertyRow.tsx
```diff
@@ -21,7 +21,28 @@ export const SelectedPropertyRow: React.FunctionComponent<ISelectedPropertyRowPr
   index,
   property,
 }) => {
-  const propertyIdentifier = getPropertyIdentifier(property);
+  const propertyName = getPropertyName(property);
+  let propertyIdentifier = '';
+  switch (propertyName.label) {
+    case NameSourceType.PID:
```
devinleighsmith, line 27: you can collapse all of the cases to reduce code duplication.

frontend/src/features/properties/map/research/add/SelectedPropertyRow.tsx
```diff
@@ -21,7 +21,28 @@ export const SelectedPropertyRow: React.FunctionComponent<ISelectedPropertyRowPr
...
+      propertyIdentifier = `${propertyName.label}: ${propertyName.value}`;
+      break;
+    case NameSourceType.PIN:
+      propertyIdentifier = `${propertyName.label}: ${propertyName.value}`;
+      break;
+    case NameSourceType.PLAN:
+      propertyIdentifier = `${propertyName.label}: ${propertyName.value}`;
+      break;
+    case NameSourceType.LOCATION:
```
devinleighsmith, line 36: and then collapse these two.

devinleighsmith (approved):

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.tsx
FuriousLlama: Yeah I checked that logic and updated the generation of id's in the table and in the selection as well as tests

frontend/src/features/properties/selector/search/PropertySearchSelectorFormView.test.tsx
FuriousLlama: for sure!

frontend/src/features/properties/selector/components/MapClickMonitor.tsx
FuriousLlama: The definition of the IMapProperty defines the pid to be a string or undefined. Pid number was returning a number instead breaking the type promise for the interface and causing some issues downstream. i.e. checks for length of a string

FuriousLlama: We should choose if we want to handle the pid as a string or a number in the frontend and abstract the conversions to the lowest level. That way we can have some guarantees when coding against that piece of info

frontend/src/features/properties/map/research/add/SelectedPropertyRow.tsx
FuriousLlama: sure can do

frontend/src/features/properties/selector/utils.ts
FuriousLlama: We could, but perhaps that is another conversation about the expectations for a plan number. Maybe empty string is valid? (very odd)

PR #1897, opened 2022-08-16
HOTFIX: Psp 3983

devinleighsmith:
pims production deployment git action(s)

### DIFF
--- a/.github/workflows/ci-cd-pims-master.yml
+++ b/.github/workflows/ci-cd-pims-master.yml
@@ -81,6 +81,28 @@ jobs:
         run: |
           ./openshift/4.0/player.sh build api -apply
 
+  tag-images:
+    name: Tag frontend and api images with version number
+    needs: [build-frontend, build-api]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Source Code
+        uses: actions/checkout@v2
+        with:
+          ref: master
+      - name: Fetch Tags
+        run: git fetch --unshallow --tags
+      - name: Login to OpenShift
+        uses: redhat-actions/oc-login@v1
+        with:
+          openshift_server_url: ${{ env.OPENSHIFT_SERVER }}
+          openshift_token: ${{ env.OPENSHIFT_TOKEN }}
+          insecure_skip_tls_verify: true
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-cd-pims-master.yml
```diff
@@ -81,6 +81,28 @@ jobs:
         run: |
           ./openshift/4.0/player.sh build api -apply
 
+  tag-images:
```
devinleighsmith, line 84: uat hasn't been tagging its images, not a big deal, but it should be doing this going forward.

.github/workflows/deploy-prod-end.yml
```diff
@@ -0,0 +1,47 @@
+name: Hide prod maintenance page after deployment.
```
devinleighsmith, line 1: The idea behind splitting this up is that there are manual steps involved in prod deployments, so it doesn't make sense to burn through all of the steps and then take down the maintenance page and post a completion notification as that would likely finish before the manual steps are completed.

Instead, the deployer would run the prod deploy start script, wait for it to finish, then run any manual steps, then run the prod deploy end script.

.github/workflows/depoy-prod-start.yml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - name: Checkout Source Code
+        uses: actions/checkout@v2
+        with:
+          ref: master
+          fetch-depth: 0
+      - name: Display the maintenance page instead of the app
+        run: |
+          ./maintenance/maintenance.sh prod on
+      - name: "Get Previous tag"
```
devinleighsmith, line 58: this mimics the get previous tag functionality from the pipeline.

maintenance/maintenance.sh
```diff
@@ -58,18 +58,6 @@ then
   exit
 fi
 
-
```
devinleighsmith, line 61: This check doesn't work for our use case - ie. a service account with access to only the tools project but with rbac permissions that allow interaction with resources in other projects (ie. dev).

FuriousLlama (approved):
If we are deploying on OS, why is it necessary for it to have downtime?

devinleighsmith:
@FuriousLlama for now we have manual steps for every prod release that must be completed before the environment can be made available.

PR #1905, opened 2022-08-17
psp-3982 modify fa url, update timeouts.

devinleighsmith:


frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.test.tsx
```diff
@@ -59,7 +59,11 @@ describe('PropertySelectorSearchContainer component', () => {
       )
       .reply(200, mockPropertyLayerSearchResponse)
       // parcel map layer (fully attributed)
-      .onGet(new RegExp('ogs-internal/ows'))
+      .onGet(
+        new RegExp(
+          'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW/ows',
+        ),
+      )
```
FuriousLlama, line 66: Does this mean we are no longer using/using less of our geoserver?

FuriousLlama (approved):

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.test.tsx
devinleighsmith: Yes, there is now a public version of the FA data that we can use that doesn't require local geoserver. I'm sad that we weren't aware of that because we could have just waited and saved ourselves some time.

PR #1913, opened 2022-08-17
Psp 3910

devinleighsmith:


backend/api/Areas/Activities/Controllers/ActivityController.cs
```diff
@@ -0,0 +1,116 @@
...
+        #endregion
+
+        #region Endpoints
+
+        /// <summary>
+        /// Add the specified activity.
+        /// </summary>
+        /// <param name="activityModel">The activity to add.</param>
+        /// <returns></returns>
+        [HttpPost("")]
```
FuriousLlama, line 53: no need for the parenthesis given there is no specific route

backend/api/Areas/Activities/Controllers/ActivityController.cs
```diff
@@ -0,0 +1,116 @@
...
+        [HasPermission(Permissions.NoteAdd)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(ActivityInstanceModel), 200)]
+        [SwaggerOperation(Tags = new[] { "activity" })]
+        public IActionResult AddActivity(NoteType type, [FromBody] ActivityInstanceModel activityModel)
```
FuriousLlama, line 58: NoteType?

backend/api/Services/ActivityService.cs
```diff
@@ -0,0 +1,122 @@
...
+            _noteService = noteService;
+            _documentService = documentService;
+        }
+
+        public PimsActivityInstance GetById(long id)
+        {
+            _logger.LogInformation("Getting activity {id}", id);
+            this.User.ThrowIfNotAuthorized(Permissions.NoteView);
+
+            // _user.ThrowIfNotAuthorized(Permissions.ResearchFileView);
```
FuriousLlama, line 41: ?

backend/api/Services/ActivityService.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            return activityInstance;
+        }
+
+        public IList<PimsActivityInstance> GetAllByResearchFileId(long researchFileId)
+        {
+            _logger.LogInformation("Getting activities by research id {researchFileId}", researchFileId);
+            this.User.ThrowIfNotAuthorized(Permissions.NoteView);
+
+            // _user.ThrowIfNotAuthorized(Permissions.ResearchFileView);
```
FuriousLlama, line 52: Permissions are not correct. Thoroughout.

backend/api/Services/ActivityService.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public PimsActivityInstance Add(PimsActivityInstance instance)
+        {
+            _logger.LogInformation("Adding activity instance ...");
+            this.User.ThrowIfNotAuthorized(Permissions.NoteAdd);
+
+            // _user.ThrowIfNotAuthorized(Permissions.ResearchFileAdd);
+            var newActivityInstance = _activityRepository.Add(instance);
+            _activityRepository.CommitTransaction();
+            return _activityRepository.GetById(newActivityInstance.ActivityInstanceId);
```
FuriousLlama, line 76: we should not need to do a secondary retrieval after the commit.

backend/api/Services/ActivityService.cs
```diff
@@ -0,0 +1,122 @@
...
+            foreach (PimsActivityInstanceDocument activityInstanceDocument in activityDocuments)
+            {
+                _documentService.DeleteActivityDocumentAsync(activityInstanceDocument);
+            }
+
+            var notes = _noteService.GetNotes(Constants.NoteType.Activity, activityId);
+            foreach (PimsNote note in notes)
+            {
+                _noteService.DeleteNote(Constants.NoteType.Activity, note.Id);
+            }
```
FuriousLlama, line 107: Just to confirm. This is to delete the notes for the given activity right?

backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -41,13 +41,13 @@ public PimsActivityInstance GetById(long id)
             return this.Context.PimsActivityInstances
                 .Include(r => r.ActivityTemplate).ThenInclude(y => y.ActivityTemplateTypeCodeNavigation)
                 .AsNoTracking()
-                .FirstOrDefault(x => x.ActivityInstanceId == id);
+                .FirstOrDefault(x => x.ActivityInstanceId == id) ?? throw new KeyNotFoundException();
```
FuriousLlama, line 44: Returning an empty object if not found sounds reasonable. The consumer could check for null or if the returned one has an id. As it is, it forces the consumer to handle a throw.

frontend/src/App.tsx
```diff
@@ -60,6 +61,7 @@ const App = () => {
               draggable
               pauseOnHover
             />
+            <ModalContainer />
```
FuriousLlama, line 64: is this part of the story or a tad of tech-debt

FuriousLlama: What would happen if there are multiple modals being stacked? I think the Lease->Deposit->Add Contact workflow will do that

FuriousLlama:
There are a ton of changes to the front end that do not seem to be related to the story. If that is related to the modal enhancement, can that be moved to a separate pr?

backend/api/Areas/Activities/Controllers/ActivityController.cs
devinleighsmith: ugh, will fix.

backend/api/Services/ActivityService.cs
devinleighsmith: This was from when I was using placeholder permissions, will remove

backend/api/Services/ActivityService.cs
devinleighsmith: yes, when an activity is deleted all corresponding notes should be deleted.

backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: I don't disagree, but this is not the pattern that we have been using. Also, I already had to fix a related issue in the research file where an empty response was being returned but the consumer forgot to check for it and just assumed the response was valid. I'd rather show this as an error, because a consumer should never request an id that doesn't exist, that indicates something went wrong. These endpoints aren't designed to be for searches.

frontend/src/App.tsx
devinleighsmith: it is not possible to stack multiple modals, they would have to just be floating popup windows which is not what this container was built for. If you want multiple stacking popups you'll need to do things the old way.

backend/dal/Repositories/ActivityRepository.cs
FuriousLlama: I see. Lets have a chat on friday to figure out which pattern we want to follow

devinleighsmith:
@FuriousLlama fair feedback, I'll close this and create a new PR that is just the modal change to start, and then another for 3910

backend/api/Services/ActivityService.cs
devinleighsmith: If you don't do this the concurrency id won't get updated and subsequent requests will fail.

PR #1915, opened 2022-08-18
Mapview sidebar router

FuriousLlama:
Refactored approute to take custom renders. 
Refactored the mapview to use a router instead of parsing the url


frontend/src/AppRouter.tsx
```diff
@@ -243,17 +243,17 @@ const AppRouter: React.FC = () => {
...
-          component={TestFileManagement}
+          customComponent={TestFileManagement}
           layout={AuthLayout}
         />
         <AppRoute
           exact
           path="/test/notes"
           title={getTitle('Test Notes')}
-          component={TestNotes}
+          customComponent={TestNotes}
```
devinleighsmith, line 253: is there a clear reason why we should rename this from component -> customcomponent? I'm not a big fan of renames like this because I feel like it's just replacing dev A's personal preference with dev B's personal preference and we end up with bloated PRs.

frontend/src/features/leases/detail/LeaseRouter.tsx
```diff
@@ -14,7 +14,7 @@ export const LeaseRouter: React.FunctionComponent<ILeaseRouterProps> = React.mem
       {Array.from(leasePages.entries()).map(([pageName, page]) => (
         <AppRoute
           path={`${path}/${pageName}`}
-          component={page.component ?? React.Fragment}
+          customComponent={page.component}
```
devinleighsmith, line 17: different change here then what you did for the ContactRouter. It seems like passing null/undefined for customcomponent isn't valid so perhaps this is the correct version of this change?

frontend/src/features/properties/map/MapRouter.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <AcquisitionContainer acquisitionFileId={Number(match.params.id)} onClose={onClose} />
+        )}
+        claim={Claims.ACQUISITION_VIEW}
+        exact
+        key={'Aquisition'}
+        title={'Acquisition File'}
+      />
+      <AppRoute
+        path="*"
+        customRender={({ match }) => <>Well this sucks{match.path}</>}
```
devinleighsmith, line 108: We do have a 404 page, lets maybe use that instead as amusing as that is.

frontend/src/features/properties/map/MapRouter.tsx
```diff
@@ -0,0 +1,117 @@
...
+        exact
+        key={'Aquisition'}
```
devinleighsmith, line 110: the key and title for this page don't seem to make sense.

frontend/src/features/properties/map/MapRouter.tsx
```diff
@@ -0,0 +1,117 @@
...
+        path={`/mapview/sidebar/research/new`}
+        customRender={() => <AddResearchContainer onClose={onClose} />}
+        claim={Claims.RESEARCH_ADD}
+        exact
+        key={'NewResearch'}
+        title={'Create Research File'}
+      />
+      <AppRoute
+        path={`/mapview/sidebar/research/:researchId`}
+        customRender={({ match }) => (
```
devinleighsmith, line 55: ok, looking at this now I think previously we would have handled this with a <Page><Container /></Page> strictly speaking I don't think I care either way on this one, it did always seem like <Page> components never had much to do.

frontend/src/utils/PrivateRoute.tsx
```diff
@@ -1,52 +1,68 @@
...
-            return (
+  let { ...rest } = props;
+
+  if (!!keycloak.obj?.authenticated) {
+    if (
+      (!rest.role && !rest.claim) ||
+      keycloak.hasRole(rest.role) ||
+      keycloak.hasClaim(rest.claim)
+    ) {
+      if (isRenderRoute(props)) {
```
devinleighsmith, line 39: I'd lean towards throwing an error if both customComponent and customRender are defined, but otherwise is ok.

frontend/src/utils/PrivateRoute.tsx
```diff
@@ -1,52 +1,68 @@
...
-        }
-      }}
-    />
-  );
+            )}
+          />
+        );
+      }
+    } else {
+      return <Redirect to={{ pathname: '/forbidden', state: { referer: location } }} />;
```
devinleighsmith, line 55: hmm, I'm worried that this may not always be equivalent. routeProps.location = the location object from the component. location = the most up to date location object.

frontend/src/utils/PrivateRoute.tsx
```diff
@@ -1,52 +1,68 @@
...
+    }
+  } else {
+    if (location.pathname !== '/login') {
+      const redirectTo = encodeURI(`${location.pathname}${location.search}`);
+      return <Redirect to={`/login?redirect=${redirectTo}`} />;
+    } else {
+      console.error('Navigation: location was expected to be login');
+      return <></>;
```
devinleighsmith, line 63: maybe the error page instead?

frontend/src/utils/PrivateRoute.tsx
```diff
@@ -1,52 +1,68 @@
...
+        return <Route {...rest} render={props.customRender} />;
+      } else {
+        const { customComponent: Component, layout: Layout } = props;
+        return (
+          <Route
+            {...rest}
+            render={props => (
               <Layout>
-                <Component {...routeProps} {...rest.componentProps} />
+                <Component {...props} />
```
devinleighsmith, line 48: hmm, we used to pass the route props as well. I wonder if this will break anything?

devinleighsmith (approved):

frontend/src/AppRouter.tsx
FuriousLlama: The reason is that the underlying Router that this is wrapping has a component parameter that was being hidden by this name. This was causing some confusion on type checking between the passed component and the component that Router has

frontend/src/features/leases/detail/LeaseRouter.tsx
FuriousLlama: I think the React.Fragment part was just being used in case the page.component was not returning a valid component. My guess is that we used that during development when some pages were not returning a valid component. Now all pages should return a valid one, so that extra code is no longer needed,

frontend/src/features/properties/map/MapRouter.tsx
FuriousLlama: This is only here for development. Will remove it once I checked the routes work as expected

frontend/src/features/properties/map/MapRouter.tsx
FuriousLlama: yep, just a placeholder

frontend/src/features/properties/map/MapRouter.tsx
FuriousLlama: I thought about that too, but given this is not really a page and more like a widget I decided to not go that route. But it would be totally valid for a page

frontend/src/utils/PrivateRoute.tsx
FuriousLlama: Uhm, I dont know either. will maintain the same for now

frontend/src/utils/PrivateRoute.tsx
FuriousLlama: We could revisit it. As it is it maintains the same functionality as before. i.e. not showing anything

PR #1916, opened 2022-08-18
global modal component

devinleighsmith:
Added a global modal to handle some of our simple modal use cases (cancel logic for example).

### DIFF
--- a/frontend/src/contexts/modalContext.tsx
+++ b/frontend/src/contexts/modalContext.tsx
@@ -18,7 +18,9 @@ export const ModalContext = React.createContext<IModalContext>({
   },
 });
 
-export const ModalContextProvider = (props: IModalContext & { children: React.ReactChildren }) => {
+export const ModalContextProvider = (
+  props: IModalContext & { children: React.ReactChild | React.ReactChild[] | React.ReactNode },
+) => {
   const [modalProps, setModalProps] = useState<ModalProps | undefined>(undefined);
   const displayFunction = React.useCallback(
     (displayModal: boolean) => {
--- a/frontend/src/contexts/modalContext.tsx
+++ b/frontend/src/contexts/modalContext.tsx
@@ -1,5 +1,4 @@
 import { ModalProps } from 'components/common/GenericModal';
-import noop from 'lodash/noop';
 import * as React from 'react';
 import { useState } from 'react';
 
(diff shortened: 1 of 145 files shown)


frontend/src/contexts/modalContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+        display: updatedModalProps?.display ?? modalProps?.display ?? false,
+        setDisplay: displayFunction,
+      }),
+    [displayFunction, modalProps],
+  );
+
+  return (
+    <ModalContext.Provider
+      value={{
+        // if user info is not available when authenticated, then the auth state is not ready
```
FuriousLlama, line 40: I don't understand this comment. Are the modalProps related to the user info?

frontend/src/contexts/modalContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+export interface IModalContext {
+  modalProps?: ModalProps;
+  setModalProps: (modalProps?: ModalProps) => void;
+  setDisplayModal: (display: boolean) => void;
+}
+
+export const ModalContext = React.createContext<IModalContext>({
+  modalProps: undefined,
+  setModalProps: noop,
+  setDisplayModal: noop,
```
FuriousLlama, line 15: If is not set and gets called, should it throw an error or log something instead?

frontend/src/contexts/modalContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+});
+
+export const ModalContextProvider = (props: IModalContext & { children: any }) => {
```
FuriousLlama, line 18: Thoughts on limiting the children to be a JSX or string instead of any?

frontend/src/hooks/useModalContext.ts
```diff
@@ -0,0 +1,30 @@
...
+import { ModalProps } from './../components/common/GenericModal';
+import { ModalContext } from './../contexts/modalContext';
+export const useModalContext = (newModalProps?: ModalProps) => {
+  const { modalProps, setModalProps, setDisplayModal } = useContext(ModalContext);
+  useEffect(() => {
+    if (newModalProps !== undefined) {
+      setModalProps({ ...newModalProps });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
FuriousLlama, line 13: Still learning react, so this might be dumb. Is the newModalProps and setModalProps not part of the dependencies of this useEffect?

FuriousLlama (approved):

frontend/src/contexts/modalContext.tsx
devinleighsmith: copy pasta, will remove

frontend/src/contexts/modalContext.tsx
devinleighsmith: Yes, there is actually a reactChild type that I can use here, more copy pasta

frontend/src/contexts/modalContext.tsx
devinleighsmith: sure, couldn't hurt. They are always being set currently but would be nice to throw an error if someone messes this up.

frontend/src/hooks/useModalContext.ts
devinleighsmith: yes, maybe I need a comment here. The intent here is to force the setModalProps logic to only run once during component initialization.

PR #1934, opened 2022-08-19
Features/psp 3908

FuriousLlama:
Updated Research File header font size & spacing
image


devinleighsmith (changes requested):
The ac for this story states that we need to handle long file names gracefully but we aren't at the moment:

image

devinleighsmith:
likely the css you are using for the name only breaks lines if spaces are present.

devinleighsmith:
Dave and I talked about this, the use case I show above is not valid, passing this.

devinleighsmith (approved):

PR #1941, opened 2022-08-20
HOTFIX: Psp 4198

devinleighsmith:
When attempting to sync keycloak users, ignore roles failures for roles the exist in KC but not keycloak.

### DIFF
--- a/backend/api/Pims.Api.csproj
+++ b/backend/api/Pims.Api.csproj
@@ -2,8 +2,8 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <UserSecretsId>0ef6255f-9ea0-49ec-8c65-c172304b4926</UserSecretsId>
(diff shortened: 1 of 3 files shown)


backend/api/Pims.Api.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <UserSecretsId>0ef6255f-9ea0-49ec-8c65-c172304b4926</UserSecretsId>
-    <Version>2.2.0-33.24</Version>
+    <Version>2.2.0-33.25</Version>
```
asanchezr, line 5: Don't want to be stingy with version numbers but wouldn't a hotfix trigger a different change in the version number?

`2.2.1-33.25` vs `2.2.0-33.25`

asanchezr (approved):

PR #1942, opened 2022-08-23
PSP-3541 Logging reference implementation

asanchezr:


backend/api/Areas/Acquisition/Controllers/AcquisitionFileController.cs
```diff
@@ -53,6 +59,16 @@ public AcquisitionFileController(IAcquisitionFileService acquisitionService, IMa
         [SwaggerOperation(Tags = new[] { "acquisitionfile" })]
         public IActionResult GetAcquisitionFile(long id)
         {
+            // RECOMMENDED - Add valuable metadata to logs
+            _logger.LogInformation("Request received by Controller: {Controller}, Action: {ControllerAction}, User: {User}, DateTime: {DateTime}",
```
devinleighsmith, line 63: Thanks for this, do both of these seem like info level logs to you or should either of these be a debug maybe?

devinleighsmith:
Hmm, I only see logging added to the controller level, are we happy with the logging we currently have at the service and repository level? I know that we have some logging at the service level but almost none at the repository level. Given that some services call other services, I'd think we'd at least want some additional logging there.

devinleighsmith:
Ah, I see the confusion with the original story definition. My thought was to add logging for the entire lifecycle of those two endpoints, not just the controller. If you don't think you have time for that this sprint feel free to log a follow up issue.

devinleighsmith (approved):

asanchezr:
I debated whether to include logging in the repositories as EF already logs the SQL statements. I can add a log like:

```cs
# repo 
public GetBlahById()
{
  _logger.Log('Getting Blah by ID')
}
```

But seemed a little repetitive - I do see a point where a service calls multiple repositories within a single service calls. Toughts?

devinleighsmith:
Yeah, I like the service idea, especially if the service is doing something non-trivial. I think I'm ok with skipping the repo logs, although I believe that currently that sql logging only occurs when the app is in debug mode (so at this point local dev only).

backend/api/Areas/Acquisition/Controllers/AcquisitionFileController.cs
asanchezr: As per MS recommended levels [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-6.0#log-level) I believe Information would be a better fit than Debug. Also Debug is ignored on live environments and the above statement gives us the ability to answer questions like: "I want to know which actions a particular user XYZ did on this day that resulted in an error"

image

PR #1955, opened 2022-08-25
scaffold for 35.01

devinleighsmith:


Makefile
```diff
@@ -274,7 +274,7 @@ db-upgrade: ## Upgrade an existing database to the TARGET_VERSION (if passed) or
 
 db-scaffold: ## Requires local install of sqlcmd
 	@echo "$(P) regenerate ef core entities from database"
-	@cd backend/entities; eval $(grep -v '^#' .env | xargs) dotnet ef dbcontext scaffold Name=PIMS Microsoft.EntityFrameworkCore.SqlServer -o ../entities/ef --schema dbo --context PimsBaseContext --context-namespace Pims.Dal --context-dir . --no-onconfiguring --namespace Pims.Dal.Entities --data-annotations -v -f
```
devinleighsmith, line 277: unless you have an env var or secret kicking around named PIMS, this command won't work. This update allows the scaffold to pick up the connection string from your existing .env file

asanchezr (approved):

PR #1958, opened 2022-08-25
PSP-3362 Acquisition file header

asanchezr:
acq-file-header


devinleighsmith (approved):
Acquisition Container tests coming later perhaps?

asanchezr:
I thought about AcquisitionContainer but tests might be more relevant when proper Details form gets implemented. Not sure if we are going with the new pattern there or keeping old pattern of in-lining the details form...

devinleighsmith:
thats up to you really, I like the new pattern so have started using it in my stories.

PR #1961, opened 2022-08-29
PSP-4067:Tests for research file backend

Maheshbabu28:


devinleighsmith (changes requested):
For the repo, I'd like to see tests for the sub entities as well. In this case the research file purpose list. We should test that the add/remove/update functionality works as expected.

devinleighsmith (approved):

PR #1968, opened 2022-08-29
PSP-4182:The Property Info left slider does not reliably display the …

Maheshbabu28:
…PIMS Files tab

### DIFF
--- a/frontend/src/features/properties/map/MapRouter.tsx
+++ b/frontend/src/features/properties/map/MapRouter.tsx
@@ -91,9 +91,14 @@ export const MapRouter: React.FunctionComponent<IMapRouterProps> = React.memo(pr
         title={'Property Information'}
       />
       <AppRoute
-        path={`/mapview/sidebar/non-inventory-property/:pid`}
+        path={`/mapview/sidebar/property/:id/non-inventory-property/:pid`}
         customRender={({ match }) => (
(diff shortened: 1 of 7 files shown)


frontend/src/features/properties/map/MapRouter.tsx
```diff
@@ -91,9 +91,14 @@ export const MapRouter: React.FunctionComponent<IMapRouterProps> = React.memo(pr
         title={'Property Information'}
       />
       <AppRoute
-        path={`/mapview/sidebar/non-inventory-property/:pid`}
+        path={`/mapview/sidebar/property/:id/non-inventory-property/:pid`}
```
devinleighsmith, line 94: This doesn't make sense unfortunately.

non-inventory-property means that the property is not in our database. As a result, the property:id must always be null. If we are being navigated to this url when a property is in our inventory something is wrong. This route should only be used when we are viewing a property that is not in our inventory on the map.

devinleighsmith: What should be happening here is the frontend should be smart enough to either route to the regular property endpoint: (ie. mapview/property/24324) when the property is in our inventory, and to /mapview/sidebar/non-inventory-property/:pid when the property is not in our inventory.

If it is in our inventory we should have an id and a marker on the map.

devinleighsmith (changes requested):
see comments.

devinleighsmith:
Closing as requires more discussion with BA group

PR #1971, opened 2022-08-30
correct issue with realm admin group being removed from kc.

devinleighsmith:
Finally tracked this down, this is why we need to get a better handle on our to-dos


devinleighsmith:
For reference, also confirmed that the kc sync will never remove any groups from existing users. This is also why previous attempts to fix this in the kc sync failed, and why only some users were affected. Only users who were updated in the admin console would have their realm-admin removed.

asanchezr (approved):
Thanks for fixing this annoying bug!

PR #1993, opened 2022-09-01
psp-4174

devinleighsmith:
stop mocking the activity list view.


frontend/src/features/properties/map/research/ResearchTabsContainer.tsx
```diff
@@ -34,7 +35,9 @@ export const ResearchTabsContainer: React.FunctionComponent<IResearchTabsContain
 
   if (researchFile?.id && hasClaim(Claims.ACTIVITY_VIEW)) {
     tabViews.push({
-      content: <ActivityListView fileId={researchFile.id}></ActivityListView>,
+      content: (
+        <ActivityListView fileId={researchFile.id} fileType={FileTypes.Research}></ActivityListView>
```
devinleighsmith, line 39: activity list view supports being attached to different types of files.

backend/api/Areas/Activities/Controllers/ActivityController.cs
```diff
@@ -46,17 +49,24 @@ public ActivityController(IActivityService activityService, IMapper mapper)
...
         {
-            var activityInstance = _mapper.Map<PimsActivityInstance>(activityModel);
-            var createdActivity = _activityService.Add(activityInstance);
+            var activityInstance = _mapper.Map<PimsActivityInstance>(activityFileModel.Activity);
+            PimsActivityInstance createdActivity;
+            createdActivity = fileType switch
+            {
+                FileType.Research => _activityService.AddResearchActivity(activityInstance, activityFileModel.FileId),
+                FileType.Acquisition => _activityService.AddAcquisitionActivity(activityInstance, activityFileModel.FileId),
+                _ => throw new BadRequestException("File type must be a known type"),
```
FuriousLlama, line 68: fancy syntax 🥇

backend/api/Areas/Activities/Controllers/ActivityController.cs
```diff
@@ -76,6 +86,29 @@ public IActionResult GetActivityById(long activityId)
             return new JsonResult(_mapper.Map<ActivityInstanceModel>(activity));
         }
 
+        /// <summary>
+        /// Retrieves the activity with the specified id.
+        /// </summary>
+        /// <param name="fileType">The type of file the fileId belongs to.</param>
+        /// <param name="fileId">the id of the file.</param>
+        /// <returns></returns>
+        [HttpGet("{fileType}/file/{fileId:long}")]
```
FuriousLlama, line 95: This looks like it should retrieve the activities for the specified file id. Is that the intent?

FuriousLlama, line 95: nit: the path looks a tad redundant 
```
[GET] activity/research/file/1
```
Could be instead
```
[GET] activity/researchFile/1
```
or just 
```
[GET] activity/research/1
```

FuriousLlama: On that same vein, would it not be better to have the file controller handle that?
```
[GET] {fileType}/{fileId}/activities/

File: backend/entities/Partials/ActivityTemplateType.cs [hunk 4]
```diff
@@ -3,14 +3,14 @@
 namespace Pims.Dal.Entities
 {
     /// <summary>
-    /// PimsVolumetricType class, provides an entity for the datamodel to manage a list of volumetric types.
```
FuriousLlama, line 6: +1

backend/api/Areas/Activities/Controllers/ActivityController.cs
devinleighsmith: I think our pattern has been, get {resource} by x where {resource} generally determines what controller the function should belong to (in this case, the activities controller).

devinleighsmith: I think this also helps the file controller get out of control with its responsibilities, otherwise it would need to handle all documents, notes, activities, and other child relationships.

devinleighsmith:
@FuriousLlama did you have any other comments or is this one good?

FuriousLlama (approved):

PR #2032, opened 2022-09-06
psp-4182 allow clicking on a properties boundary in addition to its m…

devinleighsmith:
…arker.

### DIFF
--- a/frontend/src/features/properties/map/MapView.tsx
+++ b/frontend/src/features/properties/map/MapView.tsx
@@ -43,8 +43,10 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
     history.push(`/mapview/sidebar/property/${property.id}?pid=${property.pid}`);
   };
 
-  const onPropertyViewClicked = (pid?: string | null) => {
-    if (pid !== undefined && pid !== null) {
+  const onPropertyViewClicked = (pid?: string | null, id?: number) => {
+    if (id !== undefined) {
+      history.push(`/mapview/sidebar/property/${id}?pid=${pid}`);
+    } else if (pid !== undefined && pid !== null) {
       const parsedPid = pidParser(pid);
       history.push(`/mapview/sidebar/non-inventory-property/${parsedPid}`);
     } else {
--- a/geoserver/geoserver_data/psp/mssql/PIMS_PROPERTY_BOUNDARY_VW/featuretype.xml
+++ b/geoserver/geoserver_data/psp/mssql/PIMS_PROPERTY_BOUNDARY_VW/featuretype.xml
@@ -1,0 +1,64 @@
+<featureType>
+  <id>FeatureTypeInfoImpl--679c7b53:182fb30d43f:-7215</id>
+  <name>PIMS_PROPERTY_BOUNDARY_VW</name>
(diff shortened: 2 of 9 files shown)


frontend/src/features/properties/map/MapView.tsx
```diff
@@ -42,8 +42,10 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
     history.push(`/mapview/sidebar/property/${property.id}?pid=${property.pid}`);
   };
 
-  const onPropertyViewClicked = (pid?: string | null) => {
-    if (pid !== undefined && pid !== null) {
+  const onPropertyViewClicked = (pid?: string | null, id?: number) => {
+    if (id !== undefined) {
+      history.push(`/mapview/sidebar/property/${id}?pid=${pid}`);
```
FuriousLlama, line 47: pid is not guaranteed to be non null here

devinleighsmith: having a null pid works here, I've tested it. it just results in the ltsa screen not being shown.

geoserver/geoserver_data/psp/mssql/PIMS_PROPERTY_BOUNDARY_VW/featuretype.xml
```diff
@@ -0,0 +1,64 @@
...
+  </namespace>
+  <title>PIMS_PROPERTY_BOUNDARY_VW</title>
+  <keywords>
+    <string>features</string>
+    <string>PIMS_PROPERTY_BOUNDARY_VW</string>
+  </keywords>
+  <nativeCRS class="projected">PROJCS[&quot;NAD83 / BC Albers&quot;, 
+  GEOGCS[&quot;NAD83&quot;, 
+    DATUM[&quot;North American Datum 1983&quot;, 
+      SPHEROID[&quot;GRS 1980&quot;, 6378137.0, 298.257222101, AUTHORITY[&quot;EPSG&quot;,&quot;7019&quot;]],
```
asanchezr, line 16: Thanks for adding the new map layer to our local geoserver. Could you please also update **make env** to match these changes?

frontend/src/features/properties/map/MapView.tsx
devinleighsmith: the url ends up with pid=

geoserver/geoserver_data/psp/mssql/PIMS_PROPERTY_BOUNDARY_VW/featuretype.xml
devinleighsmith: @asanchezr my understanding of make .env is that is creates the datastore.xml files needed to connect to a database. In this case, the same datastore previously used for the location view can also be used for the boundary view (note the co-location of both views under the same parent folder).

FuriousLlama (approved):

geoserver/geoserver_data/psp/mssql/PIMS_PROPERTY_BOUNDARY_VW/featuretype.xml
asanchezr: Right. I missed that detail. Thanks for clarifying

asanchezr (approved):

PR #2035, opened 2022-09-07
Added new search filter and result column | psp-3904

FuriousLlama:
image


backend/api/Areas/Acquisition/Models/AcquisitionFilterModel.cs
```diff
@@ -10,7 +10,22 @@ public class AcquisitionFilterModel : PageFilter
         #region Properties
 
         /// <summary>
-        /// get/set - The status of the acquisition file,.
+        /// get/set - The pid identifier to search by.
+        /// </summary>
+        public string Pid { get; set; }
+
+        /// <summary>
+        /// get/set - The pin identifier to serach by.
```
devinleighsmith, line 18: nit: serach

backend/dal/Helpers/Extensions/AcquisitionExtensions.cs
```diff
@@ -40,6 +40,27 @@ private static IQueryable<Entities.PimsAcquisitionFile> GenerateCommonAcquisitio
         {
             filter.ThrowIfNull(nameof(filter));
 
+            if (!string.IsNullOrWhiteSpace(filter.Pid))
+            {
+                var pidValue = filter.Pid.Replace("-", string.Empty).Trim().TrimStart('0');
```
devinleighsmith, line 45: This would be in conflict with the pid search that we use on the mapview. Do the requirements mention anything about removing any leading zeroes?

frontend/src/features/acquisition/list/AcquisitionSearchResults/AquisitionProperties.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export interface IAcquisitionPropertiesProps {
+  aquisitionProperties?: Api_AcquisitionFileProperty[];
+  maxDisplayCount: number;
+}
+
+const AcquisitionProperties: React.FunctionComponent<IAcquisitionPropertiesProps> = props => {
+  const [isExpanded, setExpanded] = useState(false);
+
+  const aquisitionProperties = props.aquisitionProperties || [];
```
devinleighsmith, line 16: nit: aquisition -> acquisition

frontend/src/features/acquisition/list/AcquisitionSearchResults/AquisitionProperties.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Col, Row } from 'react-bootstrap';
+import styled from 'styled-components';
+import { formatApiAddress } from 'utils';
+
+export interface IAcquisitionPropertiesProps {
+  aquisitionProperties?: Api_AcquisitionFileProperty[];
+  maxDisplayCount: number;
+}
+
+const AcquisitionProperties: React.FunctionComponent<IAcquisitionPropertiesProps> = props => {
```
devinleighsmith, line 13: This seems like a good candidate for a common component given that LeaseProperties.tsx is nearly identical except for the property list (although both lists share pid, pin, address. What was your thinking on keeping the two components separate?

frontend/src/utils/propertyUtils.ts
```diff
@@ -65,6 +66,21 @@ export const formatAddress = (address?: IAddress) => {
...
+ * @returns Civic address string value.
+ */
+export const formatApiAddress = (address?: Api_Address) => {
+  const values = [
+    address?.streetAddress1 ?? '',
+    address?.streetAddress2 ?? '',
+    address?.streetAddress3 ?? '',
+    address?.province ?? '',
+  ];
+  return values.join(' ') + (address?.postal ? ', ' + (address?.postal ?? '') : '');
```
devinleighsmith, line 81: some redundancy in this ternary. if address?.postal is falsey the ternary will return '' so there should be no need for (address?.postal ?? '')

frontend/src/features/acquisition/list/AcquisitionSearchResults/AquisitionProperties.tsx
devinleighsmith: I'd be more ok with duplicating this component if we had test coverage on the original. Since we don't we are losing some test coverage % by duplicating here.

FuriousLlama: That is a good observation, I thought it too. For now, I believe it would be better to keep the duplicated component and in the future remove it if there are no additional changes needed.
The problem with refactoring now is that I am not certain that the business wants/will want the same information on both sides

devinleighsmith (approved):

backend/dal/Helpers/Extensions/AcquisitionExtensions.cs
FuriousLlama: This is for search on the DB where the values for both pin and pid are integers. Note that we do the same thing when searching leases by pid/pin

```
CREATE TABLE [dbo].[PIMS_PROPERTY](
...
[PID] [int] NULL,
[PIN] [int] NULL,
...
```

frontend/src/utils/propertyUtils.ts
FuriousLlama: True. We can revisit this when we standardize the address model given the exact same ternary is used on the display of that one as well.

PR #2040, opened 2022-09-07
Removed images and schema compares from script migrations

FuriousLlama:
Reduced DB migrations folder from ~400Mb to ~100Mb.

Note that the Latest migration still contains both images and a schema compare


devinleighsmith:
@FuriousLlama have you informed @dfilteau that we don't want the schema compares in the named IS* folders anymore?

devinleighsmith:
@FuriousLlama should we remove those files from history as well to reduce the size of the repository (in addition to the size of the cloned repo)?

.github/workflows/api-dotnetcore.yml
```diff
@@ -7,7 +7,24 @@ on:
     branches: [master, dev, test]
 
 jobs:
+  # JOB to run change detection
```
devinleighsmith, line 10: This is nice, thanks for this.

.github/workflows/api-dotnetcore.yml
```diff
@@ -7,7 +7,24 @@ on:
...
+  check-changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      backend: ${{ steps.filter.outputs.backend }}
+    steps:
+      # For pull requests it's not necessary to checkout the code
```
devinleighsmith, line 17: I don't see that this is only affecting pull requests, am I missing something?

.github/workflows/api-dotnetcore.yml
devinleighsmith: Could we do something like this for Openshift as well do you think? Ie. skipping builds if not required or only checking out the parts of the repo that we need.

.github/workflows/api-dotnetcore.yml
FuriousLlama: This was a comment on the sample for that filter. I believe it will work properly on pushes too. On that matter, do you know when that would be the case? AFAIK changes to dev, test or master can only be done though PRs

FuriousLlama:
> @FuriousLlama have you informed @dfilteau that we don't want the schema compares in the named IS* folders anymore?
Not yet, will chat with him later today once this is approved.

> @FuriousLlama should we remove those files from history as well to reduce the size of the repository (in addition to the size of the cloned repo)?
Not sure, if we are not having problems with github running out of space we could just leave it as is. The future commits will be cleaner

.github/workflows/api-dotnetcore.yml
devinleighsmith: Yeah, the comment confused me because it is possible to restrict what workflows trigger an action. So you could restrict this to run on PRs but not pulls for example. I don't see that happening here however.

devinleighsmith (approved):

devinleighsmith:
Approved, thanks. I think we can leave off from removing history for now, revisit if our build times become more of a concern.

.github/workflows/api-dotnetcore.yml
FuriousLlama: I believe so. We can chat some more about to improve it

PR #2054, opened 2022-09-09
psp-3827 ensure that adding a property to a file from the map populat…

devinleighsmith:
…es all property attributes.

### DIFF
--- a/frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
+++ b/frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
@@ -3,7 +3,7 @@ import { pidFormatter } from 'utils';
 export const MUNICIPALITY_LAYER_URL =
   'https://openmaps.gov.bc.ca/geo/pub/WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP';
 export const PARCELS_LAYER_URL =
-  'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW/wfs?service=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW';
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW/wfs?service=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW';
 export const ELECTORAL_LAYER_URL =
   'https://openmaps.gov.bc.ca/geo/pub/WHSE_ADMIN_BOUNDARIES.EBC_ELECTORAL_DISTS_BS10_SVW/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_ADMIN_BOUNDARIES.EBC_ELECTORAL_DISTS_BS10_SVW';
 export const REGIONAL_LAYER_URL =
--- a/frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
+++ b/frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
@@ -27,6 +27,7 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> = 
   const { selectedResearchFeature, setSelectedResearchFeature } = React.useContext(
     SelectedPropertyContext,
   );
+  console.log(selectedResearchFeature);
 
   const initialForm = useMemo(() => {
     const researchForm = new AcquisitionForm();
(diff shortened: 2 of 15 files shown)


frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -33,7 +62,7 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> =
     history.replace(`/mapview/sidebar/acquisition/${acqFile.id}`);
   };
 
-  const helper = useAddAcquisitionFormManagement({ onSuccess });
+  const helper = useAddAcquisitionFormManagement({ onSuccess, initialForm });
```
FuriousLlama, line 65: this name does not give much information. Can the logic from the hook be put here or is reused somewhere else?

frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -21,6 +24,32 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> =
   const formikRef = useRef<FormikProps<AcquisitionForm>>(null);
 
   const close = useCallback(() => onClose && onClose(), [onClose]);
+  const { selectedResearchFeature, setSelectedResearchFeature } = React.useContext(
+    SelectedPropertyContext,
+  );
+  console.log(selectedResearchFeature);
```
FuriousLlama, line 30: loggerino

frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
```diff
@@ -3,7 +3,7 @@ import { pidFormatter } from 'utils';
 export const MUNICIPALITY_LAYER_URL =
   'https://openmaps.gov.bc.ca/geo/pub/WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP';
 export const PARCELS_LAYER_URL =
-  'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW/wfs?service=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW';
+  'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW/wfs?service=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW';
```
FuriousLlama, line 6: is this change related to the fix?

frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
asanchezr: This comes down to developer preference. I like to "hide" business logic in forms in single-use hooks. Looks cleaner to me rather than inlining 20 lines of code but I'm open to hear if others dislike that approach and we can decide as a team what we prefer. These hooks are not meant to be reused but rather to hide complexity from a rather large form. I try to think of them as "helpers" or static classes in C#

frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
asanchezr: This might also be copy pasta from research file as this form is all about Acquisition File

frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
asanchezr: I know we wanted to use the recently published FA parcel layer for all layer queries but would think that is a new feature request, not a bug? Unless relevant to this PR

frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -21,6 +24,32 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> =
...
+
+  const initialForm = useMemo(() => {
+    const researchForm = new AcquisitionForm();
```
asanchezr, line 33: seems copy pasta. please rename to acquisitionForm

frontend/src/features/properties/selector/components/MapClickMonitor.tsx
```diff
@@ -46,7 +46,7 @@ export const mapFeatureToProperty = (
     longitude: selectedFeature?.properties?.CLICK_LAT_LNG?.lng ?? latLng.lng ?? undefined,
     planNumber: selectedFeature?.properties?.PLAN_NUMBER ?? undefined,
     address: 'placeholder', //todo: need alternate source for this
-    legalDescription: 'placeholder', //todo: need access to fully attributed parcelmap bc layer,
+    legalDescription: selectedFeature?.properties?.LEGAL_DESCRIPTION,
```
asanchezr, line 49: Ah I see why you changed the parcel map URL earlier. We do need FA parcel map to get legal description

asanchezr: Probably relevant to alert testers of the change in map layer so they run regression against everything that uses parcel map layer queries...

frontend/src/features/properties/selector/map/PropertyMapSelectorFormView.tsx
```diff
@@ -5,24 +5,33 @@ import {
...
-    undefined,
+    initialSelectedProperty,
   );
 
+  useEffect(() => {
+    if (initialSelectedProperty !== undefined) {
+      setSelectedProperty(initialSelectedProperty);
+    }
+  }, [initialSelectedProperty]);
+
```
asanchezr, line 34: question - why do we need this effect here? Wouldn't **selectedProperty** always take the value of **prop.initialSelectedProperty** ?

asanchezr (approved):

frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
devinleighsmith: in order to populate the data from the map, we need to use the FA layer instead of the parcel layer. They both have similar performance and are publicly available so we don't lose any information by making this change.

PR #2063, opened 2022-09-12
psp-3846 allow adding/editing/deleting file properties to activity.

devinleighsmith:


backend/dal/Helpers/Extensions/EntityExtensions.cs
```diff
@@ -136,8 +136,11 @@ public static void SetOriginalConcurrencyControlNumber<T>(this T source, DbConte
                 }
                 else
                 {
-                    context.Entry(oldItem).CurrentValues.SetValues(item);
-                    dbItemsMap.Remove(item.Id);
+                    if (updateChildValues)
```
devinleighsmith, line 139: provide more control over the updateChild logic, to only allow updating child associations and not the contents of children.

backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -87,22 +90,26 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
         public PimsActivityInstance Update(PimsActivityInstance instance)
         {
             instance.ThrowIfNull(nameof(instance));
-            GetById(instance.ActivityInstanceId);
-            if (instance.ActivityTemplate != null)
```
devinleighsmith, line 91: rather then individually blacklisting all of the related properties that shouldn't be updated (which is error prone as it is easy to forget to add a new property here). Decided to change this to a whitelist model instead, where by default only the base entity is updated.

frontend/src/components/Table/Table.tsx
```diff
@@ -204,35 +202,24 @@ const IndeterminateCheckbox = React.forwardRef(
       rest.onChange && !allDataRef && rest.onChange(e);
       const currentSelected = selectedRef?.current ? [...selectedRef?.current] : [];
       if (isHeaderCheck) {
-        setSelected([]);
-      } else {
-        if (isSingleSelect === true) {
-          currentSelected.splice(0, currentSelected.length);
-          currentSelected.push(row.original);
+        if (e.target.checked && !indeterminate) {
```
devinleighsmith, line 205: I was able to greatly simplify this logic. Need to add unit tests here as this logic has been unstable in the past.

frontend/src/components/common/GenericModal.tsx
```diff
@@ -189,6 +190,10 @@ const StyledModal = styled(Modal)`
     .close {
       color: white;
       opacity: 1;
+      font-size: 3rem;
```
devinleighsmith, line 193: corrections to the "x" icon on the modal.

backend/api/Constants/FileType.cs
```diff
@@ -10,5 +10,7 @@ public enum FileType
         Acquisition,
         [EnumMember(Value = "research")]
         Research,
+        [EnumMember(Value = "unknown")]
+        Unknown,
```
FuriousLlama, line 14: I am guessing this is not really an unknown file type but a file type of type "unknown". If that's the case please add some comments so is not as confussing

backend/api/Models/Concepts/ActivityInstance/PimsActInstPropFile.cs
```diff
@@ -0,0 +1,11 @@
+namespace Pims.Api.Models.Concepts
+{
+    public class PimsActInstPropFile : BaseAppModel
```
FuriousLlama, line 3: PimsActInstPropFile reads like an Entity object. Why not the more readable version? Same for the map

backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -87,22 +90,26 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
...
-            {
-                Context.Entry(instance.ActivityTemplate).State = EntityState.Unchanged;
-            }
-            foreach (PimsActivityInstanceDocument activityInstanceDocument in instance.PimsActivityInstanceDocuments)
-            {
-                Context.Entry(activityInstanceDocument).State = EntityState.Unchanged;
-            }
+            var currentActivity = this.Context.PimsActivityInstances
+                .FirstOrDefault(x => x.ActivityInstanceId == instance.Id) ?? throw new KeyNotFoundException();
+            this.Context.Entry(currentActivity).CurrentValues.SetValues(instance);
```
FuriousLlama, line 95: it is more readable, but I guess it does not update the related properties if we would want that.

backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -135,6 +144,16 @@ public bool Delete(long activityId)
                 this.Context.PimsActivityInstanceNotes.Remove(activityNote);
             }
 
+            foreach (var propAcqFile in instance.PimsActInstPropAcqFiles)
```
FuriousLlama, line 147: nit, no need for tiny names

backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -135,6 +144,16 @@ public bool Delete(long activityId)
...
+            {
+                this.Context.PimsActInstPropAcqFiles.Remove(propAcqFile);
+            }
+
+            foreach (var propRsrchFile in instance.PimsActInstPropRsrchFiles)
```
FuriousLlama, line 152: Not necessarily applicable to the links you changed, but would document, leave orphaned documents if that's the only one they are linked too? maybe the same with notes

frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
```diff
@@ -19,8 +17,11 @@ export interface IDocumentUploadContainerProps {
 
 export const DocumentUploadContainer: React.FunctionComponent<IDocumentUploadContainerProps> = props => {
   const isMounted = useIsMounted();
-  const { getDocumentTypes, getDocumentTypeMetadata } = useApiDocuments();
-  const { retrieveDocumentMetadataLoading } = useDocumentProvider();
+  const {
```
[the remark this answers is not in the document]
FuriousLlama: that would be great. This changed quite a bit

FuriousLlama (approved):

backend/api/Constants/FileType.cs
devinleighsmith: Yes, fair, will do.

backend/api/Models/Concepts/ActivityInstance/PimsActInstPropFile.cs
devinleighsmith: Sure, I can remove the Pims bit, add the Model

backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: yes, that would need to be done explicitly (perhaps more akin to what I'm doing with property updates here).

backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: Sure, I was following the naming convention of the entity but I don't mind lengthening these.

backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: Need to think about this one for a bit.

devinleighsmith: I think I'm ok with this. the service enforces the business logic to ensure that any related entities (documents, notes) are deleted/disabled when an activity is deleted, whereas this repository is just responsible for making sure that an activity and all of its fk entities are removed properly. The repository shouldn't need to worry about other entities (documents, notes) other then relationships that cause the delete to fail.

frontend/src/components/common/GenericModal.tsx
asanchezr: yeah - that icon always seemed a bit small. thanks for fixing

PR #2064, opened 2022-09-12
Fixes Document Metadata | psp-4377 & psp-4378 & psp-4280

FuriousLlama:
image

### DIFF
--- a/backend/api/Services/DocumentService.cs
+++ b/backend/api/Services/DocumentService.cs
@@ -94,18 +94,23 @@ namespace Pims.Api.Services
             this.User.ThrowIfNotAuthorized(Permissions.DocumentAdd);
 
             ExternalResult<DocumentDetail> externalResult = await UploadDocumentAsync(uploadRequest.DocumentTypeMayanId, uploadRequest.File);
-            DocumentUploadResponse response = new DocumentUploadResponse() { ExternalResult = externalResult };
+            DocumentUploadResponse response = new DocumentUploadResponse()
+            {
+                DocumentExternalResult = externalResult,
+                MetadataExternalResult = new List<ExternalResult<DocumentMetadata>>(),
+            };
+
             if (externalResult.Status == ExternalResultStatus.Success)
             {
+
                 var externalDocument = externalResult.Payload;
 
-                // Save metadata of document
-                IList<Task<ExternalResult<DocumentMetadata>>> metadataTasks = new List<Task<ExternalResult<DocumentMetadata>>>();
-                foreach (var metadata in uploadRequest.DocumentMetadata)
+                // Create metadata of document
+                if (uploadRequest.DocumentMetadata != null && uploadRequest.DocumentMetadata.Count > 0)
                 {
-                    metadataTasks.Add(documentStorageRepository.CreateDocumentMetadataAsync(externalDocument.Id, metadata.MetadataTypeId, metadata.Value));
+                    response.MetadataExternalResult = await CreateMetadata(uploadRequest.DocumentTypeMayanId, uploadRequest.DocumentMetadata);
                 }
-                await Task.WhenAll(metadataTasks.ToArray());
+
                 // Create the pims document
                 PimsDocument newPimsDocument = new PimsDocument()
                 {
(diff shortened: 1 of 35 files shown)


backend/api/Services/DocumentService.cs
```diff
@@ -129,29 +134,88 @@ public async Task<DocumentUploadResponse> UploadActivityDocumentAsync(long activ
...
-            if (existingDocument is not null)
+            DocumentUpdateResponse response = new DocumentUpdateResponse()
             {
-                existingDocument.DocumentStatusTypeCode = updateRequest.DocumentStatusCode;
-                documentRepository.Update(existingDocument);
+                MetadataExternalResult = new List<ExternalResult<DocumentMetadata>>(),
+            };
+
+
+            var metadataUpdateSucessfull = false;
```
devinleighsmith, line 157: nit: full

backend/api/Services/DocumentService.cs
```diff
@@ -129,29 +134,88 @@ public async Task<DocumentUploadResponse> UploadActivityDocumentAsync(long activ
...
+                    if (existing == null)
+                    {
+                        creates.Add(updateMetadata);
+                    }
+                    else if (existing.Value != updateMetadata.Value && !string.IsNullOrEmpty(updateMetadata.Value))
+                    {
+                        updateMetadata.Id = existing.Id;
+                        updates.Add(updateMetadata);
+                    }
+                    else if (string.IsNullOrEmpty(updateMetadata.Value))
```
devinleighsmith, line 180: hmm, just want to double check that this works. So when we want to delete something we send an empty array element from the frontend? Would it be better to check if there are items in the existingMetadata that aren't in update request instead? that would be more in line with the pattern we are using in the repository.

backend/api/Services/DocumentService.cs
```diff
@@ -129,29 +134,88 @@ public async Task<DocumentUploadResponse> UploadActivityDocumentAsync(long activ
...
+                response.MetadataExternalResult.AddRange(await CreateMetadata(updateRequest.MayanDocumentId, creates));
+
+                // Delete metadata of document
+                response.MetadataExternalResult.AddRange(await DeleteMetadata(updateRequest.MayanDocumentId, deletes));
+
+                // Add the metadata response
+                foreach (var task in response.MetadataExternalResult)
+                {
+                    // Flag to know if at least one call was successful.
+                    metadataUpdateSucessfull = metadataUpdateSucessfull || task.Status == ExternalResultStatus.Success;
```
devinleighsmith, line 200: hmm, aren't we more concerned if one or more failed to report that to the user? I'd rather get an error if I had 99 successes and one failures just so that I know something didn't complete as expected.

backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -67,14 +67,8 @@ public PimsDocument Update(PimsDocument document, bool commitTransaction = true)
...
             this.User.ThrowIfNotAuthorized(Permissions.DocumentEdit);
-            var existingDocument = this.Context.PimsDocuments.Where(l => l.DocumentId == document.DocumentId).FirstOrDefault()
-                 ?? throw new KeyNotFoundException();
-            Context.Entry(existingDocument).CurrentValues.SetValues(document);
-            if (commitTransaction)
-            {
-                this.Context.CommitTransaction();
-            }
-            return existingDocument;
+            document = Context.Update(document).Entity;
```
devinleighsmith, line 70: This doesn't allow for related entities to be modified does it?

(one nice thing about context.entry(blah).currentvalues.setValues is that you can only load the relationships that are allowed to be modified whereas Update() will affect anything in the passed object.

frontend/src/features/documents/commonStyles.tsx
```diff
@@ -0,0 +1,36 @@
+import { Scrollable } from 'components/common/Scrollable/Scrollable';
+import styled from 'styled-components';
+
+export const StyledGreySection = styled.div`
+  padding: 1rem;
+  background-color: ${({ theme }) => theme.css.filterBackgroundColor};
+`;
+
+export const StyledH2 = styled.h2`
```
devinleighsmith, line 9: I'm always a bit worried when I see new h* definitions. Do our existing centralized h2, h3 definitions not work for this page? If not we may want to discuss that with Bobbi, as we are trying to make our headers consistent.

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -37,6 +53,26 @@ export const DocumentDetailContainer: React.FunctionComponent<IDocumentDetailCon
     fetch();
   }, [props.pimsDocument.mayanDocumentId, retrieveDocumentMetadata]);
 
+  useEffect(() => {
+    const fetch = async () => {
+      if (props.pimsDocument.mayanDocumentId !== undefined) {
```
devinleighsmith, line 58: If this if statements fail, do we need error handling of some kind? it seems to me that the form would potentially become unusable without this data.

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -47,12 +83,28 @@ export const DocumentDetailContainer: React.FunctionComponent<IDocumentDetailCon
...
-      onUpdate={onUpdateDocument}
-      onCancel={props.onCancel}
-    />
-  );
+
+  const onCancelUpdate = () => {
+    setIsEditable(false);
+  };
+
+  if (!isEditable) {
```
devinleighsmith, line 91: nit: if(truthy) else
vs
if(falsey) else

frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <StyledScrollable>
+              <Formik<DocumentUpdateFormData>
+                initialValues={initialFormState}
+                onSubmit={async (values: DocumentUpdateFormData, { setSubmitting }) => {
+                  if (props.document?.pimsDocument?.id && values.documentStatusCode !== undefined) {
+                    var request = values.toRequestApi();
+                    await props.onUpdate(request);
+                    setSubmitting(false);
+                  } else {
+                    console.error('Selected document status is not valid');
```
devinleighsmith, line 73: is this something that we should display to the user so they can correct their mistake? or just something we may run into during development.

frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
```diff
@@ -69,10 +69,32 @@ export const getDocumentColumns = ({
...
+function renderDocumentType({ value }: CellProps<Api_Document, Api_DocumentType>) {
   return value?.documentType ?? '';
 }
 
+const renderFileName = (onViewDetails: (values: Api_Document) => void) => {
+  return function(cell: CellProps<Api_Document, string>) {
+    const { hasClaim } = useKeycloakWrapper();
+    return (
+      <>
+        {hasClaim(Claims.DOCUMENT_VIEW) && (
```
devinleighsmith, line 81: nit: collapse these two statements into one ternary.

devinleighsmith (approved):

backend/api/Services/DocumentService.cs
FuriousLlama: We actually do send every outcome status to the front end, is just that for the purpose of updating the DB, if at least one is successful we proceed with updating our side. We can switch it around to be an all or none case if necessary

backend/dal/Repositories/DocumentRepository.cs
FuriousLlama: I don't believe it will. At the moment the document is quite simple and only has the status as a related table

frontend/src/features/documents/commonStyles.tsx
FuriousLlama: Not 100% sure, but given this is a bugfix I rather not make that change here. As it is, is just a refactor of what was already used in the styles.

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
FuriousLlama: That is correct, the data is necessary for the form to make sense. I am not sure what can be done for handling. We do not have a consistent way of re-fetching data in other parts of the application so re-load seems the way to go for the user.

frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
FuriousLlama: It should never happen for the user. The status is set in a Combobox without the option of not selecting it. If it reaches that error is that something went pretty bad.

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
FuriousLlama: a yes, the javascript

backend/api/Services/DocumentService.cs
FuriousLlama: I disagree that this does not align with the current pattern we are using. From the form point of view, the way of deleting something is just deleting the text. That does create an entry of the metadata with an empty string. The previous if in this line checks if there is already an element matching in the existing metadata and this if checks if the expected update is the empty string and if so it should be deleted. That leaves the form logic to the minimum necessary.
Changing it so that the front end removes the element if empty would be moving the logic to the frontend, that is what we agreed not to do.

frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
devinleighsmith: In that case a toast here might be nice, since I wouldn't expect users to check the console.

devinleighsmith:
@FuriousLlama can this be merged? That way I can start resolving conflicts.

PR #2072, opened 2022-09-14
PSP-4074 Review todos on backend and frontend

asanchezr:


frontend/src/components/Table/Table.tsx
```diff
@@ -125,7 +125,7 @@ export interface TableProps<T extends object = {}, TFilter extends object = {}>
   showSelectedRowCount?: boolean;
   hideHeaders?: boolean;
   onRequestData?: (props: { pageIndex: number; pageSize: number }) => void;
-  loading?: boolean; // TODO: Show loading indicator while fetching data from server
+  loading?: boolean; // TODO: PSP-4394 Show loading indicator while fetching data from server
```
devinleighsmith, line 128: @asanchezr I think this already works does it not? At least, I've seen tables with a loading indicator.

devinleighsmith (approved):

frontend/src/components/Table/Table.tsx
asanchezr: Let me double check

asanchezr: Removed obsolete TODO

FuriousLlama (approved):

PR #2075, opened 2022-09-14
Psp 3946

devinleighsmith:
Correct error in delete behaviour when using multiple documents with the same status code.


backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -80,6 +80,11 @@ public bool Delete(PimsDocument document)
         {
             document.ThrowIfNull(nameof(document));
 
+            foreach (var pimsActivityInstanceDocument in document.PimsActivityInstanceDocuments)
```
devinleighsmith, line 83: changed my mind on this, if we are deleting a document we should delete any dependent relationships. This is a stand-in for proper cascades.

backend/api/Services/DocumentService.cs
```diff
@@ -74,10 +74,9 @@ public async Task<bool> DeleteActivityDocumentAsync(PimsActivityInstanceDocument
             this.Logger.LogInformation("Deleting PIMS document for single activity");
             this.User.ThrowIfNotAuthorized(Permissions.DocumentDelete);
 
-            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespository.GetAllByDocument(activityDocument.DocumentId);
+            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespository.GetAllByDocument(activityDocument.DocumentId, true);
```
devinleighsmith, line 77: Need to track here, otherwise when we delete we will load each disconnected entity, which will result in the document status code value being loaded multiple times, which will cause an error if the same status code is loaded.

backend/api/Services/DocumentService.cs
```diff
@@ -254,7 +253,10 @@ public async Task<bool> DeleteDocumentAsync(PimsDocument document)
                 if (result.Status == ExternalResultStatus.Success || result.HttpStatusCode == HttpStatusCode.NotFound)
                 {
                     documentRepository.Delete(document);
-                    documentRepository.CommitTransaction();
+                    if (commitTransaction)
```
devinleighsmith, line 256: allow deleting multiple documents in a single transaction.

devinleighsmith:
For context, this story provides the ability to easily inject a form into the activity container based on the activity template type.

backend/dal/Repositories/DocumentActivityRepository.cs
```diff
@@ -34,32 +34,38 @@ public DocumentActivityRepository(PimsContext dbContext, ClaimsPrincipal user, I
...
                 .Include(ad => ad.Document)
                     .ThenInclude(d => d.DocumentType)
-                .Where(ad => ad.DocumentId == documentId)
-                .AsNoTracking()
-                .ToList();
+                .Where(ad => ad.DocumentId == documentId);
+            if (tracking)
+            {
+                query = query.AsNoTracking();
+            }
```
FuriousLlama, line 48: we should standardize this so the repositories are agnostic about the usage

frontend/src/features/properties/map/activity/ActivityTray/ActivityTray.tsx
```diff
@@ -9,43 +9,47 @@ import styled from 'styled-components';
...
+  currentFormContent,
+}) => {
   const [show, setShow] = useState(true);
   let trayContent = (
     <HalfHeightDiv>
       {!!activity?.id && (
         <ActivityForm
           activity={{ ...activity, id: +activity.id }}
-          file={file}
+          file={file as ActivityFile}
```
FuriousLlama, line 47: why is the type coercion necessary?

frontend/src/features/properties/map/activity/detail/ActivityContainer.test.tsx
```diff
@@ -55,7 +57,7 @@ describe('Activity Container', () => {
         }}
       >
         <ActivityContainer
-          View={ActivityView}
+          View={ActivityView as FunctionComponent<IActivityTrayProps>}
```
FuriousLlama, line 60: ?

frontend/src/features/properties/map/activity/detail/ActivityForm.tsx
```diff
@@ -45,13 +50,41 @@ export const ActivityForm: React.FunctionComponent<IActivityFormProps> = ({
...
     }
   };
+  const EditForm = formContent?.EditForm;
+  const ViewForm = formContent?.ViewForm;
+  const initialValues = ActivityModel.fromApi(activity, file.fileType);
+  if (
+    !!initialValues.activityData?.version &&
+    formContent !== undefined &&
+    initialValues.activityData?.version !== formContent?.version
+  ) {
```
FuriousLlama, line 60: Did this happen too often?

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+      EditForm: () => (
+        <>
+          <b>TEST SURVEY FORM</b>
+          <Row>
+            <Col>
+              <Input field={withNameSpace('activityData', 'test')}></Input>
+            </Col>
+          </Row>
+        </>
+      ),
```
FuriousLlama, line 24: This is a new pattern, lets go over this on our dev meeting

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+      ViewForm: () => {
+        const { values } = useFormikContext<ActivityModel>();
```
FuriousLlama, line 26: if its a view, why do we need formik? Can we pass the view props instead?

backend/api/Models/Concepts/ActivityInstance/PimsActInstPropFile.cs
```diff
@@ -0,0 +1,11 @@
+namespace Pims.Api.Models.Concepts
+{
+    public class PimsActInstPropFile : BaseAppModel
```
asanchezr, line 3: I saw another review where we were turning these into more readable names?

backend/api/Services/DocumentService.cs
```diff
@@ -74,11 +74,10 @@ public async Task<bool> DeleteActivityDocumentAsync(PimsActivityInstanceDocument
             this.Logger.LogInformation("Deleting PIMS document for single activity");
             this.User.ThrowIfNotAuthorized(Permissions.DocumentDelete);
 
-            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespository.GetAllByDocument(activityDocument.DocumentId);
+            IList<PimsActivityInstanceDocument> existingActivityDocuments = documentActivityRespository.GetAllByDocument(activityDocument.DocumentId, true);
```
asanchezr, line 77: nit: documentActivityRe(s)pository -> documentActivityRepository

asanchezr: Also while not caused by this PR, it would be nice to remove duplicated assignment from constructor
image

backend/dal/Repositories/DocumentActivityRepository.cs
asanchezr: Also the variable name is confusing - so if tracking == true we want the query as NO TRACKING? Or is it a bug?

backend/dal/Repositories/DocumentActivityRepository.cs
```diff
@@ -34,32 +34,38 @@ public DocumentActivityRepository(PimsContext dbContext, ClaimsPrincipal user, I
...
                 .Include(ad => ad.Document)
                     .ThenInclude(d => d.DocumentType)
-                .Where(ad => ad.ActivityInstanceId == activityId)
-                .AsNoTracking()
-                .ToList();
+                .Where(ad => ad.ActivityInstanceId == activityId);
+            if (tracking)
+            {
+                query = query.AsNoTracking();
+            }
```
asanchezr, line 67: Same comment re: tracking var vs current logic

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
asanchezr: agree, kinda would like to not inline the EditForm here but load it from a file. Can discuss further in the meeting

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
asanchezr: If you look at the ActivityContainer the parent of the form wraps either the View or Edit form with formik so both share the formik context. Let's discuss on our meeting about pattern for View/Edit forms like these

asanchezr (approved):

frontend/src/features/properties/map/activity/detail/ActivityForm.tsx
devinleighsmith: It didn't happen at all, but it could happen in the future.

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
devinleighsmith: yes, loading it from a file would be my recommendation. Done here since I'm not actually implemented a real form.

frontend/src/features/properties/map/activity/detail/content/formContent.tsx
devinleighsmith: you could pass the props instead, either way you will be within the context of the parent formik instance.

frontend/src/features/properties/map/activity/ActivityTray/ActivityTray.tsx
devinleighsmith: its not, removed.

backend/dal/Repositories/DocumentActivityRepository.cs
devinleighsmith: I removed this, need to discuss our current deletion pattern on Friday, our current pattern does not work for multiple deletes in a single transaction.

PR #2099, opened 2022-09-15
Document fixes

FuriousLlama:
Added warning modal when uploading a document & editing documents. 
Fixed issue that prevented validation from working on metadata

psp-4274 | psp-4379 | psp-4416 | psp-4448


frontend/src/components/common/buttons/Button.tsx
```diff
@@ -188,6 +188,7 @@ const StyledButton = styled(BootstrapButton)`
       line-height: 3rem;
       justify-content: left;
       letter-spacing: unset;
+      text-align: left;
```
FuriousLlama, line 191: this aligns the text inside a button of variant=link to the left so it works with long text. I checked the places where we use this button variant and they looked fine. I will check with sue to make sure nothing is looking messy after this.

devinleighsmith:
Please inform Sue about any regression testing she will need to perform as a result of these changes.

frontend/src/contexts/modalContext.tsx
```diff
@@ -13,42 +13,38 @@ export const ModalContext = React.createContext<IModalContext>({
...
-    (updatedModalProps?: ModalProps) =>
+    (updatedModalContent?: ModalContent) => {
       setModalProps({
-        ...modalProps,
-        ...updatedModalProps,
-        display: updatedModalProps?.display ?? modalProps?.display ?? false,
-        setDisplay: displayFunction,
-      }),
-    [displayFunction, modalProps],
+        ...updatedModalContent,
```
devinleighsmith, line 30: right, ok, so keep the display logic separate from the other modal content since generally the display is the only thing that changes after initialization.

frontend/src/features/documents/documentUpload/DocumentUploadForm.tsx
```diff
@@ -54,15 +57,24 @@ const DocumentUploadForm: React.FunctionComponent<IDocumentUploadFormProps> = pr
     props.initialDocumentType,
     props.mayanMetadataTypes,
   );
+  useEffect(() => {
+    const isTypeDirty =
```
devinleighsmith, line 61: hmm, so it seems like your changes handle the cases where:
the file is changed
the document type is changed

What about metadata changing? or does that already work without explicit handling?

devinleighsmith (approved):

frontend/src/contexts/modalContext.tsx
FuriousLlama: that is correct. It also does not make sense to allow the consumer to change/override the display function on the props.

frontend/src/features/documents/documentUpload/DocumentUploadForm.tsx
FuriousLlama: it does that automatically given those fields live in formik natively unlike the file. The reason the type needs to be set manually is because it comes as part of the form props

PR #2111, opened 2022-09-19
Psp 3912

devinleighsmith:
Allow the list of properties on an acquisition file to be edited.
Refactor existing component to be file type agnostic.


backend/api/Models/Concepts/File/FilePropertyModel.cs
```diff
@@ -0,0 +1,45 @@
+namespace Pims.Api.Models.Concepts.File
+{
+    public class FilePropertyModel : BaseModel
```
devinleighsmith, line 3: This is the minimal set of fields that all FileProperty entities should have, regardless of the type of the file.

backend/api/Models/Concepts/Property/PropertyModel.cs
```diff
@@ -147,6 +147,16 @@ public class PropertyModel : BaseModel
         /// </summary>
         public string PlanNumber { get; set; }
 
+        /// <summary>
+        /// get/set - Whether this parcel is owned by the ministry.
+        /// </summary>
+        public bool IsOwned { get; set; }
```
devinleighsmith, line 153: These fields are required in the db but not exposed/set up until this point.

backend/api/Models/Concepts/ResearchFile/ResearchFileModel.cs
```diff
@@ -14,7 +14,7 @@ public class ResearchFileModel : FileModel
         /// <summary>
         /// get/set - A list of research property relationships.
         /// </summary>
-        public IList<ResearchFilePropertyModel> ResearchProperties { get; set; }
+        public override IList<ResearchFilePropertyModel> FileProperties { get; set; }
```
devinleighsmith, line 17: research file properties have some additional fields not part of the base FileProperties entity so overriding here.

backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
```diff
@@ -0,0 +1,93 @@
...
+                Context.Entry(propertyAcquisitionFile.Property).State = EntityState.Unchanged;
+            }
+
+            Context.PimsPropertyAcquisitionFiles.Add(propertyAcquisitionFile);
+            return propertyAcquisitionFile;
+        }
+
+        public void Delete(PimsPropertyAcquisitionFile propertyAcquisitionFile)
+        {
+            propertyAcquisitionFile.ThrowIfNull(nameof(propertyAcquisitionFile));
```
devinleighsmith, line 66: I'll update this to reflect the changes we discussed today.

backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                Context.Entry(propertyAcquisitionFile.Property).State = EntityState.Unchanged;
+            }
+
+            propertyAcquisitionFile.PimsActInstPropAcqFiles.ForEach(s => Context.PimsActInstPropAcqFiles.Remove(s));
+
+            Context.PimsPropertyAcquisitionFiles.Remove(propertyAcquisitionFile);
+        }
+
+        public PimsPropertyAcquisitionFile Update(PimsPropertyAcquisitionFile propertyAcquisitionFile)
```
devinleighsmith, line 78: same here.

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -53,7 +67,22 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
       header={<AcquisitionHeader acquisitionFile={acquisitionFile} />}
       footer={null}
     >
-      <></>
+      <>
```
devinleighsmith, line 70: There isn't really a slot for this button and creating a layout for acquisition files is outside the scope of this story. I may need to circle back later in the sprint and make sure this ends up in the right place.

frontend/src/features/properties/map/activity/detail/ActivityContainer.tsx
```diff
@@ -126,18 +123,9 @@ export const getActivityPropertiesOnFile = (
...
-      ?.filter(researchProperty => activityFilePropertyIds.includes(researchProperty.id))
-      .map(p => ({ id: p.id })) as Api_PropertyFile[];
-  } else if (file?.fileType === FileTypes.Acquisition) {
-    return (file as Api_AcquisitionFile).acquisitionProperties
-      ?.filter(acquisitionProperty => activityFilePropertyIds.includes(acquisitionProperty.id))
-      .map(p => ({ id: p.id })) as Api_PropertyFile[];
-  } else {
-    throw Error('Unexpected file type');
-  }
+  return file.fileProperties
```
devinleighsmith, line 126: creating this common fileProperties logic allows this function to be simplified.

frontend/src/features/properties/map/context/sidebarContext.tsx
```diff
@@ -54,25 +47,13 @@ export const SideBarContextProvider = (props: {
     [setFile, setStaleFile],
   );
 
-  const getFileProperties = () => {
```
devinleighsmith, line 57: same here.

frontend/src/features/properties/map/research/add/models.ts
```diff
@@ -35,76 +34,9 @@ export class ResearchForm {
...
     newForm.id = model.id;
     newForm.name = model.fileName || '';
-    newForm.properties = model.researchProperties?.map(x => PropertyForm.fromApi(x)) || [];
-    newForm.rowVersion = model.rowVersion;
-
-    return newForm;
-  }
-}
-
-export class PropertyForm {
```
devinleighsmith, line 45: This model was not doing anything specific to research files, so moved this into the common area and reused it between acquisition and research files instead of using the duplicated model.

frontend/src/features/properties/map/shared/update/properties/UpdateProperties.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+    return children;
+  });
+});
+
+const setIsShowingPropertySelector = jest.fn();
+const onSuccess = jest.fn();
+const updateFileProperties = jest.fn();
+
+describe('UpdateProperties component', () => {
```
devinleighsmith, line 29: There were no tests for this component so I added some.

frontend/src/features/properties/map/shared/update/properties/UpdateProperties.tsx
```diff
@@ -4,29 +4,27 @@ import SidebarFooter from 'features/properties/map/shared/SidebarFooter';
...
-interface IUpdatePropertiesProps {
-  researchFile: Api_ResearchFile;
+export interface IUpdatePropertiesProps {
+  file: Api_File;
   setIsShowingPropertySelector: (isShowing: boolean) => void;
   onSuccess: () => void;
+  updateFileProperties: (file: Api_File) => Promise<Api_File | undefined>;
 }
 
 export const UpdateProperties: React.FunctionComponent<IUpdatePropertiesProps> = props => {
```
devinleighsmith, line 21: This component can now operate on any file type by extracting the research specific embedded call to updateResearchFileProperties to a prop.

devinleighsmith:
I've informed sue on the story of possible regression to research file add/edit properties logic.

backend/api/Models/Concepts/AcquisitionFile/AcquisitionFilePropertyMap.cs
```diff
@@ -14,14 +14,15 @@ public void Register(TypeAdapterConfig config)
...
+                .Map(dest => dest.File, src => src.AcquisitionFile)
+                .Map(dest => dest.FileId, src => src.AcquisitionFileId)
                 .Inherits<Entity.IBaseEntity, BaseModel>();
 
             config.NewConfig<AcquisitionFilePropertyModel, Entity.PimsPropertyAcquisitionFile>()
                 .Map(dest => dest.PropertyAcquisitionFileId, src => src.Id)
                 .Map(dest => dest.Property, src => src.Property)
                 .Map(dest => dest.PropertyId, src => src.Property.Id)
-                .Map(dest => dest.AcquisitionFileId, src => src.AcquisitionFile.Id)
+                .Map(dest => dest.AcquisitionFileId, src => src.File.Id != 0 ? src.File.Id : src.FileId)
```
FuriousLlama, line 25: this looks risky. Would not be better to just use the src.FileId?

backend/api/Models/Concepts/ResearchFile/ResearchFileModel.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Pims.Api.Models.Concepts
 {
-    public class ResearchFileModel : FileModel
+    public class ResearchFileModel : FileModel<ResearchFilePropertyModel>
```
FuriousLlama, line 6: not 100% sure templates is the answer here. If we are using inheritance, why not also create an interface for the FileProperty?
That way the list below will be something like:
```
public IList<FilePropertyModel> FileProperties { get; set; }
```

I think that way the upcasting to FileModel will work without any problems.

devinleighsmith: Sorry, not sure what you mean by templates, are you referring to generics?

FuriousLlama: Sorry, yeah. Same concept different name.

devinleighsmith: The problem that I was running into is overriding FileProperties within the research file.
IList<ResearchFilePropertyModel> FileProperties
does not inherit from
IList<FilePropertyModel> FileProperties
even if ResearchFilePropertyModel extends FileProperties.
The best that I could get away with in that case was "hiding" the FileProperties field in the ResearchFileModel which generates a warning.

the reason that I need IList<ResearchFilePropertyModel> FileProperties in the research file is mostly for the mappings. I could try casting or maybe see if the mapping is smart enough to realize that a FileProperty is actually a ResearchFileProperty at runtime perhaps?

FuriousLlama: You could explicitly cast which I think it will be safe. Assuming we are not going to be doing manual modifications to the object I think it would also be safe not to have the generic either.

This compiled and worked for me. 
```


public interface IFileProperty
{
    string a { get; set; }
}

public class RFileProperty : IFileProperty
{
    private string _a;

    public string a { get => _a; set => _a = value; }

    public string rOnly { get; set; }
}

public interface IFile
{
    IList<IFileProperty> Properties { get; }
}

public class RFile : IFile
{
    private IList<IFileProperty> _properties;

    public IList<IFileProperty> Properties { get => _properties; }

    public RFile()
    {
        _properties = new List<IFileProperty>();
    }

    public void AddProperty(RFileProperty property)
    {
        _properties.Add(property);
    }
}

class Test : IProblem
{
    public override void Run()
    {
        IFile rFile = new RFile();

        RFileProperty rProperty = new RFileProperty();
        rProperty.a = "some test";
        rProperty.rOnly = "only for R";

        rFile.Properties.Add(rProperty);

        foreach (IFileProperty prop in rFile.Properties)
        {
            RFileProperty rProp = (RFileProperty)prop;
            System.Console.WriteLine(rProp.a);
            System.Console.WriteLine(rProp.rOnly);
        }
    }
}
```

FuriousLlama: I think the mapper can be specified to map between the dal and the explicit object. I am not sure how the serializer will handle that

devinleighsmith: Sure, I'm willing to try it at least.

devinleighsmith: Ok, this is a bit of a rabbit hole. Thinks work mostly fine. Mapster isn't smart enough to see the inheritence in the the nested list (even if you cast the type within the mapping). Not a big deal, can just do this:
```
.AfterMapping((src, dest) =>
                {
                    dest.FileProperties = new List<FilePropertyModel>();
                    foreach(var propertyResearchFile in src.PimsPropertyResearchFiles)
                    {
                        dest.FileProperties.Add(propertyResearchFile.Adapt<ResearchFilePropertyModel>());
                    }
                });
```

The bigger deal is the json serializer at the controller level. Outgoing it doesn't work out of the box (it only serializes the base classes properties). That can be worked around by manually serializing it using this and returning the result: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-polymorphism

When receiving data we would need to do something similar. Manually parse the json as a stream and then use the steps from the above link to force it to parse the incoming json as the derived class.

So, it works, but in my mind this solution seems like more work then the above and possibly more error prone as well (if someone just parses a researchfile as a researchfile in the controller naively they wont get any of the fields in the derived property list).

What do you think?

backend/api/Models/Concepts/Property/PropertyMap.cs
```diff
@@ -32,6 +32,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.PphStatusUpdateTimestamp, src => src.PphStatusUpdateTimestamp)
                 .Map(dest => dest.PphStatusUpdateUserGuid, src => src.PphStatusUpdateUserGuid)
                 .Map(dest => dest.Notes, src => src.Notes)
+                .Map(dest => dest.IsOwned, src => src.IsOwned)
+                .Map(dest => dest.IsVisibleToOtherAgencies, src => src.IsVisibleToOtherAgencies)
```
asanchezr, line 36: Is this field `IsVisibleToOtherAgencies` relevant to this implementation of PIMS? Seems something like from legacy CITZ PIMS...

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -53,7 +67,22 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
...
+        {hasClaim(Claims.ACQUISITION_EDIT) && acquisitionFile !== undefined ? (
+          <Button variant="success" onClick={() => setIsShowingPropertySelector(true)}>
+            <MdLocationPin size={'2.5rem'} />
+            Edit properties
+          </Button>
+        ) : null}
```
asanchezr, line 76: I know this is placeholder code but here we check the user has ACQUISITION_EDIT claim. However the corresponding code in Research File checks for RESEARCH_ADD - any reason for the discrepancy? I lean towards thinking that research file logic has a bug. 

Or should we check for ADD || EDIT ? That way if a user has only ADD or EDIT claims they can still see the Edit Properties button.

asanchezr: Sorry, disregard my previous comment. This button only shows for UPDATE so we should check for UPDATE claim - right?

frontend/src/features/properties/map/shared/update/properties/UpdateProperties.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      expect(updateFileProperties).toHaveBeenCalled();
+      expect(setIsShowingPropertySelector).toHaveBeenCalledWith(false);
+      expect(onSuccess).toHaveBeenCalled();
+    });
+  });
+
+  it('handles cancel as expected', async () => {
+    setup({});
+    expect(document.body).toMatchSnapshot();
+  });
```
asanchezr, line 137: copy/pasta?

frontend/src/features/properties/map/shared/update/properties/UpdateProperties.tsx
```diff
@@ -4,29 +4,27 @@ import SidebarFooter from 'features/properties/map/shared/SidebarFooter';
...
-  const formikRef = useRef<FormikProps<ResearchForm>>(null);
-  const formResearchFile = ResearchForm.fromApi(props.researchFile);
+  const formikRef = useRef<FormikProps<FileForm>>(null);
+  const formResearchFile = FileForm.fromApi(props.file);
```
asanchezr, line 23: nit: `formResearchFile` ->  `formFile`

frontend/src/features/properties/map/shared/update/properties/UpdateProperties.tsx
```diff
@@ -80,13 +78,13 @@ export const UpdateProperties: React.FunctionComponent<IUpdatePropertiesProps> =
...
-        <Formik<ResearchForm>
+        <Formik<FileForm>
           innerRef={formikRef}
           initialValues={formResearchFile}
           validationSchema={UpdatePropertiesYupSchema}
-          onSubmit={async (values: ResearchForm) => {
-            const researchFile: Api_ResearchFile = values.toApi();
-            await saveResearchFile(researchFile);
+          onSubmit={async (values: FileForm) => {
+            const researchFile: Api_File = values.toApi();
```
asanchezr, line 86: nit: rename to `file`

backend/api/Models/Concepts/ResearchFile/ResearchFileModel.cs
FuriousLlama: That's unfortunate that the JSON serializer is not working as we want. This is a model only for marshaling the file from and to the controller right? If that is the case maybe just specified mappings are necessary without any inheritance or generics. Otherwise, I fear we will start debugging obscure issues on the frontend when the serializer doesn't do what we want if we add new data

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
devinleighsmith: Yes, and update = edit.

devinleighsmith:
@FuriousLlama or @asanchezr any further changes or is this one good now

asanchezr:
@devinleighsmith I'm good with the changes but was leaving the final approval up to @FuriousLlama as you discussed backend serialization stuff at length

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -107,12 +129,14 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
...
-            setContainerState={setContainerState}
-          />
+          <>
+            <AcquisitionMenu
+              items={menuItems}
+              selectedIndex={containerState.selectedMenuIndex}
+              onChange={onMenuChange}
+              setContainerState={setContainerState}
+            />
+          </>
```
asanchezr, line 139: out of curiosity - why do we need the extra `<></>` wrapper if there's only a single child component?

frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -77,17 +81,35 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
...
     return <LoadingBackdrop show={true} parentScreen={true}></LoadingBackdrop>;
   }
 
+  if (containerState.activeEditForm === EditFormNames.propertySelector && acquisitionFile) {
+    return (
+      <UpdateProperties
+        file={acquisitionFile}
+        setIsShowingPropertySelector={() =>
+          setContainerState({ activeEditForm: EditFormNames.acquisitionSummary })
+        }
```
asanchezr, line 106: If I'm understanding this logic well, it would take the user to the edit summary form after finishing with adding/removing properties. I believe the pattern set in Research file is to return to read only view, right?

Something like this would do the trick (I think)

```tsx
setContainerState({
  isEditing: false,
  activeEditForm: undefined
})
```

asanchezr (approved):

FuriousLlama (approved):

PR #2115, opened 2022-09-20
Document generation integration

FuriousLlama:
Added Document Generation functionality to the backend to allow document generation through restful calls to CDOGS

### DIFF
--- a/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
+++ b/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
@@ -24,7 +24,7 @@ namespace Pims.Api.Controllers
         #region Constructors
 
         /// <summary>
-        /// Creates a new instance of a ErrorController class.
+        /// Creates a new instance of a DocumentGenerationController class.
         /// </summary>
         /// <param name="documentGenerationService"></param>
         /// <param name="mapper"></param>
--- a/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
+++ b/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
@@ -1,0 +1,69 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Pims.Api.Models;
+using Pims.Api.Services;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Pims.Api.Controllers
+{
+    /// <summary>
+    /// DocumentGenerationController class, provides endpoints to handle document generation requests.
+    /// </summary>
+    //[Authorize]
+    [ApiController]
+    [ApiVersion("1.0")]
+    [Route("v{version:apiVersion}/documentGeneration/")]
+    [Route("/documentGeneration")]
+    public class DocumentGenerationController : ControllerBase
+    {
+        #region Variables
+        private readonly IDocumentGenerationService _documentGenerationService;
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Creates a new instance of a ErrorController class.
+        /// </summary>
+        /// <param name="documentGenerationService"></param>
+        /// <param name="mapper"></param>
+        public DocumentGenerationController(IDocumentGenerationService documentGenerationService)
+        {
+            _documentGenerationService = documentGenerationService;
+        }
+        #endregion
+
+        #region Endpoints
+
+        /// <summary>
+        /// Get supported file types.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("supported_types")]
+        //[HasPermission(Permissions.DocumentView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(ExternalResult<Models.Cdogs.FileTypes>), 200)]
+        [SwaggerOperation(Tags = new[] { "document-generation" })]
+        public async Task<IActionResult> GetSupportedDocumentTypes()
+        {
+            var supportedFileTypes = await _documentGenerationService.GetSupportedFileTypes();
+            //var mappedDocumentTypes = _mapper.Map<List<Concepts.DocumentTypeModel>>(documentTypes);
+            return new JsonResult(supportedFileTypes);
+        }
+
+        /// <summary>
+        /// Uploads the passed document.
+        /// </summary>
+        [HttpPost("upload_template")]
+        //[HasPermission(Permissions.DocumentAdd)]
+        [ProducesResponseType(typeof(ExternalResult<string>), 200)]
+        [SwaggerOperation(Tags = new[] { "document-generation" })]
+        public async Task<IActionResult> UploadTemplate([FromForm] IFormFile file)
+        {
+            var result = await _documentGenerationService.UploadFileTemplate(file);
+            return new JsonResult(result);
+        }
+        #endregion
+    }
+}
--- a/backend/api/Models/Cdogs/FileTypes.cs
+++ b/backend/api/Models/Cdogs/FileTypes.cs
@@ -1,0 +1,18 @@
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Pims.Api.Models.Cdogs
+{
+    /// <summary>
+    /// Represents a dictionary of supported input template file types and output file types.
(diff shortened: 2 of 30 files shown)


backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -0,0 +1,69 @@
...
+    public class DocumentGenerationController : ControllerBase
+    {
+        #region Variables
+        private readonly IDocumentGenerationService _documentGenerationService;
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Creates a new instance of a ErrorController class.
```
devinleighsmith, line 27: copy pasta

backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -0,0 +1,69 @@
...
+        #endregion
+
+        #region Endpoints
+
+        /// <summary>
+        /// Get supported file types.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("supported_types")]
+        //[HasPermission(Permissions.DocumentView)]
```
devinleighsmith, line 44: mmm, good reminder, we'll need a claims task at some point for this.

backend/api/Models/Cdogs/FileTypes.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Pims.Api.Models.Cdogs
+{
+    /// <summary>
+    /// Represents a dictionary of supported input template file types and output file types.
+    /// </summary>
+    public class FileTypes
+    {
+        /// <summary>
+        /// get/set - The document type id.
```
devinleighsmith, line 12: this comment doesn't seem correct.

backend/api/Repositories/Cdogs/CdogsAuthRepository.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        private async Task<ExternalResult<JwtResponse>> RequestToken()
+        {
+            _logger.LogDebug("Getting authentication token...");
+
+            using HttpClient client = _httpClientFactory.CreateClient("Pims.Api.Logging");
+            client.DefaultRequestHeaders.Accept.Clear();
+
+            Uri endpoint = new Uri(_config.AuthHost, "auth/realms/jbd6rnxw/protocol/openid-connect/token");
```
devinleighsmith, line 81: I would think that the realm needs to be .env configurable. There are dev/test/prod instances of cdogs so I would assume there are dev/test/prod keycloak realms as well.

backend/api/Repositories/Cdogs/CdogsAuthRepository.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                _lastSucessfullRequest = DateTime.UtcNow;
+                _currentToken = tokenResult.Payload;
+            }
+
+            return _currentToken.AccessToken;
+        }
+
+        private bool IsValidToken()
```
devinleighsmith, line 55: Hmm, I don't see anything about the refresh token in this logic.

backend/api/Repositories/RestCommon/BaseRestRespository.cs
```diff
@@ -0,0 +1,212 @@
...
+                result.Status = ExternalResultStatus.Error;
+                result.Message = "Exception during Delete";
+                _logger.LogError("Unexpected exception during delete {e}", e);
+            }
+            return result;
+        }
+
+        private async Task<ExternalResult<T>> ProcessResponse<T>(HttpResponseMessage response)
+        {
+            ExternalResult<T> result = new ExternalResult<T>()
```
devinleighsmith, line 164: This looks like it was all done before we discussed using ProblemDetails so fine to leave until we do 4468

backend/api/Repositories/RestCommon/BaseRestRespository.cs
```diff
@@ -0,0 +1,212 @@
...
+        {
+            _logger = logger;
+            _httpClientFactory = httpClientFactory;
+        }
+
+        public abstract void AddAuthentication(HttpClient client, string authenticationToken = null);
+
+        public async Task<ExternalResult<T>> GetAsync<T>(Uri endpoint, string authenticationToken)
+        {
+            using HttpClient client = _httpClientFactory.CreateClient("Pims.Api.Logging");
```
devinleighsmith, line 38: could we put this common code into a private function to cut down on code duplication?

backend/api/Repositories/Cdogs/CdogsRepository.cs
```diff
@@ -0,0 +1,86 @@
...
+            return result;
+        }
+
+        public async Task<ExternalResult<string>> UploadTemplateAsync(IFormFile file)
+        {
+            _logger.LogDebug("Uploading document template...");
+            string authenticationToken = await _authRepository.GetTokenAsync();
+
+            byte[] fileData;
+            using var byteReader = new BinaryReader(file.OpenReadStream());
```
devinleighsmith, line 58: nice to not have to manually dispose this.

backend/api/Repositories/Cdogs/CdogsRepository.cs
```diff
@@ -0,0 +1,86 @@
...
+            IConfiguration configuration)
+            : base(logger, httpClientFactory, configuration)
+        {
+            _authRepository = authRepository;
+        }
+
+        public async Task<ExternalResult<FileTypes>> GetFileTypesAsync()
+        {
+            _logger.LogDebug("Retrieving supported file types...");
+            string authenticationToken = await _authRepository.GetTokenAsync();
```
devinleighsmith, line 42: hmm, does it make sense for the repository to ask the other repository for the token or should the service get the token?

devinleighsmith:
Would you mind updating the gen-env files script to add any (non-secret) config required by cdogs? It could also add the keys for the secrets, just not the values for those.

backend/api/Repositories/Cdogs/CdogsAuthRepository.cs
FuriousLlama: There is no refresh token when using the client credentials grant

backend/api/Repositories/RestCommon/BaseRestRespository.cs
FuriousLlama: We cant. The 'using' statement needs to be in the scope of the function that will dispose it

backend/api/Repositories/Cdogs/CdogsRepository.cs
FuriousLlama: The auth repository has a 'singleton' instantiation while the other ones just have a 'scope'. That means that the auth repository instance will be re-used during the execution of the application and reused while the other ones get created for each request.

devinleighsmith: Sorry, that wasn't my question. I was wondering if the CdogsRepository should directly get the token from the auth repository or if the service should pass it in.

backend/api/Repositories/RestCommon/BaseRestRespository.cs
FuriousLlama: I dont think is the same thing. The conversation around ProblemDetails was about handling exceptions at the application level and improving them with additional information. Results from external calls should be propagated from the repositories.

FuriousLlama:
> Would you mind updating the gen-env files script to add any (non-secret) config required by cdogs? It could also add the keys for the secrets, just not the values for those.

I will add a new PR for the code generation update once this PR completed

backend/api/Repositories/Cdogs/CdogsAuthRepository.cs
devinleighsmith: oh, I saw a refresh token in the model that you created, so that is not populated?

devinleighsmith (approved):

backend/api/Repositories/Cdogs/CdogsAuthRepository.cs
FuriousLlama: I based the JTW model on a general jtw response. If at some point we change the grant type it could be populated. I could remove the unused fields for clarity

PR #2118, opened 2022-09-20
IS-37.01 Database Seed Table Change

dfilteau:
PSP_PIMS S37.01 Sprint 37 2022-Sep-20
- Altered seed script:
  - 97_DML_PIMS_ACTIVITY_INSTANCE_STATUS_TYPE.sql


devinleighsmith:
@dfilteau  Could we also add a change to the build and alter up scripts to add a row to the PIMS_ACTIVITY_TEMPLATE table for the general activity? Currently that table is only populated in the test data which means that users are unable to create/view General Activities in non-dev environments.

FuriousLlama (approved):

PR #2124, opened 2022-09-20
Updated env generation for the api file

FuriousLlama:


scripts/gen-env-files.sh
```diff
@@ -9,7 +9,7 @@ read -p 'Username: ' varKeycloak
 echo 'Enter a username for the API database.'
 read -p 'Username: ' varApiDb
 
-passvar=$(grep -Po 'DB_PASSWORD=\K.*$' ./database/mssql/.env)
+passvar=$(grep -Po '^DB_PASSWORD=\K.*$' ./database/mssql/.env)
```
FuriousLlama, line 12: Looks for DB_PASSWORD at the start of the line. Prevents issues where there where multiple passwords commented out

scripts/gen-env-files.sh
```diff
@@ -72,11 +72,28 @@ else
...
 DB_PASSWORD=$passvar
+
 Keycloak__Secret=
-Keycloak__ServiceAccount__Secret=" >> ./backend/api/.env
+Keycloak__ServiceAccount__Secret=
+
+Ltsa__MyLtsaUsername=
+Ltsa__MyLtsaUserPassword=
+Ltsa__IntegratorUsername=
+Ltsa__IntegratorPassword=
```
FuriousLlama, line 87: @devinleighsmith These ones were not there but looked necessary given we are retrieving LTSA data. Any concerns?

Maheshbabu28 (approved):

scripts/gen-env-files.sh
```diff
@@ -72,11 +72,28 @@ else
...
+
+Mayan__BaseUri=http://localhost:7080/api/v4
+Mayan__ConnectionUser=admin
+Mayan__ConnectionPassword=
+
+Cdogs__AuthEndpoint=
+Cdogs__CDogsHost=
+Cdogs__ServiceClientId=
+Cdogs__ServiceClientSecret=" >> ./backend/api/.env
```
FuriousLlama, line 96: Followed a similar pattern as the keycloak secret account variable. I think adding those to the user input are more cumbersome than just updating the env file

scripts/gen-env-files.sh
devinleighsmith: This makes sense to me.

scripts/gen-env-files.sh
devinleighsmith: Having them here is a good idea, as long as none of them are set.

asanchezr (approved):

devinleighsmith (approved):

PR #2129, opened 2022-09-21
Psp 4494

devinleighsmith:


frontend/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      props: { pageSize: 5, manualPagination: true, onRequestData },
+    });
+
+    const page2Button = getByLabelText('Page 2');
+    userEvent.click(page2Button);
+    expect(tableRows).toHaveLength(6); //pagination is handled externally, the table component will ignore the page size prop.
+    expect(onRequestData).toHaveBeenCalled();
+  });
+
+  it('can select only one row at a time when single select prop set', async () => {
```
devinleighsmith, line 133: @Maheshbabu28 something like this was what I was hoping for when you fixed the single select pagination last sprint.

frontend/src/components/Table/Table.tsx
```diff
@@ -239,6 +239,17 @@ const validateProps = <T extends IIdentifiedObject, TFilter extends object = {}>
...
+    throw Error(
+      'When manual pagination is set to false do not pass in onRequestData or pageIndex.',
+    );
+  }
+  if (
+    (props.setSelectedRows || props.isSingleSelect || props.selectedRows) &&
+    !props.showSelectedRowCount
+  ) {
+    throw Error('Show selected row count required for table select logic');
+  }
```
FuriousLlama, line 252: How about using props type composition and only exporting the types and not the composed type.

Something like:
```
type ITableProps = HiddenPagination | SelectableRows;
```
We would need to define the clusters of props that need to go together.

I know this would be more complicated to implement, but it would give some guidance to the consumer of this component

frontend/src/components/Table/Table.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const { tableRows } = setup({ props: { pageSize: 5, manualPagination: false } });
+
+    expect(tableRows).toHaveLength(5);
+  });
+
+  it('it displays all rows if manualPagination is false', async () => {
+    const { tableRows } = setup({ props: { pageSize: 5, manualPagination: true } });
+
+    expect(tableRows).toHaveLength(6);
+  });
```
asanchezr, line 96: nit: `it("... if manualPagination is true", `

asanchezr (approved):

frontend/src/components/Table/Table.tsx
asanchezr: I agree that the table component has grown in complexity a lot and is in dire need of some refactoring but I'd rather had this specific bug fixes now and deal with the rest as tech debt ticket to be prioritized. Thoughts?

FuriousLlama: that's fair, expedient bug fixing sounds better. We can refactor it on dev rather than on test

FuriousLlama (approved):

PR #2135, opened 2022-09-23
PSP-4450: Acquisition team alignment issues

Maheshbabu28:


frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -99,13 +99,38 @@ export const ContactInput: React.FC<ContactInputProps> = ({
 
 const StyledDiv = styled.div`
   border-radius: 0.3rem;
-  padding: 12px 6px 6px 12px;
+  padding: 6px;
   min-height: 2.5em;
   background-image: none;
-  background-color: ${props => props.theme.css.filterBackgroundColor};
+  color: #495057;
```
FuriousLlama, line 105: we should use the colors in the theme. If there is no color that matches it please create one there. If there is only a small difference between the existing colors and the one specified we should chat about it. Otherwise we will get 100 shades of blue

FuriousLlama:
@Maheshbabu28 can you add a screen shot with the fix given this is a UI bug?

frontend/src/components/common/form/ContactInput.tsx
Maheshbabu28: Will look for the color in theme and try to use it.

Maheshbabu28:
> 

@FuriousLlama 
image

frontend/src/components/common/form/ContactInput.tsx
```diff
@@ -99,13 +99,38 @@ export const ContactInput: React.FC<ContactInputProps> = ({
...
 
   &.is-invalid {
     border: ${props => props.theme.css.dangerColor} solid 0.1rem;
   }
 `;
+
+export const StyledRemoveLinkButton = styled(LinkButton)`
+  &&.btn {
+    float: right;
+    color: #aaaaaa;
```
asanchezr, line 116: same comment re: hardcoded colors vs trying one from CSS theme

Maheshbabu28: removed and used from themes

frontend/src/components/common/form/ContactInput.tsx
Maheshbabu28: removed and used from themes

FuriousLlama (approved):

PR #2138, opened 2022-09-23
psp-4510 correct display of property name on activity property selector.

devinleighsmith:


frontend/src/features/properties/selector/utils.ts
```diff
@@ -37,17 +37,22 @@ export const getPropertyName = (property: IMapProperty): PropertyName => {
   return { label: NameSourceType.NONE, value: '' };
 };
 
-export const getResearchPropertyName = (
-  researchProperty?: Api_ResearchFileProperty,
+export const getFilePropertyName = (
+  fileProperty?: Api_PropertyFile,
+  skipName: boolean = false,
```
FuriousLlama, line 42: We really should try not to have special rules like these, otherwise we can get a lot of different variants of how to display a file property. The idea of this function is that the business said that they wanted standard displaying of properties.

devinleighsmith: What is your suggestion for this case then? basically in this case the business has asked for the identifier to display that is not the descriptive name (because on this screen the descriptive name will already be displayed if it exists).

devinleighsmith: @FuriousLlama

FuriousLlama: Earlier we had a bug logged in because the display did not match the logic (the bug was around the property selection vs the research property list differences). As a result we agreed that we were going to have an standard way of displaying property names. By adding special logic for special views we are breaking that contract.

I suggest that we talk to the BAs about adding special display logic. Specially given that is more gold plating IMO

FuriousLlama (approved):

PR #2143, opened 2022-09-23
Updated search of properties and improved aggregation of the data

FuriousLlama:


frontend/src/hooks/pims-api/useApiRequestWrapper.ts
```diff
@@ -6,18 +6,18 @@ import { useDispatch } from 'react-redux';
 import { handleAxiosResponse } from 'utils';
 
 export interface IResponseWrapper<
-  FunctionType extends (...args: any[]) => Promise<AxiosResponse<unknown | undefined>>
+  FunctionType extends (...args: any) => Promise<AxiosResponse<unknown | undefined>>
```
FuriousLlama, line 9: @devinleighsmith I did not not find it necessary to make the params an array. Was there a strong reason why this was an any[] before?

frontend/src/hooks/useLayerQueryProvider.ts
```diff
@@ -0,0 +1,68 @@
+import { AxiosResponse } from 'axios';
+import { useCallback } from 'react';
+import { toast } from 'react-toastify';
+
+import { useApiGeocoder } from './pims-api/useApiGeocoder';
+import { useApiRequestWrapper } from './pims-api/useApiRequestWrapper';
+import { IGeocoderPidsResponse, IGeocoderResponse } from './useApi';
+
+export const useGeocoderRepository = () => {
+  const { getSitePidsApi, searchAddressApi, getNearestToPointApi } = useApiGeocoder();
```
asanchezr, line 10: This file looks like a duplicate of `useGeocoderRepository.ts` ...

FuriousLlama: You are right. I actually used that on the container

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.tsx
```diff
@@ -66,16 +77,25 @@ export const PropertySelectorSearchContainer: React.FunctionComponent<IPropertyS
...
       // match the region and district for all found properties
       if (result?.features?.length !== undefined && result?.features?.length <= 15) {
-        await Promise.all(
-          foundProperties.map(p => matchRegionAndDistrict(p, regionService, districtService)),
+        var matchTask = foundProperties.map(p =>
+          matchRegionAndDistrict(p, findOneWhereContainsRegion, findOneWhereContainsDistrict),
         );
 
-        await Promise.all(foundProperties.map(p => matchAddress(p, getNearestToPoint)));
+        var getAddressTasks = foundProperties.map(p => getPropertyAddress(p, getNearestToPoint));
```
devinleighsmith, line 92: nit: might be able to increase performance here by using a promise chain on each getPropertyAddress call instead of getting all addresses in a loop and then looping again through the responses.

frontend/src/hooks/pims-api/useApiRequestWrapper.ts
devinleighsmith: to me it indicates an expectation of multiple params instead of just one. essentially args is an array of 0 or more elements

frontend/src/hooks/pims-api/useApiRequestWrapper.ts
```diff
@@ -107,7 +107,8 @@ export const useApiRequestWrapper = <
 
   useEffect(() => {
     if (invoke) {
-      wrappedApiRequest();
+      // If invoke is set then the function is parameterless.
+      wrappedApiRequest(undefined as Parameters<FunctionType>);
```
devinleighsmith, line 111: this seems undesirable to me calling a function with undefined != calling with no parameters

frontend/src/hooks/useGeocoderRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+  const { execute: executeSitePids, loading: isLoadingSitePids } = useApiRequestWrapper<
+    (siteId: string) => Promise<AxiosResponse<IGeocoderPidsResponse>>
+  >({
+    requestFunction: useCallback(async (siteId: string) => await getSitePidsApi(siteId), [
+      getSitePidsApi,
+    ]),
+    requestName: 'getSitePids',
+    onError: useCallback(axiosError => {
+      toast.error(
+        `Failed to get PIMS property data. error from backend: ${axiosError?.response?.data.error}`,
```
devinleighsmith, line 22: I wonder if we should indicate a failure in geocoder here or if that would confuse the users?

devinleighsmith (approved):

frontend/src/features/properties/selector/search/PropertySelectorSearchContainer.tsx
FuriousLlama: we could revisit it. At this point I rather just wrap up the bug and do some more research on the chaining later.

frontend/src/hooks/pims-api/useApiRequestWrapper.ts
FuriousLlama: It only needs to happen here. In other cases the function parameters need to match the specified FunctionType, otherwise, it will not compile. In this case, the FunctionType is still generic so there is no way to know if it should work

frontend/src/hooks/useGeocoderRepository.ts
FuriousLlama: That is something we will need to check with the BA team. At some point, the users will need to report errors to someone. 
"What is the best way to show those so that the issues can be found quickly?" answering that will give us a clear idea of how to proceed.

PR #2144, opened 2022-09-23
PSP-4376: Create ACQ - Default region and give message when removing …

Maheshbabu28:
…staff

### DIFF
--- a/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
+++ b/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
@@ -28,7 +28,7 @@ export const AcquisitionProperties: React.FunctionComponent<AcquisitionPropertie
             <Col>
               <MapSelectorContainer
                 onSelectedProperty={(newProperty: IMapProperty) => {
-                  const formProperty = AcquisitionPropertyForm.fromMapProperty(newProperty);
+                  const formProperty = PropertyForm.fromMapProperty(newProperty);
                   if (values.properties?.length === 0) {
                     formikProps.setFieldValue(`region`, formProperty.region);
                   }
--- a/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
+++ b/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
@@ -1,12 +1,18 @@
 import MapSelectorContainer from 'features/properties/selector/MapSelectorContainer';
 import { IMapProperty } from 'features/properties/selector/models';
-import { FieldArray, useFormikContext } from 'formik';
+import { FieldArray, FormikProps, useFormikContext } from 'formik';
 import { Col, Row } from 'react-bootstrap';
 
 import { PropertyForm } from '../../shared/models';
 import { AcquisitionForm } from './models';
 
-export const AcquisitionProperties: React.FunctionComponent = () => {
+interface AcquisitionPropertiesProp {
+  formikProps: FormikProps<AcquisitionForm>;
+}
+
+export const AcquisitionProperties: React.FunctionComponent<AcquisitionPropertiesProp> = ({
+  formikProps,
+}) => {
   const { values } = useFormikContext<AcquisitionForm>();
 
   return (
@@ -24,6 +30,9 @@ export const AcquisitionProperties: React.FunctionComponent = () => {
                 onSelectedProperty={(newProperty: IMapProperty) => {
                   const formProperty = PropertyForm.fromMapProperty(newProperty);
                   push(formProperty);
+                  if (values.properties?.length === 0) {
+                    formikProps.setFieldValue(`region`, formProperty.region);
+                  }
                 }}
                 existingProperties={values.properties}
                 onRemoveProperty={remove}
--- a/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
+++ b/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
@@ -1,6 +1,6 @@
 import MapSelectorContainer from 'features/properties/selector/MapSelectorContainer';
 import { IMapProperty } from 'features/properties/selector/models';
-import { FieldArray, FormikProps, useFormikContext } from 'formik';
+import { FieldArray, FormikProps } from 'formik';
 import { Col, Row } from 'react-bootstrap';
 
 import { PropertyForm } from '../../shared/models';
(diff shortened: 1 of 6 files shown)


Maheshbabu28:
image

image

frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
```diff
@@ -1,11 +1,17 @@
...
 
-export const AcquisitionProperties: React.FunctionComponent = () => {
+interface AcquisitionPropertiesProp {
+  formikProps: FormikProps<AcquisitionForm>;
+}
+
+export const AcquisitionProperties: React.FunctionComponent<AcquisitionPropertiesProp> = ({
+  formikProps,
+}) => {
   const { values } = useFormikContext<AcquisitionForm>();
```
devinleighsmith, line 15: @Maheshbabu28 please either pass the formik props or use useFormikContext, not both.

frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -97,6 +119,21 @@ export const AddAcquisitionForm = React.forwardRef<
             when={formikProps.dirty && formikProps.submitCount === 0}
             message="You have made changes on this form. Do you wish to leave without saving?"
           />
+
+          <AcquisitionFormModal
+            message="Selected Ministry region is different from that of one or more selected properties. Do you wish to continue?"
+            title="Different Ministry region"
+            display={showDiffMinistryRegionModal}
+            handleOk={() => {
+              setIgnoreRegionWarning(true);
```
devinleighsmith, line 128: I'm not convinced this is safe. Essentially what you are doing here is
1) starting a new render because of a state change
2) starting another new render because of a state change
3) submitting the formik form.

I'm not convinced that there are any hard guarantees here that the above two renders will take place before the form submission which may result in a race condition. What I would recommend instead is that rather then using the modal to resubmit the form, store the form values either in a scoped arrow function or in state, and then submit those form values when the modal is confirmed.

devinleighsmith (changes requested):
changes to modal logic required

frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -34,14 +35,35 @@ export const AddAcquisitionForm = React.forwardRef<
...
+  const [ignoreRegionWarning, setIgnoreRegionWarning] = React.useState<boolean>(false);
+  const isMinistryRegionDiff = (values: AcquisitionForm): boolean => {
+    const selectedPropRegions = values.properties.map(x => x.region);
+    return (
+      (selectedPropRegions.length &&
+        (selectedPropRegions.indexOf(Number(values.region)) === -1 ||
+          !selectedPropRegions.every(e => e === selectedPropRegions[0]))) ||
+      false
+    );
+  };
```
asanchezr, line 50: @devinleighsmith and @Maheshbabu28 - can we double check this is the intended logic as laid out in Confluence. I recall that BAs mentioned that the region check should be against the 1st property (not all). If that's the case then this logic could be simplified/reworked. Comments?

frontend/src/features/properties/map/acquisition/add/AddAcquisitionTeamForm.tsx
```diff
@@ -67,9 +69,21 @@ export const AddAcquisitionTeamForm: React.FunctionComponent<AddAcquisitionTeamF
...
+                      setShowRemoveTeamMemberModal(true);
                     }}
                   ></Button>
+                  <AcquisitionFormModal
+                    message="Are you sure you want to remove this row?"
+                    title="Remove Team Member"
+                    display={showRemoveTeamMemberModal}
+                    handleOk={() => {
+                      arrayHelpers.remove(index);
+                      setShowRemoveTeamMemberModal(false);
```
asanchezr, line 81: This comes down to dev preference but I personally see low value in creating an extra class for a modal that is never going to be shared. That is, this could be refactored to just use the base **GenericModal** and pass "Ok" and "Cancel" as buttons text. Thoughts?

frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
```diff
@@ -23,6 +29,9 @@ export const AcquisitionProperties: React.FunctionComponent = () => {
                 onSelectedProperty={(newProperty: IMapProperty) => {
                   const formProperty = AcquisitionPropertyForm.fromMapProperty(newProperty);
                   push(formProperty);
+                  if (values.properties?.length === 0) {
+                    formikProps.setFieldValue(`region`, formProperty.region);
+                  }
```
asanchezr, line 34: Can we double check this logic as well? The **push()** call in line 31 should modify `values.properties` by inserting the new property so I find it odd that we could ever check for `values.properties.length === 0` the line after...

asanchezr: unless the update occurs asynchronously and then we risk a race condition

frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
Maheshbabu28: image
As per confluence this is for multiple property(ies)

frontend/src/features/properties/map/acquisition/add/AddAcquisitionTeamForm.tsx
Maheshbabu28: This is the pattern used across application so followed the same. I was more inclined to use the GenericModal but did'nt want to deviate from existing norm. Please let me know your preference.

frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
Maheshbabu28: We want to use region of only first property to be set in ministry region so the check. The push() will happen every time but can be moved to after the check and setting the region.

frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
Maheshbabu28: I over complicated it, we have to any way hide the modal using the state which will start a render. But I have removed the "ignoreRegionWarning" state change as it was unnecessary and directly used the onSubmit call. Form will submitted only once in this case.

frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
Maheshbabu28: done

devinleighsmith (approved):

PR #2214, opened 2022-09-30
psp-4505 add claims for document generation.

devinleighsmith:


asanchezr (approved):

FuriousLlama (changes requested):
Will need to be the added to 38_01 build

devinleighsmith:
> Will need to be the added to 38_01 build

Will do, I need to make some minor updates here based on requirements confirmation with the BAs.

source/backend/dal/Security/Permissions.cs
```diff
@@ -123,16 +123,19 @@ public enum Permissions
         [Display(GroupName = "document", Name = "document-admin", Description = "Can perform admin functions on documents.")]
         DocumentAdmin = 39,
 
+        [Display(GroupName = "document", Name = "generate-documents", Description = "Can generate documents.")]
+        GenerateDocuments = 40,
```
FuriousLlama, line 127: Also, we need one for an Admin to add template documents.

devinleighsmith: I think for the story for that one I just said that simply checking for the admin role is good enough for adding template documents. I don't see a case where another user type would be given access to that page.

devinleighsmith:
This one is on hold until 38.01 is merged.

source/backend/dal/Security/Permissions.cs
FuriousLlama: Do you mean what is stated in the story? It mentions that the user has the role of Admin. There is no mention of the claims. For synchronization with Mayan, there is document-admin permission that we could use if we want to group both admin responsibilities together.

devinleighsmith: yep, My idea is that we wouldn't use a claim at all and just use role membership.

devinleighsmith:
@FuriousLlama given the decision we came to on the document admin claim, can you remove the changes requested flag? (once I fix the conflict of course).

FuriousLlama (approved):

PR #2223, opened 2022-10-04
PSP-4131: Auto capture Activity Status Change as a Note

Maheshbabu28:


source/backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -91,14 +93,34 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
...
 
-        public PimsActivityInstance Update(PimsActivityInstance instance)
+        public Tuple<PimsActivityInstance, PimsActivityInstanceNote> Update(PimsActivityInstance instance)
         {
             instance.ThrowIfNull(nameof(instance));
             var currentActivity = this.Context.PimsActivityInstances
                 .FirstOrDefault(x => x.ActivityInstanceId == instance.Id) ?? throw new KeyNotFoundException();
+
+            PimsActivityInstanceNote note = null;
+            if (currentActivity.ActivityInstanceStatusTypeCode!=instance.ActivityInstanceStatusTypeCode && this.User.HasPermission(Permissions.NoteAdd))
```
asanchezr, line 103: nit: fix spacing around `!=`

asanchezr: Also, come to think about - not sure about the permission check for current user. I believe it's the "system" creating a note automatically so it would be irrelevant if the user has claims for notes or not. What are your thoughts on this @devinleighsmith ?

My understanding is that we would want to log notes per status change regardless of the claims of the user who triggered this change - is it not?

source/backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -91,14 +93,34 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
...
+                    AppCreateTimestamp = System.DateTime.Now,
+                    AppCreateUserid = instance.AppCreateUserid,
+                    Note = new PimsNote()
+                    {
+                        IsSystemGenerated = true,
+                        NoteTxt = $"Activity status changed from { currentActivity.ActivityInstanceStatusTypeCode } to { instance.ActivityInstanceStatusTypeCode }",
+                        AppCreateTimestamp = System.DateTime.Now,
+                        AppCreateUserid = instance.AppCreateUserid,
+                    },
+                };
```
asanchezr, line 118: Also want to double check what should be set for `AppCreateUserid`. You are setting it to the user that originally created the activity - which might not be the same as the current user making a status change, right?

The AC on JIRA says:

```
For the note: 
Record the audit fields same as for any other note (created by/on)
 - Created by - user who did the status change
 - Created on - the date on which the change happens
```

So I think this would record the wrong user. Say Bob created the activity but then Mary changes the status to "Archived". With the current logic the note would show as created by Bob when Mary was the one that did the actual change

asanchezr: I think `AppCreateUserid` should be set to `this.User.Userid` to reflect who made the change...

source/backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -91,14 +93,34 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
             return instance;
         }
 
-        public PimsActivityInstance Update(PimsActivityInstance instance)
+        public Tuple<PimsActivityInstance, PimsActivityInstanceNote> Update(PimsActivityInstance instance)
```
devinleighsmith, line 96: yeah, I'm not a big fan of this. I think the service should be responsible for updating the activity and then also adding a note (in one transaction).

Creating repository methods that combine operations like this mean we can't compose individual operations anymore. Ie: it is no longer possible to update an activity without generating a note.

Also in general repositories should not affect external tables.

Thankfully this shouldn't be a big change. There is already a repository to add a note, so you can just revert this method and make the service responsible for composing the update activity and add note repository calls.

devinleighsmith (changes requested):
see comments

source/backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: Yes, in this case as long as the user has the ability to update an activity the system should be able to generate the note, regardless of their note permissions.

source/backend/dal/Repositories/ActivityRepository.cs
```diff
@@ -91,14 +93,34 @@ public PimsActivityInstance Add(PimsActivityInstance instance)
...
         {
             instance.ThrowIfNull(nameof(instance));
             var currentActivity = this.Context.PimsActivityInstances
                 .FirstOrDefault(x => x.ActivityInstanceId == instance.Id) ?? throw new KeyNotFoundException();
+
+            PimsActivityInstanceNote note = null;
+            if (currentActivity.ActivityInstanceStatusTypeCode!=instance.ActivityInstanceStatusTypeCode && this.User.HasPermission(Permissions.NoteAdd))
+            {
+                // add note
+                note = new PimsActivityInstanceNote()
```
devinleighsmith, line 106: as stated above, the note should not be created here. In fact, I don't understand why the note is being created here since it is not being saved to the db. It looks like is is created here just so it can be passed back to the service, but why not just create it in the service instead?

source/backend/dal/Repositories/ActivityRepository.cs
Maheshbabu28: @devinleighsmith this was a precondition in the story that user needs to have a Create permission on the note.
I will remove it as per above discussion.

source/backend/dal/Repositories/ActivityRepository.cs
Maheshbabu28: The repository is only responsible for providing a note instance as the decision to whether create a note or not depends on the current activity status and updated activity status which is only available in  repository. 
In the code the service is only responsible for updating the activity and is also adding a note (in one transaction) using the entity note repo.

Please check the ActivityService code and let me know if still needs changes.

source/backend/dal/Repositories/ActivityRepository.cs
Maheshbabu28: makes sense, using this.User.GetUsername() to capture current user trying to update.

source/backend/dal/Repositories/ActivityRepository.cs
Maheshbabu28: As the decision to create a note depends on earlier status and status changed by the user which is only known by the repository and not service.

asanchezr: @Maheshbabu28 as an alternative you could implement a method in the repository that validates if an activity status is about to be changed. That way the service could call it and also be responsible of deciding if a note is required. You can see this pattern being used in **AcquisitionFileService** and repository where whenever a user tries to change the region of an acquisition file we ask the user for confirmation first (via an exception if the user didn't confirm). 

In your case you wouldn't be throwing an exception but instead creating a note if the repo tells the service that the status of the incoming activity doesn't match the one stored in the DB

Maheshbabu28: Thanks @asanchezr  for detailed explanation, I have made the changes.

source/backend/dal/Repositories/ActivityRepository.cs
Maheshbabu28: changed as per feedback.

source/backend/dal/Repositories/ActivityRepository.cs
devinleighsmith: @Maheshbabu28 can we see those changes? I don't see them here.

Maheshbabu28: @devinleighsmith https://github.com/bcgov/PSP/pull/2223/commits/20efbcb1d9e8698cda64117683ef00164f054114

source/backend/api/Services/ActivityService.cs
```diff
@@ -187,5 +192,27 @@ private void ValidateVersion(long activityId, long activityVersion)
...
+            {
+                PimsActivityInstanceNote note = new PimsActivityInstanceNote()
+                {
+                    ActivityInstanceId = currentActivity.ActivityInstanceId,
+                    AppCreateTimestamp = System.DateTime.Now,
+                    AppCreateUserid = instance.AppCreateUserid,
+                    Note = new PimsNote()
+                    {
+                        IsSystemGenerated = true,
+                        NoteTxt = $"Activity status changed from { currentActivity.ActivityInstanceStatusTypeCode } to { instance.ActivityInstanceStatusTypeCode }",
```
devinleighsmith, line 209: @Maheshbabu28 this should be ActivityInstanceStatusTypeCodeNavigation.Description, so that we don't return an internal code representation when the user is expecting a human-readable representation.

devinleighsmith: @Maheshbabu28 this is the only change that is missing before this can be merged.

Maheshbabu28: @devinleighsmith  Done

devinleighsmith (approved):

PR #2224, opened 2022-10-05
PSP-4413 Acquisition file number create and update

asanchezr:
Also, PSP-4508 Remove unnecessary code when creating ACQ File

image

### DIFF
--- a/source/backend/dal/Repositories/AcquisitionFileRepository.cs
+++ b/source/backend/dal/Repositories/AcquisitionFileRepository.cs
@@ -35,32 +35,6 @@ namespace Pims.Dal.Repositories
         #region Methods
 
         /// <summary>
-        /// Generates a new Acquisition Number in the following format.
-        /// </summary>
-        /// <remarks>
-        /// <list type="bullet">
-        /// <item>
-        /// Format: YY-XXXXXXXXXXXX-ZZ
-        /// </item>
-        /// <item>
-        /// Prefix - (YY above) The prefix numbers for an Acquisition file correspond with the MoTI region
-        /// </item>
-        /// <item>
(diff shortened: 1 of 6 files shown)


source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -25,14 +25,41 @@ public class AcquisitionFileRepository : BaseRepository<PimsAcquisitionFile>, IA
...
+        /// File # - (XXXXX... above) Acquisition File number is created and each file number should increase in increments of 1.
+        /// The digit base number is unique to the file. Do not pad the number with zeros.
+        /// </item>
+        /// <item>
+        /// Suffix - (ZZ above) The suffix numbers for an Acquisition file defaults to 01.
+        /// </item>
+        /// </list>
+        /// </remarks>
+        /// <returns>The formatted Acquisition File Number.</returns>
+        public static string GenerateAcquisitionNumber(short prefix, long fileNumber, short suffix = 1)
```
devinleighsmith, line 58: interesting, I take it that we don't have a method of creating these with anything other then a suffix of 1 for the time being?

devinleighsmith (approved):

source/backend/dal/Repositories/AcquisitionFileRepository.cs
asanchezr: @devinleighsmith - yes as per the business rules listed in the method comments they want to use "01" as suffix for all acquisition files (at this time). I just future-proofed the method a bit.  If they want to change that in the future, it supports other suffices 

@FuriousLlama I'm fine making it private since no one is calling it outside of this class

PR #2228, opened 2022-10-06
Psp-3984 cleanup frontend to use Api_Property. Remove references to old IApiPropertyModel. Introduce a hook to facilitate loading composed properties.

devinleighsmith:


source/frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -339,22 +296,18 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+                      cluster.properties,
+                      latitude,
+                      longitude,
+                    );
+                    convertedProperty && onMarkerClick(convertedProperty); //open information slideout
+                    setCurrentSelected(convertedProperty);
+
                     selectedPropertyContext.setPropertyInfo(convertedProperty);
+                  } catch (err) {
+                    console.log(err);
```
FuriousLlama, line 310: should this be console.error?

source/frontend/src/features/mapSideBar/MotiInventoryContainer.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('MotiInventoryContainer component', () => {
 
   it('shows the property information tab for inventory properties', async () => {
     mockAxios.onPost().reply(200, {});
-    mockAxios.onGet(new RegExp('/properties/*')).reply(200, { id: 9212434 });
+    mockAxios.onGet(new RegExp('/properties/concept/*')).reply(200, { id: 9212434 });
```
FuriousLlama, line 79: if we are cleaning up the properties API we could also remove the '/concept' part for properties given we would not need that differentiation anymore

source/frontend/src/models/api/Property.ts
```diff
@@ -43,10 +43,12 @@ export interface Api_Property extends Api_ConcurrentVersion, Api_AuditFields {
 
   isSensitive?: boolean;
   isVisibleToOtherAgencies?: boolean;
+  isProvincialPublicHwy?: boolean;
```
FuriousLlama, line 46: Good to have the complete set of members on the front end too

source/frontend/src/models/api/Property.ts
```diff
@@ -43,10 +43,12 @@ export interface Api_Property extends Api_ConcurrentVersion, Api_AuditFields {
...
   isOwned?: boolean;
   isPropertyOfInterest?: boolean;
   pphStatusTypeCode?: string;
   isRwyBeltDomPatent?: boolean;
+  isALR?: boolean;
```
FuriousLlama, line 51: I cant find this field on the backend model. Lets keep the Api_* models a 1:1 to the backend

FuriousLlama (changes requested):
Cannot find 'isALR' on the backend model. Is that a calculated field?

asanchezr:
> Cannot find 'isALR' on the backend model. Is that a calculated field?

Yes, this is populated from one of the map layer queries

source/frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: it shouldn't be here at all really, was added for debugging purposes. Will remove.

devinleighsmith:
@FuriousLlama anything else on this one or can you clear your changes requested flag?

FuriousLlama (approved):

PR #2232, opened 2022-10-06
correct test api/keycloak settings.

devinleighsmith:
now that test is hosted in the dev.oidc keycloak, update relevant settings to point to the updated location.

### DIFF
--- a/tools/keycloak/sync/realm.test.json
+++ b/tools/keycloak/sync/realm.test.json
@@ -5,7 +5,7 @@
     "DisplayNameHtml": "<a>PIMS</a>",
     "Clients": [
       {
-        "ClientId": "pims-api",
+        "ClientId": "pims-api-test",
         "Name": "PIMS API",
         "Description": "PIMS API authentication client",
         "Enabled": true,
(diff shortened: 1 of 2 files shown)


FuriousLlama (approved):

tools/keycloak/sync/realm.test.json
```diff
@@ -39,15 +39,15 @@
         "ImplicitFlowEnabled": false,
         "DirectAccessGrantsEnabled": true,
         "AuthorizationServicesEnabled": false,
-        "RootUrl": "https://pims-app-test-3cd915-dev.apps.silver.devops.gov.bc.ca",
+        "RootUrl": "https://pims-app-3cd915-dev.apps.silver.devops.gov.bc.ca",
         "RedirectUris": [
-          "https://pims-app-test-3cd915-dev.apps.silver.devops.gov.bc.ca/*",
-          "https://tst-pims.th.gov.bc.ca/*"
+          "https://pims-app-3cd915-dev.apps.silver.devops.gov.bc.ca/*",
+          "https://dev-pims.th.gov.bc.ca/*"
```
asanchezr, line 45: @devinleighsmith just confirming that we want to point the TEST kc realm to the DEV api instance? It would seem that only the service account needs to be changed...

PR #2278, opened 2022-10-13
Increment DEV  version to IS39

asanchezr:


source/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "2.2.0-38.21",
```
FuriousLlama, line 3: package-lock file needs also to be updated

asanchezr: updated

devinleighsmith (approved):

PR #2297, opened 2022-10-18
Psp 4495 fix property summary display of address.

devinleighsmith:


source/frontend/src/utils/propertyUtils.ts
```diff
@@ -81,6 +81,20 @@ export const formatApiAddress = (address?: Api_Address) => {
...
+/**
+ * Provides a formatted address as a string.
+ * @param address Address object from property.
+ * @returns Civic address string value.
+ */
+export const formatApiSummaryAddress = (address?: Api_Address) => {
+  const values = [
+    address?.streetAddress1 ?? '',
+    address?.streetAddress2 ?? '',
+    address?.streetAddress3 ?? '',
```
asanchezr, line 93: This method seems to be replacing **formatApiAddress** - any reason to create a new method instead of updating the old one?

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -82,6 +82,9 @@ public PimsAcquisitionFile GetById(long id)
                 .Include(r => r.PimsPropertyAcquisitionFiles)
                     .ThenInclude(rp => rp.Property)
                     .ThenInclude(p => p.DistrictCodeNavigation)
+                .Include(r => r.PimsPropertyAcquisitionFiles)
+                    .ThenInclude(rp => rp.Property)
```
FuriousLlama, line 86: These look like a good place for using the property API and concatenate the results. That way we would have a central location for retrieving property data without having to modify multiple quieries

source/frontend/src/features/properties/selector/utils.ts
```diff
@@ -1,7 +1,9 @@
 import { IMapProperty } from 'features/properties/selector/models';
 import { compact } from 'lodash';
 import { Api_PropertyFile } from 'models/api/PropertyFile';
-import { formatApiAddress, pidFormatter } from 'utils';
+import { pidFormatter } from 'utils';
+
+import { formatApiSummaryAddress } from './../../../utils/propertyUtils';
```
FuriousLlama, line 6: Nit : we should get a root folder character and go from there instead of relative navigation for things like this

source/frontend/src/utils/propertyUtils.ts
FuriousLlama: Agree, the more formats we have the more likely we are to make a mistake

source/backend/dal/Repositories/AcquisitionFileRepository.cs
devinleighsmith: I'll log tech debt for that, but I'm not comfortable doing that for this bugfix since this repository was already pulling the property before this bugfix.

source/frontend/src/features/properties/selector/utils.ts
devinleighsmith: agreed, my auto importer likely did this, will correct.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
devinleighsmith: https://jira.th.gov.bc.ca/browse/PSP-4741

asanchezr (approved):

PR #2299, opened 2022-10-18
psp-4672 correct POI legend label.

devinleighsmith:


source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -56,7 +56,7 @@ export const parcelIconSelect = L.icon({
  * @param properties
  */
 export const createPoints = (properties: IProperty[], type: string = 'Point') =>
-  properties.map(x => {
+  properties.map((x) => {
```
asanchezr, line 59: Please fix lint error above
image

asanchezr (approved):
Approved but need to fix linting errors

PR #2311, opened 2022-10-20
clamav auto update template, instructions.

devinleighsmith:


devinleighsmith:
Some context for the reviewers, the clamav image that we are using is provided by the lab. It automatically updates its virus definitions on startup. However, by default the image will startup once and will not restart unless a crash or manual restart occurs. This change introduces a new cronjob that boots a pod that has the oc-cli installed. That pod will trigger a rollout in the clamav pod.

asanchezr (approved):

PR #2313, opened 2022-10-21
Updated area and volume calculators

FuriousLlama:
image

image


source/frontend/src/components/measurements/AreaContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  isEdditable?: never;
+  onChange?: never;
+}
+
+interface EditableProps {
+  landArea?: number;
+  unitCode?: string;
+
+  isEdditable: boolean;
```
devinleighsmith, line 16: Edditable

source/frontend/src/components/measurements/AreaContainer.tsx
```diff
@@ -0,0 +1,37 @@
+import { AreaForm } from './AreaForm';
+import AreaView from './AreaView';
+
+interface ViewOnlyProps {
+  landArea?: number;
+  unitCode?: string;
+
+  isEdditable?: never;
```
devinleighsmith, line 8: Edditable

source/frontend/src/components/measurements/AreaContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  landArea?: number;
+  unitCode?: string;
+
+  isEdditable: boolean;
+  onChange: (landArea: number, areaUnitTypeCode: string) => void;
+}
+
+type IAreaContainerProps = EditableProps | ViewOnlyProps;
+
+function isViewOnly(areaComponent: ViewOnlyProps | EditableProps): areaComponent is ViewOnlyProps {
```
devinleighsmith, line 22: Is using a cast in the function definition necessary? (areaComponent is ViewOnlyProps).

source/frontend/src/components/measurements/AreaContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  var editable = areaComponent as EditableProps;
```
devinleighsmith, line 23: Something funky is going on, why allow both ViewOnlyProps and EditableProps but then force the type to be an EditableProps via a cast?

source/frontend/src/components/measurements/styles.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const StyledGreenCol = styled.div`
+  max-width: 25rem;
+  padding: 1rem;
+  background-color: ${props => props.theme.css.selectedColor};
+  border: 1px solid ${props => props.theme.css.completedColor};
+  border-radius: 0.5rem;
+`;
+
+export const StyledGreenGrey = styled.div`
```
devinleighsmith, line 11: Where does the 'green' come into this? isn't this grey and blue?

source/frontend/src/components/measurements/VolumeView.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <Row>
+            <Col className="text-right">{formatNumber(meters)}</Col>
+            <Col>
+              <span>
+                metres<sup>3</sup>
+              </span>
+            </Col>
+          </Row>
+          <Row>
+            <Col className="text-right">{formatNumber(feets)}</Col>
```
devinleighsmith, line 29: nit: feets vs feet. the singular of feet is foot, the plural is feet. ie: 1 foot 3 feet

source/frontend/src/components/measurements/VolumeContainer.tsx
```diff
@@ -0,0 +1,43 @@
...
+type IVolumeContainerProps = EditableProps | ViewOnlyProps;
+
+function isViewOnly(areaComponent: ViewOnlyProps | EditableProps): areaComponent is ViewOnlyProps {
+  var editable = areaComponent as EditableProps;
+  return editable.onChange === undefined || editable.isEdditable === false;
+}
+
+const VolumeContainer: React.FunctionComponent<IVolumeContainerProps> = props => {
+  if (isViewOnly(props)) {
+    return <VolumeView volume={props.volumetricMeasurement} unitCode={props.volumetricUnit} />;
```
devinleighsmith, line 31: Good usage of view/form. This is easy to read so I'm glad we are doing this going forward.

devinleighsmith (approved):

source/frontend/src/components/measurements/AreaContainer.tsx
asanchezr: The way I understand it, I believe this is an (advanced) feature of typescript called type narrowing where the function makes an assertion of the end type based on the shape of the object passed in as a parameter. A more contrived example would be:

```tsx
function isFish(pet: Fish | Bird): pet is Fish {
  return (pet as Fish).swim !== undefined;
}
``` 

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

source/frontend/src/components/measurements/AreaForm.tsx
```diff
@@ -1,17 +1,17 @@
...
 
-import { StyledTable } from '../styles';
+import { StyledGreenCol, StyledGreenGrey, StyledInput } from './styles';
 
-export interface IUpdateLandMeasurementEditTableProps {
+export interface IAreaFormProps {
   area?: number;
   areaUnitTypeCode?: string;
-  onChange?: (landArea: number, areaUnitTypeCode: string) => void;
+  onChange: (landArea: number, areaUnitTypeCode: string) => void;
```
asanchezr, line 11: Not a big deal but this change is forcing the consumer to **always** pass an _onChange_ event handler. There might be cases where the consumer does not care about the changes, it only wants to render the table

asanchezr (approved):

source/frontend/src/components/measurements/AreaContainer.tsx
FuriousLlama: Good point. no need for the extra casting

source/frontend/src/components/measurements/AreaForm.tsx
FuriousLlama: I think this is better given the parent using the container can just pass an empty function. Otherwise the child needs to know how it could be used instead of assuming the parent will figure out

PR #2314, opened 2022-10-21
PSP-4408: Deprecate "useApi" hook

Maheshbabu28:


source/frontend/src/components/maps/leaflet/Map.test.tsx
```diff
@@ -157,7 +156,7 @@ xdescribe('MapProperties View', () => {
     );
     mockGetParcel = jest.fn(async () => ({} as IProperty));
 
-    (useApi as unknown as jest.Mock<Partial<typeof useApi>>).mockReturnValue({
+    (useApiGeocoder as unknown as jest.Mock<Partial<typeof useApiGeocoder>>).mockReturnValue({
```
devinleighsmith, line 159: this seems a bit odd since loadProperties and getProperty are not something that was returned from useApi in the past. Perhaps this isn't being used?

devinleighsmith (approved):

source/frontend/src/components/maps/leaflet/Map.test.tsx
Maheshbabu28: Yes looks like, all the test cases in this file are also skipped.

PR #2326, opened 2022-10-26
Fix map markers in dev

devinleighsmith:


devinleighsmith:
Looks like the format of requires changed with the react-scripts PR. only svgs require the .default after the require.

FuriousLlama (approved):

asanchezr (approved):
This branch is out of date, please update it

devinleighsmith:
image

PR #2367, opened 2022-11-02
PSP-4183: BVT: From the Property List view, should be able to select …

Maheshbabu28:
…and access a property


Maheshbabu28:
image

source/frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -53,7 +52,19 @@ const PropertyListView: React.FC = () => {
...
-
+  const history = useHistory();
+  const onShowPropertyInfo = useCallback(
+    (property: IProperty, inNewTab: boolean) => {
+      !inNewTab && history.push(`/mapview/sidebar/property/${property.id}?pid=${property.pid}`);
+      inNewTab &&
+        window.open(`/mapview/sidebar/property/${property.id}?pid=${property.pid}`, '_blank');
+    },
+    [history],
+  );
```
asanchezr, line 63: This implementation is correct - however I would recommend a simpler approach where no callback is needed. Look at how **LeaseSearchResult** was implemented for an existing pattern on links for each table row. Example:

```tsx
// in colums.tsx
Cell: (props: CellProps<ILeaseSearchResult>) => (
      <Link to={`/lease/${props.row.original.id}`}>{props.row.original.lFileNo}</Link>
    ),
```

Notice that you already have the current property in context for each table row, no need to pass that to a callback function. Inlining a simple router **<Link ... />** is sufficient

asanchezr: As for the link to open in a new tab:

```tsx
<Link to="route" target="_blank" rel="noopener noreferrer" />
```

source/frontend/src/features/properties/list/columns.tsx
```diff
@@ -18,59 +23,108 @@ const NumberCell = ({ cell: { value } }: CellProps<IProperty, number>) => format
...
+        return (
+          <StyledDiv>
+            {hasClaim(Claims.PROPERTY_VIEW) && (
+              <StyledIconButton
+                data-testid="view-prop-tab"
+                title="View Property Info"
+                variant="light"
+                onClick={() => {
+                  onShowPropertyInfo(cellProps.row.original, false);
+                }}
```
asanchezr, line 102: See my previous comment re: using router links instead of callback functions here...

source/frontend/src/features/properties/list/PropertyListView.tsx
Maheshbabu28: yes this is much better, changed.

source/frontend/src/features/properties/list/columns.tsx
```diff
@@ -73,4 +79,47 @@ export const columns = ({ municipalities }: Props): ColumnWithProps<IProperty>[]
...
+          {hasClaim(Claims.PROPERTY_VIEW) && (
+            <StyledIconButton data-testid="view-prop-ext" title="View Property" variant="light">
+              <Link
+                to={`/lease/${cellProps.row.original.id}`}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                <FaExternalLinkAlt />
+              </Link>
+            </StyledIconButton>
```
asanchezr, line 114: Are we sure these are the right links? Older code had different urls...

image

Maheshbabu28: my bad got confused with other story, this is corrected now. Thank you!

source/frontend/src/features/properties/list/columns.tsx
```diff
@@ -73,4 +79,49 @@ export const columns = ({ municipalities }: Props): ColumnWithProps<IProperty>[]
...
+      return (
+        <StyledDiv>
+          {hasClaim(Claims.PROPERTY_VIEW) && (
+            <StyledIconButton
+              data-testid="view-prop-tab"
+              title="View Property Info"
+              variant="light"
+            >
+              <Link
+                to={`/mapview/sidebar/property/${cellProps.row.original.id}?pid=${cellProps.row.original.pid}`}
```
devinleighsmith, line 100: @Maheshbabu28 this isn't working currently as the side bar is expecting a pid that does not contain '-' characters.
Either update the sidebar code to format the pid as necessary or update your code to send the pid in the expected format.

devinleighsmith (changes requested):

source/frontend/src/features/properties/list/columns.tsx
Maheshbabu28: Fixed

devinleighsmith (approved):

PR #2369, opened 2022-11-02
psp-4829 The FA parcel map layer does not require the Siteminder cookie anymor…

devinleighsmith:
…e, as we are using the public layer. This should fix the "header too large" error being returned by the PIMS geoserver in DEV.


asanchezr (approved):

devinleighsmith:
This is the error in the dev geoserver: 
image

PR #2383, opened 2022-11-04
Lease Documents | psp 4712

FuriousLlama:
image


source/frontend/src/features/leases/detail/LeaseContainer.tsx
```diff
@@ -40,6 +41,7 @@ export interface ILeasePage {
   header?: ReactNode;
   description?: string;
   validation?: Yup.ObjectSchema<any>;
+  claims?: string[] | string;
```
devinleighsmith, line 44: string [] or string? if you had just one entry wouldn't [claim] make sense?

devinleighsmith (approved):

source/frontend/src/features/leases/detail/LeaseContainer.tsx
FuriousLlama: I agree, I just made it match the router claims type

PR #2387, opened 2022-11-04
PSP-3988 Upgrade to React 18, Leaflet to 1.9

asanchezr:


source/frontend/config-overrides.js
```diff
@@ -14,11 +14,19 @@ const addCompressionPlugin = config => {
...
+    config.resolve.fallback = fallback;
+    addCompressionPlugin(config);
+    config.ignoreWarnings = [/Failed to parse source map/, /autoprefixer/];
+    return config;
+  },
+  // The Jest config to use when running your jest tests
+  jest: function override(config) {
+    config.transformIgnorePatterns = ['node_modules/(?!(react-leaflet|@react-leaflet/core)/)'];
+    return config;
+  },
```
asanchezr, line 31: This change is required to make jest aware that the new react-leaflet library has no CommonJS exports (like most libraries out there)

Not sure why react-leaflet would choose to only release in the newer ES module (ESM) format that is not as widely supported by existing tooling. But we don't control how the library author publishes his library

source/frontend/package.json
```diff
@@ -141,6 +133,9 @@
   "jest": {
     "resetMocks": false,
     "clearMocks": true,
+    "moduleNameMapper": {
+      "\\.(css|less|sass|scss)$": "<rootDir>/src/mocks/styleMock.js"
+    },
```
asanchezr, line 138: I had to add this configuration to avoid Jest getting stuck with a few of our files that import css or scss directly; ie

```tsx
// fake example
import 'button.css';

export class Button {
...
}
```

We still have a few of these. Why it wasn't an issue before? beats me

source/frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -41,7 +41,11 @@ const InputContainer = styled('div')`
 /**
  * Component to display column filter component, used internally by the table
  */
-const ColumnFilter: React.FC<IColumnFilterProps> = ({ column, onFilter, children }) => {
+const ColumnFilter: React.FC<React.PropsWithChildren<IColumnFilterProps>> = ({
```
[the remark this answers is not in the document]
asanchezr: https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc

source/frontend/src/components/Table/DateCell.tsx
```diff
@@ -1,7 +1,8 @@
...
+import { prettyFormatDate, prettyFormatDateTime, stringToFragment } from 'utils';
 
-export const DateCell = ({ cell: { value } }: CellProps<Date, string>) => prettyFormatDate(value);
+export const DateCell = ({ cell: { value } }: CellProps<any, string | Date | undefined>) =>
+  stringToFragment(prettyFormatDate(value));
 
-export const DateTimeCell = ({ cell: { value } }: CellProps<Date, string>) =>
-  prettyFormatDateTime(value);
+export const DateTimeCell = ({ cell: { value } }: CellProps<any, string | Date | undefined>) =>
+  stringToFragment(prettyFormatDateTime(value));
```
asanchezr, line 8: Upgrading typescript exposed a series of typing errors. This one is a common-one. All table cell renderers should return a valid JSX element. For some reason old typescript was happy with just strings, new typescript complains. So I created a helper method to return a React fragment wrapping the string. Basically `<>YOUR STRING HERE</>`

source/frontend/src/components/Table/MoneyCell.tsx
```diff
@@ -41,10 +41,10 @@ export const EditableMoneyCell = ({
   );
 };
 
-export const MoneyCell = ({ cell: { value } }: CellProps<IProperty, number | ''>) => (
+export const MoneyCell = ({ cell: { value } }: CellProps<any, number | '' | undefined>) => (
```
asanchezr, line 44: We use MoneyCell on tables that do not include IProperty data so typescript was complaining. Since I couldn't find a common type I had to use `any`

source/frontend/tsconfig.json
```diff
@@ -17,7 +17,10 @@
     "jsx": "react-jsx",
     "downlevelIteration": true,
     "noFallthroughCasesInSwitch": true,
-    "typeRoots": ["src/typings", "node_modules/@types"]
+    "typeRoots": ["src/typings", "node_modules/@types"],
+    "paths": {
+      "react-draggable": ["typings/react-draggable"]
+    }
```
asanchezr, line 23: I had to do this change to force react-draggable to use updated typescript tytpings

devinleighsmith:
Text on 
image
is now very tiny. Does not match current dev.

devinleighsmith:
Something kinda wonky going on with the datepicker styling, not a big deal imo.
image

devinleighsmith:
'Current' instead of Current on property values tab

devinleighsmith:
Marker legend on map will not open.

devinleighsmith:
the behaviour of saving the lease improvements form has changed. Now when saving the 'are you sure' dialog pops up and the workflow gets disrupted.

devinleighsmith (approved):

FuriousLlama (approved):

PR #2388, opened 2022-11-07
PSP-4763: Plan # in pop-up doesn't match plan # in Property Info pane

Maheshbabu28:


source/frontend/src/hooks/useComposedProperties.ts
```diff
@@ -52,18 +52,35 @@ export const useComposedProperties = ({
...
+        PROPERTY_TYPES.PARCEL_MAP,
+      );
+    }
   }, [
     executeGetApiProperty,
     executeGetPropertyAssociations,
     executeGetPropertyWfs,
+    executeGetAllFeatures,
     id,
+    retrievedPid,
```
devinleighsmith, line 82: @Maheshbabu28 This is not a good idea unfortunately. What this will result in is:
1) the first three requests will be executed because the id is defined.
2) those requests will complete which will populate the pid/pin.
3) the first three requests will execute again for the same id, then the other two requests will execute.

The useffect for the later two requests should be split out. to avoid this. Actually, why wouldn't this be included in the useeffect below which is already using the retrievedPid/pin?

devinleighsmith (changes requested):
see requested changes. Please confirm that no duplicate requests are being made.

source/frontend/src/hooks/useComposedProperties.ts
Maheshbabu28: pid is accepting as prop but missed also extracted from api response. You are right its better places in the useeffect below, moved.

devinleighsmith:
@Maheshbabu28 unfortunately something still doesn't appear correct. Here is an example of a property that does have a Plan # (as can be seen from the popup). However, the header for that property is not displaying a plan #.
image

devinleighsmith (approved):
see requested changes. Please confirm that no duplicate requests are being made.

source/frontend/src/hooks/useComposedProperties.ts
```diff
@@ -81,8 +82,14 @@ export const useComposedProperties = ({
         () => executeBcAssessmentSummary(retrievedPid),
         PROPERTY_TYPES.BC_ASSESSMENT,
       );
+      typeCheckWrapper(
+        () =>
+          executeGetAllFeatures({ PID: retrievedPid }, { forceSimplePid: true, timeout: 30000 }),
+        PROPERTY_TYPES.PARCEL_MAP,
+      );
+      typeCheckWrapper(() => findByPid(retrievedPid, true), PROPERTY_TYPES.PARCEL_MAP);
```
asanchezr, line 90: @Maheshbabu28 could you please clarify if these two requests are duplicate? they seem to be. 
If not, please add a comment explaining why for other devs looking at this code - thanks!

Maheshbabu28: It's duplicate, removed(missed during rebase)

PR #2391, opened 2022-11-07
PSP-4711:Add link to property info pop-up to create lease/license fro…

Maheshbabu28:
…m the home screen


devinleighsmith (changes requested):
@Maheshbabu28 this doesn't appear to work. When I click the lease and license - create new screen I am taken to the lease and license creation page, but the property that I selected is not populated on the form (see how the research file/acquisition file works for comparison).

Maheshbabu28:
@devinleighsmith I am able to populate the PID, PIN(not available in this screen shot), lease area in my local.
image
image

Maheshbabu28:
whatever the fields exists we are populating those. 
As per the Bobbi's comment
image

devinleighsmith (approved):
@Maheshbabu28 this doesn't appear to work. When I click the lease and license - create new screen I am taken to the lease and license creation page, but the property that I selected is not populated on the form (see how the research file/acquisition file works for comparison).

PR #2399, opened 2022-11-08
Psp 2378 bc assessment tab(s)

devinleighsmith:


source/frontend/src/features/mapSideBar/tabs/bcAssessment/BcAssessmentTabView.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { IFormLease } from 'interfaces';
+import { render, RenderOptions } from 'utils/test-utils';
+
+import BcAssessmentTabView, { IBcAssessmentTabViewProps } from './BcAssessmentTabView';
+
+const history = createMemoryHistory();
+
+describe('BcAssessmentTabView component', () => {
+  const setup = (
+    renderOptions: RenderOptions & IBcAssessmentTabViewProps & { lease?: IFormLease } = {
```
asanchezr, line 12: do we need **lease* here in this type? seems like copy/pasta

source/frontend/src/hooks/useBcAssessmentLayer.ts
```diff
@@ -0,0 +1,393 @@
...
+      }
+      if (process.env.NODE_ENV === 'development') {
+        return {
+          data: mockBcAssessmentSummary,
+          status: 200,
+          statusText: 'Success',
+          headers: {},
+          config: {},
+        };
+      }
```
asanchezr, line 95: this is certainly a new pattern for local development - I gather we didn't want to setup a local WFS layer in geoserver for this?

asanchezr (approved):

source/frontend/src/hooks/useBcAssessmentLayer.ts
devinleighsmith: we don't have the underlying data in this case.

devinleighsmith: Logged https://jira.th.gov.bc.ca/browse/PSP-4852 as a follow up, should be possible to request access to the underlying layers.

PR #2403, opened 2022-11-09
psp-4818 link properties of interest by coordinates.

devinleighsmith:


source/backend/api/Areas/Leases/Controllers/LeaseController.cs
```diff
@@ -53,11 +57,27 @@ public LeaseController(ILeaseRepository leaseRepository, IMapper mapper)
...
         }
 
+        /// <summary>
+        /// Get the lease for the specified primary key 'id'.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("concept/{id:long}")]
+        [HasPermission(Permissions.LeaseView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(IEnumerable<Api.Models.Concepts.LeaseModel>), 200)]
```
asanchezr, line 72: nit: I believe the return type shouldn't be an **IEnumerable** if we only return one lease?

source/backend/api/Areas/Leases/Controllers/LeaseController.cs
```diff
@@ -67,12 +87,12 @@ public IActionResult GetLease(int id)
         [Produces("application/json")]
         [ProducesResponseType(typeof(IEnumerable<Models.Lease.LeaseModel>), 200)]
```
asanchezr, line 88: nit: return type should be updated to the new concept model

source/backend/api/Models/Concepts/Lease/LeaseMap.cs
```diff
@@ -0,0 +1,95 @@
...
+                .Map(dest => dest.TfaFileNumber, src => src.TfaFileNumber)
+                .Map(dest => dest.PsFileNo, src => src.PsFileNo)
+                .Map(dest => dest.MotiName, src => src.GetMotiName())
+                .Map(dest => dest.ExpiryDate, src => src.GetExpiryDate())
+                .Map(dest => dest.StartDate, src => src.OrigStartDate)
+                .Map(dest => dest.ProgramName, src => src.GetProgramName())
+                .Map(dest => dest.OtherCategoryType, src => src.LeaseCategoryOtherDesc)
+                .Map(dest => dest.OtherProgramType, src => src.OtherLeaseProgramType)
+                .Map(dest => dest.OtherPurposeType, src => src.OtherLeasePurposeType)
+                .Map(dest => dest.OtherPurposeType, src => src.LeasePurposeOtherDesc)
```
asanchezr, line 29: could you please confirm these two mappings are correct? they look duplicates to me...

```cs
  .Map(dest => dest.OtherPurposeType, src => src....)
  .Map(dest => dest.OtherPurposeType, src => src....)
```

devinleighsmith: Looks like there are actually two database fields referring to the same logical field from the API. I've updated this to write to both and read from the newer field or if not set the older field.

asanchezr (approved):

PR #2441, opened 2022-11-18
Psp 4827

devinleighsmith:
add enforced idle timeout to application. This prevents the keycloak token from being refreshed past the siteminder cookie expiry.


asanchezr (approved):

source/frontend/src/tenants/ITenantConfig.ts
```diff
@@ -23,6 +23,10 @@ export interface ITenantConfig {
   // configuration pertaining the Fully Attributed Parcel Map layer
   parcelMapFullyAttributed: IFullyAttributedParcelLayerConfig;
   bcAssessment: IBcAssessmentLayerConfig;
+  // the amount of time it takes to time out the idle prompt (in seconds)
+  idlePromptTimeout: number;
+  // the amount of time it takes to display the idle prompt (in seconds)
+  idleTimeout: number;
```
asanchezr, line 29: I think the code is using this prompt number in **minutes** but comment here says otherwise - could you confirm or update comment? thanks

PR #2444, opened 2022-11-18
Added documents to the file tabs | psp-4766

FuriousLlama:
image

Note: Also fixes PSP-4854


devinleighsmith:
Noticing that the only way to get document updates to the documents tab is a hard refresh or by navigating away/back to the file. This may be a good use for the staleFile/setStaleFile property of the SideBarContext (That is how the activity updates are displayed).

devinleighsmith:
Requirement: **Default is the most recent document uploaded** doesn't seem to be working.

devinleighsmith:
Also the filter doesn't match the mockup: 
image

devinleighsmith:
The font size also doesn't match up for the column text (16 vs 14), I wonder if that is a defect from the original document list view story or if Bobbi changed that for this story? might be worth checking with her.

source/backend/api/Areas/Documents/DocumentRelationshipController.cs
```diff
@@ -120,39 +124,6 @@ public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipT
             }
         }
 
-        /// <summary>
```
devinleighsmith, line 123: yeah, I think the document controller makes more sense then the relationship controller.

source/backend/api/Services/DocumentActivityService.cs
```diff
@@ -37,6 +39,22 @@ public class DocumentActivityService : BaseService, IDocumentActivityService
...
+            this.User.ThrowIfNotAuthorized(Permissions.DocumentView);
+
+            switch (fileType)
+            {
+                case FileType.Research:
+                    return documentActivityRepository.GetAllByResearchFile(fileId);
+                case FileType.Acquisition:
+                    return documentActivityRepository.GetAllByAcquisitionFile(fileId);
+                default:
+                    throw new BadRequestException("FileT type not valid to get documents.");
```
devinleighsmith, line 54: FileT

source/backend/api/Services/DocumentService.cs
```diff
@@ -123,39 +123,45 @@ public async Task<DocumentUpdateResponse> UpdateDocumentAsync(DocumentUpdateRequ
...
+                else if (string.IsNullOrEmpty(updateMetadata.Value))
+                {
+                    updateMetadata.Id = existing.Id;
+                    deletes.Add(updateMetadata);
+                }
+            }
+
+            var metadataUpdateSucessful = false;
 
+            // If there are no changes, mark the external update as successfull
```
devinleighsmith, line 158: github got a bit confused formatting these changes, but it seems like the main change is that the if statement at line 127 in the previous version was removed in favour of this if statement below. Can you help me understand the reasoning for that change? it seems like that would cause documentStorageRepository.GetDocumentMetadataAsync to get executed where it wouldn't have been previously.

source/backend/dal/Repositories/DocumentActivityRepository.cs
```diff
@@ -46,6 +46,40 @@ public IList<PimsActivityInstanceDocument> GetAllByDocument(long documentId)
...
+        /// <summary>
+        /// Get a list of all the document activity relationships for a a given aquisition file.
+        /// </summary>
+        /// <returns></returns>
+        public IList<PimsActivityInstanceDocument> GetAllByAcquisitionFile(long fileId)
+        {
+            return this.Context.PimsActivityInstanceDocuments
+                .Include(ad => ad.Document)
+                    .ThenInclude(d => d.DocumentStatusTypeCodeNavigation)
+                .Include(ad => ad.Document)
```
devinleighsmith, line 74: duplicated include maybe?

source/frontend/src/features/documents/hooks/useDocumentProvider.ts
```diff
@@ -90,6 +95,31 @@ export const useDocumentProvider = () => {
...
+    ),
+    requestName: 'updateDocumentMetadataApiCall',
+    onSuccess: useCallback(() => {
+      toast.success('Updated document metadata');
+    }, []),
+    onError: useCallback((axiosError: AxiosError<IApiError>) => {
+      if (axiosError?.response?.status === 400) {
+        toast.error(axiosError?.response.data.error);
+      } else {
+        toast.error('Update document error. Check responses and try again.');
```
devinleighsmith, line 118: Update document error or Update document metadata error?

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
```diff
@@ -94,6 +80,25 @@ const renderFileName = (onViewDetails: (values: Api_Document) => void) => {
...
+
+function renderUploaded(cell: CellProps<Api_Document, string>) {
+  return (
+    <Row className="no-gutters">
+      <Col>{prettyFormatDate(cell.row.original.appCreateTimestamp)}</Col>
+      <Col xs="auto">
+        <StyledIcon>
+          <TooltipIcon
+            toolTipId="initiator-tooltip"
+            toolTip={cell.row.original.appCreateUserid}
```
devinleighsmith, line 93: I'd think this should be the last updated not the creator, but good to check requirements here.

devinleighsmith (approved):
No major issues, I did leave some comments that may need to be addressed before merging.

source/backend/dal/Repositories/DocumentActivityRepository.cs
FuriousLlama: good catch!

source/frontend/src/features/documents/hooks/useDocumentProvider.ts
FuriousLlama: Document. Although the call has metadata on its name, it is possible that the status is the only part being updated

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
FuriousLlama: I think is the creator given that the uploaded column is showing the date of creation

source/backend/api/Services/DocumentService.cs
FuriousLlama: This was the cause of the bug. 
Thge line " foreach (var updateMetadata in updateRequest.DocumentMetadata)" was expecting that sometimes the updateRequest would not be empty or null before checking the metadata differences. This is a wrong assumption. The front end always sends a documentMetadata field array even if there are no changes.

Now it will perform the proper diff on the set and get the correct added/modified/deleted count. If the count of any of those  are not zero, then update them on Mayan and use that flag to mark the update to continue.

The GetDocumentMetadataAsync call is necessary regardless given our pattern is not to trust the front end to mark changes

source/backend/api/Areas/Documents/DocumentRelationshipController.cs
FuriousLlama: The clear indication of this was that the relationshiptype was not being used at all in the call

FuriousLlama:
> Noticing that the only way to get document updates to the documents tab is a hard refresh or by navigating away/back to the file. This may be a good use for the staleFile/setStaleFile property of the SideBarContext (That is how the activity updates are displayed).

That worked like a charm. Thanks!

FuriousLlama:
@devinleighsmith Regarding the size, default drop size and Filter by. I did not change those given it could possibly affect all instances of the component of the document. I will change it with Bobbie to make sure that is understood. Alternatively, doing changes that are only needed on a single screen would increase the complexity of the component without (maybe) a strong reason for it.

devinleighsmith:
> @devinleighsmith Regarding the size, default drop size and Filter by. I did not change those given it could possibly affect all instances of the component of the document. I will change it with Bobbie to make sure that is understood. Alternatively, doing changes that are only needed on a single screen would increase the complexity of the component without (maybe) a strong reason for it.

The above makes sense to me, but we want to make sure that we get signoff from Bobbi or we at least declare these known issues to Sue, otherwise these decisions will come back as bugs.

PR #2446, opened 2022-11-22
Feature/psp 4757

Maheshbabu28:


devinleighsmith:
The search MUST be doing an exact search for this logic to work. It doesn't seem to be given that I typed 895030 and the result it gave back was:
image

devinleighsmith:
clicking on an existing marker incorrectly displays the new untitled map marker. That marker should only display when the search result is not in our inventory, and should certainly never display when clicking on an existing marker.

devinleighsmith:
There seems to be a bug where the zoom to result functionality doesn't work on a freshly refreshed app. Ie. the first time a property is searched for the map will not zoom to the result. Subsequent searches do jump to the result.

devinleighsmith:
pin searches don't seem to be working. I've tried two different examples (such as 4667020) and the search says no results found even though that pin is one that I found on the map. Also, the loading logic for the map doesn't appear to be hooked up properly to the new search, I've had instances where the map shows old data while a search is running in the background but no loading indicator is visible.

devinleighsmith:
Also seems like the latitude/longitude is being added to the url bar and then it can't be cleared by changing the search filter or resetting the search filter. see useRouterFilter for that.

devinleighsmith (changes requested):
Need the following at a bare minimum:
pin searches need to work and only accept exact matches
pid searches must only accept exact matches
clicking on an existing map marker should not display an untitled property marker
when a search is loading the loading spinner must be displayed

Maheshbabu28:
@devinleighsmith 
PIN & PID searches are working and also added loader for non inventory search.
On reset lat/long is cleared from the url now. 
Clicking on existing marker will not display orange marker. Orange marker will be shown if property id does not exist.

devinleighsmith:
@Maheshbabu28 
Getting closer, the following is not working still:
clicking on the orange marker replaces the marker with a selected inventory marker. This is incorrect, it should just highlight the orange marker (bobbi provided a highlighted orange marker for this purpose).

Maheshbabu28:
@devinleighsmith forgot to push the changes. Can you please review and approve this pr

devinleighsmith:
@Maheshbabu28 unfortunately this is still not quite there. now when conducting a search when an existing search is displayed, the map zooms into the previous search result and then does not zoom to the new search result. The user has to manually zoom the map out to see where their search result is located.

devinleighsmith (approved):

PR #2453, opened 2022-11-23
psp-4929 do not attempt to delete a property if it is in a separate r…

devinleighsmith:
…esearch file/acquisition file.

### DIFF
--- a/source/backend/api/Services/AcquisitionFileService.cs
+++ b/source/backend/api/Services/AcquisitionFileService.cs
@@ -140,8 +140,11 @@ namespace Pims.Api.Services
                 _acquisitionFilePropertyRepository.Delete(deletedProperty);
                 if (deletedProperty.Property.IsPropertyOfInterest.HasValue && deletedProperty.Property.IsPropertyOfInterest.Value)
                 {
-                    int propertyCount = _acquisitionFilePropertyRepository.GetAcquisitionFilePropertyRelatedCount(deletedProperty.PropertyId);
-                    if (propertyCount == 1 && deletedProperty?.Property?.IsPropertyOfInterest == true)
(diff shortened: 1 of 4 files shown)


source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -140,8 +140,11 @@ public PimsAcquisitionFile UpdateProperties(PimsAcquisitionFile acquisitionFile)
                 _acquisitionFilePropertyRepository.Delete(deletedProperty);
                 if (deletedProperty.Property.IsPropertyOfInterest.HasValue && deletedProperty.Property.IsPropertyOfInterest.Value)
                 {
-                    int propertyCount = _acquisitionFilePropertyRepository.GetAcquisitionFilePropertyRelatedCount(deletedProperty.PropertyId);
```
FuriousLlama, line 143: if the GetResearchFilePropertyRelatedCount and acquisition sibling function are not used lets remove them

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -140,8 +140,11 @@ public PimsAcquisitionFile UpdateProperties(PimsAcquisitionFile acquisitionFile)
...
-                    if (propertyCount == 1 && deletedProperty?.Property?.IsPropertyOfInterest == true)
+                    PimsProperty propertyWithAssociations = _propertyRepository.GetAssociations(deletedProperty.PropertyId);
+                    var leaseAssociationCount = propertyWithAssociations.PimsPropertyLeases.Count;
+                    var researchAssociationCount = propertyWithAssociations.PimsPropertyResearchFiles.Count;
+                    var acquisitionAssociationCount = propertyWithAssociations.PimsPropertyAcquisitionFiles.Count;
```
FuriousLlama, line 146: maybe we need a method on the property repository that returns a total count of each association.

PR #2486, opened 2022-11-30
psp-4951 keycloak sync sso gold updates.

devinleighsmith:
keycloak sync changes to allow one-way sync of pims database to keycloak. Synchronizes:
PIMS claims
PIMS roles
PIMS user roles (provided the user already exists in keycloak, there is no way to add users to gold at this time).
Allows keycloak sync tool to simultaneously authorize to the sso gold management api as well as the PIMS api integration within gold.

Also includes changes to PIMS api to handle authorizing the pims service account.

Finally, includes updates to logging/reporting of keycloak sync to assist in finding and debugging issues.


source/database/mssql/scripts/dbscripts/PSP_PIMS_S42_00/Test Data/38_PIMS_USER.sql
```diff
@@ -0,0 +1,276 @@
+DECLARE @devinPersonId bigint;
```
devinleighsmith, line 1: So, this sets up your dev environment so that you can log into the application using your idir. By default you will have the System Admin role assigned and nothing else.

tools/keycloak/sync/SyncFactory.cs
```diff
@@ -5,43 +5,46 @@
 using System.Net.Http;
```
devinleighsmith, line 5: @asanchezr @FuriousLlama this is the class that needs the most inspection. The workflow is:
1) Adding all claims from PIMS to keycloak as keycloak roles
2) Delete any claims from PIMS that are in keycloak but not in PIMS
3) Add all roles from PIMS to keycloak as keycloak composite roles
4) Delete any roles from PIMS that are in keycloak but not in PIMS
5) For every user in PIMS, ensure that their user role mappings are the same as they are in keycloak.

tools/.editorconfig
```diff
@@ -0,0 +1,257 @@
+# Editor configuration, see https://editorconfig.org
```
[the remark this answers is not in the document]
FuriousLlama: we can totally do that. I'll create a ticket for that and asign it to me

tools/keycloak/sync/SyncFactory.cs
```diff
@@ -67,34 +71,62 @@ public async Task<int> SyncAsync()
...
-                var krole = await GetKeycloakRoleAsync(claim);
-                claim.Key = krole.Id;
-                claim.KeycloakRoleId = krole.Id;
+                claimsPage = await _pimsClient.HandleRequestAsync<Api.Models.PageModel<PModel.ClaimModel>>(HttpMethod.Get, $"admin/claims?page={page++}&quantity=50");
+                claims = claims.Concat(claimsPage.Items);
+            }
+            while (claimsPage != null && claimsPage.Items.Any());
+
+            var allKeycloakRoles = await _keycloakManagementClient.HandleRequestAsync<ResponseWrapperModel<RoleModel[]>>(HttpMethod.Get, $"{_keycloakManagementClient.GetIntegrationEnvUri()}/roles");
+            var keycloakRoles = allKeycloakRoles.Data.Where(x => x.Composite.HasValue && x.Composite.Value == false);
```
FuriousLlama, line 109: q: If it wants to get roles should the Composite.Value be true instead if it is not null?

devinleighsmith: Hmm, I'm worried I'm not understanding, this is checking for Composite.Value to be false and not null.

Maybe I can provide a description of composites and roles in PSP vs keycloak.

So keycloak doesn't support claims or roles in the same way that we do. They just support roles. Logically, a keycloak role is most similar to our claims, as it is basically just a string. If the user has that string, they have that claim. Where it gets confusing is that a keycloak role may also be a composite of some other keycloak roles. This is close to being equivalent to our roles in PSP. So what I'm doing here is:
if a keycloak role exists, and it has at least one composite child it should be treated as a PSP role.
If a keycloak role exists and it doesn't have any composite children it should be treated as a PSP claim.

Any ideas on how to make the separation between the two or the naming more robust are welcome.

devinleighsmith:
@FuriousLlama it looks like I can save a lease without providing a region right now which is a required field - likely not related to your changes but still needs fixing.

devinleighsmith:
@FuriousLlama were there more comments that need addressing or is this one ready to be merged now?

FuriousLlama (approved):

PR #2502, opened 2022-12-06
PSP-2136: Contacts for Lease/License File (Tenant and/or Representati…

Maheshbabu28:
…ves)


Maheshbabu28:
image

image

devinleighsmith:
this header should be bc-sans bold:
image

devinleighsmith:
The x icon to the left of a selected row does not work - it should remove the row.

devinleighsmith:
Clicking on a contact, ie: 
image

should open that contact in a new tab. However, currently it does nothing.

devinleighsmith:
@Maheshbabu28 I'm seeing something weird with all of the dropdowns on this page, I'm not able to choose a value, when I click the dropdown it opens and closes immediately.

source/frontend/src/features/leases/detail/LeasePages/tenant/TenantOrganizationContactInfo.tsx
```diff
@@ -26,57 +25,38 @@ export const TenantOrganizationContactInfo: React.FunctionComponent<
...
   const tenant: FormTenant = getIn(values, nameSpace);
+  console.log('values', values);
   let primaryContact = tenant?.initialPrimaryContact;
   if (primaryContact?.id !== tenant?.primaryContactId) {
     primaryContact = tenant?.primaryContactId
       ? getPrimaryContact(tenant?.primaryContactId, tenant)
       : undefined;
   }
   const primaryContactName = formatApiPersonNames(primaryContact);
+  //TODO: Change link to
```
devinleighsmith, line 36: ?

source/frontend/src/features/leases/detail/LeasePages/tenant/TenantOrganizationContactInfo.tsx
```diff
@@ -26,57 +25,38 @@ export const TenantOrganizationContactInfo: React.FunctionComponent<
 > = ({ nameSpace, disabled }) => {
   const { values } = useFormikContext<IFormLease>();
   const tenant: FormTenant = getIn(values, nameSpace);
+  console.log('values', values);
```
devinleighsmith, line 28: remove

Maheshbabu28:
@devinleighsmith x icon is working before in my local. And I am able to open the contact info to new tab before. I tested all these before pushing my changes to github. Let me pull this pr into my local and look into it.

source/frontend/src/features/leases/detail/LeasePages/tenant/TenantOrganizationContactInfo.tsx
Maheshbabu28: removed

Maheshbabu28:
> this header should be bc-sans bold: image

done

devinleighsmith:
Waiting for PR fixes based on the issues in todays demo as well as fixes to action failure.

devinleighsmith:
Couple of things I notice:
1) In the demo, cancelling the modal would remove all of the selected items. This is the expected behaviour. In the version that I just tried, if you select 3 contacts in the modal and then click cancel all of the selections remain in the list view.
2) Select 3 items in the modal, click the select button, and then remove all of those items from the list view using the x buttons. Next, reopen the modal. All 3 previously selected items are already selected. Now click select again, nothing happens. Whenever the select button is clicked the currently selected items in the modal must be added to the list view.
3) Open the modal, select two items. Go to the second page of results, select 2 more items. the original 2 selections get cleared (removed). The user should be able to navigate between pages without their selection being cleared, or the paging should be removed entirely.
4) The same logic occurs if the search criteria is changed, please check with Bobbi on what the expected behaviour should be if the user selects some results and then changes their search criteria and selects some different results.
5) The right hand side select appears to be choosing the property manager by default. However, on selection, all of the rows that were listed as the default end up being grouped as unknown.

devinleighsmith:
Reviewed again, seeing the following issues:

1. The right hand side select appears to be choosing the property manager by default. However, on selection, all of the rows that were listed as the default end up being grouped as unknown (unchanged from last review).
2. It is currently not possible to select results from multiple pages at the same time. Selecting a result on page 1, then selecting another result on page 2, and clicking select will only select the result from page 2. Please confirm with the BA that created this story if that behaviour is acceptable.

devinleighsmith:
@Maheshbabu28 please do some more manual testing the next time you submit changes to look for any obvious issues.

If the user selects one or more tenants, but does not select a tenant type, the system will still allow them to save those tenants. Those tenants are also added to the db. However, in the view screen they are not visible at all.

devinleighsmith:
@Maheshbabu28 in my previous list of errors I asked you to confirm with the BA's if this page needs to be able to select results from multiple pages of results at the same time. I don't see that on the story or documented here. Can you please either allow results from multiple pages to be selected or confirm that functionality is not required?

Maheshbabu28:
@devinleighsmith ContactManager internally uses Table component and in this on page change or search selection is not presereved.
 This will be lot of changes and had discussion with Bobbi this can be taken as a BUG . Please suggest.

devinleighsmith (approved):

PR #2503, opened 2022-12-06
PSP-4946 View Lease/License File in the half screen file format

asanchezr:
view-lease-file

### DIFF
--- a/source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
+++ b/source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
@@ -1,0 +1,26 @@
+import ExpandableTextList from 'components/common/ExpandableTextList';
+import { getPropertyName } from 'features/properties/selector/utils';
+import { ILease, IProperty } from 'interfaces';
+import { Api_Person } from 'models/api/Person';
+import * as React from 'react';
+import { formatNames } from 'utils/personUtils';
+
+export interface ILeaseHeaderTenantsProps {
+  lease?: ILease;
+}
+
+export const LeaseHeaderTenants: React.FC<ILeaseHeaderTenantsProps> = ({ lease }) => {
+  return (
+    <ExpandableTextList<Api_Person>
+      items={lease?.persons ?? []}
+      keyFunction={(p: Api_Person, index: number) => `lease-tenant-${p?.id ?? index}`}
+      renderFunction={(p: Api_Person) => (
+        <>{formatNames([p?.firstName, p?.middleNames, p?.surname])}</>
+      )}
+      delimiter="; "
+      maxCollapsedLength={2}
+    />
+  );
+};
+
+export default LeaseHeaderTenants;
--- a/source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
+++ b/source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
@@ -6,9 +6,15 @@ import { formatNames } from 'utils/personUtils';
 
 export interface ILeaseHeaderTenantsProps {
   lease?: ILease;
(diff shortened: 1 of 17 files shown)


source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
```diff
@@ -0,0 +1,31 @@
...
+  lease,
+  delimiter = '; ',
+  maxCollapsedLength = 2,
+}) => {
+  return (
+    <ExpandableTextList<Api_Person>
+      items={lease?.persons ?? []}
+      keyFunction={(p: Api_Person, index: number) => `lease-tenant-${p?.id ?? index}`}
+      renderFunction={(p: Api_Person) => (
+        <>{formatNames([p?.firstName, p?.middleNames, p?.surname])}</>
```
devinleighsmith, line 23: There is no confluence page for this story that I could see, but noticed that in the mockup all of the names were formatted as last, first ie: Smart, Jenny

This seems out of line with what we've discussed in the past and decided upon for names on other screens, such as contacts. Worth checking with Bobbi and then putting a note on the story so we don't get a bug for this.

devinleighsmith:
On the header are the lease/license numbers supposed to be prepended with L- ?

devinleighsmith:
not a big deal but these look horizontally aligned in the mockup. Maybe worth getting confirmation from Bobbi: 
image

source/frontend/src/features/properties/map/lease/detail/LeaseTabsContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+    name: 'Payments',
+  });
+
+  tabViews.push({
+    content: (
+      <LeaseTab
+        lease={lease}
+        setLease={setLease}
+        refreshLease={refreshLease}
+        leasePage={leasePages.get(LeasePageNames.SURPLUS)}
```
devinleighsmith, line 114: nit, do we need both leasepagenames and leasefiletabnames?

devinleighsmith:
What is the test coverage on this code looking like? do we need tech debt to cover gaps or are tests on the way?

source/frontend/src/features/properties/map/lease/common/LeaseHeaderTenants.tsx
asanchezr: I took this code verbatim from old Lease Header. I would expect this story would be mostly about UI re-arrangement, not "logic changes". I'll ask Bobbi anyways but would lean to keep the existing logic (in the new UI) and file a change request if needed.

asanchezr:
> not a big deal but these look horizontally aligned in the mockup. Maybe worth getting confirmation from Bobbi: image

In this case it was hard to align these as in code we are constrained by the bootstrap column widths (ie col=2 or 3, etc) vs Bobbi that just moves things around to fit whatever spacing. I tried pushing the property values to the right but then it clashed with the right hand section of the header. 

I'm going to add a comment for Bobbi in the story but implementing pixel perfect alignment here would involve re-work of the whole header, not sure if worth it. Also notice in the original mockup she DID have the Lease Licence # field out of alignment so the pattern is already broken a bit. Happy to discuss if there's easy ways to keep everything aligned (given that text length on the header "fields" are so varied)

asanchezr:
> On the header are the lease/license numbers supposed to be prepended with L- ?

I double-checked and this is data-driven; ie we display whatever is in **lease.lFileNo** - it just happens that my local test data has no L- in it. Again, I just ported the existing logic over to a new UI. I think the local test data should be updated to include L- with all lease numbers

FuriousLlama (approved):

FuriousLlama:
We should reorganize the features into folders that make more sense. Right now things under map are starting to get crowded

PR #2505, opened 2022-12-06
npm version bumps.

devinleighsmith:


source/frontend/package.json
```diff
@@ -112,6 +112,10 @@
     "sass": "1.43.2",
     "typescript": "4.8.4"
   },
+  "overrides": {
```
devinleighsmith, line 115: note, since we are 8.5.5 this should be supported by our builds (supported as of 8.3) we use node 16.15.0 which provides npm 8.5.5

asanchezr: I like this. This is now a native npm feature - no need for that external library for overrides anymore

asanchezr (approved):

PR #2506, opened 2022-12-07
Lease Properties selector (+ general properties refactor) | psp 4726

FuriousLlama:
image


devinleighsmith:
going to try to review this between meetings today.

devinleighsmith:
There seems to be a bug with adding a property that is already in the inventory (ie, dropping a pin on a parcel that already has a green marker).

The first time I drop the pin, nothing happens. The second time I drop the pin it works.
@FuriousLlama

devinleighsmith:
Currently the create lease is adding new properties as blue pins, but I think the story is asking for green pins?
**if user clicks on the parcel not owned - need to get info from that parcel, and that to the inventory show - green pin**

although the wording seems unnecessarily convoluted.

devinleighsmith:
@FuriousLlama it looks like I can save a lease without providing a region right now which is a required field - likely not related to your changes but still needs fixing.

devinleighsmith:
All of the fields appear to be populated in the DB correctly.

source/backend/api/Services/LeaseService.cs
```diff
@@ -71,38 +80,11 @@ public PimsLease Update(PimsLease lease, bool userOverride = false)
         /// <returns></returns>
         private PimsLease AssociatePropertyLeases(PimsLease lease, bool userOverride = false)
         {
+            MatchProperties(lease);
```
devinleighsmith, line 83: Good yes, we do need to start doing this now.

source/backend/api/Services/LeaseService.cs
```diff
@@ -120,10 +102,80 @@ private PimsLease AssociatePropertyLeases(PimsLease lease, bool userOverride = f
...
+            property.PropertyClassificationTypeCode = "UNKNOWN";
+            property.PropertyDataSourceEffectiveDate = System.DateTime.Now;
+            property.PropertyDataSourceTypeCode = "PMBC";
+
+            property.PropertyTypeCode = "UNKNOWN";
+
+            property.PropertyStatusTypeCode = "UNKNOWN";
+            property.SurplusDeclarationTypeCode = "UNKNOWN";
+
+            property.IsPropertyOfInterest = true;
```
devinleighsmith, line 170: yes, so I'm guessing you are updating this to false.

source/frontend/src/components/common/styles.ts
```diff
@@ -102,13 +102,6 @@ export const Tabs = styled(BsTabs)`
   }
 `;
 
-export const NoPaddingRow = styled(Row)`
```
devinleighsmith, line 105: looks like dead code, nice.

source/frontend/src/features/leases/add/AddLeaseContainer.tsx
```diff
@@ -36,21 +38,28 @@ export const AddLeaseContainer: React.FunctionComponent<
   const { search } = useMapSearch();
 
   useEffect(() => {
+    if (!!selectedFileFeature && !!formikRef.current) {
+      formikRef.current.resetForm();
+      formikRef.current?.setFieldValue('properties', [
+        PropertyForm.fromMapProperty(mapFeatureToProperty(selectedFileFeature)),
+      ]);
+    }
     return () => {
```
devinleighsmith, line 47: Hmm, I think this should still only fire once, but am not sure. Actually I guess it isn't a big deal if this fires more then once as long as the create file from map functionality still works. Carry on.

source/frontend/src/features/leases/models.ts
```diff
@@ -130,6 +128,14 @@ export class LeaseFormModel {
       otherType: stringToNull(this.otherLeaseTypeDescription),
     };
   }
+
+  public getPropertiesAsForm(): PropertyForm[] {
+    return this.properties
+      .map(property => {
+        return property.property;
+      })
+      .filter((item): item is PropertyForm => !!item);
```
devinleighsmith, line 137: ooh, I see. like this.

source/frontend/src/features/leases/models.ts
```diff
@@ -160,74 +166,43 @@ export class LeaseDetails {
 
 export class FormLeaseProperty {
   id?: number;
-  property?: FormProperty;
+  property?: PropertyForm;
   leaseId?: number;
   rowVersion?: number;
-  landArea?: number;
-  areaUnitType?: Api_TypeCode<string>;
+  landArea: string;
```
devinleighsmith, line 172: is landArea required? or is this taking into account formik's restriction that undefined is not allowed.

source/frontend/src/features/leases/models.ts
```diff
@@ -160,74 +166,43 @@ export class LeaseDetails {
...
 
   static fromApi(apiPropertyLease: Api_PropertyLease): FormLeaseProperty {
     const model = new FormLeaseProperty(apiPropertyLease.lease?.id);
-    model.property = FormProperty.fromApi(apiPropertyLease.property ?? {});
+    model.property = PropertyForm.fromApi(apiPropertyLease.property ?? {});
     model.id = apiPropertyLease.id;
     model.rowVersion = apiPropertyLease.rowVersion;
-    model.landArea = apiPropertyLease.leaseArea;
-    model.areaUnitType = apiPropertyLease.areaUnitType;
+    model.landArea = apiPropertyLease.leaseArea?.toString() || '0';
```
devinleighsmith, line 186: hmm, I slightly worry that a 0 is not equal to not set in this case. We had a similar problem in CITZ pims, where setting a value of zero had significance vs. not setting at all.

source/frontend/src/features/leases/models.ts
```diff
@@ -160,74 +166,43 @@ export class LeaseDetails {
...
+    model.areaUnitTypeCode = apiPropertyLease.areaUnitType?.id || PropertyAreaUnitTypes.Meter;
```
devinleighsmith, line 187: This on the other hand is the default in the mockup so I think this is fine.

source/frontend/src/features/leases/models.ts
```diff
@@ -160,74 +166,43 @@ export class LeaseDetails {
 
 export class FormLeaseProperty {
   id?: number;
-  property?: FormProperty;
+  property?: PropertyForm;
```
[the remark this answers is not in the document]
FuriousLlama: Totally agree, thats just the name of the form used by research and acq. For now I wont rename it

source/frontend/src/features/leases/models.ts
FuriousLlama: It does not look like is required on the mockups

source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
```diff
@@ -0,0 +1,167 @@
...
+      display: false,
+      closeButton: false,
+      okButtonText: 'Add',
+      cancelButtonText: 'Cancel',
+      handleOk: confirmAdd,
+      handleCancel: cancelAdd,
+    };
+  }, [confirmAdd, cancelAdd]);
+
+  useDeepCompareEffect(() => {
```
devinleighsmith, line 81: is the deep compare required here?

source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
```diff
@@ -0,0 +1,167 @@
...
+      <FieldArray
+        name="properties"
+        render={arrayHelpers => {
+          arrayHelpersRef.current = arrayHelpers;
+          return (
+            <>
+              <Row className="py-3 no-gutters">
+                <Col>
+                  <MapSelectorContainer
+                    addSelectedProperties={async (newProperties: IMapProperty[]) => {
```
devinleighsmith, line 101: This is a pretty big function to have embedded in jsx, would lean towards having this in a function outside of the jsx

source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
```diff
@@ -0,0 +1,167 @@
...
+                        const property = newProperties[i];
+                        const formProperty = FormLeaseProperty.fromMapProperty(property);
+
+                        // Retrieve the pims id of the property if it exists
+                        if (
+                          formProperty.property !== undefined &&
+                          formProperty.property.apiId === undefined
+                        ) {
+                          const result = await searchProperty(property);
+                          if (result?.length === 1) {
```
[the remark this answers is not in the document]
FuriousLlama: in that case thoughts on just adding the id that patches the first pid/pin?

source/frontend/src/features/properties/map/shared/update/properties/UpdateProperties.tsx
```diff
@@ -92,18 +95,36 @@ export const UpdateProperties: React.FunctionComponent<
...
-                      onSelectedProperty={(newProperty: IMapProperty) => {
-                        const formProperty = PropertyForm.fromMapProperty(newProperty);
-                        push(formProperty);
-                      }}
-                      existingProperties={formikProps.values.properties}
-                      onRemoveProperty={remove}
-                    />
-                  </Col>
-                </Row>
+                <>
```
devinleighsmith, line 98: mmm, good to know that the update properties for the various files uses the same component under the hood. forgot about that.

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -87,3 +90,48 @@ export const getApiPropertyName = (property?: Api_Property): PropertyName => {
   };
   return getPropertyName(mapProperty);
 };
+
```
devinleighsmith, line 93: centralizing these makes sense.

source/frontend/src/features/leases/models.ts
devinleighsmith: so... maybe should be nullable then?

source/frontend/src/features/leases/models.ts
devinleighsmith: kk

source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const addProperty = useCallback(() => {
+    if (arrayHelpersRef.current !== null && propertyToAdd !== undefined) {
+      arrayHelpersRef.current.push(propertyToAdd);
+
+      setPropertyToAdd(undefined);
+    }
+  }, [setPropertyToAdd, arrayHelpersRef, propertyToAdd]);
+
+  const searchProperty = async (newProperty: IMapProperty): Promise<IProperty[] | undefined> => {
```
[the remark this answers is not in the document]
devinleighsmith: right, so I guess our property exists checking is limited to just pid or pin for now, I think that is ok, we have that other bug already logged to deal with similar lat/lngs.

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: This is what consumed me the most amount of time. We are not being totally thorough with the ts typing causing some issues where data is allegedly a type but internally is a different one

source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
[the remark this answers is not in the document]
devinleighsmith: Yes, I think that works.

devinleighsmith (approved):

devinleighsmith (approved):

PR #2532, opened 2022-12-10
PSP-5033: Cancel warning pop-up should only appear when user made changes to the form

asanchezr:


asanchezr:
Closing PR as this needs to targed the **test ** branch

PR #2536, opened 2022-12-14
Added marker logic for payable leases

FuriousLlama:
image


source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -114,13 +137,22 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
...
  */
 export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
+  if (feature.properties.IS_PAYABLE_LEASE || feature.properties.isPayableLease) {
+    if (selected) {
+      return propertyWithLeaseIconSelect;
+    } else {
+      return propertyWithLeaseIcon;
+    }
+  }
+
```
asanchezr, line 147: Did you update the marker legend to include this new pin? I mean this map component...
image

FuriousLlama: uh nope. Never pressed that button before. Guess thats what they meant by 'label' :)

source/frontend/src/components/maps/leaflet/Legend/Legend.tsx
```diff
@@ -17,6 +17,10 @@ export const Legend = () => {
         pin: require('assets/images/pins/land-poi.svg').default,
         label: 'Property of Interest',
       },
+      {
+        pin: require('assets/images/pins/land-lease.svg').default,
+        label: 'Payable lease/license',
+      },
```
asanchezr, line 23: not a big deal but I see four (4) images added on this PR - 2 svgs and 2 pngs but I only see the svgs being used in code. Are the pngs used or could they be safely removed?

asanchezr (approved):

source/frontend/src/components/maps/leaflet/Legend/Legend.tsx
FuriousLlama: My guess is that they could be removed if not used. But I did not check if the other marker images have the same usage.

PR #2542, opened 2022-12-14
Added warning mesage when removing properties from leases

FuriousLlama:
image


source/frontend/src/features/leases/shared/propertyPicker/LeasePropertySelector.tsx
```diff
@@ -137,14 +177,15 @@ export const LeasePropertySelector: React.FunctionComponent<
                 </Col>
               </Row>
               <Section header="Selected properties">
+                {propertyIndexToRemove}
```
asanchezr, line 180: was this here for debugging? or is the story calling for showing the 0-based index of the property on the ui?

asanchezr (approved):

PR #2554, opened 2022-12-16
PSP-4926: FT: Activities - When changing the activity to Complete or …

Maheshbabu28:
…Cancelled, PIMS still allows to link/unlink properties to the activity

### DIFF
--- a/source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
+++ b/source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
@@ -61,8 +61,9 @@ export const ActivityPropertyModal: React.FunctionComponent<
   };
 
   const handleOk = async () => {
-    if (disableSave) setDisplay(false);
-    else {
+    if (disableSave) {
+      setDisplay(false);
+    } else {
       if (activityModel && allProperties?.length > 0) {
         activityModel.actInstPropFiles =
           selectedFileProperties.map(p => ({
@@ -94,6 +95,7 @@ export const ActivityPropertyModal: React.FunctionComponent<
           allProperties={allProperties}
           selectedFileProperties={selectedFileProperties}
           setSelectedFileProperties={setSelectedFileProperties}
+          disableSelection={disableSave}
         />
       }
     />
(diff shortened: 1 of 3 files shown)


source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
```diff
@@ -58,16 +61,19 @@ export const ActivityPropertyModal: React.FunctionComponent<
...
-      activityModel.actInstPropFiles =
-        selectedFileProperties.map(p => ({
-          activityId: activityModel?.id,
-          propertyFileId: p?.id,
-        })) ?? [];
-      const updatedActivity = await onSave(activityModel);
-      // if the activity was updated successfully, hide the modal.
-      if (!!updatedActivity) {
-        setDisplay(false);
+    if (disableSave) setDisplay(false);
```
asanchezr, line 64: I think we wrap `if` statements with braces, even for one-liners; ie

```tsx
if (...) {
 setDisplay(false)
}
```

source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
```diff
@@ -58,16 +61,19 @@ export const ActivityPropertyModal: React.FunctionComponent<
...
+        activityModel.actInstPropFiles =
+          selectedFileProperties.map(p => ({
+            activityId: activityModel?.id,
+            propertyFileId: p?.id,
+          })) ?? [];
+        const updatedActivity = await onSave(activityModel);
+        // if the activity was updated successfully, hide the modal.
+        if (!!updatedActivity) {
+          setDisplay(false);
+        }
```
asanchezr, line 76: Looking at the whole logic this "stops" the bug by closing the dialog without saving but this might be confusing to the user who was able to interact with the property list via checkboxes. This is kinda of a gray area but I would lean towards implementing what Bobbi suggested in JIRA:

> I don't think we want that as a fix, because the users will still need to click that to view what properties are included in the activity.
> We could turn it into a view-only kind of button though, by not allowing them to select/deselect. (ie: no checkboxes)

What are your thoughts @devinleighsmith ?

source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
devinleighsmith: correct

source/frontend/src/features/properties/map/activity/detail/ActivityPropertyModal.tsx
devinleighsmith: agreed, this doesn't fix the issue in my mind as the user will still see those property rows as selected even if they leave and go back to the modal. Unless the user is savvy and notices that no success message is displayed they will have no idea that their changes were not saved.

I would say either the checkboxes should be disabled but still visible as suggested.

Maheshbabu28: Added disables boolean to table to allow disable ability of checkbox.

devinleighsmith (approved):

PR #2556, opened 2022-12-16
PSP-5113: FT-REG: Property Inventory - List View: When an inserted ad…

Maheshbabu28:
…dress is not in PIMS inventory, the seach by Address feature doesn't work


source/frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -57,6 +60,19 @@ export const PropertyFilter: React.FC<React.PropsWithChildren<IPropertyFilterPro
...
+    exact: true,
+    strict: true,
+  });
+  React.useEffect(() => {
+    if (matched !== null) {
+      setListView(true);
+    } else {
+      setListView(false);
+    }
+  }, [matched, setListView]);
```
asanchezr, line 75: It looks like this component **PropertyFilter** is being embedded in two different places and we want it to behave differently depending on where is hosted - right? If that is the case I think it would be a bit cleaner to have this component accept a boolean property to control the desired behaviour instead of relying on the control figuring out where it is being hosted via the current route URL. 

That way the caller/host of this component would call it with `<PropertyFilter someBooleanProp={true} ... />` or `<PropertyFilter someBooleanProp={false} ... />` instead of having to rely of hard-coded route urls.

Thoughts @devinleighsmith ?

devinleighsmith: I agree, matching here makes us more vulnerable in the case we change any of our routes.

Maheshbabu28: added boolean property

devinleighsmith (approved):

PR #2562, opened 2022-12-17
add mayan sync endpoints to postman collection.

devinleighsmith:
mayan sync endpoints referenced by documentation not included in current instance of postman, added.


asanchezr (approved):
There seems to be more than mayan endpoint additions here but I guess the collection was outdated?

PR #2569, opened 2022-12-21
psp-4929 correct delete behaviour when file is related to another fil…

devinleighsmith:
…e (lease, research, etc).

### DIFF
--- a/source/backend/dal/Repositories/PropertyRepository.cs
+++ b/source/backend/dal/Repositories/PropertyRepository.cs
@@ -324,6 +324,7 @@ namespace Pims.Dal.Repositories
                 .Where(p => p.PropertyId == property.Id)
                 .Include(p => p.PimsPropertyResearchFiles)
                 .Include(p => p.PimsPropertyAcquisitionFiles)
+                .Include(p => p.PimsPropertyLeases)
                 .FirstOrDefault();
 
             existingProperty.PimsPropertyResearchFiles.ForEach(pr => this.Context.Remove(new PimsPropertyResearchFile() { Id = pr.Id }));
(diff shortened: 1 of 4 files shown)


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -326,8 +326,9 @@ public void Delete(PimsProperty property)
                 .Include(p => p.PimsPropertyAcquisitionFiles)
                 .FirstOrDefault();
 
-            existingProperty.PimsPropertyResearchFiles.ForEach(pr => this.Context.Remove(pr));
-            existingProperty.PimsPropertyAcquisitionFiles.ForEach(pa => this.Context.Remove(pa));
+            existingProperty.PimsPropertyResearchFiles.ForEach(pr => this.Context.Remove(new PimsPropertyResearchFile() { Id = pr.Id }));
+            existingProperty.PimsPropertyAcquisitionFiles.ForEach(pa => this.Context.Remove(new PimsPropertyAcquisitionFile() { Id = pa.Id }));
+            existingProperty.PimsPropertyLeases.ForEach(l => this.Context.Remove(new PimsPropertyLease() { Id = l.Id }));
```
asanchezr, line 331: It's a bit odd that the "existingProperty" EF query doesn't include the related data for leases yet we try to delete them. Have you tested that "PimsPropertyLeases" is populated with the above query?

Github doesn't let me highlight the relevant code so here is a screenshot:

image

devinleighsmith: In my tests it was populated, but I agree that for consistency it should be included, will add that.

asanchezr (approved):

PR #2572, opened 2022-12-21
Feature/psp 4735

Maheshbabu28:


devinleighsmith:
@Maheshbabu28 unfortunately I think there was a fundamental understanding of what this story is asking for.

We are completely decommissioning all /lease endpoints. The only lease pages going forward are the new lease pages on the mapview. The ask of this story was to make sure all of the edit functions on the new mapview sidebar lease pages worked as expected. As a result, I've closed this PR as it does not appear to address the story adequately. Let's discuss what to do here tomorrow.

PR #2602, opened 2023-01-04
psp-3394 add address to properties when added from property selector.

devinleighsmith:


source/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
```diff
@@ -32,11 +34,17 @@ export const AcquisitionProperties: React.FunctionComponent<
...
                       const formProperty = PropertyForm.fromMapProperty(property);
                       if (values.properties?.length === 0 && index === 0) {
                         formikProps.setFieldValue(`region`, formProperty.region);
                       }
+                      const address = property?.pid
+                        ? await getPrimaryAddressByPid(property.pid)
+                        : undefined;
+                      formProperty.address = address
+                        ? AddressForm.fromBcaAddress(address)
+                        : undefined;
```
asanchezr, line 47: There is a subtle change in behaviour here by making the forEach callback async. There's the chance that the original order of the properties is not preserved in the case of an async promise to call BC Assessment say for property #2 takes way longer than property #10. 

In that case property-10 would be inserted in the formik array before property-2

If we really care for the order (I think we do) we should change this code to something that "serializes" the promises. Something like https://stackoverflow.com/questions/50263209/how-do-i-serialize-a-javascript-array-of-functions-returning-promises

source/frontend/src/features/properties/map/research/add/ResearchProperties.tsx
```diff
@@ -28,8 +30,14 @@ const ResearchProperties: React.FunctionComponent<React.PropsWithChildren<unknow
...
                   addSelectedProperties={(newProperties: IMapProperty[]) => {
-                    newProperties.forEach(property => {
+                    newProperties.forEach(async property => {
                       const formProperty = PropertyForm.fromMapProperty(property);
+                      const address = property?.pid
+                        ? await getPrimaryAddressByPid(property.pid)
+                        : undefined;
+                      formProperty.address = address
+                        ? AddressForm.fromBcaAddress(address)
+                        : undefined;
```
asanchezr, line 40: Same comment re: async callbacks

source/frontend/src/features/properties/map/shared/models.ts
```diff
@@ -1,10 +1,14 @@
 import { IMapProperty } from 'components/propertySelector/models';
+import { IBcAssessmentSummary } from 'hooks/useBcAssessmentLayer';
+import { Api_Address } from 'models/api/Address';
 import { Api_File } from 'models/api/File';
 import { Api_Property } from 'models/api/Property';
 import { Api_PropertyFile } from 'models/api/PropertyFile';
 import { formatApiAddress, pidParser } from 'utils';
 import { toTypeCode } from 'utils/formUtils';
 
+import { formatBcaAddress } from './../../../../utils/propertyUtils';
```
asanchezr, line 10: nit: would be a bit better to include this import in the 'utils' import 2 lines above

source/frontend/src/features/properties/map/shared/update/properties/UpdateProperties.tsx
```diff
@@ -100,8 +110,14 @@ export const UpdateProperties: React.FunctionComponent<
...
                         addSelectedProperties={(newProperties: IMapProperty[]) => {
-                          newProperties.forEach(property => {
+                          newProperties.forEach(async property => {
                             const formProperty = PropertyForm.fromMapProperty(property);
+                            if (property.pid) {
+                              const address = await getPrimaryAddressByPid(property.pid);
+                              formProperty.address = address
+                                ? AddressForm.fromBcaAddress(address)
+                                : undefined;
+                            }
```
asanchezr, line 120: same comment re: async callbacks

source/frontend/src/features/properties/map/acquisition/add/AcquisitionProperties.tsx
```diff
@@ -32,13 +34,19 @@ export const AcquisitionProperties: React.FunctionComponent<
...
+                        if (property.pid) {
+                          const bcaSummary = await getPrimaryAddressByPid(property.pid);
+                          formProperty.address = bcaSummary?.address
+                            ? AddressForm.fromBcaAddress(bcaSummary?.address)
+                            : undefined;
+                          formProperty.legalDescription = bcaSummary?.legalDescription?.LEGAL_TEXT;
+                        }
+                        push(formProperty);
+                      });
+                    }, Promise.resolve());
```
asanchezr, line 49: The new code seems to have discarded some previous functionality to set the region based on the first added property... could you please check?

```tsx
  if (values.properties?.length === 0 && index === 0) {
    formikProps.setFieldValue(`region`, formProperty.region);
  }
```

devinleighsmith: hmm, I'll add this back.

source/frontend/src/hooks/useBcAssessmentLayer.ts
```diff
@@ -46,29 +59,29 @@ export const useBcAssessmentLayer = (
...
     withCredentials: true,
   });
   const getValuesWrapper = useWfsLayer(url, {
-    name: names['VALUE'],
+    name: names[BC_ASSESSMENT_TYPES.VALUES],
     withCredentials: true,
   });
   const getChargesWrapper = useWfsLayer(url, {
-    name: names['CHARGE'],
+    name: names[BC_ASSESSMENT_TYPES.CHARGES],
```
asanchezr, line 76: Also want to confirm that the changes from hardcoded singular 'ADDRESS' / 'VALUE' etc to the enum values (which are all plural) would not affect the layer query params?

asanchezr (approved):

source/frontend/src/hooks/useBcAssessmentLayer.ts
devinleighsmith: Correct, the query params are just for the folio id and roll number. These enum values just correspond to the keys in the response object.

asanchezr: okay but I mean that the final text is different. It used to be 'CHARGE' and now it will be 'CHARGE(S)' - would that still work?

asanchezr: I'm just comparing the old hardcoded strings with the string values of the new enum and they don't match

PR #2636, opened 2023-01-13
Added Project info to Acq | psp-3859

FuriousLlama:
image


source/backend/api/Services/ProjectService.cs
```diff
@@ -0,0 +1,42 @@
...
+            ClaimsPrincipal user,
+            ILogger<ProjectService> logger,
+            IProjectRepository projectRepository)
+            : base(user, logger)
+        {
+            _logger = logger;
+            _projectRepository = projectRepository;
+        }
+
+        public IList<PimsProject> GetProjectPredictions(string filter, int maxResult)
```
devinleighsmith, line 31: Can you explain the use of predictions here? Are we using this somewhere else maybe? This doesn't seem to describe search functionality but maybe I don't understand the context.

source/backend/api/Services/ProjectService.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            _logger.LogInformation("Getting all projects");
```
devinleighsmith, line 33: Are we getting all projects or just matching projects?

source/backend/api/Services/ProjectService.cs
```diff
@@ -0,0 +1,42 @@
...
+            /*this.User.ThrowIfNotAuthorized(Permissions.ActivityView);
```
devinleighsmith, line 34: Need a todo here I guess until Eduardo merges his changes.

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <param name="input"></param>
+        /// <param name="top"></param>
+        /// <returns>An array of contacts matching the filter.</returns>
+        [HttpGet("search={input}&top={top}")]
+        //[HasPermission(Permissions.ContactView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(List<ProjectModel>), 200)]
+        [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public IActionResult GetProjectPredictions(string input, int top)
```
devinleighsmith, line 57: needs a todo for claims or wait for eduardo's PR.

source/backend/dal/Repositories/ProjectRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+        public ProjectRepository(PimsContext dbContext, ClaimsPrincipal user, ILogger<ProjectRepository> logger)
+            : base(dbContext, user, logger)
+        {
+        }
+        #endregion
+
+        #region Methods
+
+        /// <summary>
+        /// Retrieves the activities with the specified research file id.
```
devinleighsmith, line 32: nit: activities -> projects

source/frontend/src/components/common/form/Select.tsx
```diff
@@ -89,6 +89,7 @@ export const Select: React.FC<React.PropsWithChildren<SelectProps>> = ({
   };
 
   const onSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    console.log(e);
```
devinleighsmith, line 92: probably meant to remove this.

source/backend/api/Services/ProjectService.cs
FuriousLlama: Well, basically I used the same ideas used by the person/organization autocomplete. It could be renamed to search/filter or anything else to make it more reusable.

source/backend/api/Services/ProjectService.cs
FuriousLlama: Just matching the projects on the filter. This text remained from an earlier version where it was retrieving all projects and filtering on the front end

source/frontend/src/components/common/form/Select.tsx
FuriousLlama: Yep!

source/frontend/src/features/properties/map/acquisition/add/models.ts
```diff
@@ -65,6 +73,12 @@ export class AcquisitionForm implements WithAcquisitionTeam {
     newForm.region = fromTypeCode(model.regionCode)?.toString();
     // ACQ file properties
     newForm.properties = model.fileProperties?.map(x => PropertyForm.fromApi(x)) || [];
+    newForm.fundingTypeCode = model.fundingTypeCode?.id;
+    newForm.fundingTypeOtherDescription = model.fundingOther || '';
+    newForm.project =
+      model.project !== undefined
```
devinleighsmith, line 79: Hmm, if a model coming from the backend has a null or undefined id I'm not sure we want to set it to 0 here.

Bit of a weird case, this should never happen, but if it does happen and we just call context.Update(acqFile) on the backend we might start writing projects with id 0 to the db.

source/frontend/src/features/properties/map/acquisition/common/update/acquisitionTeam/UpdateAcquisitionTeamYupSchema.ts
```diff
@@ -1,6 +1,13 @@
 import * as Yup from 'yup';
 
 export const UpdateAcquisitionTeamYupSchema = Yup.object().shape({
+  fundingTypeOtherDescription: Yup.string().when('fundingTypeCode', {
+    is: (fundingTypeCode: string) => fundingTypeCode && fundingTypeCode === 'OTHER',
+    then: Yup.string()
+      .required('Other Description required')
```
devinleighsmith, line 7: should this be Other Description or Other description.

Or perhaps it should refer to the field name which is: **Other funding**

source/frontend/src/features/properties/map/acquisition/detail/AcquisitionSummaryView.tsx
```diff
@@ -23,13 +23,26 @@ const AcquisitionSummaryView: React.FunctionComponent<
   const keycloak = useKeycloakWrapper();
   const detail: DetailAcquisitionFile = DetailAcquisitionFile.fromApi(acquisitionFile);
 
+  const projectName =
+    acquisitionFile?.project !== undefined
+      ? acquisitionFile?.project?.code + ' - ' + acquisitionFile?.project?.description
```
devinleighsmith, line 28: hmm, if the code or description is null/undefined will that get written out here as a string? hmm... nvm they are both required fields in the db.

Mild nit I guess since this only works if the frontend knows these fields are non-nullable.

source/frontend/src/features/properties/map/acquisition/hooks/useProjectTypeahead.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { getProjectPredictions } = useApiProjects();
+  const [isTypeaheadLoading, setIsTypeaheadLoading] = useState(false);
+  const [matchedProjects, setMatchedProjects] = useState<IAutocompletePrediction[]>([]);
+
+  // fetch autocomplete suggestions from server
+  const handleTypeaheadSearch = async (query: string) => {
+    try {
+      setIsTypeaheadLoading(true);
+      const { data } = await getProjectPredictions(query);
+      console.log(data);
```
devinleighsmith, line 23: still need this?

source/frontend/src/features/properties/map/acquisition/hooks/useProjectTypeahead.ts
```diff
@@ -0,0 +1,42 @@
...
+      setMatchedProjects(
```
devinleighsmith, line 24: if you aren't using our api wrapper you will need to use isMounted here.

devinleighsmith, line 24: admittedly an edge case, but if the user keeps typing I don't see that we are enforcing search order or handling cancellations here, so in theory if the requests complete out of order you may get results from a previously typed string.

source/frontend/src/features/properties/map/acquisition/update/summary/UpdateAcquisitionForm.tsx
```diff
@@ -69,6 +73,41 @@ export const UpdateAcquisitionForm = React.forwardRef<
...
+              <SectionField label="Product">
+                <Select field="product" options={[]} placeholder="Select..." />
+              </SectionField>
+              <SectionField label="Funding">
+                <Select
+                  field="fundingTypeCode"
+                  options={acquisitionFundingTypes}
+                  placeholder="Select..."
+                  onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+                    const selectedValue = [].slice
```
devinleighsmith, line 95: had to look this one up but seems fine.

source/frontend/src/mocks/mockProjects.ts
```diff
@@ -0,0 +1,159 @@
+import { Api_AcquisitionFile } from 'models/api/AcquisitionFile';
+
+export const mockProjects = (
+  id = 1,
+  name = 'Test ACQ File',
+  rowVersion = 1,
+): Api_AcquisitionFile => ({
```
devinleighsmith, line 7: hmm, if the function is labelled mockProjects I would expect an array of projects here not an acq file that has projects attached.

source/frontend/src/features/properties/map/acquisition/common/update/acquisitionTeam/UpdateAcquisitionTeamYupSchema.ts
FuriousLlama: could be. Just followed what we do now

source/frontend/src/features/properties/map/acquisition/hooks/useProjectTypeahead.ts
FuriousLlama: Thats right. However this is almost the same code as we have in the other autocomplete and it seems to work fine

source/frontend/src/features/properties/map/acquisition/update/summary/UpdateAcquisitionForm.tsx
FuriousLlama: Yeh, rather convoluted way of updating the value

source/frontend/src/mocks/mockProjects.ts
FuriousLlama: I think the last update with the tests didn't get pushed :P

source/frontend/src/features/properties/map/acquisition/add/models.ts
FuriousLlama: That should not happen. In any case the toApi would clear it and be safe again

FuriousLlama: > @FuriousLlama what is the plan for unit test coverage on this one? Do we need a follow up story?

Given that this story only modifies existing screens the current tests would work to get the snaps. Regarding the actual input of the forms. Do we want to test that using unittests? IMO that looks better suited for integration tests.

source/backend/api/Services/ProjectService.cs
devinleighsmith: Ok, if that is already in use this is fine.

source/frontend/src/features/properties/map/acquisition/add/models.ts
devinleighsmith: Don't we have an entirely new api on the backend and frontend for the autocomplete though? Shouldn't we have some tests for that?

source/frontend/src/features/properties/map/acquisition/hooks/useProjectTypeahead.ts
devinleighsmith: Yep, you would probably need to force the requests to return in a different order. If you want to leave this as is maybe log some low-priority tech debt? unlikely to see problems from this but strictly speaking the code isn't handling network latency.

source/frontend/src/features/properties/map/acquisition/add/models.ts
FuriousLlama: Ah Backend tests, sure!

devinleighsmith (approved):

PR #2637, opened 2023-01-13
PSP-4459 Financial Codes list view

asanchezr:
Screenshot 2023-01-11 174947


devinleighsmith:
reviewing

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
```diff
@@ -0,0 +1,131 @@
...
+          return true;
+        });
+      }
+
+      if (sort) {
+        const sortFields = Object.keys(sort);
+        if (sortFields?.length > 0) {
+          const sortBy = sortFields[0] as keyof Api_FinancialCode;
+          const sortDirection = sort[sortBy];
+          return orderBy(records, sortBy, sortDirection);
```
devinleighsmith, line 79: This isn't working correctly for the code value column as it is sorting those values as if they were strings not numbers: 
image

asanchezr: This the result of the code property on the JSON model being a string. If you look at the DB there are some tables that had number codes and others had string codes. One had to be picked up and I believe Doug's latest DB change (in review) will change all codes to be string

asanchezr: If they want to sort as numbers (never specified in AC) we would need a different approach (most likely a new list view for codes that are numeric, not sure)

source/frontend/src/constants/claims.ts
```diff
@@ -15,7 +15,8 @@ export enum Claims {
   PROPERTY_DELETE = 'property-delete',
   ADMIN_USERS = 'admin-users',
   ADMIN_ROLES = 'admin-roles',
-  ADMIN_DOCUMENT_GENERATION = 'admin-roles', // Todo: set the propper role
+  ADMIN_DOCUMENT_GENERATION = 'admin-roles', // TODO: use proper claim
+  ADMIN_FINANCIAL_CODES = 'admin-roles', // TODO: use proper claim
```
devinleighsmith, line 19: Don't think this is used anywhere, so you may as well take this out.

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const fetchData = useCallback(async () => {
+    await getFinancialCodes();
+  }, [getFinancialCodes]);
+
+  useEffect(() => {
+    if (financialCodeResults === undefined) {
+      fetchData();
+    }
+  }, [fetchData, financialCodeResults]);
```
devinleighsmith, line 44: is this necessary? if the undefined check is removed this should only request once since fetchData isn't changing. Is there a time where we manually set financialCodeResults to undefined?

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
```diff
@@ -0,0 +1,131 @@
...
+  defaultFinancialCodeFilter,
+  IFinancialCodeFilter,
+} from './FinancialCodeFilter/FinancialCodeFilter';
+import { FinancialCodeResults } from './FinancialCodeResults/FinancialCodeResults';
+import * as Styled from './styles';
+
+/**
+ * Page that displays acquisition files information.
+ */
+export const FinancialCodeListView: React.FC = () => {
```
devinleighsmith, line 25: test coverage on sorting/filtering could use some improvement.

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
asanchezr: Happy to change but was following pattern set in **ActivityList**

image

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
asanchezr: Will have to see if there's a pattern already for testing sorting and filtering on list views. I think what I've seen is that we test the sub-components (results and filter) but will take a look

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
```diff
@@ -0,0 +1,131 @@
...
+            return false;
+          }
+          // Finally return true because record has satisfied all conditions
+          return true;
+        });
+      }
+
+      if (sort) {
+        const sortFields = Object.keys(sort);
+        if (sortFields?.length > 0) {
```
devinleighsmith, line 76: This doesn't seem to correspond to this requirement: Display the results sorted by Code type, sort order and in absence of sort order Code value

devinleighsmith (changes requested):
two comments require being addressed:
sort rules missing from AC.
sort of code when numeric?

asanchezr:
> two comments require being addressed: sort rules missing from AC. sort of code when numeric?

I will check into the sort code AC.
The sorting when values "look like numeric values" needs a larger discussion as the DB is being modified to all be string values

source/frontend/src/features/admin/financial-codes/list/FinancialCodeListView.tsx
asanchezr: I checked across all existing list views and couldn't find tests covering sorting/filtering (at the list view level). I agree this area should be improved but given that there's no existing pattern to follow this should go into a tech-debt task to add coverage for all list views not just this one

devinleighsmith (approved):

PR #2649, opened 2023-01-16
Fixed selected feature file property

FuriousLlama:


source/frontend/src/features/properties/map/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -48,10 +48,7 @@ export const AddAcquisitionContainer: React.FC<
         PropertyForm.fromMapProperty(mapFeatureToProperty(selectedFileFeature)),
       ]);
     }
-    return () => {
-      setState({ type: MapStateActionTypes.SELECTED_FILE_FEATURE, selectedFileFeature: null });
```
devinleighsmith, line 52: With this removed, if you:
1) create an acq from the map
2) either create or cancel that acq file creation.
3) create a new acq file using the left hand bar (not on the map), the selectedFileFeature from the step #1 is still present and so the new acq file that should have no properties has the property from #1

devinleighsmith (changes requested):
this fixes the bug but introduces another.

devinleighsmith (approved):

devinleighsmith:
seems like a good solution.

PR #2662, opened 2023-01-17
psp-3986 cache list of lookups.

devinleighsmith:


source/frontend/src/customAxios.ts
```diff
@@ -42,6 +42,7 @@ export const CustomAxios = ({
     baseURL,
     headers: {
       'Access-Control-Allow-Origin': '*',
+      'Cache-Control': process.env.NODE_ENV === 'development' ? 'no-cache' : '',
```
devinleighsmith, line 45: In development, don't cache anything as we may be changing lookup codes around a bunch.

source/backend/api/Controllers/LookupController.cs
```diff
@@ -76,102 +83,112 @@ public IActionResult GetPropertyClassificationTypes()
...
-            codes.AddRange(documentStatusTypes);
-            codes.AddRange(acquisitionFileStatusTypes);
-            codes.AddRange(acquisitionPhysFileStatusTypes);
-            codes.AddRange(acquisitionTypes);
-            codes.AddRange(activityTemplateTypes);
-            codes.AddRange(activityStatusTypes);
-            codes.AddRange(acqFilePersonProfileTypes);
-            codes.AddRange(tenantTypes);
-            codes.AddRange(acqFundingTypes);
+            if (!_memoryCache.TryGetValue(CacheKeys.Lookup, out JsonResult cachedLookupResponse))
```
devinleighsmith, line 90: Store this result in memory, this should make all future requests for the lookup codes very fast and reduce need for db transactions until the server is restarted.

source/backend/api/Controllers/LookupController.cs
```diff
@@ -76,102 +83,112 @@ public IActionResult GetPropertyClassificationTypes()
         [Produces("application/json")]
         [ProducesResponseType(typeof(IEnumerable<Model.LookupModel>), 200)]
         [SwaggerOperation(Tags = new[] { "lookup" })]
+        [ResponseCache(Duration = OneHourInSeconds)]
```
devinleighsmith, line 86: Could use some feedback here. My worry is that a user could: 1) Log into the app 2) get a list of lookup codes 3) we do a database update 4) the list of codes is now out of date.

The risk of this in prod seem low but I can see this happening in our lower environments. Perhaps this is more trouble then its worth?

source/backend/api/Controllers/LookupController.cs
```diff
@@ -76,102 +83,112 @@ public IActionResult GetPropertyClassificationTypes()
...
+                codes.AddRange(acquisitionTypes);
+                codes.AddRange(activityTemplateTypes);
+                codes.AddRange(activityStatusTypes);
+                codes.AddRange(acqFilePersonProfileTypes);
+                codes.AddRange(tenantTypes);
+                codes.AddRange(acqFundingTypes);
+
+                var response = new JsonResult(codes);
+
+                _memoryCache.Set(CacheKeys.Lookup, response);
```
devinleighsmith, line 187: store the result in memory if not already present.

source/backend/api/Controllers/LookupController.cs
FuriousLlama: It really depends on how often the codes get updated. AFAIK they require a migration, so not very often. I think it should be ok to cache them for an hour.

FuriousLlama (approved):

asanchezr (approved):

PR #2684, opened 2023-01-19
make all ltsa options configurable via environment variables. (#2667)

devinleighsmith:


asanchezr:
@devinleighsmith - don't we also need to bump the hotfix number for this?

asanchezr (approved):

asanchezr (approved):

PR #2699, opened 2023-01-20
PSP-4460 create financial code

asanchezr:
**NOTE** - working on unit tests. Will add them to this pr

image

### DIFF
--- a/source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
+++ b/source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
@@ -1,8 +1,12 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
+using Pims.Core.Exceptions;
+using Pims.Core.Extensions;
+using Pims.Dal.Constants;
 using Pims.Dal.Entities;
 
 namespace Pims.Dal.Repositories
@@ -36,6 +40,27 @@ namespace Pims.Dal.Repositories
         {
             return this.Context.PimsBusinessFunctionCodes.AsNoTracking()
                 .ToList();
+        }
+
+        public PimsBusinessFunctionCode Add(PimsBusinessFunctionCode pimsCode)
+        {
+            pimsCode.ThrowIfNull(nameof(pimsCode));
+
+            // Check for uniqueness on save. At any given point an active code should be unique per code type.
+            var existingCodes = Context.PimsBusinessFunctionCodes.Where(
+                    c => EF.Functions.Collate(c.Code, SqlCollation.LATIN_GENERAL_CASE_INSENSITIVE) == pimsCode.Code)
+                .ToList();
+
+            // Active codes have no expiry date or they expire in the future.
+            var now = DateTime.UtcNow;
+            var isDuplicate = existingCodes.Any(c => !c.ExpiryDate.HasValue || c.ExpiryDate?.Date > now.Date);
+            if (isDuplicate)
+            {
+                throw new DuplicateEntityException("Duplicate business function code found");
+            }
+
+            Context.PimsBusinessFunctionCodes.Add(pimsCode);
+            return pimsCode;
         }
 
         #endregion
--- a/source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
+++ b/source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
@@ -51,10 +51,12 @@ namespace Pims.Dal.Repositories
                     c => EF.Functions.Collate(c.Code, SqlCollation.LATIN_GENERAL_CASE_INSENSITIVE) == pimsCode.Code)
                 .ToList();
 
-            // Active codes have no expiry date or they expire in the future.
             var now = DateTime.UtcNow;
+            var newCodeIsActive = !pimsCode.ExpiryDate.HasValue || pimsCode.ExpiryDate?.Date > now.Date;
+
+            // Active codes have no expiry date or they expire in the future.
             var isDuplicate = existingCodes.Any(c => !c.ExpiryDate.HasValue || c.ExpiryDate?.Date > now.Date);
-            if (isDuplicate)
+            if (newCodeIsActive && isDuplicate)
             {
                 throw new DuplicateEntityException("Duplicate business function code found");
             }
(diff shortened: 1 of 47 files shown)


devinleighsmith:
reviewing

source/backend/tests/unit/api/Services/FinancialCodeServiceTest.cs
```diff
@@ -0,0 +1,173 @@
...
+            repo5.Setup(x => x.GetAllFinancialActivityCodes()).Returns(new List<PimsFinancialActivityCode>());
+            repo6.Setup(x => x.GetAllWorkActivityCodes()).Returns(new List<PimsWorkActivityCode>());
+            repo7.Setup(x => x.GetAllResponsibilityCodes()).Returns(new List<PimsResponsibilityCode>());
+
+            // Act
+            Action act = () => service.GetAllFinancialCodes();
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+            repo1.Verify(x => x.GetAllBusinessFunctionCodes(), Times.Never);
```
devinleighsmith, line 92: with the new pattern we are moving away from mocking/verifying when we are throwing an exception that would cause a mock not to be used.

source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
```diff
@@ -38,6 +42,29 @@ public IList<PimsBusinessFunctionCode> GetAllBusinessFunctionCodes()
...
+                .ToList();
+
+            var now = DateTime.UtcNow;
+            var newCodeIsActive = !pimsCode.ExpiryDate.HasValue || pimsCode.ExpiryDate?.Date > now.Date;
+
+            // Active codes have no expiry date or they expire in the future.
+            var isDuplicate = existingCodes.Any(c => !c.ExpiryDate.HasValue || c.ExpiryDate?.Date > now.Date);
+            if (newCodeIsActive && isDuplicate)
+            {
+                throw new DuplicateEntityException("Duplicate business function code found");
```
devinleighsmith, line 61: Is there a repo test that checks for this?

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeContainer.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+      addFinancialCode: mockApi,
+    };
+  },
+}));
+
+const history = createMemoryHistory();
+
+let viewProps: IAddFinancialCodeFormProps | undefined;
+
+const TestView: React.FC<IAddFinancialCodeFormProps> = props => {
```
devinleighsmith, line 27: nice.

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeContainer.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    expect(getByText(/Cannot create duplicate financial code/)).toBeVisible();
+  });
+
+  it('displays a toast for generic server errors', async () => {
+    setup();
+
+    await act(async () => {
+      const error500 = createAxiosError(500);
+      viewProps?.onError(error500);
```
devinleighsmith, line 67: having direct access to the props seems nice, looks like it is easier to write concise tests.

source/backend/tests/unit/api/Services/FinancialCodeServiceTest.cs
asanchezr: Can you please point to specific example of new pattern? I remember we talked about this but haven't seen the new pattern in action. In my tests I have tried to trim down the mocking/checking but seemed useful in this case to check that no underlying repos where called.

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeForm.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    expect(mockProps.onCancel).toHaveBeenCalled();
+    expect(description).toHaveTextContent('');
+  });
+
+  it('does not call onCancel when form is dirty, cancel is clicked but modal is not confirmed', async () => {
+    setup();
+
+    const textbox = document.querySelector(`input[name="description"]`) as HTMLInputElement;
+    const cancelButton = screen.getByText('Cancel');
+    await act(() => userEvent.paste(textbox, `another description`));
```
devinleighsmith, line 99: haven't seen this before, whats the advantage of using paste here? Should we update our fill input method to use this?

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeContainer.test.tsx
asanchezr: Yeah, I like this new pattern. It allows to "ignore" which button was clicked in the underlying form as that is no the concern of the container. Happy to demo it this Friday as I still see new PRs not using this pattern

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeForm.test.tsx
asanchezr: Personal preference here. I kinda dislike the fill method a bit as I have found issues in the past or if you forget to specify "textarea" as the last parameter it doesn't work for text areas. I would prefer if we could try to use userEvent instead. 
And yes, we should update/refactor our userEvent

asanchezr: Also I use "paste" instead of `userEvent.type` because I'm not interested in waiting for 10 (or 20) keystrokes for eg: "sample description".

source/backend/dal/Repositories/BusinessFunctionCodeRepository.cs
asanchezr: I tried to create a test but could not get it to work because the InMemory database used in tests does not support EF functions such as Collate (which we use in various repos). 

See: https://stackoverflow.com/questions/70693625/the-collate-method-is-not-supported-because-the-query-has-switched-to-client-e

asanchezr: We could refactor this repo (and others that use EF functions) to be more testable but I wonder about the ROI

source/backend/tests/unit/api/Services/FinancialCodeServiceTest.cs
devinleighsmith: have a look at my PR, most of the usages of Times.Never have been removed.

source/frontend/src/features/admin/financial-codes/add/AddFinancialCodeContainer.test.tsx
devinleighsmith: If you see any new containers not using dependency injection please block them from being merged. (changes required)

PR #2712, opened 2023-01-24
Psp 5267 - unit tests

devinleighsmith:
Based on our discussion two weeks ago, this PR removes unnecessary asserts, objects and mocks that aren't strictly required for a test. Also some minor fixes to correct service vs repository tests that were in the wrong location or were named incorrectly.


source/backend/tests/unit/api/Services/AcquisitionServiceTest.cs
```diff
@@ -51,42 +62,30 @@ public void Add_Success()
...
-            var repository = helper.GetService<Mock<IAcquisitionFileRepository>>();
-            repository.Setup(x => x.Add(It.IsAny<PimsAcquisitionFile>())).Returns(acqFile);
-
             // Act
             Action act = () => service.Add(acqFile);
 
             // Assert
             act.Should().Throw<NotAuthorizedException>();
-            repository.Verify(x => x.Add(It.IsAny<PimsAcquisitionFile>()), Times.Never);
         }
```
asanchezr, line 74: I see now what the new pattern is regarding not mocking repos when exceptions are expected to be thrown but I believe there is a subtle case of a bug that would be not longer covered by this check if the service code throws **after** calling the repo. This could happen due to developer error.

Say, this contrived example:

```cs
// poorly written service code
public PimsEntity AddWithBug(entity)
{
   var added = _repo.Add(entity);
   User.ThrowIfNotAuthorized(Permissions.SystemAdmin);
}
```

In that case the tests above in the new pattern would all pass because the service eventually throws but wouldn't we want to also make sure that no repo is called as well?

asanchezr: By not checking for repo calls we are implicitly trusting the developers to write the auth check (or any other check that results in an exception) before any calls to the repos

source/backend/tests/unit/api/Services/AcquisitionServiceTest.cs
```diff
@@ -95,13 +94,12 @@ public void Add_ThrowIfNull()
         public void GetById_Success()
         {
             // Arrange
-            var helper = new TestHelper();
             var user = PrincipalHelper.CreateForPermission(Permissions.AcquisitionFileView);
-            var service = helper.Create<AcquisitionFileService>(user);
+            var service = _helper.Create<AcquisitionFileService>(user);
```
asanchezr, line 98: shouldn't we use **CreateAcquisitionServiceWithPermissions** here instead?

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -27,18 +27,31 @@ namespace Pims.Api.Test.Services
...
+            return _helper.Create<ProjectService>();
+        }
+
         [Fact]
         public void Search_Success()
         {
             // Arrange
-            var helper = new TestHelper();
+            var service = CreateProjectServiceWithPermissions(Permissions.ProjectView);
             var user = PrincipalHelper.CreateForPermission(Permissions.ProjectView);
```
asanchezr, line 49: seems like we need to remove user here - right?

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -52,18 +65,13 @@ public void Search_Success()
...
-
-            var repository = helper.GetService<Mock<IProjectRepository>>();
+            var service = CreateProjectServiceWithPermissions();
 
             // Act
             Action act = () => service.SearchProjects("some string", 1);
 
             // Assert
             act.Should().Throw<NotAuthorizedException>();
-            repository.Verify(x => x.SearchProjects(It.IsAny<string>(), It.IsAny<int>()), Times.Never);
```
asanchezr, line 66: There are a few tests below that would need to follow new pattern. They are still asserting **Times.Never** on repo usage when exceptions are thrown

source/backend/tests/unit/api/Services/PropertyServiceTest.cs
```diff
@@ -23,22 +23,33 @@ namespace Pims.Api.Test.Services
...
     public class PropertyServiceTest
     {
+        private TestHelper _helper;
+
+        public PropertyServiceTest()
+        {
+            _helper = new TestHelper();
+        }
+
+        private PropertyService CreateProjectServiceWithPermissions(params Permissions[] permissions)
```
asanchezr, line 33: nit: please rename to appropriate class **CreatePropertyServiceWithPermissions** or what do you think of using the  same generic name on all service tests - **CreateWithPermissions** ?

I personally like the 2nd option a bit better and there's no risk of name override because the method is private - thoughts?

source/backend/tests/unit/api/Services/ResearchFileServiceTest.cs
```diff
@@ -23,21 +24,31 @@ namespace Pims.Api.Test.Services
...
     public class ResearchFileServiceTest
     {
+        private TestHelper _helper;
+
+        public ResearchFileServiceTest()
+        {
+            _helper = new TestHelper();
+        }
+
+        private ResearchFileService CreateProjectServiceWithPermissions(params Permissions[] permissions)
```
asanchezr, line 34: please update method name

asanchezr (approved):

source/backend/tests/unit/api/Services/AcquisitionServiceTest.cs
devinleighsmith: This is fair, there is some value to having the non-call check in the case of writes, I'll update this to keep those checks for writes but leave them out for reads as that will always be redundant.

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
devinleighsmith: Right, these likely came from the merge. Will update.

source/backend/tests/unit/api/Services/PropertyServiceTest.cs
devinleighsmith: No strong preference.

PR #2719, opened 2023-01-25
psp-5258 reduce toast messages.

devinleighsmith:


source/frontend/src/constants/toasts.ts
```diff
@@ -23,74 +23,6 @@ export const user = {
   USER_ERROR,
 };
 
-/** These toasts are used by the parcel detail api */
```
devinleighsmith, line 26: dead code

asanchezr (approved):

PR #2726, opened 2023-01-26
psp-5284 make all layer urls configurable.

devinleighsmith:


source/frontend/src/features/mapSideBar/hooks/usePropertyDetails.ts
```diff
@@ -16,14 +8,22 @@ import {
   IPropertyDetailsForm,
   toFormValues,
 } from '../tabs/propertyDetails/detail/PropertyDetailsTabView.helpers';
+import { useTenant } from './../../../tenants/useTenant';
```
asanchezr, line 11: nit: please use `import { useTenant } from 'tenants';`

source/frontend/src/hooks/useQueryMapLayersByLocation.ts
```diff
@@ -32,11 +26,19 @@ export interface IMapLayerResults {
...
+    electoralLayerUrl,
+    alrLayerUrl,
+    reservesLayerUrl,
+  } = useTenant();
+
+  const motiRegionService = useLayerQuery(motiRegionLayerUrl);
+  const highwaysService = useLayerQuery(hwyDistrictLayerUrl);
+  const electoralService = useLayerQuery(electoralLayerUrl);
+  const alrService = useLayerQuery(alrLayerUrl);
+  const firstNationsService = useLayerQuery(reservesLayerUrl);
```
asanchezr, line 41: This whole sections is been highlighted by codecov as not covered by tests. I recall we discussed trying to use pr review for adding tests where they are valuable. Are they valuable here?

asanchezr: image

source/frontend/public/tenants/tenant.json
```diff
@@ -14,7 +14,7 @@
   "layers": [],
   "propertiesUrl": "ogs-internal/ows?service=wfs&request=GetFeature&typeName=PIMS_PROPERTY_LOCATION_VW&outputformat=json&srsName=EPSG:4326&version=2.0.0&",
   "parcelMapFullyAttributed": {
-    "url": "https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW/ows",
+    "url": "https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW_TEST/ows",
     "name": "pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW"
   }
```
asanchezr, line 19: I believe this whole json file gets replaced in live environments via a config-map but shouldn't this placeholder json file have the new urls too?

asanchezr (approved):

source/frontend/src/hooks/useQueryMapLayersByLocation.ts
devinleighsmith: Test coverage by this story doesn't seem appropriate since this story is enhancing existing functionality that has no tests. Will log test coverage story since that coverage has never existed.

source/frontend/public/tenants/tenant.json
devinleighsmith: Any missing keys get filled in by the default object, so note required here.

PR #2727, opened 2023-01-26
correct note delete behaviour.

devinleighsmith:


source/backend/dal/Repositories/EntityNoteRepository.cs
```diff
@@ -87,7 +87,7 @@ public bool DeleteActivityNotes(long entityId)
 
         public bool DeleteAcquisitionFileNotes(long entityId)
         {
-            var acquisitionFileNotes = this.Context.PimsAcquisitionFileNotes.Include(ai => ai.Note).Where(x => x.AcquisitionFileId == entityId).ToList();
+            var acquisitionFileNotes = this.Context.PimsAcquisitionFileNotes.Include(ai => ai.Note).Where(x => x.NoteId == entityId).ToList();
```
asanchezr, line 90: nit: given that it wasn't clear which entity we were targeting with the parameter **entityId**, a comment or function documentation would be appreciated to avoid future mistakes

asanchezr (approved):

PR #2729, opened 2023-01-26
PSP-5037 : View / Read Project Details

eddherrera:
image


source/frontend/src/features/projects/detail/projectPages/details/ProjectDetailInformation.tsx
```diff
@@ -0,0 +1,17 @@
...
+  nameSpace?: string;
+}
+
+export const ProjectDetailInformation: React.FunctionComponent<
+  React.PropsWithChildren<IProjectDetailInformationProps>
+> = ({ nameSpace }) => {
+  const { values } = useFormikContext<IProjectForm>();
+  const summary = getIn(values, withNameSpace(nameSpace, 'note'));
+
+  return <Section>This works!{summary}</Section>;
```
asanchezr, line 16: Are we sure we want to display "This works!" ?

source/frontend/src/features/projects/detail/projectPages/details/ProjectDetailsForm.tsx
```diff
@@ -0,0 +1,30 @@
...
+export interface IProjectDetailsFormProps {
+  isEditing?: boolean;
+}
+
+const ProjectDetailsForm: React.FunctionComponent<
+  React.PropsWithChildren<IProjectDetailsFormProps>
+> = () => {
+  const { project } = React.useContext(ProjectStateContext);
+
+  console.log(project);
```
asanchezr, line 18: please remove console.log

source/frontend/src/features/properties/map/MapRouter.tsx
```diff
@@ -182,6 +183,16 @@ export const MapRouter: React.FunctionComponent<IMapRouterProps> = memo(props =>
...
+      <AppRoute
+        path={`/mapview/sidebar/project/:id`}
+        customRender={({ match }) => (
+          <ProjectContainer projectId={Number(match.params.id)} onClose={onClose} />
+        )}
+        claim={Claims.PROJECT_VIEW}
+        exact
+        key={'Project'}
+        title={'Project'}
+      />
```
asanchezr, line 195: Please use the dependency injection pattern for Containers -> Views
See an example in **AddFinancialCodeContainer**. We are moving to a new pattern where the container gets the inner view injected via props to help with testing and decoupling.

```tsx
          <AppRoute
            protected
            exact
            path="/admin/financial-code/new"
            customRender={() => (
              <AuthLayout>
                <AddFinancialCodeContainer View={AddFinancialCodeForm} />
              </AuthLayout>
            )}
            role={Roles.SYSTEM_ADMINISTRATOR}
            title={getTitle('Create Financial Code')}
          ></AppRoute>
```

Notice how we "inject" the formik form into the container above. This allows us to create simpler forms for testing so the container tests don't have to rely on knowing about the inner children

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -40,6 +40,28 @@ public ProjectController(IProjectService projectService, IMapper mapper)
...
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public async Task<IActionResult> GetById([FromRoute] long id)
+        {
+            var project = await _projectService.GetById(id);
+            if (project is null)
+            {
+                return NotFound();
+            }
+
+            return Ok(_mapper.Map<ProjectModel>(project));
```
asanchezr, line 62: Just double-checking that calling the method **Ok()** here is equivalent to returning **JsonResult** ?

asanchezr: According to this [link](https://learn.microsoft.com/en-us/aspnet/core/web-api/advanced/formatting?view=aspnetcore-7.0) there is a subtle difference in returning Ok in that the client can negotiate the return type via the **Accept** header in the request. I'm not opposed to start using Ok but would like that we all agree in a common pattern as I see a mix of Ok and JsonResult even in this same file

source/backend/dal/Repositories/ProjectRepository.cs
```diff
@@ -72,17 +72,17 @@ public Task<Paged<PimsProject>> GetPageAsync(ProjectFilter filter)
...
-            User.ThrowIfNotAuthorized(Permissions.PropertyView);
+            User.ThrowIfNotAuthorized(Permissions.ProjectView);
 
             return await Context.PimsProjects
                     .AsNoTracking()
                     .Include(x => x.ProjectStatusTypeCodeNavigation)
                     .Include(x => x.RegionCodeNavigation)
-                    .Where(x => x.Id == id)
+                    .Where(x => x.Id == projectId)
                     .FirstOrDefaultAsync();
```
asanchezr, line 86: Also not opposed to using **async** DB operations but I think we were not using that previously. Is performance better or is just dev preference to use async EF methods? Would like to sync up with other devs so we so the same

asanchezr (changes requested):
Please update frontend container to follow new practice of dependency injection

source/frontend/src/features/projects/detail/projectPages/details/ProjectDetailInformation.tsx
eddherrera: Removed

source/frontend/src/features/projects/detail/projectPages/details/ProjectDetailsForm.tsx
eddherrera: removed

source/frontend/src/features/properties/map/MapRouter.tsx
eddherrera: Updated

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
eddherrera: I think we should be making an explicit response for the Api, not the content type.

source/backend/dal/Repositories/ProjectRepository.cs
eddherrera: thread should perform a little better but overall, not a big win. Async feels a little more modern approach, but we could keep not using it overall.

devinleighsmith:
@asanchezr can you re-review these changes today and clear your flag please?

asanchezr:
> @asanchezr can you re-review these changes today and clear your flag please?

Will do

source/backend/dal/Repositories/ProjectRepository.cs
FuriousLlama: Let's not do this. The request is spawned on its own thread already. So far all calls to this method are just waiting on it, making the async, not a worth having

FuriousLlama: @eddherrera My PR will change this so no need to worry as we chatted earlier

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -196,5 +196,42 @@ public async void Add_Project_Success()
...
+            var service = helper.Create<ProjectService>(user);
+
+            var repository = helper.GetService<Mock<IProjectRepository>>();
+
+            // Act
+            Func<Task> act = async () => await service.GetById(It.IsAny<long>());
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+            repository.Verify(x => x.Get(It.IsAny<long>()), Times.Never);
```
asanchezr, line 215: We are no longer asserting that repos are never called when exception was thrown by service, please remove this line

source/frontend/src/features/properties/map/project/add/models.ts
```diff
@@ -10,13 +10,13 @@ export class ProjectForm {
   region?: NumberFieldValue;
   summary?: string;
 
-  toApi(): Api_Project {
+  toApi(id: number = 1): Api_Project {
```
asanchezr, line 13: sorry, not sure I understand the need to pass the Id here.

source/frontend/src/features/properties/map/project/ProjectContainer.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import MockAdapter from 'axios-mock-adapter';
+import { mockLookups } from 'mocks';
+import { mockProjectGetResponse } from 'mocks/mockProjects';
+import { lookupCodesSlice } from 'store/slices/lookupCodes';
+import { render, RenderOptions } from 'utils/test-utils';
+
+import { SideBarContextProvider } from '../context/sidebarContext';
+import ProjectContainer, { IProjectContainerViewProps } from './ProjectContainer';
+
+const mockAxios = new MockAdapter(axios);
```
asanchezr, line 11: Please do not mock axios. In new code we should mock the providers instead (useProjectProvider)

source/frontend/src/features/properties/map/project/ProjectContainerView.tsx
```diff
@@ -0,0 +1,59 @@
...
+import MapSideBarLayout from 'features/mapSideBar/layout/MapSideBarLayout';
+import { useCallback } from 'react';
+import { FaBriefcase } from 'react-icons/fa';
+import styled from 'styled-components';
+
+import ProjectHeader from './common/ProjectHeader';
+import { IProjectContainerViewProps } from './ProjectContainer';
+import ViewSelector from './ViewSelector';
+
+const ProjectContainerView: React.FC<IProjectContainerViewProps> = ({
```
asanchezr, line 11: This view seems to need unit tests

source/frontend/src/features/properties/map/project/add/AddProjectContainer.test.tsx
```diff
@@ -78,42 +84,73 @@ describe('AddProjectContainer component', () => {
...
+    const { getByText, getNameTextbox, getRegionDropdown, getNumberTextbox, getStatusDropdown } =
+      setup();
 
     const formTitle = getByText(/Create Project/i);
-    const input = getNameTextbox();
-    const select = getRegionDropdown();
+    const nameInput = getNameTextbox();
+    const numberInput = getNumberTextbox();
+    const selectRegion = getRegionDropdown();
+    const selectSatus = getStatusDropdown();
```
asanchezr, line 94: nit: please fix typo selectS(t)atus

source/frontend/src/features/properties/map/project/common/ProjectHeader.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Api_Project } from 'models/api/Project';
+import { Col, Row } from 'react-bootstrap';
+import styled from 'styled-components';
+import { prettyFormatDate } from 'utils';
+
+export interface IProjectHeaderProps {
+  project?: Api_Project;
+}
+
+const ProjectHeader: React.FunctionComponent<React.PropsWithChildren<IProjectHeaderProps>> = ({
```
asanchezr, line 12: Header component should have tests

source/frontend/src/features/properties/map/project/update/UpdateProjectContainer.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export interface IUpdateProjectContainerProps {
+  initialValues: ProjectForm;
+  projectStatusOptions: SelectOption[];
+  projectRegionOptions: SelectOption[];
+  validationSchema?: any | (() => any);
+  onSubmit: (values: ProjectForm, formikHelpers: FormikHelpers<ProjectForm>) => void | Promise<any>;
+}
+
+const UpdateProjectContainer = React.forwardRef<FormikProps<any>, IUpdateProjectContainerProps>(
```
asanchezr, line 19: This doesn't seem relevant to this PR but I guess is fine to keep it since you will working with this in your next story/PR

source/frontend/src/mocks/mockProjects.ts
```diff
@@ -33,22 +33,52 @@ export const mockProjects = (): Api_Project[] => [
 
 export const mockProjectPostResponse = (
   id: number = 1,
```
asanchezr, line 35: Not sure we want to send Id = 1 to create a new project. I believe the existing pattern is to leave Id undefined for POSTs

source/frontend/src/models/api/Project.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export interface Api_Project extends Api_ConcurrentVersion, Api_AuditFields {
   id?: number;
   projectStatusTypeCode?: Api_TypeCode<string>;
-  businessFunctionCode?: any; // TODO: Match new code types
-  costTypeCode?: any; // TODO: Match new code types
-  workActivityCode?: any; // TODO: Match new code types
+  businessFunctionCode?: Api_CodeType;
+  costTypeCode?: Api_CodeType;
+  workActivityCode?: Api_CodeType;
```
asanchezr, line 11: They look almost identical but I believe Api_CodeType is specific to Country and Provinces. It looks that here the more relevant model would be Api_FinancialCode which was created for business function, cost type, etc

asanchezr (approved):

source/frontend/src/features/properties/map/project/common/ProjectHeader.tsx
asanchezr: a single snapshot test would be sufficient

PR #2732, opened 2023-01-27
ETL: PSP - 5273,5374,5275,5276,5277,5278,5279 - Getting all Codes from PAIMS to PIMS

PryancaJSharma:
ETL: PSP - 5273,5374,5275,5276,5277,5278,5279 - Getting all Codes from PAIMS to PIMS


PryancaJSharma:
Created another PR to fix this one

PR #2750, opened 2023-01-31
PSP-4462 Update financial code

asanchezr:
image


source/backend/api/Services/FinancialCodeService.cs
```diff
@@ -133,7 +152,93 @@ public FinancialCodeModel Add(FinancialCodeTypes type, FinancialCodeModel model)
...
+        }
+
+        public FinancialCodeModel Update(FinancialCodeTypes type, FinancialCodeModel model)
+        {
+            model.ThrowIfNull(nameof(model));
+
+            _logger.LogInformation("Updating financial code with type {type} and model {model}", type, model);
+            User.ThrowIfNotAuthorized(Permissions.SystemAdmin);
+
+            ValidateVersion(type, model.Id, model.RowVersion);
```
devinleighsmith, line 166: What is the benefit of checking this manually? The db will throw a dbconcurrency exception if the rowversion is not updated.

devinleighsmith:
Looks like this is missing tests on the backend. Either need that for this story or need tech debt logged to cover missing coverage.

asanchezr:
> Looks like this is missing tests on the backend. Either need that for this story or need tech debt logged to cover missing coverage.

As mentioned in standup I'm working in them but the rest of the code is reviewable

source/backend/api/Services/FinancialCodeService.cs
asanchezr: Well, this pattern is pervasive across other services. If we want to abandon it then we should do it across all of them. Topic for Friday chat I think

asanchezr: This is a snap of another service doing the check.
image

FuriousLlama: I think the original intent of that method was to be used by child entities to make sure the parent was in sync

devinleighsmith (approved):

PR #2762, opened 2023-02-01
Psp 3914 documents for research and acquisition files.

devinleighsmith:


source/frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -47,8 +48,7 @@ const DocumentListContainer: React.FunctionComponent<
     if (documents !== undefined && isMounted()) {
       setDocumentResults([
         ...documents
-          .map(x => x.document)
-          .filter((x): x is Api_Document => !!x)
+          .filter((x): x is Api_DocumentRelationship => !!x?.document)
```
devinleighsmith, line 51: @FuriousLlama this seems odd to me. The onDelete and other methods seem to be assuming that an Api_DocumentRelationship is what each row is based on, but it used to be that each DocumentRow was equivalent to an Api_Document, and there is no way of going from a DocumentRow to an Api_DocumentRelationship

FuriousLlama: Looks like the types for these are not being validated.
image

My guess is that the original list only cared about documents and not document relationships which makes sense. So if we want to keep that distinction the onDelete callback would need to find the documentrelationship based on the document id. Otherwise need to convert the list to work with relationshipdocs

devinleighsmith: hmm, ok. I think right now we never delete a document without a relationship but maybe we will need this in the future. Sounds like a tech debt item.

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -0,0 +1,156 @@
...
+{
+    /// <summary>
+    /// DocumentFileService implementation provides document managing capabilities for files.
+    /// </summary>
+    public class DocumentFileService : BaseService, IDocumentFileService
+    {
+        private readonly IAcquisitionFileDocumentRepository acquisitionFileDocumentRepository;
+        private readonly IResearchFileDocumentRepository researchFileDocumentRepository;
+        private readonly IDocumentService documentService;
+        private readonly IMapper mapper;
```
FuriousLlama, line 26: very nit: underscore for local private variables.

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -0,0 +1,156 @@
...
+            where T : PimsFileDocument
+        {
+            this.Logger.LogInformation("Retrieving PIMS documents related to the file of type $fileType", fileType);
+            this.User.ThrowIfNotAuthorized(Permissions.DocumentView);
+
+            switch (fileType)
+            {
+                case FileType.Research:
+                    this.User.ThrowIfNotAuthorized(Permissions.ResearchFileView);
+                    return researchFileDocumentRepository.GetAllByResearchFile(fileId).Select(f => f as T).ToArray();
```
FuriousLlama, line 53: looks fine, but I wonder if this violates the type check safety

source/backend/tests/unit/dal/Repositories/AcquisitionFileDocumentRepositoryTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var user = PrincipalHelper.CreateForPermission(Permissions.DocumentView);
+
+            var acquisitionFileDocument = new PimsAcquisitionFileDocument();
+
+            var repository = helper.CreateRepository<AcquisitionFileDocumentRepository>(user);
+
+            // Act
+            var result = repository.AddAcquisition(acquisitionFileDocument);
+
+            // Assert
```
FuriousLlama, line 31: no assert :(

source/backend/tests/unit/dal/Repositories/AcquisitionFileDocumentRepositoryTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var user = PrincipalHelper.CreateForPermission(Permissions.DocumentView);
+
+            var acquisitionFileDocument = new PimsAcquisitionFileDocument();
+
+            var repository = helper.CreateRepository<AcquisitionFileDocumentRepository>(user);
+
+            // Act
+            var result = repository.DeleteAcquisition(acquisitionFileDocument);
+
+            // Assert
```
FuriousLlama, line 48: not asserting anything

source/backend/tests/unit/dal/Repositories/ResearchFileRepositoryTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var user = PrincipalHelper.CreateForPermission(Permissions.DocumentView);
+
+            var ResearchFileDocument = new PimsResearchFileDocument();
+
+            var repository = helper.CreateRepository<ResearchFileDocumentRepository>(user);
+
+            // Act
+            var result = repository.AddResearch(ResearchFileDocument);
+
+            // Assert
```
FuriousLlama, line 31: same as above

source/backend/tests/unit/dal/Repositories/ResearchFileRepositoryTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var user = PrincipalHelper.CreateForPermission(Permissions.DocumentView);
+
+            var ResearchFileDocument = new PimsResearchFileDocument();
+
+            var repository = helper.CreateRepository<ResearchFileDocumentRepository>(user);
+
+            // Act
+            var result = repository.DeleteResearch(ResearchFileDocument);
+
+            // Assert
```
FuriousLlama, line 48: same as above

source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -28,27 +30,40 @@ export class DocumentRow {
   isFileAvailable?: boolean;
   appCreateTimestamp?: string;
   appCreateUserid?: string;
+  relationshipId?: number;
+  relationshipType?: DocumentRelationshipType;
+  parentId?: number;
 
-  public static fromApi(document: Api_Document): DocumentRow {
+  public static fromApi(relationship: Api_DocumentRelationship): DocumentRow {
```
FuriousLlama, line 37: just to keep in mind. Not sure if this is leaking abstraction given the row would only care about the document information vs also having the relationship information.

source/frontend/src/features/properties/map/acquisition/detail/AcquisitionDocumentsTab.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <>
+      <DocumentListContainer
+        title="File Documents"
+        parentId={acquisitionFileId}
+        relationshipType={DocumentRelationshipType.ACQUISITION_FILES}
+      />
+      <DocumentListContainer
+        title="Activity Documents"
+        parentId={acquisitionFileId}
+        relationshipType={DocumentRelationshipType.ACQUISITION_FILE_ACTIVITIES}
```
FuriousLlama, line 22: flexible!

FuriousLlama (approved):
Some tests not asserting things, otherwise looks good!

source/backend/tests/unit/dal/Repositories/AcquisitionFileDocumentRepositoryTest.cs
devinleighsmith: oh whoops

devinleighsmith:
@FuriousLlama nooo I hadn't fixed those tests up yet :-( thats ok, will make another PR for that, doesn't really need to get merged tonight.

PR #2764, opened 2023-02-01
Product Create and view |  psp-4515

FuriousLlama:
image

image


source/backend/api/Areas/Products/Controllers/ProductController.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="mapper"></param>
+        ///
+        public ProductController(IProjectService projectService, IMapper mapper)
+        {
+            _projectService = projectService;
+            _mapper = mapper;
+        }
+
+        /// <summary>
+        /// Retrieves a the acquisition file for the given product.
```
devinleighsmith, line 41: nit: a the

devinleighsmith: Should this require view acq file claim as well?

source/backend/api/Areas/Products/Controllers/ProductController.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        /// <param name="productId"></param>
+        /// <returns>An array of contacts matching the filter.</returns>
+        [HttpGet("{productId}/acquisitionFiles")]
+        [HasPermission(Permissions.ProjectView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(AcquisitionFileModel), 200)]
+        [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "product" })]
+        public IActionResult GetFiles(int productId)
```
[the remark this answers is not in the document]
devinleighsmith: since this method appears to be specific to acq files you may want to change this method name, otherwise we may get collisions when we start adding other types of files.

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -96,15 +92,10 @@ public IActionResult SearchProjects(string input, int top)
         [ProducesResponseType(typeof(ProjectModel), 200)]
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "project" })]
-        public async Task<IActionResult> AddProject(ProjectModel projectModel)
+        public IActionResult AddProject(ProjectModel projectModel)
         {
-            var newProject = await _projectService.Add(_mapper.Map<Dal.Entities.PimsProject>(projectModel));
-            if(newProject is null)
```
devinleighsmith, line 102: fair, we normally throw exceptions for these cases.

source/backend/api/Areas/Products/Controllers/ProductController.cs
FuriousLlama: not sure, this is required for the project view on the FE so if the user does not have the acq claim that page will not work correctly

source/backend/api/Areas/Products/Controllers/ProductController.cs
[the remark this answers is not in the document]
FuriousLlama: I think is a long. Will align the controller

source/backend/api/Services/ProjectService.cs
```diff
@@ -26,16 +27,19 @@ public class ProjectService : BaseService, IProjectService
...
             IProjectRepository projectRepository,
-            IProductRepository productRepository)
+            IProductRepository productRepository,
+            IAcquisitionFileRepository acquisitionFileRepository)
             : base(user, logger)
         {
             _logger = logger;
             _projectRepository = projectRepository;
             _productRepository = productRepository;
+            _aquisitionFileRepository = acquisitionFileRepository;
```
devinleighsmith, line 42: nit: aquisition -> acquisition

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -176,6 +176,12 @@ public short GetRegion(long id)
                 .FirstOrDefault() ?? throw new KeyNotFoundException();
         }
 
+        public PimsAcquisitionFile GetByProductId(long productId)
```
devinleighsmith, line 179: This would fall into the TryGet pattern I believe since it returns null when a result is not found.
Also, what happens if you pass a null object to the mapper? it looks like this gets passed all the way back to the controller with no null checking, not sure if that would result in a null pointer or not.

source/backend/dal/Repositories/ProductRepository.cs
```diff
@@ -28,11 +28,12 @@ public ProductRepository(PimsContext dbContext, ClaimsPrincipal user, ILogger<Pr
...
         /// <param name="projectId"></param>
         /// <returns></returns>
-        public IList<PimsProduct> GetByProject(int projectId)
+        public IList<PimsProduct> GetByProject(long projectId)
         {
             return this.Context.PimsProducts.AsNoTracking()
-                .Where(o => o.ParentProjectId == projectId)
-                .OrderBy(a => a.Code)
+                .Where(p => p.ParentProjectId == projectId)
+                .Include(p => p.PimsAcquisitionFiles)
```
devinleighsmith, line 35: ok, yes, this is fine by me, looks like we aren't digging into the acq file itself which is in line with our best practices on nested entity loading.

source/backend/dal/Repositories/ProjectRepository.cs
```diff
@@ -74,28 +74,35 @@ public Task<Paged<PimsProject>> GetPageAsync(ProjectFilter filter)
...
-            return await Context.PimsProjects
+            return Context.PimsProjects
                     .AsNoTracking()
+                    .Include(x => x.PimsProducts)
                     .Include(x => x.ProjectStatusTypeCodeNavigation)
                     .Include(x => x.RegionCodeNavigation)
-                    .Where(x => x.Id == projectId)
-                    .FirstOrDefaultAsync();
+                    .Where(x => x.Id == id)
+                    .FirstOrDefault();
```
devinleighsmith, line 88: same comment here, I think you removed the null checking on the controller but I worry that the mapper doesn't handle nulls.

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -208,30 +208,68 @@ public void Get_ProjectById_ShouldFail_NotAuthorized()
             var repository = helper.GetService<Mock<IProjectRepository>>();
 
             // Act
-            Func<Task> act = async () => await service.GetById(It.IsAny<long>());
+            Assert.Throws<NotAuthorizedException>(() => service.GetById(1));
```
devinleighsmith, line 211: personally I have a mild preference for using a function here and asserting the throws check in the assert area to get a clean split between act and assert, but interested to understand your preference here.

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -208,30 +208,68 @@ public void Get_ProjectById_ShouldFail_NotAuthorized()
...
+        [Fact]
+        public void Get_Products_ShouldFail_NotAuthorized()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission();
+            var service = helper.Create<ProjectService>(user);
+
+            var repository = helper.GetService<Mock<IProductRepository>>();
+            //repository.Setup(x => x.GetByProject(It.IsAny<long>())).Returns(new List<PimsProduct>() { new PimsProduct() });
```
devinleighsmith, line 245: can we delete this?

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
```diff
@@ -208,30 +208,68 @@ public void Get_ProjectById_ShouldFail_NotAuthorized()
...
+        [Fact]
+        public void Get_ProductFile_ShouldFail_NotAuthorized()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission();
+            var service = helper.Create<ProjectService>(user);
+
+            var repository = helper.GetService<Mock<IAcquisitionFileRepository>>();
+            //repository.Setup(x => x.GetByProject(It.IsAny<long>())).Returns(new List<PimsProduct>() { new PimsProduct() });
```
devinleighsmith, line 264: same here.

source/frontend/src/features/mapSideBar/tabs/Section.tsx
```diff
@@ -20,7 +20,7 @@ export const Section: React.FC<
     <StyledFormSection className={clsx('form-section', className)}>
       {header && (
         <StyledSectionHeader>
-          <Row>
+          <Row className="no-gutters">
```
devinleighsmith, line 23: Seems unlikely but just in case, can you notify Sue of this change? I feel like I've seen changes like this introduce scroll bars unexpectedly, but that may not be a problem here.

source/frontend/src/features/properties/map/project/add/AddProjectFileYupSchema.ts
```diff
@@ -9,4 +10,33 @@ export const AddProjectYupSchema = Yup.object().shape({
...
+      description: Yup.string()
+        .required('Product Description is required')
+        .max(200, 'Product description must be at most ${max} characters'),
+      costEstimate: Yup.lazy(value =>
+        value === ''
+          ? Yup.string()
+          : Yup.number().typeError('Cost estimate must be a number').max(MAX_SQL_MONEY_SIZE),
+      ),
+      startDate: Yup.date(),
+      /*costEstimateDate: Yup.date().when('costEstimate', (costEstimate: string, schema: any) =>
```
devinleighsmith, line 27: hmm, what's the plan with these commented out validations? are these for a future story or something?

source/frontend/src/features/properties/map/project/add/AddProjectForm.tsx
```diff
@@ -19,66 +22,82 @@ export interface IAddProjectFormProps {
...
+const StyledRow = styled(Row)`
+  border-bottom-style: solid;
+  border-bottom-color: grey;
+  border-bottom-width: 0.1rem;
+  text-align: left;
+`;
+
 export const MediumTextArea = styled(TextArea)`
-  textarea.form-control {
+  /*textarea.form-control {
```
devinleighsmith: can we remove this? also surprised this compiles lol, aren't those comments within the ``?

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+  nameSpace: string;
+  formikProps: FormikProps<ProjectForm>;
+}
+
+export const ProductSubForm: React.FunctionComponent<IProductSubFormProps> = ({
+  index,
+  nameSpace,
+  formikProps,
+}) => {
+  const costEstimate = formikProps.values.products[index].costEstimate;
```
devinleighsmith, line 22: might want to use withNameSpace here as well?

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+          </SectionField>
+        </Col>
+      </Row>
+      <SectionField label="Start date" labelWidth="2" contentWidth="4">
+        <FastDatePicker field={withNameSpace(nameSpace, 'startDate')} formikProps={formikProps} />
+      </SectionField>
+      <Row>
+        <Col>
+          <SectionField label="Cost estimate">
+            <Input field={withNameSpace(nameSpace, 'costEstimate')} />
```
devinleighsmith, line 43: hmm, pretty sure we have a special type of input for money, let me check real quick

devinleighsmith: yep, FastCurrencyInput

devinleighsmith: assuming this is a currency value which I think it is.

source/backend/tests/unit/api/Services/ProjectServiceTest.cs
FuriousLlama: fair, ill update the other tests as well so we get consistency

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <SectionField label="Scope" labelWidth="12">
+        <MediumTextArea field={withNameSpace(nameSpace, 'scope')} />
+      </SectionField>
+    </>
+  );
+};
+
+export default ProductSubForm;
+
+export const MediumTextArea = styled(TextArea)`
```
devinleighsmith, line 69: maybe a candidate for a shared style?

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const [showModal, setShowModal] = useState(false);
+  const [showFileModal, setShowFileModal] = useState(false);
+  const [rowToDelete, setRowToDelete] = useState<number | undefined>(undefined);
+
+  const { retrieveProductFile, retrieveProductFileLoading } = useProductProvider();
+
+  const products = values.products || [];
+
+  const handleRemove = async (index: number) => {
```
devinleighsmith, line 35: huh, interesting. Think I need to look at the story to understand what is going on here.

devinleighsmith: oh... ok. So this relates to: "Important: you cannot delete this product unless you remove it from associated files first."

source/frontend/src/features/properties/map/project/add/AddProjectFileYupSchema.ts
FuriousLlama: There is a discrepancy between confluence and the UX/UI page on the estimatedate being requried. I will remove the comments.

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+  };
+
+  const showDeleteWarning = (index: number) => {
+    setShowModal(true);
+    setRowToDelete(index);
+  };
+
+  const confirmDelete = () => {
+    if (rowToDelete !== undefined) {
```
devinleighsmith, line 55: hmm, I feel like this is one of those weird error states that should never occur but if it does occur its hard to know what to do. Do we need a warning to the user here or is that a waste?

source/backend/dal/Repositories/ProjectRepository.cs
FuriousLlama: no problem if null

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Section header="Associated products">
+        <LoadingBackdrop show={retrieveProductFileLoading}></LoadingBackdrop>
+        <FieldArray
+          name={field}
+          render={arrayHelpers => {
+            arrayHelpersRef.current = arrayHelpers;
+            return (
+              <>
+                {products.map((product, index, array) => (
+                  <div key={`product-${index}`}>
```
devinleighsmith, line 72: can we use something other then the index here? React doesn't recommend using the index unless there is no way to form a stable key:
https://reactjs.org/docs/lists-and-keys.html#keys
I think we could at least use both the id and the index to make keys stable if they do have an id right?

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+                </LinkButton>
+              </>
+            );
+          }}
+        ></FieldArray>
+      </Section>
+
+      <GenericModal
+        display={showFileModal}
+        title="Remove Product wtih Files"
```
devinleighsmith, line 104: wtih = with

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
FuriousLlama: imo this is cleaner and type-checkeable. But no prob switching to the formik way if you believe strongly on it

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
FuriousLlama: Yes, not sure if the new mockups are coming with this or is it just a thing for these forms but it could be moved into the general forms section

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
FuriousLlama: IMO it would be a waste. The only chance this happens is if the state setting didnt happen (got to confirm without seen the prompt first)

source/frontend/src/features/properties/map/project/add/models.ts
```diff
@@ -1,23 +1,50 @@
...
+  toApi(): Api_Project {
     return {
-      id: id,
+      id: this.id,
       code: this.projectNumber,
       description: this.projectName,
       projectStatusTypeCode: toTypeCode<string>(this.projectStatusType),
       regionCode: this.region ? toTypeCode<number>(+this.region) : undefined,
       note: this.summary,
+      products: this.products?.map(x => x.toApi(this.id)),
```
devinleighsmith, line 47: I think you always want an array here right? this returns undefined if products is undefined. might want to add ?? []

source/frontend/src/features/properties/map/project/detail/ProjectProductView.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const ProjectProductView: React.FunctionComponent<
+  React.PropsWithChildren<IProjectProductViewProps>
+> = ({ project }) => {
+  const productCount = project?.products?.length || 0;
+
+  return (
+    <StyledSummarySection>
+      <Section header="Associated Products" isCollapsable initiallyExpanded>
+        {project?.products?.map((a: Api_Product, b: number) => (
```
devinleighsmith, line 19: could we rename a to something like product or currentProduct and b to index or similar?

source/frontend/src/features/properties/map/project/detail/ProjectProductView.tsx
```diff
@@ -0,0 +1,58 @@
...
+  );
+};
+
+export default ProjectProductView;
+
+const StyledSummarySection = styled.div`
+  background-color: ${props => props.theme.css.filterBackgroundColor};
+`;
+
+const StyledProductWrapper = styled.div``;
```
devinleighsmith, line 50: nit: could clean this up from the looks of it.

source/frontend/src/hooks/providers/useProductProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { execute: retrieveProductFile, loading: retrieveProductFileLoading } =
+    useApiRequestWrapper<
+      (projectId: number) => Promise<AxiosResponse<Api_AcquisitionFile | null, any>>
+    >({
+      requestFunction: useCallback(
+        async (productId: number) => await getProductFile(productId),
+        [getProductFile],
+      ),
+      requestName: 'retrieveProductFile',
+      onSuccess: useCallback(() => toast.success('File for project retrieved'), []),
```
devinleighsmith: I think this falls into the new category of a TMI toast. The BVT feedback we received was that they like getting feedback on save, but in cases such as load this can be redundant since the user can tell if their project file was retrieved just by looking at if there is data on the screen.

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
FuriousLlama: No ID at this point :( and the other fields could also be empty. Could generate a GUID or prefix it with -create-

source/frontend/src/features/properties/map/project/add/models.ts
FuriousLlama: It would be fine if its undefined

source/frontend/src/hooks/providers/useProductProvider.ts
FuriousLlama: Would be a good idea to introduce more robust logging if we are replacing the toasts everywhere

source/frontend/src/features/properties/map/project/detail/ProjectProductView.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { SectionField } from 'features/mapSideBar/tabs/SectionField';
+import { Api_Product, Api_Project } from 'models/api/Project';
+import styled from 'styled-components';
+import { formatMoney, prettyFormatDate } from 'utils';
+
+export interface IProjectProductViewProps {
+  project?: Api_Project;
+}
+
+const ProjectProductView: React.FunctionComponent<
```
devinleighsmith, line 11: I guess the idea here is that this view is covered by existing snapshots?

source/frontend/src/features/properties/map/project/add/ProductsArrayForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import { ProductForm, ProjectForm } from './models';
+import ProductSubForm from './ProductSubForm';
+
+export interface IProductsArrayFormProps {
+  field: string;
+  formikProps: FormikProps<ProjectForm>;
+}
+
+export const ProductsArrayForm: React.FunctionComponent<IProductsArrayFormProps> = ({
```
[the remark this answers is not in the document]
FuriousLlama: I do agree. I'll open a tech debt for it given this story is running out of time

source/frontend/src/features/properties/map/project/add/models.ts
```diff
@@ -1,23 +1,50 @@
-import { Api_Project } from 'models/api/Project';
+import { Api_Product, Api_Project } from 'models/api/Project';
 import { NumberFieldValue } from 'typings/NumberFieldValue';
 import { toTypeCode } from 'utils/formUtils';
 
+export class ProductForm {
```
devinleighsmith, line 5: Would this fall into our new category of needing a unit test?

source/frontend/src/features/properties/map/project/add/ProductSubForm.tsx
devinleighsmith: Not that strongly, just noticing it is being used below but not here. Was wondering if there was something I was missing.

source/frontend/src/hooks/providers/useProductProvider.ts
devinleighsmith: I think all success/errors are already logged as long as you use the apiRequestWrapper (via redux).

devinleighsmith:
Just going to check this out locally really quick and then will finalize review.

source/frontend/src/features/properties/map/project/add/models.ts
FuriousLlama: It definitely would. I think once the fromApi is set it would be good to add it

devinleighsmith:
Hey, so I think this story is supposed to include update product functionality, is that the part of this PR blocked by Eduardo?

FuriousLlama:
> Hey, so I think this story is supposed to include update product functionality, is that the part of this PR blocked by Eduardo?

That is correct. The update functionality depends on the project update, which is not ready.

devinleighsmith:
@FuriousLlama looks like status is labelled as required but the ui is allowing save without it being set.

FuriousLlama:
> @FuriousLlama looks like status is labelled as required but the ui is allowing save without it being set.

On the Project? That is not part of this PR directly, but I can change it.

devinleighsmith:
hmm, I tried to enter a product code that already existed and the db threw this error: ---> Microsoft.Data.SqlClient.SqlException (0x80131904): *** [PIMS_PRODCT_I_S_I_TR], Line 9. Errno 2601: Cannot insert duplicate key row in object 'dbo.PIMS_PRODUCT' with unique index 'PRODCT_CODE_IDX'. The duplicate key value is (2).

devinleighsmith:
also the estimate date is listed as required on the ui if the cost estimate is non-empty but the form does not enforce that currently.

FuriousLlama:
> hmm, I tried to enter a product code that already existed and the db threw this error: ---> Microsoft.Data.SqlClient.SqlException (0x80131904): *** [PIMS_PRODCT_I_S_I_TR], Line 9. Errno 2601: Cannot insert duplicate key row in object 'dbo.PIMS_PRODUCT' with unique index 'PRODCT_CODE_IDX'. The duplicate key value is (2).

Yes, this is something we need to chat with the BA's. If the code is unique we need to add some sort of validation to make sure is different when getting inserted. Right now there is nothing there in the acceptance criteria or confluence that implies that it should be unique.

devinleighsmith:
> > @FuriousLlama looks like status is labelled as required but the ui is allowing save without it being set.
> 
> On the Project? That is not part of this PR directly, but I can change it.

fair, up to you then. If not I can log a bug against that other story.

devinleighsmith:
> > hmm, I tried to enter a product code that already existed and the db threw this error: ---> Microsoft.Data.SqlClient.SqlException (0x80131904): *** [PIMS_PRODCT_I_S_I_TR], Line 9. Errno 2601: Cannot insert duplicate key row in object 'dbo.PIMS_PRODUCT' with unique index 'PRODCT_CODE_IDX'. The duplicate key value is (2).
> 
> Yes, this is something we need to chat with the BA's. If the code is unique we need to add some sort of validation to make sure is different when getting inserted. Right now there is nothing there in the acceptance criteria or confluence that implies that it should be unique.

Ok, I made a follow up story for that so that we don't lose track of this or the above comes back as a bug: https://jira.th.gov.bc.ca/browse/PSP-5389

devinleighsmith (approved):

source/backend/dal/Repositories/AcquisitionFileRepository.cs
FuriousLlama: it does check for nulls. If null, undefined or empty it assumes there are no files associated with the product

PR #2771, opened 2023-02-02
PSP-5038 : Edit Project Details

eddherrera:
image


source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -107,6 +113,29 @@ public async Task<IActionResult> AddProject(ProjectModel projectModel)
...
+            _logger.LogInformation(
+                "Request received by Controller: {Controller}, Action: {ControllerAction}, User: {User}, DateTime: {DateTime}",
+                nameof(ProjectController),
+                nameof(UpdateProject),
+                User.GetUsername(),
+                DateTime.Now);
+
+            var updatedProject = _projectService.Update(id, _mapper.Map<Dal.Entities.PimsProject>(model));
+
+            return Ok(updatedProject);
```
devinleighsmith, line 136: please use JsonResult, as it is more consistent with our existing patterns (or provide a reason why this is a better alternative).

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -107,6 +113,29 @@ public async Task<IActionResult> AddProject(ProjectModel projectModel)
...
+        /// </summary>
+        /// <returns></returns>
+        [HttpPut("{id:long}")]
+        [HasPermission(Permissions.ProjectEdit)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(ProjectModel), 200)]
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public IActionResult UpdateProject([FromRoute]long id, [FromBody] ProjectModel model)
+        {
+            _logger.LogInformation(
```
devinleighsmith, line 127: Would be good to come up with some logging standards, as everything is ad-hoc at the moment.

asanchezr: This log statement is based off that research spike I did a while back on structured logging best practices. Where we are lacking is in following that for evert controller/method. I also wish there were a way to automate/eliminate this boilerplate

source/frontend/src/features/properties/map/project/ProjectContainer.tsx
```diff
@@ -45,19 +47,21 @@ export enum ProjectPageNames {
...
-  // [
-  //   ProjectPageNames.DETAILS,
-  //   {
-  //     component: UpdateProjectContainer,
-  //     title: 'Project details',
-  //     validation: AddProjectYupSchema,
-  //   },
-  // ],
-]);
+// export const projectPages: Map<ProjectPageNames, IProjectPage> = new Map<
```
devinleighsmith, line 50: ?

source/frontend/src/features/properties/map/project/ProjectContainer.tsx
```diff
@@ -107,6 +111,11 @@ const ProjectContainer: React.FunctionComponent<
 
   useEffect(() => setProjectLoading(loadingProject), [loadingProject, setProjectLoading]);
 
+  const onSuccess = () => {
+    fetchProject();
```
devinleighsmith, line 115: does this need to be an await so that the state is only updated after the fetch is complete?

source/frontend/src/features/properties/map/project/ProjectContainerView.tsx
```diff
@@ -31,14 +68,41 @@ const ProjectContainerView: React.FC<IProjectContainerViewProps> = ({
       title={viewTitle}
       icon={<FaBriefcase className="mr-2 mb-2" size={32} />}
       header={<ProjectHeader project={project} />}
+      footer={
+        isEditing && (
```
devinleighsmith, line 72: hmm, ok, so the user can get to this form when not editing?

source/frontend/src/features/properties/map/project/add/models.ts
```diff
@@ -9,15 +9,30 @@ export class ProjectForm {
...
       code: this.projectNumber,
       description: this.projectName,
       projectStatusTypeCode: toTypeCode<string>(this.projectStatusType),
       regionCode: this.region ? toTypeCode<number>(+this.region) : undefined,
       note: this.summary,
+      rowVersion: this.rowVersion,
     };
   }
+
+  static fromApi(model: Api_Project): ProjectForm {
```
devinleighsmith, line 26: If this has a to and from api we can add a test for isomorphism (ie input -> output -> input is the same)

source/frontend/src/features/properties/map/project/update/UpdateProjectContainer.tsx
```diff
@@ -1,74 +1,93 @@
...
+  const projectStatusTypeCodes = getOptionsByType(API.PROJECT_STATUS_TYPES);
+  const regionTypeCodes = getByType(API.REGION_TYPES).map(c => mapLookupCode(c));
+
+  const handleSubmit = async (values: ProjectForm, formikHelpers: FormikHelpers<ProjectForm>) => {
+    try {
+      const updatedProject = values.toApi();
+      const response = await updateProject(updatedProject);
+
+      if (!!response?.id) {
+        formikHelpers?.resetForm();
```
devinleighsmith, line 43: hmm, you may want to reset the form the value returned by the response, otherwise the form may flicker during the transition from edit back to view mode.

devinleighsmith:
@eddherrera I believe that project name + number is supposed to be unique, as on https://jira.th.gov.bc.ca/browse/PSP-4514

not seeing handling for that on either create or update atm.

devinleighsmith (approved):

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
FuriousLlama: This does not look addressed

source/frontend/src/features/properties/map/project/update/UpdateProjectContainer.tsx
```diff
@@ -1,74 +1,93 @@
...
+                <Select
+                  field="projectStatusType"
+                  options={projectStatusTypeCodes}
+                  placeholder="Select..."
+                />
+              </SectionField>
+              <SectionField label="MoTI region" required={true}>
+                <Select field="region" options={regionTypeCodes} placeholder="Select..." />
+              </SectionField>
+              <SectionField label="Project summary">
```
FuriousLlama, line 81: Containers should not have UI elements when possible. Given these fields are mostly the same with the create section, reusing them would be preferable

FuriousLlama (changes requested):

source/backend/api/Services/ProjectService.cs
```diff
@@ -112,5 +113,28 @@ private async Task<Paged<PimsProject>> GetPageAsync(ProjectFilter filter)
...
+            }
+        }
+        
+        public PimsProject Update(long id, PimsProject project)
+        {
+            _user.ThrowIfNotAuthorized(Permissions.ProjectEdit);
+            project.ThrowIfNull(nameof(project));
+            _logger.LogInformation($"Updating project with id ${project.Id}");
+
+            ValidateVersion(id, project);
```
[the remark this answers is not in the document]
asanchezr: I agree this might not needed but I think we need to discuss what the pattern should be moving forward.

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
eddherrera: IMHO: .Net already returns JSON format by default. Returning a new instance of a JsonResult class seems redundant and not usefull. 
I would rather focus on returning the proper response than type/format and let the API configured options to handle/negotiate the response format/type.

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
eddherrera: removed for now.

source/backend/api/Services/ProjectService.cs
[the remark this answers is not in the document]
eddherrera: I've seen this pattern in a couple more places now. The only benefit would be to short circuit the update before reaching the repository. I'll remove it for now.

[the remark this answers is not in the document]
eddherrera: Just to clarify, how does EF check for this?

source/frontend/src/features/properties/map/project/ProjectContainerView.tsx
eddherrera: Tested again, haven't found a way to access the form when NOT editing.

devinleighsmith:
@eddherrera when you move this from draft to ready for review please send a note on teams as well so we don't miss the change.

source/frontend/src/features/properties/map/project/add/models.ts
eddherrera: Added to UpdateProjectContainer.test.tsx

source/frontend/src/features/properties/map/project/update/UpdateProjectContainer.tsx
eddherrera: Project in context get's updated after the update, which is later passed to the Form.

source/frontend/src/features/properties/map/project/ProjectContainer.tsx
eddherrera: removed

source/backend/api/Services/ProjectService.cs
```diff
@@ -108,6 +109,18 @@ public PimsProject Add(PimsProject project)
             return _projectRepository.Get(newProject.Id);
         }
 
+        public PimsProject Update(long id, PimsProject project)
```
FuriousLlama, line 112: the id parameter is not being used.

source/frontend/src/features/properties/map/project/ProjectContainerView.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const { asFragment } = setup({ ...mockProps });
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  // it('displays modal when form is dirty and cancel is clicked', async () => {
+  //   setup({ ...mockProps, showConfirmModal: true });
+
+  //   expect(screen.getByText('Confirm changes')).toBeVisible();
+  // });
+});
```
FuriousLlama, line 49: can this be deleted?

source/frontend/src/features/properties/map/project/ProjectContainerView.tsx
```diff
@@ -31,14 +68,41 @@ const ProjectContainerView: React.FC<IProjectContainerViewProps> = ({
...
+          <SidebarFooter
+            isOkDisabled={formikRef?.current?.isSubmitting}
```
FuriousLlama, line 74: this will not work given the formik ref change does not cause re-renders

source/frontend/src/features/properties/map/project/update/UpdateProjectContainerView.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <Formik<ProjectForm>
+      enableReinitialize
+      innerRef={props.formikRef}
+      initialValues={initialValues}
+      validationSchema={AddProjectYupSchema}
+      onSubmit={onSubmit}
+    >
+      {formikProps => (
+        <>
```
FuriousLlama, line 28: nit: empty brackets not needed here given the container would count as the single child

FuriousLlama (approved):

PR #2786, opened 2023-02-02
psp-5391 (git action performance) jest update, github action processor usage improvement.

devinleighsmith:


.github/workflows/app-react.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - run: npm run build --if-present
         working-directory: ${{env.working-directory}}
 
-      - run: npm run coverage
+      - run: npm run coverage -- --maxWorkers=2
```
devinleighsmith, line 55: this works because npm by default detects available processors and uses one less, to try and be less intrusive:
https://jestjs.io/docs/configuration#maxworkers-number--string
As a result this only uses one of our gha's two available cores.

FuriousLlama (approved):

source/frontend/package.json
```diff
@@ -138,6 +140,9 @@
     "pretty-quick": "pretty-quick"
   },
   "jest": {
+    "transform": {
```
devinleighsmith, line 143: This helps with ts compilation within jest, in my testing gives a 10 - 15% performance improvement.

.github/workflows/app-react.yml
devinleighsmith: On my fork this cuts test execution time by 50%

asanchezr (approved):

PR #2814, opened 2023-02-09
psp-5457 psp-5388 fix edit action on documents. Update interfaces to …

devinleighsmith:
…enforce typings.

### DIFF
--- a/source/frontend/src/features/documents/ComposedDocument.ts
+++ b/source/frontend/src/features/documents/ComposedDocument.ts
@@ -23,16 +23,16 @@ export interface ComposedDocument {
 
 export class DocumentRow {
   id?: number;
-  mayanDocumentId?: number;
-  documentType?: Api_DocumentType;
-  statusTypeCode?: Api_TypeCode<string>;
-  fileName?: string;
-  isFileAvailable?: boolean;
+  mayanDocumentId: number | undefined;
+  documentType: Api_DocumentType | undefined;
+  statusTypeCode: Api_TypeCode<string> | undefined;
+  fileName: string | undefined;
+  isFileAvailable: boolean | undefined;
   appCreateTimestamp?: string;
   appCreateUserid?: string;
-  relationshipId?: number;
-  relationshipType?: DocumentRelationshipType;
-  parentId?: number;
+  relationshipId: number | undefined;
+  relationshipType: DocumentRelationshipType | undefined;
+  parentId: number | undefined;
 
   public static fromApi(relationship: Api_DocumentRelationship): DocumentRow {
     const row: DocumentRow = new DocumentRow();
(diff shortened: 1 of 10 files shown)


source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -23,16 +23,16 @@ export interface ComposedDocument {
 
 export class DocumentRow {
   id?: number;
-  mayanDocumentId?: number;
-  documentType?: Api_DocumentType;
-  statusTypeCode?: Api_TypeCode<string>;
-  fileName?: string;
-  isFileAvailable?: boolean;
+  mayanDocumentId: number | undefined;
```
devinleighsmith, line 26: This article does a decent job of describing the differences, but essentially having an interface with all optional fields is the same as {}. Optional fields are not validated by ts.

this article discusses the differences
https://kate-dev.medium.com/when-to-use-typescript-optional-property-how-is-it-different-from-declaring-property-as-undefined-2319a0ee1f07

Essentially having an interface with all optional fields isn't much better than any. We probably need some tech debt to go back and fix this, and maybe a discussion on Friday about how/when we want ? vs | undefined

source/frontend/src/utils/test-utils.tsx
```diff
@@ -22,23 +22,25 @@ export const mockKeycloak = (
...
-  const { claims, roles, organizations, authenticated = true } = props;
+  const { claims, roles, organizations, authenticated = true, userInfo } = props;
   // mock what would be returned by keycloak userinfo endpoint
-  const userInfo = {
+  const defaultUserInfo = {
     organizations: organizations ?? [1],
     client_roles: [...(claims ?? []), ...(roles ?? [])] ?? [],
     email: 'test@test.com',
     name: 'Chester Tester',
+    idir_user_guid: '7DB280070D474EF0BB46C365A4B95A73',
```
asanchezr, line 35: The secret scanning action is not happy with this string - please change it to something else; e.g. "foobar"

source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -50,6 +50,22 @@ export class DocumentRow {
...
+    row.documentType = document?.documentType;
+    row.mayanDocumentId = document?.mayanDocumentId;
+    row.statusTypeCode = document?.statusTypeCode;
+    row.fileName = document?.fileName;
+    row.appCreateTimestamp = document?.appCreateTimestamp;
+    row.appCreateUserid = document?.appCreateUserid;
+
+    row.relationshipId = undefined;
+    row.relationshipType = undefined;
+    row.parentId = undefined;
```
FuriousLlama, line 65: this looks like the doc row might be a Document Relationship representation rather than just a document. Do you know if passing a document relationship works better?

devinleighsmith: yes, this was just added for compatibility purposes. In some cases a row is built from a Document Relationship and sometimes from a document. Previously we were getting away with this because of the loose type checking.

devinleighsmith: I have some time now since my other stories are done, can probably go through and try and make the usage consistent.

asanchezr (approved):

PR #2815, opened 2023-02-09
PSP-5368 FT: Projects - Some Project Results columns should be sortable

eddherrera:
- Added sort option for colums Code/Name/LastUpdatedBy/LastUdpatedDate
- Placeholder for status dropdown list.
image


asanchezr (commented):
This is a bug fix that should go into the **test** branch I believe but this pr is targeting DEV

PR #2816, opened 2023-02-10
Bug fixes: PSP-5354, PSP-5370

asanchezr:
- PSP-5364: Financial Code List - page size selector not shown
- PSP-5370: Limit project selection to regions associated to current user

### DIFF
--- a/source/backend/api/Services/ProjectService.cs
+++ b/source/backend/api/Services/ProjectService.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -18,6 +19,8 @@ namespace Pims.Api.Services
         private readonly IProjectRepository _projectRepository;
         private readonly IProductRepository _productRepository;
         private readonly IAcquisitionFileRepository _acquisitionFileRepository;
+        private readonly IUserRepository _userRepository;
+        private readonly ILookupRepository _lookupRepository;
         private readonly ClaimsPrincipal _user;
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


source/backend/api/Services/ProjectService.cs
```diff
@@ -28,29 +31,43 @@ public class ProjectService : BaseService, IProjectService
...
+            // Limit search results to user's assigned region(s), but always include "Cannot determine" region
+            var pimsUser = _userRepository.GetUserInfoByKeycloakUserId(_user.GetUserKey());
+            var userRegions = pimsUser.PimsRegionUsers.Select(r => r.RegionCode).ToHashSet();
 
-            return projects;
+            var noRegion = _lookupRepository.GetAllRegions().FirstOrDefault(r => r.RegionName == "Cannot determine")?.Code;
+            if (noRegion.HasValue)
+            {
+                userRegions.Add(noRegion.Value);
+            }
```
asanchezr, line 68: There can be projects with "Cannot determine" as their region so the assumption here is that we want to include those as well - in addition to the one matching the user region(s).

This is not specified in the AC of the original story so open to hear thoughts on whether to make this assumption or not

devinleighsmith: On acq files the missing requirement was that cannot determine is not a valid option when selecting a region. May be worth asking if that is the case here as well.

devinleighsmith (approved):

source/backend/api/Services/ProjectService.cs
asanchezr: Well, ACQ and projects can be brought in via ETL and there's a valid case where region code would be 4 == Cannot determine.

I agree with your other PR that limits the regions upon creation so you cannot select region 4 but for searches I think we should still support region 4. Need a definitive answer from BAs

PR #2828, opened 2023-02-11
Document generation backend |  psp-5249

FuriousLlama:


source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -13,7 +14,7 @@ namespace Pims.Api.Controllers
     /// <summary>
     /// DocumentGenerationController class, provides endpoints to handle document generation requests.
     /// </summary>
-    [Authorize]
+    //[Authorize]
```
FuriousLlama, line 17: Claims are missing for document generation. I will get another PR to fix that

source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -16,7 +15,7 @@ import Api_TypeCode from 'models/api/TypeCode';
 
 export interface ComposedDocument {
   mayanMetadata?: Api_Storage_DocumentMetadata[];
-  pimsDocument?: Api_Document;
+  pimsDocument?: Api_DocumentRelationship;
   documentDetail?: Api_Storage_DocumentDetail;
```
FuriousLlama, line 19: This will change based on @devinleighsmith fix of the type

source/backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -90,16 +90,36 @@ public IActionResult GetDocumentTypes()
...
         {
             var result = await _documentService.DownloadFileAsync(mayanDocumentId, mayanFileId);
-            return new FileStreamResult(result.Payload.FilePayload, result.Payload.Mimetype) { FileDownloadName = result.Payload.FileName };
+
+            if (result?.Payload == null)
+            {
+                return new NotFoundResult();
+            }
+
+            byte[] fileBytes = System.Convert.FromBase64String(result.Payload.FilePayload);
```
devinleighsmith, line 121: This will force the entire stream to be stored in memory before it can be returned. It also forces the json parser to parse the entire object instead of parsing a stream. In the short term this will probably cause file downloads of >10mb to fail.

Can we come up with a way of using file streams instead? to me, a file stream has significantly lower impact of the memory usage of the server.

devinleighsmith (changes requested):
see comment about file streams

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -54,17 +55,50 @@ public async Task<IActionResult> GetSupportedDocumentTypes()
...
+        public async Task<IActionResult> UploadTemplateAndDownload([FromBody] DocumentGenerationRequest request)
+        {
+            var result = await _documentGenerationService.GenerateDocument(request.TemplateType, request.TemplateData);
+            if (result?.Payload == null)
+            {
+                return new NotFoundResult();
+            }
+
+            byte[] fileBytes = System.Convert.FromBase64String(result.Payload.FilePayload);
+            return new FileContentResult(fileBytes, result.Payload.Mimetype) { FileDownloadName = result.Payload.FileName };
```
devinleighsmith, line 99: See my other comment around file streams, this will have a large impact on the memory usage of the backend pods.

source/backend/api/Areas/Documents/DocumentController.cs
FuriousLlama: Yes we can. We could keep the stream from the repository and have the service be smarter to if it needs to be serialized to base64.

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
FuriousLlama: There are two API's now. One does return a direct download link for the stream while the other one converts it to a base64 string with more information for the encoding.

devinleighsmith (approved):

source/frontend/src/features/documents/documentDetail/DocumentDetailView.tsx
```diff
@@ -33,6 +33,8 @@ export const DocumentDetailView: React.FunctionComponent<
 > = props => {
   const { hasClaim } = useKeycloakWrapper();
 
+  console.log(props.document);
```
devinleighsmith, line 36: console.log.

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -54,17 +55,50 @@ public async Task<IActionResult> GetSupportedDocumentTypes()
...
         }
+
+        /// <summary>
+        /// Renders the given template with the request data and returns the result as wrapped encoded base64 file.
+        /// </summary>
+        [HttpPost("render_template/download-wrapped")]
+        //[HasPermission(Permissions.GenerateDocuments)]
+        [ProducesResponseType(typeof(ExternalResult<FileDownload>), 200)]
+        [SwaggerOperation(Tags = new[] { "document-generation" })]
+        public async Task<IActionResult> UploadTemplateAndDownloadWrapepd([FromBody] DocumentGenerationRequest request)
```
devinleighsmith, line 77: Wrapepd

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -13,7 +14,7 @@ namespace Pims.Api.Controllers
     /// <summary>
     /// DocumentGenerationController class, provides endpoints to handle document generation requests.
     /// </summary>
-    [Authorize]
+    //[Authorize]
     [ApiController]
     [ApiVersion("1.0")]
     [Route("v{version:apiVersion}/documentGeneration/")]
```
devinleighsmith, line 20: to follow our other format this may have been better as /document/generation/... instead

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -54,17 +55,50 @@ public async Task<IActionResult> GetSupportedDocumentTypes()
...
         public async Task<IActionResult> UploadTemplate([FromForm] IFormFile file)
         {
             var result = await _documentGenerationService.UploadFileTemplate(file);
             return new JsonResult(result);
         }
+
+        /// <summary>
+        /// Renders the given template with the request data and returns the result as wrapped encoded base64 file.
+        /// </summary>
+        [HttpPost("render_template/download-wrapped")]
```
devinleighsmith, line 73: why mix x_y and x-y formatting in the url?

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
devinleighsmith: Might be worth thinking through the usage of that. One api will be usable across the application while the other will only be useable in areas where we can guarantee the file size is small. I think small file size guarantees may be possible for document template uploads/downloads, is that the use case that you were expecting for that API?

FuriousLlama: Ideally, we would have the front end download directly from the backend API using the stream and setting the correct headers. That way we would be leveraging the browser's capabilities to stream data as well as the backend's. However, we are doing extra processing on the front end that might be preventing that (I am not 100% sure it does, but if not we should switch to the streaming API).

From the looks on the front end is because we cannot guarantee that the file is available, so we do some checks before prompting the download. We could have this revisited once we have that flag on the backend to know if the file has been processed and create a download link on the fly once the file has been fetched.

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
FuriousLlama: yikes

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
FuriousLlama: I think is a sufficiently distinct functionality to warrant its own route. But not hard set on it.

PR #2829, opened 2023-02-11
PSP-5049 Research file - add/update/view project details

asanchezr:
view-research-file

create-research-file

update-research-file


source/backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -125,9 +130,11 @@ public PimsResearchFile Update(PimsResearchFile researchFile)
                 Context.Entry(deletedPurpose).State = EntityState.Deleted;
             }
 
-            researchFile.PimsResearchFilePurposes = purposes;
+            existingResearchFile.PimsResearchFilePurposes = purposes;
+
+            Context.Entry(existingResearchFile).CurrentValues.SetValues(researchFile);
```
devinleighsmith, line 135: I prefer this method of saving updates, but it would be worth mentioning to Sue to run regression on research file update to make sure everything works as it did before.

asanchezr: Good idea. I had to change the update method because the other way was not compatible with our "UpdateChild" method - was getting errors from EF

source/frontend/src/features/properties/map/research/detail/ResearchSummaryView.tsx
```diff
@@ -90,6 +96,13 @@ const ResearchSummaryView: React.FunctionComponent<
           />
         ) : null}
       </StyledEditWrapper>
+      <Section header="Project">
+        <SectionField label="Ministry project">
+          {detail.researchFileProjects.map(formattedName => (
```
devinleighsmith, line 101: seems a bit odd to display this section empty when there are no projects, but that is what the requirements ask for so /shrug

devinleighsmith (approved):

FuriousLlama (approved):

PR #2848, opened 2023-02-15
Test console output

devinleighsmith:
this doesn't fix all of the test console output, but removes some clutter.


source/frontend/package.json
```diff
@@ -26,7 +26,7 @@
     "polylabel": "1.1.0",
     "query-string": "6.12.1",
     "react": "18.2.0",
-    "react-bootstrap": "1.5.2",
+    "react-bootstrap": "1.6.6",
```
asanchezr, line 29: Is the react-bootstrap upgrade relevant to fix the test warnings?

PR #2851, opened 2023-02-15
mock bc assessment with msw instead of an inline response.

devinleighsmith:


source/frontend/src/mocks/msw.ts
```diff
@@ -0,0 +1,30 @@
...
+      return res(ctx.delay(500), ctx.status(200), ctx.json(getMockLegalDescriptions()));
+    } else if (req.url.search.includes('BCA_FOLIO_GNRL_PROP_VALUES_SV')) {
+      return res(ctx.delay(500), ctx.status(200), ctx.json(getMockValues()));
+    } else if (req.url.search.includes('BCA_FOLIO_SALES_SV')) {
+      return res(ctx.delay(500), ctx.status(200), ctx.json(getMockSales()));
+    } else if (req.url.search.includes('BCA_FOLIO_DESCRIPTIONS_SV')) {
+      return res(ctx.delay(500), ctx.status(200), ctx.json(getMockDescription()));
+    } else if (req.url.search.includes('BCA_FOLIO_LAND_CHARS_SV')) {
+      return res(ctx.delay(500), ctx.status(200), ctx.json(getMockLandChars()));
+    }
```
asanchezr, line 28: We will need to do the same setup for jest tests because jest runs in Node and there we don't have service workers. 
Ideally we would split the setupWorker/setupServer calls from the handlers to be able to reuse the URL mocking in both the browser APP and jest.

Please see this sample repo: https://github.com/mswjs/examples/tree/master/examples/rest-react/src/mocks

image

asanchezr: We would also need to update **setupTests.ts** to launch msw in jest. Like this: https://github.com/mswjs/examples/blob/master/examples/rest-react/src/setupTests.js

devinleighsmith: True, but I think that is outside the scope of this task: https://jira.th.gov.bc.ca/browse/PSP-5496 (setup the mock to replace existing bc assessment mock).

asanchezr: fair enough, then we need a tech-debt task to add msw to jest. We will benefit from having the same mocking behaviour in tests as local dev

asanchezr (approved):

PR #2855, opened 2023-02-15
psp-5249 document generation POC frontend.

devinleighsmith:


source/frontend/src/features/properties/map/shared/detail/models.ts
```diff
@@ -0,0 +1,42 @@
...
+  deposit_amount: string | null;
+  purchase_price: string;
+  owners: [
+    {
+      lastname: string;
+      firstname: string;
+      organization_name: string;
+      incorporation_number: string;
+      reg_no: string;
+    },
```
FuriousLlama, line 16: for the non poc it would be better if they are defined on a different interface and just referenced here. Same with agents and properties

FuriousLlama (approved):

PR #2887, opened 2023-02-21
PSP-4629 : Acquisition - property owners - update

eddherrera:
image


source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -143,7 +143,9 @@ public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile)
             }
 
             this.Context.Entry(existingAcqFile).CurrentValues.SetValues(acquisitionFile);
-            this.Context.UpdateChild<PimsAcquisitionFile, long, PimsAcquisitionFilePerson>(p => p.PimsAcquisitionFilePeople, acquisitionFile.Internal_Id, acquisitionFile.PimsAcquisitionFilePeople.ToArray());
+            this.Context.UpdateChild<PimsAcquisitionFile, long, PimsAcquisitionFilePerson>(p => p.PimsAcquisitionFilePeople, acquisitionFile.Internal_Id, acquisitionFile.PimsAcquisitionFilePeople.ToArray(), true);
```
devinleighsmith, line 146: I don't think you want to set true here, that would allow the system to modify the person, which is not desirable. I think in this case we are just updating the reference to the entities, not the entities themselves.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -143,7 +143,9 @@ public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile)
...
+            this.Context.UpdateChild<PimsAcquisitionFile, long, PimsAcquisitionOwner>(p => p.PimsAcquisitionOwners, acquisitionFile.Internal_Id, acquisitionFile.PimsAcquisitionOwners.ToArray(), true);
```
devinleighsmith, line 147: In this case passing the true flag makes since since the ref to the owner may need to be updated as well as the owner itself.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -143,7 +143,9 @@ public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile)
...
+            this.Context.UpdateGrandchild<PimsAcquisitionFile, long, PimsAcquisitionOwner>(o => o.PimsAcquisitionOwners, oa => oa.Address, acquisitionFile.Internal_Id, acquisitionFile.PimsAcquisitionOwners.ToArray());
```
devinleighsmith, line 148: Actually I don't think this is correct to chain update child and updategrandchild like this, I think you should be using one or the other here.

devinleighsmith: Yeah, I checked this out and tried it, throws an exception when adding a net new owner during an update.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
eddherrera: Updated

source/backend/dal/Repositories/AcquisitionFileRepository.cs
eddherrera: Updated

source/backend/dal/Repositories/AcquisitionFileRepository.cs
eddherrera: Updated

source/frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -19,8 +19,8 @@ import React from 'react';
 import { Prompt } from 'react-router-dom';
 import styled from 'styled-components';
 
+import UpdateAcquisitionOwnersSubForm from '../common/update/acquisitionOwners/UpdateAcquisitionOwnersSubForm';
```
devinleighsmith, line 22: huh, I thought our linter enforced import order, I'll have a look at this later.

source/frontend/src/features/properties/map/acquisition/common/models.ts
```diff
@@ -95,8 +95,24 @@ export class OwnerAddressFormModel {
...
+  static addressLines(apiAddress: OwnerAddressFormModel | undefined): number {
+    if (!apiAddress) {
+      return 1;
+    }
+
+    if (apiAddress.streetAddress3?.trim() !== '') {
+      return 3;
+    }
+
+    if (apiAddress.streetAddress2?.trim() !== '') {
```
devinleighsmith, line 110: hmm, not sure if this is a realistic concern but null or undefined are not equal to '', so those would trigger an additional address line

source/frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
devinleighsmith: Oh, I see, this is a rename. NVM.

source/frontend/src/features/properties/map/acquisition/detail/models.ts
```diff
@@ -47,4 +52,65 @@ export class DetailAcquisitionFileOwner {
...
+
+  if (streetAddress2) {
+    addressDisplay = addressDisplay.concat(streetAddress2, '\n');
+  }
+
+  if (streetAddress3) {
+    addressDisplay = addressDisplay.concat(streetAddress3, '\n');
+  }
+
+  if (streetAddress4) {
```
devinleighsmith, line 99: do we have a streetAddress4?

source/frontend/src/features/properties/map/acquisition/detail/models.ts
```diff
@@ -47,4 +52,65 @@ export class DetailAcquisitionFileOwner {
...
+  let addressDisplay = '';
+  let streetAddress1 = address?.streetAddress1 ? address?.streetAddress1.trim() : null;
+  let streetAddress2 = address?.streetAddress2 ? address?.streetAddress2.trim() : null;
+  let streetAddress3 = address?.streetAddress3 ? address?.streetAddress3.trim() : null;
+  let streetAddress4 = concatValues(
+    [address?.municipality, address?.province?.description, address?.postal],
+    ', ',
+  );
+
+  if (streetAddress1) {
```
devinleighsmith, line 87: Hmm, here you are using a truthy check instead of a non-empty check. Why the difference in this context?

devinleighsmith (approved):

PR #2896, opened 2023-02-22
Wava Scan Improvements | PSP-5118

FuriousLlama:
Updated CSP settings for local dev enhanced api CSP settings. 
Updated nginx cipher suits template. Cleaned up manifest.
Added headers to both backend and frontend recomended by the scan


source/frontend/public/manifest.json
```diff
@@ -7,7 +7,7 @@
       "sizes": "17x17"
     },
     {
-      "src": "tenants/moti/PIMS-logo.png",
+      "src": "tenants/MOTI/PIMS-logo.png",
```
FuriousLlama, line 10: not sure this file is used at all. AFAIK the manifest is only needed if the app is configured to be used as a PWA

openshift/s2i/nginx-runtime/nginx.conf.template
```diff
@@ -39,6 +39,10 @@ http {
         underscores_in_headers on;
         ignore_invalid_headers off;
 
+        ssl_protocols               TLSv1.2 TLSv1.3;
```
devinleighsmith, line 42: I'm curious if this'll work, we have more strict policies on our Openshift server right now but they appear to be getting overwritten by the SSL ciphers on the reverse proxy server (ie. vanity url).

source/frontend/config-overrides.js
```diff
@@ -29,4 +29,33 @@ module.exports = {
     config.transformIgnorePatterns = ['node_modules/(?!(react-leaflet|@react-leaflet/core)/)'];
     return config;
   },
+
+  devServer: function (configFunction) {
```
devinleighsmith, line 33: Huh, I don't see this called anywhere explicitly, is this function called automatically by some react-app-rewired process?

source/backend/api/Startup.cs
```diff
@@ -342,6 +343,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
 
             app.UseRouting();
             app.UseCors();
+
+            // Set responses secure headers.
+            ConfigureSecureHeaders(app, Configuration);
```
devinleighsmith, line 348: So, from what I can see this will result in two CSPs, one from the api and one from nginx. The one on the index has been removed.

I'm still having a hard time conceptualizing how the two CSPs will interact. presumably the nginx one will take effect first since the user will hit the nginx box before they are even served the react app and therefore also before they get this csp from the api. If that is the case, what is the point of this CSP? does this CSP override the nginx one because it happens later in the startup process? What is the benefit to having two CSPs? can we get away with just having one?

sorry, a bunch of questions I know but still confused about this situation.

openshift/s2i/nginx-runtime/nginx.conf.template
devinleighsmith: For reference:
image

devinleighsmith: and vanity url route: https://www.ssllabs.com/ssltest/analyze.html?d=dev-pims.th.gov.bc.ca

source/frontend/config-overrides.js
FuriousLlama: Yeah, it's just part of the react dev server hosting

source/backend/api/Startup.cs
FuriousLlama: I think it depends on what is serving the response[1]. If the response is being served by Nginx then the CSP policy on the headers will be what was set for the front end. If the reply is coming from the backend, then the backend will set the response headers. I am not 100% sure about the mechanisms we are using for proxying the backend calls, but it does not look like the headers are being overridden.

Also, I don't think the CSP is necessary for APIs given that there is no rendering on the front end. But having it set explicitly makes it more clear.

In [1] you can see the last request with the changes from this PR and the explicit CSP from the backend api.
[1] 
image

openshift/s2i/nginx-runtime/nginx.conf.template
FuriousLlama: ah, that explains some things. I will dig some more on the cipher's part. But at this point, I believe the spike cap has been reached.

devinleighsmith (approved):

PR #2903, opened 2023-02-23
PSP-5504 : FT: Acquisition Files - Notes Tab: Changes on File status …

eddherrera:
- Add note when Acquisition File Status is updated.

### DIFF
--- a/source/backend/api/Services/AcquisitionFileService.cs
+++ b/source/backend/api/Services/AcquisitionFileService.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
@@ -25,6 +26,7 @@ namespace Pims.Api.Services
         private readonly IPropertyRepository _propertyRepository;
         private readonly ICoordinateTransformService _coordinateService;
         private readonly ILookupRepository _lookupRepository;
+        private readonly IEntityNoteRepository _entityNoteRepository;
 
         public AcquisitionFileService(
             ClaimsPrincipal user,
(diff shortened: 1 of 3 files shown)


source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -299,5 +305,34 @@ private void ValidateMinistryRegion(long acqFileId, short updatedRegion)
                 throw new BusinessRuleViolationException("The Ministry region has been changed, this will result in a change to the file's prefix. This requires user confirmation.");
             }
         }
+
+        private void CheckAcquisitionFileStatusChanged(PimsAcquisitionFile updateAcquisitionFile)
```
devinleighsmith, line 309: based on the name of this function I'd be tempted to call this something like:
addNoteIfFileStatusChanged

as is this appears to be something that checks if the file status has changed, not a function that adds a note if the file status has changed.

source/backend/tests/unit/api/Services/AcquisitionServiceTest.cs
```diff
@@ -219,14 +223,51 @@ public void Update_Success_UserOverride()
...
+                Id = acqFile.AcquisitionFileStatusTypeCodeNavigation.Id,
+                Description = acqFile.AcquisitionFileStatusTypeCodeNavigation.Description,
+            }});
+
+            // Act
+            var result = service.Update(acqFile, true);
+
+            // Assert
+            repository.Verify(x => x.Update(It.IsAny<PimsAcquisitionFile>()), Times.Once);
+            noteRepository.Verify(x => x.Add(It.IsAny<PimsAcquisitionFileNote>()), Times.Once);
```
devinleighsmith, line 268: In this instance I'd be tempted to check that the note that is going to be added is as expected.

devinleighsmith (approved):

PR #2904, opened 2023-02-24
Lease Notes | psp-4807

FuriousLlama:
Added the note tab to the frontend leases and wired up backend


source/backend/api/Services/LeaseService.cs
```diff
@@ -84,7 +111,7 @@ private PimsLease AssociatePropertyLeases(PimsLease lease, bool userOverride = f
         {
             MatchProperties(lease);
 
-            lease.PimsPropertyLeases.ForEach(propertyLease =>
+            foreach (var propertyLease in lease.PimsPropertyLeases)
```
devinleighsmith, line 114: What is the benefit of making this change?

source/backend/api/Services/LeaseService.cs
```diff
@@ -64,6 +71,26 @@ public PimsLease Add(PimsLease lease, bool userOverride = false)
...
         {
+            var currentLease = _leaseRepository.Get(lease.LeaseId);
+
+            if (currentLease.LeaseStatusTypeCode != lease.LeaseStatusTypeCode)
+            {
+                _entityNoteRepository.Add<PimsLeaseNote>(
+                    new PimsLeaseNote()
+                    {
+                        LeaseId = currentLease.LeaseId,
+                        AppCreateTimestamp = System.DateTime.Now,
```
devinleighsmith, line 82: is this necessary? I know we do some automatic insertion of these audit values in most scenarios. Not sure if that is applicable here.

source/backend/api/Services/NoteService.cs
```diff
@@ -58,21 +58,29 @@ public EntityNoteModel Add(NoteType type, EntityNoteModel model)
             switch (type)
             {
                 case NoteType.Activity:
-                    var pimsEntity = _mapper.Map<PimsActivityInstanceNote>(model);
+                    PimsActivityInstanceNote pimsEntity = _mapper.Map<PimsActivityInstanceNote>(model);
```
devinleighsmith, line 61: Should we talk about best practices around usage of var?

source/backend/api/Services/LeaseService.cs
FuriousLlama: The ForEach was conflicting with the LINQ namespace and the extensions we have

source/backend/dal/Repositories/EntityNoteRepository.cs
```diff
@@ -100,6 +109,24 @@ public bool DeleteAcquisitionFileNotes(long entityId)
...
+        {
+            var leaseFileNotes = this.Context.PimsLeaseNotes.
+                                        Include(ln => ln.Note).
+                                        Where(x => x.NoteId == noteId).ToList();
+
+            if (leaseFileNotes.Any())
+            {
+                foreach (var leaseFileNote in leaseFileNotes)
+                {
+                    this.Context.PimsLeaseNotes.Remove(leaseFileNote);
```
devinleighsmith, line 122: I wonder if we need this now or if our cascades or covering this. Maybe we need to go back now that we have some cascade logic and clean some of these up? certainly not something you need to worry about this PR but I should probably log tech debt for it.

source/backend/dal/Repositories/Interfaces/IEntityNoteRepository.cs
```diff
@@ -14,10 +14,14 @@ IEnumerable<T> GetAll<T>(Func<T, bool> predicate)
 
         IEnumerable<PimsNote> GetAllActivityNotesById(long entityId);
 
-        IEnumerable<PimsNote> GetAllAcquisitionNotesById(long entityId);
+        IEnumerable<PimsNote> GetAllAcquisitionNotesById(long acquisitionId);
```
devinleighsmith, line 17: I like that you've made this more readable but kinda surprised you just targeted AcqNotes?

devinleighsmith (approved):

source/backend/dal/Repositories/Interfaces/IEntityNoteRepository.cs
FuriousLlama: Didn't want to update Activity related code :)

devinleighsmith: lol, makes sense.

PR #2906, opened 2023-02-24
psp-5271 add poi removal to lease/license.

devinleighsmith:


source/backend/dal/Repositories/LeaseRepository.cs
```diff
@@ -188,6 +188,124 @@ public PimsLease Get(long id)
...
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.LeaseView);
+
+            PimsLease lease = this.Context.PimsLeases.AsSplitQuery()
+                .Include(l => l.PimsPropertyLeases)
+                .ThenInclude(p => p.Property)
+                    .ThenInclude(p => p.Address)
+                    .ThenInclude(p => p.Country)
+                .Include(l => l.PimsPropertyLeases)
+                    .ThenInclude(p => p.Property)
```
asanchezr, line 201: question - what's the difference with the **Get** method? I see no "AsNoTracking()" here. 
Am I missing something?

asanchezr (approved):

source/backend/dal/Repositories/LeaseRepository.cs
devinleighsmith: Just asNoTracking() is the only change. I tried adding asNoTracking to the pre-existing method but immediately saw two workflows immediately start to fail so unable to change the existing method, however I need asNoTracking otherwise I am unable to delete the related lease entities due to multiple tracked entity changes. I'll add a TODO here because ideally this would be fixed in the long term.

source/backend/dal/Repositories/PropertyLeaseRepository.cs
```diff
@@ -38,6 +38,25 @@ public IEnumerable<PimsPropertyLease> GetAllByPropertyId(long propertyId)
...
+        /// </summary>
+        /// <param name="leaseId"></param>
+        /// <returns></returns>
+        public IEnumerable<PimsPropertyLease> GetAllByLeaseId(long leaseId)
+        {
+            return this.Context.PimsPropertyLeases.AsNoTracking().Include(p => p.Property).Include(l => l.Lease).Where(p => p.LeaseId == leaseId);
+        }
+
+        /// <summary>
+        /// Delete a propertyLeaseFile. Note that this method will fail unless all dependencies are removed first.
```
asanchezr, line 52: Seems like we could use some unit tests to cover this methods

devinleighsmith: These methods don't actually fall into our minimal requirement for unit tests: https://confluence.th.gov.bc.ca/display/PIMS/Best+Practices#BestPractices-UnitTests

So nice to have for sure, but minimal value since we would mostly just be testing the underlying EF methods.

source/frontend/src/features/leases/detail/LeasePages/improvements/ImprovementsContainer.tsx
```diff
@@ -17,7 +17,7 @@ export const ImprovementsContainer: React.FunctionComponent<
 > = ({ isEditing, formikRef, onEdit }) => {
   const { lease } = useContext(LeaseStateContext);
 
-  if (!lease?.improvements?.length) {
+  if (!lease?.improvements?.length && !isEditing) {
```
asanchezr, line 20: Thanks for fixing this. I noticed it my local and was going to ask

asanchezr (approved):

PR #2917, opened 2023-02-24
psp-4629 migrate projects api/frontend to return nullable fields from…

devinleighsmith:
… backend.

probably requires a meeting, putting up for review to allow for discussion


source/backend/core/Json/CustomJsonResultFilterAttribute.cs
```diff
@@ -0,0 +1,34 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.Options;
+
+namespace Pims.Core.Json
+{
+    public class NullJsonResultFilter : IResultFilter
```
asanchezr, line 9: mild preference for filename matching name of the class

source/backend/core/Pims.Core.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <PackageReference Include="ClosedXML" Version="0.95.4" />
     <PackageReference Include="CsvHelper" Version="27.1.0" />
+    <PackageReference Include="Macross.Json.Extensions" Version="2.2.0" />
```
asanchezr, line 13: Do we need this extension for the JsonFilter to work? I don't see it referenced anywhere. Perhaps a left-over of previous work?

asanchezr:
@devinleighsmith I like the new approach with explicit null/undefined values but perhaps we should setup a meeting with all devs to discuss adoption etc

source/frontend/src/models/api/Project.ts
```diff
@@ -1,31 +1,58 @@
...
+  costTypeCode: null,
+  regionCode: null,
+  description: null,
+  note: null,
+  products: [],
+  rowVersion: null,
+};
+
+export interface Api_Product extends Api_ConcurrentVersion_Null, Api_AuditFields {
+  id?: number | null | undefined;
```
FuriousLlama, line 35: Q: Why is it necessary that the IDs support undefined?

source/frontend/src/features/properties/map/project/models.ts
```diff
@@ -1,81 +1,90 @@
...
-  region?: NumberFieldValue;
-  summary?: string;
-  rowVersion?: number;
+  id: number | undefined;
+  projectName: string | undefined = undefined;
+  projectNumber: string | undefined = undefined;
+  projectStatusType: string | undefined = undefined;
+  region: NumberFieldValue | undefined = undefined;
+  summary: string | undefined = undefined;
+  rowVersion: number | undefined = undefined;
```
FuriousLlama, line 60: Would it be better if the forms also use null instead of undefined?

devinleighsmith: I had the same thought, my reasoning here was more to do with backwards compatibility. We have to change all of the API* fields to support nulls, the same is not true on the forms. We already frequently reformat the data from the API to be easier to consume by the forms, so this seemed like a suitable way to reduce risk.

source/frontend/src/models/api/Project.ts
devinleighsmith: It doesn't need to explicitly, I'll change this.

source/backend/core/Pims.Core.csproj
devinleighsmith: I'll double check, I originally added this for the new filter attribute but have since made changes that may no longer require this to be necessary.

source/backend/core/Json/CustomJsonResultFilterAttribute.cs
devinleighsmith: will fix.

source/frontend/src/features/properties/map/project/models.ts
devinleighsmith: The other reason being that the backend requires nulls since it does not support undefined. The frontend has no such limitations. Forcing the datatype of form fields to be x | null because of limitations on the backend seems like an abstraction leak.

PR #2933, opened 2023-02-28
Psp 5503 display warning when removing a property from an acq file that has takes/activities associated.

devinleighsmith:


source/frontend/src/features/properties/map/acquisition/AcquisitionContainer.tsx
```diff
@@ -1,28 +1,19 @@
...
-import AcquisitionMenu from './common/AcquisitionMenu';
+import { IAcquisitionViewProps } from './AcquisitionView';
 import { EditFormNames } from './EditFormNames';
 import { useAcquisitionProvider } from './hooks/useAcquisitionProvider';
-import ViewSelector from './ViewSelector';
 
 export interface IAcquisitionContainerProps {
   acquisitionFileId: number;
   onClose: () => void;
+  View: React.FunctionComponent<React.PropsWithChildren<IAcquisitionViewProps>>;
```
asanchezr, line 16: Thanks for refactoring this container into the new dependency injection pattern!

source/frontend/src/features/properties/map/acquisition/AcquisitionView.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+    return {
+      ...utils,
+      getCloseButton: () => utils.getByTitle('close'),
+    };
+  };
+
+  beforeEach(() => {
+    mockAxios.onGet(new RegExp('users/info/*')).reply(200, {});
+    mockAxios.onGet(new RegExp('notes/*')).reply(200, mockNotesResponse());
```
asanchezr, line 93: ideally we would like to move away from mockAxios for new tests but given this is a refactor I guess it's fine

source/frontend/src/features/properties/map/acquisition/ViewSelector.tsx
```diff
@@ -17,6 +17,7 @@ export interface IViewSelectorProps {
   selectedMenuIndex: number;
   setContainerState: (value: Partial<AcquisitionContainerState>) => void;
   onSuccess: () => void;
+  ref: React.RefObject<FormikProps<any>>;
```
asanchezr, line 20: nit: I think this is not needed here because ViewSelector is created with React.forwardRef(...) that appends that ref to the props, but doesn't hurt to also having it here I think

asanchezr (approved):

PR #2939, opened 2023-02-28
PSP-2194 : SPIKE: Specify string culture

eddherrera:
Added HealthCheck for Database Collation to an expected value for String Comparison purposes.

### DIFF
--- a/source/backend/api/Helpers/Healthchecks/PimsDatabaseHealtcheck.cs
+++ b/source/backend/api/Helpers/Healthchecks/PimsDatabaseHealtcheck.cs
@@ -1,0 +1,61 @@
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Pims.Api.Helpers.Healthchecks
+{
+    public class PimsDatabaseHealtcheck : IHealthCheck
+    {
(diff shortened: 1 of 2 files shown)


FuriousLlama:
Looks like the changes validate the DB string settings. How about the CA1305 warning for general string comparisons in c#?

source/backend/api/Helpers/Healthchecks/PimsDatabaseHealtcheck.cs
```diff
@@ -0,0 +1,61 @@
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Pims.Api.Helpers.Healthchecks
+{
+    public class PimsDatabaseHealtcheck : IHealthCheck
```
devinleighsmith, line 9: type on class/filename

devinleighsmith (approved):

source/backend/api/Startup.cs
```diff
@@ -241,6 +242,13 @@ public void ConfigureServices(IServiceCollection services)
                 .AddCheck("liveliness", () => HealthCheckResult.Healthy())
                 .AddSqlServer(csBuilder.ConnectionString, tags: new[] { "services" });
 
+            services.AddHealthChecks()
```
devinleighsmith, line 245: would be good to test this out in dev, but for that we may need logging as we are unlikely to be able to change the collation on the db, any ideas @eddherrera ?

PR #2943, opened 2023-02-28
Fixes to the Lease view screen

FuriousLlama:


source/frontend/src/features/leases/detail/LeasePages/details/DetailAdministration.tsx
```diff
@@ -65,10 +67,7 @@ export const DetailAdministration: React.FunctionComponent<
           />
         </SectionField>
         <SectionField label="Effective date" labelWidth="3">
-          <Input
-            disabled={disabled}
-            field={withNameSpace(nameSpace, 'responsibilityEffectiveDate')}
-          />
+          {prettyFormatDate(responsabilityDate)}
```
devinleighsmith, line 70: nit: responsibilityDate

devinleighsmith:
Not seeing this requirement satisfied: PID should have it's own space to be displayed, not following Legal description content

devinleighsmith:
@FuriousLlama this doesn't seem fixed either unless I'm missing something: **Initiator data is not displayed even an option has been selected.**

devinleighsmith:
Did this get fixed? **Responibility field displays "Headquarter" on the View Form if the user doesn't select any option in particular (Option left with default value "Select group responsible")**

devinleighsmith:
@FuriousLlama this still seems to be a problem as well: **In case an address is not present, the "Address not available in PIMS" should be located next to the Address label**

devinleighsmith (changes requested):
see comments, not all AC items addressed yet.

devinleighsmith (approved):
Based on discussion giving approval so that if you want to merge once you fix the actual bugs you can.

source/frontend/src/features/leases/detail/LeasePages/details/LeaseDetailsForm.tsx
```diff
@@ -12,21 +12,19 @@ import styled from 'styled-components';
...
   return (
     <Formik initialValues={{ ...defaultLease, ...lease }} enableReinitialize={true} onSubmit={noop}>
       <StyledDetails>
         <DetailTermInformation />
-        <PropertiesInformation disabled={!isEditing} />
-        <DetailAdministration disabled={!isEditing} />
-        <DetailDocumentation disabled={!isEditing} />
+        <PropertiesInformation disabled={true} />
+        <DetailAdministration disabled={true} />
+        <DetailDocumentation disabled={true} />
```
FuriousLlama, line 27: This needs to be refactored given the underlying components are only used for view

PR #2978, opened 2023-03-08
Added modal warning if there is a missmatch between claims

FuriousLlama:


source/backend/dal/Repositories/UserRepository.cs
```diff
@@ -116,6 +116,50 @@ public PimsUser Activate()
             return user;
         }
 
+        public bool ValidateClaims(PimsUser user)
```
devinleighsmith, line 119: This seems like a good candidate for unit tests.

devinleighsmith (approved):

PR #2980, opened 2023-03-09
PSP-5633 : FT: Acquisition File - Informative text is not populating under Owners header

eddherrera:


source/frontend/src/features/properties/map/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -248,3 +251,9 @@ const Container = styled.div`
     max-width: 25rem;
   }
 `;
+
+const StyledParagraph = styled.p`
+  color: #494949;
+  font-size: 1.6rem;
+  text-decoration: none;
```
FuriousLlama, line 258: We really should not be using one-off colors. Also, would it be possible to standardize this style so we can reuse it? I am thinking there might be other places where this kind of 'hint' is placed

devinleighsmith: Yeah, the color should be something like: border-color: ${({ theme }) => theme.css.textColor};

eddherrera: moved to the common styles to be reused in other places.

FuriousLlama (approved):

PR #2981, opened 2023-03-09
psp-5457 update document type naming.

devinleighsmith:
Think I finally sorted this one out.

### DIFF
--- a/source/frontend/src/features/documents/ComposedDocument.ts
+++ b/source/frontend/src/features/documents/ComposedDocument.ts
@@ -16,7 +16,7 @@ import Api_TypeCode from 'models/api/TypeCode';
 
 export interface ComposedDocument {
   mayanMetadata?: Api_Storage_DocumentMetadata[];
-  pimsDocument?: Api_DocumentRelationship;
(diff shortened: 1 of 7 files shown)


source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -16,7 +16,7 @@ import Api_TypeCode from 'models/api/TypeCode';
 
 export interface ComposedDocument {
   mayanMetadata?: Api_Storage_DocumentMetadata[];
-  pimsDocument?: Api_DocumentRelationship;
+  pimsDocumentRelationship?: Api_DocumentRelationship;
```
FuriousLlama, line 19: get the feeling I have seen this code before :)

FuriousLlama (approved):

source/frontend/src/features/documents/ComposedDocument.ts
devinleighsmith: Yeah, I've struggled with refactoring this code without unit tests or enforced typings.

PR #2983, opened 2023-03-09
psp-5679 product duplication fix.

devinleighsmith:


source/backend/dal/Repositories/ProductRepository.cs
```diff
@@ -42,11 +42,12 @@ public IList<PimsProduct> GetByProject(long projectId)
         /// </summary>
         /// <param name="incomingProducts"></param>
         /// <returns></returns>
-        public IEnumerable<PimsProduct> GetByProductBatch(IEnumerable<PimsProduct> incomingProducts)
+        public IEnumerable<PimsProduct> GetByProductBatch(IEnumerable<PimsProduct> incomingProducts, long projectId)
```
FuriousLlama, line 45: maybe we need to change the name of the method or change the name of the parameter or add more documentation about this function. As it is, its a tad obscure what it is doing, and why it needs a project id

devinleighsmith: fair, I'll add some more doc.

FuriousLlama (approved):

PR #2986, opened 2023-03-10
Psp 5626 create geoserver map proxy within the PIMS api

devinleighsmith:
Based on pattern in the bcgov/crt project
use basic auth and internal route to the geoserver instead of webADE

### DIFF
--- a/source/backend/api/Controllers/PimsGeoserverController.cs
+++ b/source/backend/api/Controllers/PimsGeoserverController.cs
@@ -1,0 +1,65 @@
+using System;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+using AspNetCore.Proxy;
+using AspNetCore.Proxy.Options;
+using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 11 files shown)


source/backend/api/Controllers/PimsGeoserverController.cs
```diff
@@ -0,0 +1,65 @@
...
+        #endregion
+
+        #region Endpoints
+
+        [Route("{**rest}")]
+        [HasPermission(Permissions.PropertyView)]
+        public Task ProxyCatchAll(string rest)
+        {
+            // If you don't need the query string, then you can remove this.
+            var queryString = this.Request.QueryString.Value;
```
FuriousLlama, line 50: do we need the query string?

devinleighsmith:
Sure @FuriousLlama I'll add some detail to the controller

source/backend/api/Controllers/PimsGeoserverController.cs
devinleighsmith: Yes, I'll remove the comment.

PR #2989, opened 2023-03-10
PSP-5583 : Project selection lookahead - they need to search in both project number and name

eddherrera:


source/frontend/src/hooks/useProjectTypeahead.ts
```diff
@@ -27,7 +27,8 @@ export const useProjectTypeahead = (): IProjectTypeahead => {
       if (isMounted()) {
         setMatchedProjects(
           data.map<IAutocompletePrediction>(x => {
-            return { id: x.id || 0, text: x.description?.toString() || '' };
+            let projectText = `${x.code || ''}  ${x.description}`;
```
devinleighsmith, line 30: umm, do we need || '' for description?

eddherrera: probably is not needed since description is never null. Will add it for the interface definition.

devinleighsmith (approved):

PR #2995, opened 2023-03-13
Update TermForm.test.tsx.snap

stairaku:
Changing the html values for isGstEligible since we have boeth radio buttons with "false" value


devinleighsmith:
@stairaku snapshots are automatically generated by the code for unit testing purposes, they should never be manually changed. You likely meant to change an underlying component here.

PR #2997, opened 2023-03-14
psp-5476 allow document type to be specified in Mayan sync JSON.

devinleighsmith:


source/backend/api/Models/Mayan/PimsSync/DocumentTypeModel.cs
```diff
@@ -10,5 +10,8 @@ public class DocumentTypeModel
 
         [JsonPropertyName("metadata_types")]
         public IList<DocumentMetadataTypeModel> MetadataTypes { get; set; }
+
+        [JsonPropertyName("display_order")]
+        public int DisplayOrder { get; set; }
```
FuriousLlama, line 15: looks good. The document type code will be pretty much the same.

PR #2999, opened 2023-03-14
Psp 3686 takes

devinleighsmith:


devinleighsmith:
> It would seem the DBScripts are out of scope of the ticket.

thanks, will have a look and correct.

PR #3015, opened 2023-03-20
PSP-3478 Acquisition File Checklist READ/UPDATE

asanchezr:
checklist-read

checklist-update


asanchezr:
> Tried this out and the checklist does not appear to work on pre-existing acq files. Don't remember if we decided if that was ok or not given acq files haven't hit production yet. At the very least will need to be called out on the jira.

Good question. I take the following AC to mean that for the initial items (introduced with this pr) you would need to create a new ACQ file to see the checklist. To me retro-fitting existing ACQ files is an ETL task but we could discuss tomorrow on standup.

> Operational requirement -
if a new checklist item is added to the database it should show up on the UI for all files that are created after it has been added.
Any checklist item that is disabled should not be available on the UI for files created after the checklist item has been disabled

devinleighsmith:
Can we get bobbi to update this on her mockup? I can totally see FT writing a bug for this, but I'm assuming we can't do this kind of styling since this is generated: 
image

asanchezr:
> nit: seems a tad odd that the lines are breaking text where they currently are. For example, it seems like there is lots of space for Spending Authority Approcal(s) -SAA: i but currently the i is being shifted down to the next line.
> 
> image

The space is actually ocupied by the "updated by" column - which we don't show for initial values. Try updating a few values and you'll see it (also shown on mockup)

asanchezr:
> Can we get bobbi to update this on her mockup? I can totally see FT writing a bug for this, but I'm assuming we can't do this kind of styling since this is generated: image

I brought it up with her and others and she said she was ok with the implementation as we did it. Not sure if she has cycles to update all mockups but I will ask

devinleighsmith:
**Based on the acquisition type user may have additional sections to consider on the list (sec3/6)** Are we able to support this requirement? This seems like something that would be difficult for generation without some kind of database attribute to group those optional items.

devinleighsmith:
You may get a bug logged for when you save the checklist, the screen resets to the acq file details. I came up with a solution that I didn't love that involved manipulating the default tab within the current file (https://github.com/bcgov/PSP/pull/2999/files, see defaultPropertyTab). The other option would be that we start allowing route based navigation of the tabs (currently tracked by state only).

asanchezr:
> You may get a bug logged for when you save the checklist, the screen resets to the acq file details. I came up with a solution that I didn't love that involved manipulating the default tab within the current file (https://github.com/bcgov/PSP/pull/2999/files, see defaultPropertyTab). The other option would be that we start allowing route based navigation of the tabs (currently tracked by state only).

I guess this is open to interpretation but I did ask the question about whether the default tab should be **Checklist** or keep it at **File Details** and BAs said keep File details as default. 

Also there is no specific AC saying: "After save go to the read version of the checklist" so it goes to the default tab

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -206,6 +219,33 @@ public PimsAcquisitionFile UpdateProperties(PimsAcquisitionFile acquisitionFile)
...
+            acquisitionFile.ThrowIfNull(nameof(acquisitionFile));
+            _logger.LogInformation("Updating acquisition file checklist with AcquisitionFile id: {id}", acquisitionFile.Internal_Id);
+            _user.ThrowIfNotAuthorized(Permissions.AcquisitionFileEdit);
+
+            // Get the current checklist items for this acquisition file.
+            var currentItems = _checklistRepository.GetChecklistItemsByAcquisitionFileId(acquisitionFile.Internal_Id).ToDictionary(ci => ci.Internal_Id);
+
+            foreach (var incomingItem in acquisitionFile.PimsAcquisitionChecklistItems)
+            {
+                if (!currentItems.TryGetValue(incomingItem.Internal_Id, out var existingItem))
```
devinleighsmith, line 233: Does this allow for this requirement to be met? 
**if a new checklist item is added to the database it should show up on the UI for all files that are created after it has been added.**

source/backend/dal/Repositories/AcquisitionFileChecklistRepository.cs
```diff
@@ -0,0 +1,67 @@
...
+        public IEnumerable<PimsAcqChklstItemType> GetAllChecklistItemTypes()
+        {
+            return Context.PimsAcqChklstItemTypes
+                .Include(it => it.AcqChklstSectionTypeCodeNavigation)
+                .OrderBy(it => it.DisplayOrder)
+                .AsNoTracking()
+                .ToArray();
+        }
+
+        public List<PimsAcquisitionChecklistItem> GetChecklistItemsByAcquisitionFileId(long acquisitionFileId)
```
devinleighsmith, line 43: nit: I think we are trying to use GetAll.... for any repo function that returns a list, so GetAllChecklistItemsByAcquisitionFileId

source/backend/api/Services/AcquisitionFileService.cs
asanchezr: We might have to discuss how that is to be achieved for updates. My interpretation is that the code would ensure that new ACQ files would have the new items. Those existing ones would need a DB ETL to bring them up to date so I didn't code any check on the update to add new entries. But if they want this handled via code then we would need to add more logic/complexity to it

source/backend/dal/Repositories/AcquisitionFileChecklistRepository.cs
asanchezr: will update

source/backend/entities/Helpers/Extensions/TypeCodeExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+        /// <summary>
+        /// Determines whether a given type is currently expired, based on its EffectiveDate and ExpiryDate values
+        /// </summary>
+        /// <param name="type">The type code.</param>
+        /// <param name="currentDate">The current date.</param>
+        /// <returns>True if the type is expired; false otherwise.</returns>
+        public static bool IsExpiredType(this IExpiringTypeEntity type, DateTime? currentDate = null)
+        {
+            DateTime now = currentDate.HasValue ? (DateTime)currentDate?.Date : DateTime.UtcNow.Date;
+            return (type.EffectiveDate.Date > now) || (type.ExpiryDate.HasValue && type.ExpiryDate?.Date <= now);
```
devinleighsmith, line 16: Hmm is > correct or >= for the effective date?

source/frontend/src/components/common/UserNameTooltip.tsx
```diff
@@ -39,12 +39,7 @@ export const UserNameTooltip: React.FunctionComponent<
       toolTipId={'userNameTooltip'}
       innerClassName={'userNameTooltip'}
       toolTip={userNameInfo}
-      customToolTipIcon={<StyledUserLabel>{userName ?? 'USER'}</StyledUserLabel>}
+      customToolTipIcon={<strong>{userName ?? 'USER'}</strong>}
```
devinleighsmith, line 42: this affects existing code I think, does the change to strong here keep the secondaryVariantColor?

source/frontend/src/features/mapSideBar/tabs/SectionField.tsx
```diff
@@ -1,4 +1,4 @@
-import clsx from 'classnames';
+import { default as clsx, default as cx } from 'classnames';
```
devinleighsmith, line 1: can you explain what is going on here?

source/frontend/src/features/properties/map/acquisition/ViewSelector.tsx
```diff
@@ -25,6 +27,16 @@ export const ViewSelector = React.forwardRef<FormikProps<any>, IViewSelectorProp
     // render edit forms
     if (props.isEditing && !!props.acquisitionFile) {
       switch (props.activeEditForm) {
+        case EditFormNames.acquisitionChecklist:
+          return (
+            <UpdateAcquisitionChecklistContainer
+              formikRef={formikRef as any}
```
devinleighsmith, line 33: Manuel would want me to ask why type coercion is necessary here.

source/frontend/src/features/properties/map/acquisition/update/checklist/models.ts
```diff
@@ -0,0 +1,118 @@
...
+  Api_AcquisitionFileChecklistItem,
+  Api_AcquisitionFileChecklistItemType,
+  lastModifiedBy,
+  sortByDisplayOrder,
+} from 'models/api/AcquisitionFile';
+import { Api_AuditFields } from 'models/api/AuditFields';
+import { ILookupCode } from 'store/slices/lookupCodes';
+import { fromTypeCode, toTypeCode } from 'utils/formUtils';
+
+export class AcquisitionChecklistFormModel {
```
devinleighsmith, line 12: this might be a candidate for basic unit tests, up to you.

devinleighsmith (approved):

devinleighsmith:
> > You may get a bug logged for when you save the checklist, the screen resets to the acq file details. I came up with a solution that I didn't love that involved manipulating the default tab within the current file (https://github.com/bcgov/PSP/pull/2999/files, see defaultPropertyTab). The other option would be that we start allowing route based navigation of the tabs (currently tracked by state only).
> 
> I guess this is open to interpretation but I did ask the question about whether the default tab should be **Checklist** or keep it at **File Details** and BAs said keep File details as default.
> 
> Also there is no specific AC saying: "After save go to the read version of the checklist" so it goes to the default tab. See this comment https://jira.th.gov.bc.ca/browse/PSP-3478?focusedCommentId=496622&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-496622

This would certainly fall under "assumed requirements" but I believe that normally saving should take you to the view version of that screen. Recommend checking with the BAs.

source/backend/api/Services/AcquisitionFileService.cs
devinleighsmith: Check with the Ba's on this one as well, but the AC seems to be stating that this story should be responsible for supporting that requirement, not a separate ETL story.

source/frontend/src/features/mapSideBar/tabs/SectionField.tsx
asanchezr: oops, will fix

source/frontend/src/features/properties/map/acquisition/ViewSelector.tsx
asanchezr: I'd have to check, I think because there are several forms that can receive the formik-ref but will see if typing can be improved

devinleighsmith:
> > Tried this out and the checklist does not appear to work on pre-existing acq files. Don't remember if we decided if that was ok or not given acq files haven't hit production yet. At the very least will need to be called out on the jira.
> 
> Good question. I take the following AC to mean that for the initial items (introduced with this pr) you would need to create a new ACQ file to see the checklist. To me retro-fitting existing ACQ files is an ETL task but we could discuss tomorrow on standup.
> 
> > Operational requirement -
> > if a new checklist item is added to the database it should show up on the UI for all files that are created after it has been added.
> > Any checklist item that is disabled should not be available on the UI for files created after the checklist item has been disabled

Can we at least make sure that the BA's are aware of this and that there is an ETL subtask for this story? my guess is that there is not which would mean that all ETL'd data would have broken checklists.

source/backend/api/Services/AcquisitionFileService.cs
asanchezr: I will confirm with BAs but reading the bolded statement above, it requires that new ACQ get the new item from DB but does not imply anything about updating old ACQ checklists

asanchezr: I take it that way because it says, "show in the UI for the files CREATED after the entry was added to the db", not for files UPDATED

source/backend/entities/Helpers/Extensions/TypeCodeExtensions.cs
asanchezr: The logic is to determine when a type code is expired (or not). Based on conversations for another entity on the system that supports the same concept of "expiry" with two dates (effective - expiryDate) this is the actual text from Confluence:

> **Effective date:** As of this date the code is available for selection in PIMS
> **Expiry date:**  As of this date the code is not available in PIMS for selection.
>
> **NOTE**: Since both effective and expiry dates are inclusive of the provided date, the expiry date should be at least one day after the effective date

So the here the code is expired if the effective date has not been reached yet, but will become active the day of so > makes sense here for effective date. Also remember that we are dealing with dates without time component (by doing UtcNow.Date)

source/frontend/src/components/common/UserNameTooltip.tsx
asanchezr: **TooltipIcon.scss** already defines that anything contained by it will have the secondary variant color, so it was redundant in the custom icon. The color is preserved.

```scss

.tooltip-icon {
  min-width: 1.4rem;
  min-height: 1.4rem;
  color: $secondary-variant-color; // <---
  margin: 0rem 0.2rem;
}

```

PR #3021, opened 2023-03-22
Lease Consultations | features/psp 4715

FuriousLlama:


source/frontend/src/features/leases/add/ConsultationSubForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  // Not all consultations might be comming from the backend. Add the ones missing.
+  if (consultations.length !== consultationTypes.length) {
+    const newConsultations: FormLeaseConsultation[] = [];
+
+    consultationTypes.forEach(x => {
+      const newConsultation = FormLeaseConsultation.fromApiLookup(values.id || 0, x);
+
+      // If there is a consultation with the type, set the status to the existing one
+      let existingConsultation = consultations.find(y => y.consultationType === x.id);
```
devinleighsmith, line 33: mild preference to use named variables instead of placeholders like x and y. placeholders are fine in simple logic but this is starting to get harder to read.

source/frontend/src/features/leases/add/ConsultationSubForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+const ConsultationSubForm: React.FunctionComponent<
+  React.PropsWithChildren<IConsultationSubFormProps>
+> = ({ formikProps }) => {
+  const { values, setFieldValue } = formikProps;
+  const { consultations } = values;
+  const { getOptionsByType, getByType } = useLookupCodeHelpers();
+  const consultationTypes = getByType(API.CONSULTATION_TYPES);
+  const consultationStatusTypes = getOptionsByType(API.CONSULTATION_STATUS_TYPES);
+
+  // Not all consultations might be comming from the backend. Add the ones missing.
```
devinleighsmith, line 25: nit: comming

source/frontend/src/features/leases/add/ConsultationSubForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+      newConsultations.push(newConsultation);
+    });
+    setFieldValue('consultations', newConsultations);
+  }
+
+  return (
+    <Section header="Consultation" isCollapsable initiallyExpanded>
+      {values.consultations.map((a, i) => (
+        <SectionField
+          key={`consultations.${i}`}
```
devinleighsmith, line 50: given that all consultations have a unique type can we use that instead of i here? (since react recommends avoiding i for mapped components in jsx where possible).

source/frontend/src/features/leases/detail/LeasePages/details/DetailConsultation.tsx
```diff
@@ -0,0 +1,66 @@
...
+      newConsultations.push(newConsultation);
+    });
+    setFieldValue('consultations', newConsultations);
+  }
+
+  return (
+    <Section header="Consultation" initiallyExpanded isCollapsable>
+      {values.consultations.map((consultation, index) => (
+        <SectionField
+          key={`consultations.${index}`}
```
devinleighsmith, line 54: same comment here re: index

source/frontend/src/features/leases/detail/LeasePages/details/DetailConsultation.tsx
```diff
@@ -0,0 +1,66 @@
...
+        id: 0,
+        parentLeaseId: values.id || 0,
+        consultationType: { id: x.id.toString(), description: x.name },
+        consultationStatusType: { id: 'UNKNOWN', description: 'Unknown' },
+        rowVersion: 0,
+      };
+
+      // If there is a consultation with the type, set the status to the existing one
+      let existingConsultation = values.consultations.find(y => y.consultationType?.id === x.id);
+      if (existingConsultation !== undefined) {
```
devinleighsmith, line 40: this seems pretty similar to the logic in the edit, would it be worth extracting this so that any fixes/enhancements can be made in one area instead of two?

source/frontend/src/features/mapSideBar/tabs/ltsa/LtsaDuplicateTitleSubForm.tsx
```diff
@@ -28,7 +28,7 @@ export const LtsaDuplicateTitleSubForm: React.FunctionComponent<
                 `duplicateCertificatesOfTitle.${index}`,
               );
               return (
-                <>
+                <React.Fragment key={`${title}-${index}`}>
```
devinleighsmith, line 31: thanks for this.

FuriousLlama:
> @FuriousLlama are tests coming in a future PR? there does seem to be some non-trivial logic in how the consultations handle new values and display existing values.

Yes, I just wanted to get confirmation that we want to have the logic on the front end vs having it on the backend.

source/frontend/src/features/leases/add/ConsultationSubForm.tsx
FuriousLlama: Sure thing!

source/frontend/src/features/leases/add/ConsultationSubForm.tsx
FuriousLlama: could do that

source/frontend/src/features/leases/add/ConsultationSubForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+              )}
+            </Col>
+          </Row>
+        </SectionField>
+      ))}
+    </Section>
+  );
+};
+
+export const isLeaseCategoryVisible = (typeId?: string) => {
```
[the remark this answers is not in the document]
FuriousLlama: yep, dead code...

source/frontend/src/features/leases/detail/LeasePages/details/DetailConsultation.tsx
FuriousLlama: I agree, they are both almost the same. The reason that I didn't do that is because the lease code could use some refactoring and this could be part of it. If you feel strongly about it I can do that now.

devinleighsmith:
> > @FuriousLlama are tests coming in a future PR? there does seem to be some non-trivial logic in how the consultations handle new values and display existing values.
> 
> Yes, I just wanted to get confirmation that we want to have the logic on the front end vs having it on the backend.

Huh, is there a realistic potential of someone calling the service method to create a lease where we would want that consultation logic? actually even if they did this frontend change would still work.

FuriousLlama:
> > > @FuriousLlama are tests coming in a future PR? there does seem to be some non-trivial logic in how the consultations handle new values and display existing values.
> > 
> > 
> > Yes, I just wanted to get confirmation that we want to have the logic on the front end vs having it on the backend.
> 
> Huh, is there a realistic potential of someone calling the service method to create a lease where we would want that consultation logic? actually even if they did this frontend change would still work right?

I don't think so. The front end handling the addition for new/older leases makes sense to me vs having the backend do that work. Just wanted to make sure that we agreed it was the best approach ;)

devinleighsmith:
> > > > @FuriousLlama are tests coming in a future PR? there does seem to be some non-trivial logic in how the consultations handle new values and display existing values.
> > > 
> > > 
> > > Yes, I just wanted to get confirmation that we want to have the logic on the front end vs having it on the backend.
> > 
> > 
> > Huh, is there a realistic potential of someone calling the service method to create a lease where we would want that consultation logic? actually even if they did this frontend change would still work right?
> 
> I don't think so. The front end handling the addition for new/older leases makes sense to me vs having the backend do that work. Just wanted to make sure that we agreed it was the best approach ;)

I'm good with it based on what I see here.

devinleighsmith (approved):

PR #3025, opened 2023-03-22
Psp 5584 form tab on acq file, ability to add a form based on type to acq file

devinleighsmith:


source/backend/api/Models/Concepts/ActivityInstance/ActivityInstanceFileModel.cs
```diff
@@ -1,6 +1,6 @@
 namespace Pims.Api.Models.Concepts
 {
-    public class ActivityInstanceFileModel
+    public class ActivityInstanceFileModel : BaseAppModel
```
FuriousLlama, line 3: is this change related to activities?

source/backend/api/Models/Concepts/Form/FileFormModel.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Pims.Api.Models.Concepts
+{
+    public class FileFormModel : BaseAppModel
+    {
+        public long Id { get; set; }
+
+        public long FileId { get; set; }
+
+        public LookupModel<string> FormTypeCode { get; set; }
```
FuriousLlama, line 11: Not sure the form type matches the lookup exactly given it also has a document id

source/frontend/src/features/properties/map/form/hooks/useFormRepository.tsx
FuriousLlama, line 1: Lets move repositories that interact with the api to the /hook/repositories folder instead of being part of the feature path

source/backend/api/Models/Concepts/Form/FileFormModel.cs
devinleighsmith: Oh, right. Its a lookup with that one extra attribute. All right if I fix that in my next PR? will probably be easier.

FuriousLlama: I will change it once it I merge it.

FuriousLlama (approved):

PR #3030, opened 2023-03-23
psp-5684 add additional feedback when selected map property is loadin…

devinleighsmith:
…g and bca data is loading.

Add timeouts to bca data load. Shorten default timeouts.


devinleighsmith:
In my testing this is not a bug, but actually just poor feedback. There are a number of external service calls here, some that can take a long time. Also, we have to chain together (1) a call to the fa parcel map layer to get the pid, and then (2) a call to bc assessment using the pid from the previous step in order to get the address.

I've set a time limit to BCA call to prevent excessive load times since that data is optional anyways.
Also I've added a spinner to the map when waiting for selected property data and to the sidebar when waiting for bca data.

PR #3034, opened 2023-03-24
Psp 5282

devinleighsmith:


codecov.yml
```diff
@@ -29,3 +29,8 @@ comment:
 
 ignore:
   - "*.ignore.*"
+  - "PimsContext.cs"
```
devinleighsmith, line 32: not sure if people agree with me on these, but to me PimsContext, PimsBaseContext and everything in the entities folder doesn't need testing.

Swagger/healthchecks are a bit more arguable.

source/backend/api/Controllers/AuthController.cs
```diff
@@ -74,20 +74,6 @@ public IActionResult Activate()
             return new JsonResult(new Model.UserModel(user.Internal_Id, user.GuidIdentifierValue.Value, hasValidClaims));
         }
 
-        /// <summary>
-        /// Redirect user to registration page.
-        /// </summary>
-        /// <returns></returns>
-        [HttpGet("register")]
-        [SwaggerOperation(Tags = new[] { "auth" })]
-        public IActionResult Register(string redirect_uri)
```
devinleighsmith, line 83: we don't have a registration page anymore.

source/frontend/src/utils/download.ts
```diff
@@ -1,81 +0,0 @@
...
-  fileName: string;
-  actionType: string;
-}
-
-/**
- * Programmatically triggers a file download with content generated through an API
- * @param filename the file name
- * @param blobData Raw blob data as returned by a file export API
- */
-export const downloadFile = (filename: string, blobData: any) => {
```
devinleighsmith, line 22: technically we were still using this but only for user management so I updated that code to follow our conventions and then was able to remove this.

codecov.yml
devinleighsmith: based on discussion, we want to ignore any frontend models, that would include legacy models as well as Concepts. Just need to make sure to only target the models, and not the mappings.

lets apply this to the frontend API models as well. also we should be ignoring mocks.

PR #3038, opened 2023-03-25
psp-5871 fix display issue with full screen sidebar.

devinleighsmith:


source/frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -237,79 +238,88 @@ const Map: React.FC<React.PropsWithChildren<MapProps>> = ({
...
-        )}
-
-        <ReactLeafletMap
-          center={[lat, lng]}
-          zoom={lastZoom}
-          maxZoom={MAP_MAX_ZOOM}
-          closePopupOnClick={true}
-          ref={handleMapCreated}
-          whenReady={handleMapReady}
+    <VisibilitySensor partialVisibility={true}>
```
devinleighsmith, line 241: when the sidebar is hidden, stop displaying map elements. Fixes issues with absolutely positioned items and causing the fullscreen sidebar to scroll.

PR #3044, opened 2023-03-28
psp-5585 add form list view, popup display of form data.

devinleighsmith:


FuriousLlama (changes requested):
This will need to be merged with the merge I am working on of the other form story. tldr; please wait that story

PR #3045, opened 2023-03-29
hide listview download icon until file detail performance is fixed.

devinleighsmith:
based on discussion, mayan file details api performs badly when many concurrent requests are made. Moving that call to just the view files popup as it performs better in that context. Download only possible from view screen now.


FuriousLlama (commented):
Looks good

FuriousLlama (approved):

PR #3046, opened 2023-03-29
File Document & Template Management |  psp-5544

FuriousLlama:
Note:
FormDocuments reads less ambiguous than Forms. But not hard set on it.


source/backend/api/Areas/Documents/DocumentRelationshipController.cs
```diff
@@ -62,45 +65,45 @@ public class DocumentRelationshipController : ControllerBase
...
         /// <param name="parentId">Used to identify document's parent entity.</param>
         /// <returns></returns>
-        [HttpGet("{relationshipType}/{parentId:long}")]
+        [HttpGet("{relationshipType}/{parentId}")]
         [Produces("application/json")]
         [HasPermission(Permissions.DocumentView)]
         [ProducesResponseType(typeof(IList<DocumentRelationshipModel>), 200)]
         [SwaggerOperation(Tags = new[] { "document" })]
-        public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipType, long parentId)
+        public IActionResult GetRelationshipDocuments(DocumentRelationType relationshipType, string parentId)
```
devinleighsmith, line 73: huh, I don't really get this. Why is this better?

source/backend/api/Areas/FormDocument/FormDocumentController.cs
```diff
@@ -0,0 +1,91 @@
...
+        #endregion
+
+        #region Endpoints
+
+        /// <summary>
+        /// Gets the form document types.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet]
+        [HasPermission(Permissions.FormView)]
```
devinleighsmith, line 56: Hmm, I think these may need to be locked down to admins too right?

devinleighsmith: Actually maybe just the add.

source/backend/api/Models/Concepts/FormDocument/FormDocumentFileModel.cs
```diff
@@ -0,0 +1,11 @@
+namespace Pims.Api.Models.Concepts
+{
+    public class FormDocumentFileModel : BaseAppModel
```
devinleighsmith, line 3: Thanks for this.

source/backend/api/Models/Concepts/FormDocument/FormDocumentTypeModel.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public class FormDocumentTypeModel : BaseAppModel
+    {
+        /// <summary>
+        /// get/set - Form type code.
+        /// </summary>
+        public string FormTypeCode { get; set; }
+
+        /// <summary>
+        /// get/set - Form type document id.
```
devinleighsmith, line 14: These comments are strictly optional in my mind, in case you want to omit them in the future. I think we really only need comments when a field is doing something unusual.

source/backend/api/Areas/Documents/DocumentRelationshipController.cs
FuriousLlama: Sadly, It is necessary. The id that the relationship has is a string (Form Type Code). Having the relationship use a string type allows both ids' as numbers and the type code as a string

source/backend/api/Services/FormDocumentService.cs
```diff
@@ -0,0 +1,136 @@
...
+        public async Task<DocumentUploadRelationshipResponse> UploadFormDocumentTemplateAsync(string formTypeCode, DocumentUploadRequest uploadRequest)
+        {
+            this.Logger.LogInformation("Uploading template for document form type");
+            this.User.ThrowIfNotAuthorized(Permissions.DocumentAdmin);
+
+            var currentFormType = _formTypeRepository.GetByFormTypeCode(formTypeCode);
+            if (currentFormType.DocumentId != null)
+            {
+                var result = await DeleteFormDocumentTemplateAsync(currentFormType);
+                if (result.Status != ExternalResultStatus.Success)
```
devinleighsmith, line 69: Is there an advantage to using ExternalResult here instead of a boolean? (asking because I used a boolean for a delete on my story but would standardize if there was a benefit).

FuriousLlama: It is more flexible. The caller would have more knowledge about why the call failed.

devinleighsmith: oh wait I'm being thick, this is a call to Mayan not to the repo, sorry.

source/backend/api/Services/FormDocumentService.cs
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    throw new InvalidOperationException("Could not remove existing template");
+                }
+                else
+                {
+                    // Retrieve the form type again to get the latest after deletion.
+                    currentFormType = _formTypeRepository.GetByFormTypeCode(formTypeCode);
```
devinleighsmith, line 76: Hmm, why do we need this? does the DeleteFormDocumentTemplateAsync call make a change to the database as well as mayan?

FuriousLlama: It follows the usual process for deleting documents; If there is only one, then it deletes both the mayan stored one and the pims reference to it,

source/backend/api/Services/FormDocumentService.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            DocumentUploadResponse uploadResult = await _documentService.UploadDocumentAsync(uploadRequest);
+
+            DocumentUploadRelationshipResponse relationshipResponse = new DocumentUploadRelationshipResponse()
+            {
+                UploadResponse = uploadResult,
+            };
+
+            if (uploadResult.DocumentExternalResult.Status == ExternalResultStatus.Success && uploadResult.Document != null && uploadResult.Document.Id != 0)
```
devinleighsmith, line 87: nit: can maybe shorten to uploadResult.Document?.Id != 0

source/backend/api/Services/FormDocumentService.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                int currentCount = _documentRepository.DocumentRelationshipCount((long)documentId);
+                if (currentCount == 1)
+                {
+                    return await _documentService.DeleteDocumentAsync(formType.Document);
+                }
+                else
+                {
+                    formType.DocumentId = null;
+                    _formTypeRepository.SetFormTypeDocument(formType);
```
devinleighsmith, line 115: nit: I think we generally use Update* instead of Set* for methods like this.

source/backend/api/Services/FormDocumentService.cs
```diff
@@ -0,0 +1,136 @@
...
+                var updatedFormType = _formTypeRepository.SetFormTypeDocument(currentFormType);
+                _formTypeRepository.CommitTransaction();
+
+                relationshipResponse.DocumentRelationship = _mapper.Map<DocumentRelationshipModel>(updatedFormType);
+            }
+
+            return relationshipResponse;
+        }
+
+        public async Task<ExternalResult<string>> DeleteFormDocumentTemplateAsync(PimsFormType formType)
```
devinleighsmith, line 99: This is an interesting case, we are only ever updating some metadata on this type rather then changing it.

source/backend/dal/Repositories/FormTypeRepository.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public PimsFormType SetFormTypeDocument(PimsFormType formType)
+        {
+            using var queryScope = Logger.QueryScope();
+
+            var existingFormType = Context.PimsFormTypes
+                .FirstOrDefault(x => x.FormTypeCode == formType.FormTypeCode) ?? throw new KeyNotFoundException();
+
+            Context.Entry(existingFormType).CurrentValues.SetValues(formType);
```
devinleighsmith, line 55: I wonder if we should just manually update the documentId and nothing else? are there any other fields user's are allowed to change?

source/backend/entities/Partials/FormType.cs
```diff
@@ -3,11 +3,15 @@
 namespace Pims.Dal.Entities
 {
     /// <summary>
-    /// PimsFormType class, provides an entity for the datamodel to manage form types.
+    /// PimsFormType class, provides an entity for the datamodel to manage a list of property road types.
```
devinleighsmith, line 6: a list of property road types?

source/backend/entities/Partials/FormType.cs
```diff
@@ -3,11 +3,15 @@
...
-    public partial class PimsFormType : ITypeEntity<string>
+    public partial class PimsFormType : StandardIdentityBaseAppEntity<string>, IBaseEntity, ITypeEntity<string>
     {
         #region Properties
+
+        [NotMapped]
+        public override string Internal_Id { get => FormTypeCode; set => FormTypeCode = value; }
+
         [NotMapped]
         public string Id { get => this.FormTypeCode; set => this.FormTypeCode = value; }
```
devinleighsmith, line 16: nit: I think if you are making this not a lookup you can take this out.

source/backend/tests/unit/api/Services/FormDocumentServiceTest.cs
```diff
@@ -0,0 +1,322 @@
...
+            // Act
+            Func<Task> act = () => service.UploadFormDocumentTemplateAsync(testTypeCode, testUploadRequest);
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+            documentService.Verify(x => x.UploadDocumentAsync(testUploadRequest), Times.Never);
+        }
+
+        [Fact]
+        public void UploadFormDocumentTemplateAsync_DeleteIfPrevious()
```
devinleighsmith, line 154: Thanks for these.

source/frontend/src/features/admin/document-template/__snapshots__/DocumentTemplateManagementContainer.test.tsx.snap
```diff
@@ -119,7 +119,7 @@ exports[`DocumentTemplateManagementContainer component matches snapshot 1`] = `
                   class="form-control"
                 >
                   <option>
-                    Select an Activity type
+                    Select an Form type
```
devinleighsmith, line 122: nit: Select a Form type

source/backend/dal/Repositories/FormTypeRepository.cs
FuriousLlama: not right now.

source/backend/dal/Constants/DocumentFormType.cs
```diff
@@ -0,0 +1,10 @@
+namespace Pims.Dal.Constants
+{
+    public static class DocumentFormType
```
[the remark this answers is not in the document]
FuriousLlama: Yep not used. Added it because I thought it might be useful later

source/frontend/src/features/documents/documentUpload/DocumentUploadForm.tsx
```diff
@@ -56,7 +52,7 @@ const DocumentUploadForm: React.FunctionComponent<
   };
 
   const initialFormData = new DocumentUploadFormData(
-    documentStatusTypes[0]?.value?.toString(),
+    props.documentStatusOptions[0]?.value?.toString(),
```
devinleighsmith, line 55: nit: would be slightly safer/more readable to search for none instead of hardcoding the first index of the array.

source/backend/entities/Partials/FormType.cs
FuriousLlama: ups bad merge

source/frontend/src/features/documents/documentUpload/DocumentUploadForm.tsx
```diff
@@ -107,6 +103,7 @@ const DocumentUploadForm: React.FunctionComponent<
                 field="documentTypeId"
                 options={documentTypes}
                 onChange={props.onDocumentTypeChange}
+                disabled={documentTypes.length === 1}
```
devinleighsmith, line 106: what is this for?

FuriousLlama: Makes the combo box a field if there is only one option. Better than clicking the combobox and not be able to change anything

source/frontend/src/hooks/repositories/useFormDocumentRepository.ts
```diff
@@ -0,0 +1,48 @@
...
+export const useFormDocumentRepository = () => {
+  const getFormDocumentTypes = useApiRequestWrapper<
+    () => Promise<AxiosResponse<Api_FormDocumentType[], any>>
+  >({
+    requestFunction: useCallback(async () => await getFormDocumentTypesApi(), []),
+    requestName: 'getFormDocumentTypes',
+    onSuccess: useAxiosSuccessHandler(),
+    invoke: true,
+    onError: useAxiosErrorHandler(
+      'Failed to load form document types. Either refresh the page to try again or try and load a different activity.',
```
devinleighsmith, line 21: this should probably not be activity.

devinleighsmith:
Hmm, on my machine I can't select a document status during upload, trying to figure out why.

devinleighsmith:
> Hmm, on my machine I can't select a document status during upload, trying to figure out why.

NVM, PEBKAC, seems fine to me. No major issues that would block merge.

devinleighsmith (approved):

PR #3052, opened 2023-03-30
psp-5722 set global mapster default to prevent recursion.

devinleighsmith:


source/backend/api/Models/Concepts/AcquisitionFile/AcquisitionFileChecklistItemMap.cs
```diff
@@ -8,7 +8,6 @@ public class AcquisitionFileChecklistItemMap : IRegister
         public void Register(TypeAdapterConfig config)
         {
             config.NewConfig<Entity.PimsAcquisitionChecklistItem, AcquisitionFileChecklistItemModel>()
-                .PreserveReference(true)
```
devinleighsmith, line 11: Doing this was strictly optional but I like having things concise where possible.

asanchezr (approved):
Thanks for making this change. Good quality of life improvement

PR #3054, opened 2023-03-31
swap boundaries layer for parcels layer to reduce cpu load.

devinleighsmith:


source/frontend/src/components/maps/hooks/useMapSearch.tsx
```diff
@@ -118,7 +118,6 @@ export const useMapSearch = () => {
 };
 
 /**
- * TODO: PSP-4390 convert polygon features into lat/lng coordinates, remove this when new Point Geoserver layer available.
  * @param feature the feature to obtain lat/lng coordinates for.
  * @returns [lat, lng]
  */
```
devinleighsmith, line 123: This is sort of the key function, it returns a point from either a polygon or a point, which is why everything works with the boundary layer which is all polygons. Swapping to points should reduce the amount of cpu required to display all of the map properties since each property would not need to be converted to a point first.

asanchezr (approved):

PR #3056, opened 2023-03-31
PSP-5643 : Enhancements - acquisition file take areas

eddherrera:
image


source/backend/api/Areas/Takes/Controllers/TakeController.cs
```diff
@@ -99,15 +99,15 @@ public IActionResult UpdateAcquisitionPropertyTakes(long acquisitionFileProperty
...
         /// </summary>
         /// <returns></returns>
-        [HttpGet("property/{propertyId:long}/count")]
+        [HttpGet("acquisition/{acquisitionFileId:long}/property/{propertyId:long}/count")]
         [HasPermission(Permissions.AcquisitionFileView, Permissions.PropertyView)]
         [Produces("application/json")]
         [ProducesResponseType(typeof(TakeModel), 200)]
         [SwaggerOperation(Tags = new[] { "take" })]
-        public IActionResult GetTakesCountByPropertyId(long propertyId)
+        public IActionResult GetTakesCountByPropertyId([FromRoute]long acquisitionFileId, [FromRoute]long propertyId)
```
devinleighsmith, line 110: Why was this change necessary? you already had the:
A)all takes for the current property.
B)all takes for the current property on the current file

So if you want the count of takes that aren't part of the acquisition file, you simply subtract b from a. Am I missing something?

devinleighsmith (changes requested):
see above comment.

source/frontend/src/features/mapSideBar/tabs/takes/update/models.ts
```diff
@@ -28,6 +28,10 @@ export const TakesYupSchema = Yup.object().shape({
           .required('Required when flag is true')
           .moreThan(0, 'Must be greater than 0 when flag is true'),
       }),
+      ltcEndDt: Yup.string().when('isLicenseToConstruct', {
```
devinleighsmith, line 31: this is the error that prints when ltc end date is not specified: Must be greater than 0 when flag is true

devinleighsmith, line 31: this isn't working, I'm able so save without providing a value when the flag is true.

devinleighsmith:
Section 16 end date should just be End date according to confluence.

source/backend/api/Areas/Takes/Controllers/TakeController.cs
eddherrera: @devinleighsmith the PropertyId would only let me know how many takes does the property has on all Acquisition File, but the new enhancement would need to know which Acquisition File is being view to discard the takes in that Acquisition File. 
B) is not true since propertyId could be on any given acquisition file.

eddherrera:
> Section 16 end date should just be End date according to confluence.

Updated

source/backend/api/Areas/Takes/Controllers/TakeController.cs
devinleighsmith: We discussed this, moving to the strategy of the frontend doing the counts. The frontend will use: 1 count endpoint as well as the actual takes response, do the subtraction, and display the final number.

eddherrera: Updated

devinleighsmith (approved):

PR #3057, opened 2023-03-31
PSP-5750 : Add "Product number and name" to acquisition file header

eddherrera:
image


source/frontend/src/features/properties/map/acquisition/common/AcquisitionHeader.test.tsx
```diff
@@ -48,4 +48,24 @@ describe('AcquisitionHeader component', () => {
...
+    ).toBeVisible();
+  });
+
+  it('renders the file Product code and description concatenated', async () => {
+    const testAcquisitionFile = mockAcquisitionFileResponse();
+    const { getByText } = setup({ acquisitionFile: testAcquisitionFile });
+
+    expect(getByText('Ministry product:')).toBeVisible();
+    expect(getByText('00048 - MISCELLANEOUS CLAIMS')).toBeVisible();
+  });
```
FuriousLlama, line 70: Please add one where the code/description is null or undefined given there is some logic there to filter those out

eddherrera: Should I still need to make that test since both fields are not nullable in the backend?

FuriousLlama: The testing code should not make any assumptions that are not explicit in the code being tested. In this case, the codes can be undefined or null according to the types used in the model.

Also, we do not use explicit nulls on the backend except for numeral primitives. That way any object reference could be null

eddherrera: Updated

FuriousLlama: Thank you for updating. I should have been more explicit; The code that does the format is what needs testing given it has some logic around it[1].
So a acq with a product, but with either code or description being null

[1]``
 {formatMinistryProduct(
                acquisitionFile?.product?.code,
                acquisitionFile?.product?.description,
              )}
``

source/frontend/src/features/mapSideBar/tabs/HeaderField.tsx
```diff
@@ -31,8 +35,18 @@ export const HeaderLabelCol: React.FC<IHeaderLabelColProps> = props => (
 
 export const HeaderContentCol: React.FC<
   React.PropsWithChildren<IHeaderContentColProps>
-> = props => (
-  <Col xs={props.contentWidth ?? 'auto'} className="pl-1 text-left">
-    <strong>{props.children}</strong>
-  </Col>
-);
+> = props => {
+  return props.valueTestId ? (
```
devinleighsmith, line 39: I don't love this, could we just set the datatestid to the prop value or empty string so we don't need a ternary?

devinleighsmith (approved):

source/frontend/src/features/properties/map/acquisition/common/AcquisitionHeader.test.tsx
eddherrera: Removed unnecessary logic.

source/frontend/src/features/mapSideBar/tabs/HeaderField.tsx
eddherrera: Removed ternary, React indeed doesn't add the property when the value passed is null or undefined.

PR #3061, opened 2023-04-05
HOTFIX: Is49 fixes

devinleighsmith:
PSP-5863
PSP-5794
PSP-4486
PSP-5821

### DIFF
--- a/source/frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
+++ b/source/frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
@@ -152,7 +152,7 @@ const useActiveFeatureLayer = ({
           : DistrictCodes.Unknown,
         DISTRICT_NAME: district.DISTRICT_NAME ?? 'Cannot determine',
       };
-      if (activeFeature.geometry.type === 'Polygon') {
+      if (activeFeature?.geometry?.type === 'Polygon') {
         activeFeatureLayer?.addData(activeFeature);
       }
       setState({ type: MapStateActionTypes.SELECTED_FEATURE, selectedFeature: activeFeature });
--- a/source/frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
+++ b/source/frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
@@ -152,7 +152,9 @@ const useActiveFeatureLayer = ({
           : DistrictCodes.Unknown,
         DISTRICT_NAME: district.DISTRICT_NAME ?? 'Cannot determine',
       };
-      activeFeatureLayer?.addData(activeFeature);
+      if (activeFeature.geometry.type === 'Polygon') {
+        activeFeatureLayer?.addData(activeFeature);
+      }
       setState({ type: MapStateActionTypes.SELECTED_FEATURE, selectedFeature: activeFeature });
     } finally {
       setState({ type: MapStateActionTypes.LOADING, loading: false });
--- a/source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
+++ b/source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
@@ -36,6 +36,12 @@ export const Address: React.FunctionComponent<React.PropsWithChildren<IAddressPr
   const { setFieldValue, values } = useFormikContext();
   const countryId = getIn(values, withNameSpace(namespace, 'countryId'));
 
+  const line1Count = !!getIn(values, withNameSpace(namespace, 'streetAddress1')) ? 1 : 0;
+  const line2Count = !!getIn(values, withNameSpace(namespace, 'streetAddress2')) ? 1 : 0;
+  const line3Count = !!getIn(values, withNameSpace(namespace, 'streetAddress3')) ? 1 : 0;
+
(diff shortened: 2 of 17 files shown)


source/frontend/src/components/maps/hooks/useActiveFeatureLayer.tsx
```diff
@@ -152,7 +152,9 @@ const useActiveFeatureLayer = ({
           : DistrictCodes.Unknown,
         DISTRICT_NAME: district.DISTRICT_NAME ?? 'Cannot determine',
       };
-      activeFeatureLayer?.addData(activeFeature);
+      if (activeFeature.geometry.type === 'Polygon') {
```
devinleighsmith, line 155: PSP-5863 basically this code is trying to add a highlighted feature even if no feature data is available, that results in a placeholder marker with a lat/lng set to the click location. Since we don't have a marker set up for that it displays as a missing placeholder image.

source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
```diff
@@ -36,6 +36,12 @@ export const Address: React.FunctionComponent<React.PropsWithChildren<IAddressPr
   const { setFieldValue, values } = useFormikContext();
   const countryId = getIn(values, withNameSpace(namespace, 'countryId'));
 
+  const line1Count = !!getIn(values, withNameSpace(namespace, 'streetAddress1')) ? 1 : 0;
```
devinleighsmith, line 39: This logic seems needlessly complex but rewriting it seems outside the scope of this hotfix.

Basically there are 2 major problems:
1) addressLines is an optional parameter that is not being set. I'm guessing the author meant to set that but forgot, so it always defaults to 1 even if multiple are present.
2) because this form displays as the data is loading, even if there are addressLines during the loading period there are not, which results in a useState in the useCounter hook being set to 1. useState initial values are not updated when they change so I had to add a useEffect to useCounter to allow the values of **initial** to propogate.

source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
```diff
@@ -51,20 +57,20 @@ export const Address: React.FunctionComponent<React.PropsWithChildren<IAddressPr
 
   // this counter determines how many address lines we render in the form; e.g. street1, street2, etc
   const { count, increment, decrement } = useCounter({
-    initial: addressLines ?? 1,
+    initial: addressLines || 1,
```
devinleighsmith, line 60: this field is what I updated in useCounter.

source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
```diff
@@ -51,20 +57,20 @@ export const Address: React.FunctionComponent<React.PropsWithChildren<IAddressPr
...
     min: 1,
     max: 3,
   });
-
   // clear extra address fields when they get removed from address form...
-  useEffect(() => {
```
devinleighsmith, line 60: this useEffect used to have a race condition where sometimes the streetAddress3 and 2 were cleared during the load. updated this to a function instead to only trigger removal in response to a click.

source/frontend/src/features/properties/map/project/hooks/useAddProjectFormManagement.tsx
```diff
@@ -14,6 +16,13 @@ export interface IUseAddProjectFormProps {
 export function useAddProjectForm(props: IUseAddProjectFormProps) {
   const { onSuccess } = props;
   const { addProject } = useProjectProvider();
+  const { getPublicByType } = useLookupCodeHelpers();
+  const initialValues = props.initialForm ?? new ProjectForm();
+  if (!initialValues.projectStatusType) {
```
devinleighsmith, line 21: want to avoid fully hardcoding this in case the 'AC' value is removed or disabled.

source/frontend/src/features/contacts/contactUtils.ts
```diff
@@ -208,7 +208,7 @@ export function formAddressToApiAddress(
   return {
     ...formAddress,
     countryId: parseInt(formAddress?.countryId.toString()) || 0,
-    provinceId: isEmpty(formAddress?.provinceId)
+    provinceId: isEmpty(formAddress?.provinceId.toString())
```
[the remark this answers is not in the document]
FuriousLlama: that makes sense. From their docs
```
 /**
         * Checks if value is empty. A value is considered empty unless it’s an arguments object, array, string, or
         * jQuery-like collection with a length greater than 0 or an object with own enumerable properties.
         *
         * @param value The value to inspect.
         * @return Returns true if value is empty, else false.
         */
         ```

source/frontend/src/features/properties/map/project/hooks/useAddProjectFormManagement.tsx
FuriousLlama: Even better would be to have the status types as enums

source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
FuriousLlama: yeah this looks quite complex for a list of addresses

FuriousLlama (approved):

PR #3069, opened 2023-04-06
psp-5861 modifying documents should not refresh file.

devinleighsmith:


source/frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -83,13 +74,8 @@ const DocumentListContainer: React.FunctionComponent<
   };
 
   const updateCallback = useCallback(() => {
-    // Check if the component is working in a File context. If it is, delegate the update.
```
devinleighsmith, line 86: @FuriousLlama I believe the reason we were doing this was because of activity documents and file documents being on the same screen so we had to force refresh the whole file when documents was changed. I don't think we need to do that anymore??? hopefully because this is much simpler.

source/frontend/src/features/properties/map/acquisition/detail/AcquisitionFileTabs.tsx
```diff
@@ -114,7 +114,10 @@ export const AcquisitionFileTabs: React.FC<IAcquisitionFileTabsProps> = ({
       tabViews={tabViews}
       defaultTabKey={defaultTab}
       activeTab={activeTab}
-      setActiveTab={setActiveTab}
+      setActiveTab={(tab: FileTabNames) => {
```
FuriousLlama, line 117: Just to note that we need to refactor the ViewSelector. It has grown in complexity before the initial implementation.

source/frontend/src/features/documents/list/DocumentListContainer.tsx
FuriousLlama: Right, this was forcing the update.

FuriousLlama (approved):

source/frontend/src/features/properties/map/acquisition/detail/AcquisitionFileTabs.tsx
devinleighsmith: Agreed, may want to just use routing so that user's can link other's to specific tabs and the back button would work.

PR #3078, opened 2023-04-12
Psp 5880 - generate letter

devinleighsmith:
I don't think it would be appropriate to include the templates in the repo, so I'm going to attach the actual (functional) template to confluence.

### DIFF
--- a/source/backend/api/Services/DocumentGenerationService.cs
+++ b/source/backend/api/Services/DocumentGenerationService.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Security.Claims;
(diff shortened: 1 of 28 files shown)


source/backend/api/Services/DocumentGenerationService.cs
```diff
@@ -74,33 +75,39 @@ public async Task<ExternalResult<FileDownload>> GenerateDocument(string template
 
             // TODO: This needs to retrieve by the passed template type. At this point that is not possible.
             PimsDocument document = _documentRepository.GetAllByDocumentType("CDOGS Template").LastOrDefault();
```
FuriousLlama, line 77: This needs to be updated now that there are proper document types

devinleighsmith: Hmm, I'll need to log a follow up story I guess, since for this story those changes aren't present yet.

FuriousLlama (approved):

PR #3086, opened 2023-04-14
psp-5702 mayan doc sync enhancements.

devinleighsmith:


source/backend/api/Models/Mayan/PimsSync/MetadataModel.cs
```diff
@@ -0,0 +1,14 @@
+using System.Text.Json.Serialization;
+
+namespace Pims.Api.Models.Mayan.Sync
+{
+    public class MetadataModel
+    {
+
```
FuriousLlama, line 7: nit: spacing

source/frontend/src/models/api/Document.ts
```diff
@@ -17,6 +17,7 @@ export interface Api_Document extends Api_ConcurrentVersion, Api_AuditFields {
 export interface Api_DocumentType extends Api_ConcurrentVersion, Api_AuditFields {
   id?: number;
   documentType: string | undefined;
+  documentTypeDescription: string | undefined;
   mayanId: number | undefined;
```
FuriousLlama, line 21: since we are making the change, these should be null

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -57,28 +57,40 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
 
             // Add the metadata types not in mayan
             IList<Task<ExternalResult<MetadataType>>> createTasks = new List<Task<ExternalResult<MetadataType>>>();
+            IList<Task<ExternalResult<MetadataType>>> updateTasks = new List<Task<ExternalResult<MetadataType>>>();
             foreach (var metadataTypeLabel in model.MetadataTypes)
```
FuriousLlama, line 61: We should also delete metadata if the remove_lingering_metadata_types is set to true

FuriousLlama: nvm, line 87 covers that.

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -57,28 +57,40 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
             {
-                if (metadata.Payload.Results.FirstOrDefault(x => x.Label == metadataTypeLabel) == null)
+                var matchingMetadataResult = metadata.Payload.Results.FirstOrDefault(x => x.Name == metadataTypeLabel.Name);
```
FuriousLlama, line 63: because it matches by name, if the old metadata types names won't be the same and will end up being removed and new ones created. If it is in prod with metadata that might cause issues.  We need a one time sync that will convert the metadata on mayan to use the names before using the regular sync process

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
 
-            ExternalBatchResult batchResult = new ExternalBatchResult();
-
-            Task<ExternalResult<QueryResult<DocumentType>>> documentTypeTask = mayanDocumentRepository.TryGetDocumentTypesAsync(pageSize: 5000);
-            documentTypeTask.Wait();
-
-            ExternalResult<QueryResult<DocumentType>> documentTypesRetrieved = documentTypeTask.Result;
+            IList<PimsDocumentTyp> pimsDocumentTypes = documentTypeRepository.GetAll();
 
             // Add the document types not in mayan
```
FuriousLlama, line 116: nit: this is adding the documents not in the JSON file request, not mayan.

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
-                IList<Task<ExternalResult<string>>> deleteTasks = new List<Task<ExternalResult<string>>>();
-                foreach (var documentType in documentTypesRetrieved.Payload.Results)
+                // Disable the document types that are not on the sync model
+                foreach (var documentType in pimsDocumentTypes)
                 {
-                    if (model.DocumentTypes.FirstOrDefault(x => x.Label == documentType.Label) == null)
+                    if (!model.DocumentTypes.Any(x => x.Name == documentType.DocumentType))
                     {
-                        deleteTasks.Add(mayanDocumentRepository.TryDeleteDocumentTypeAsync(documentType.Id));
+                        // TODO: currently there is no is_disabled field available for document types.
```
FuriousLlama, line 142: There is a is disabled now :)

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
+            ExternalResult<QueryResult<DocumentType>> mayanDocumentTypes = documentTypeTask.Result;
 
             IList<PimsDocumentTyp> pimsDocumentTypes = documentTypeRepository.GetAll();
 
-            // Add the document types not in the backend
-            IList<PimsDocumentTyp> createdDocumentTypes = new List<PimsDocumentTyp>();
-            IList<PimsDocumentTyp> updatedDocumentTypes = new List<PimsDocumentTyp>();
-            foreach (var mayanDocumentType in mayanResult.Payload.Results)
+            // Add the document types not in mayan
+            IList<AddDocumentToMayanWithNameResponseModel> createResponseTasks = new List<AddDocumentToMayanWithNameResponseModel>();
```
FuriousLlama, line 166: nit: createResponseTasks => createRequestTasks 
These tasks are to create new entries in mayan.

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
+                if (matchingTypeFromMayan == null)
                 {
-                    var newPimsDocType = new PimsDocumentTyp() { MayanId = mayanDocumentType.Id, DocumentType = mayanDocumentType.Label, DisplayOrder = matchingTypeFromModel?.DisplayOrder };
-                    createdDocumentTypes.Add(documentTypeRepository.Add(newPimsDocType));
+                    createResponseTasks.Add(new AddDocumentToMayanWithNameResponseModel() { AddDocumentTypeTask = mayanDocumentRepository.TryCreateDocumentTypeAsync(new DocumentType() { Label = pimsDocumentTyp.DocumentTypeDescription }), Name = pimsDocumentTyp.DocumentType });
                 }
-                else if (pimsDocumentTypes.FirstOrDefault(x => x.MayanId == mayanDocumentType.Id && (mayanDocumentType.Label != x.DocumentType || x.DisplayOrder != matchingTypeFromModel?.DisplayOrder)) != null)
+                else if (matchingTypeFromMayan.Label != pimsDocumentTyp.DocumentTypeDescription)
                 {
                     // if the Mayan id is the same but the label or display order has changed, update the document type in PIMS.
```
FuriousLlama, line 177: This is no longer true. If the Mayan id is the same but the label is different update the document type in mayan

FuriousLlama: just the comment, the logic is sound

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
+                    var matchingPimsType = pimsDocumentTypes.FirstOrDefault(x => x.DocumentType == task.Name);
+                    matchingPimsType.MayanId = task.AddDocumentTypeTask.Result.Payload.Id;
+                    documentTypeRepository.Update(matchingPimsType);
+                }
+                batchResult.CreatedDocumentType.Add(task.AddDocumentTypeTask.Result);
+            }
+            documentTypeRepository.CommitTransaction();
 
-            IList<PimsDocumentTyp> removedDocumentTypes = new List<PimsDocumentTyp>();
+            pimsDocumentTypes = documentTypeRepository.GetAll(); // re-retrieve the document list from pims as it may have been updated.
```
FuriousLlama, line 199: q: why is this necessary? if it has changed the previous steps might be wrong too.

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -94,113 +106,122 @@ public ExternalBatchResult SyncMayanMetadataTypes(SyncModel model)
...
+                else if (matchingTypeFromMayan.Label != pimsDocumentTyp.DocumentTypeDescription)
                 {
                     // if the Mayan id is the same but the label or display order has changed, update the document type in PIMS.
-                    var updatedPimsDocType = new PimsDocumentTyp() { MayanId = mayanDocumentType.Id, DocumentType = mayanDocumentType.Label, DisplayOrder = matchingTypeFromModel?.DisplayOrder };
-                    updatedDocumentTypes.Add(documentTypeRepository.Update(updatedPimsDocType));
+                    matchingTypeFromMayan.Label = pimsDocumentTyp.DocumentTypeDescription;
+                    updateTasks.Add(mayanDocumentRepository.TryUpdateDocumentTypeAsync(matchingTypeFromMayan));
                 }
             }
+            Task.WaitAll(createResponseTasks.Select(x => x.AddDocumentTypeTask).Concat(updateTasks).ToArray());
```
FuriousLlama, line 182: not critical, but the delete tasks could be done also concurrently with the create + update

source/backend/api/Services/DocumentSyncService.cs
```diff
@@ -223,66 +248,69 @@ private void SyncDocumentTypeMetadataTypes(SyncModel model, ref ExternalBatchRes
...
             ExternalBatchResult batchResult = new ExternalBatchResult();
 
-            var documentType = retrievedDocumentTypes.FirstOrDefault(x => x.Label == documentTypeModel.Label);
-            ExternalResult<QueryResult<DocumentTypeMetadataType>> documentTypeMetadataTypes = await mayanDocumentRepository.TryGetDocumentTypeMetadataTypesAsync(documentType.Id, pageSize: 5000);
-
-            // Update the document type's metadata types
-            IList<Task<ExternalResult<DocumentTypeMetadataType>>> documentTypeMetadataTypeTasks = new List<Task<ExternalResult<DocumentTypeMetadataType>>>();
-            var retrievedLinks = documentTypeMetadataTypes.Payload.Results;
-            foreach (var metadataTypeModel in documentTypeModel.MetadataTypes)
+            var documentType = retrievedDocumentTypes.FirstOrDefault(x => x.Id == pimsDocumentTyp.MayanId);
```
FuriousLlama, line 255: nit" for readability
documentType  -> mayanDocumentType

FuriousLlama (changes requested):
There needs to be a one-time sync for the existing metadata info that changes the name given thats the field used for matching now. Otherwise existing metadata will be removed for every doctype

devinleighsmith:
@FuriousLlama I believe I've addressed your comments.

FuriousLlama:
The migrate metadata function seemed to work without any problems. The mayan migrate errors out if there where no documents initially on PIMS but they where on mayan. 
Not necessary now, but the return info should be enhanced to have which doc_type failed instead of only what mayan returned:

```
{
            "status": "Error",
            "message": "{\"label\":[\"Document type with this Label already exists.\"]}",
            "httpStatusCode": "BadRequest"
        },
        ```

FuriousLlama:
Oh, and don't forget to update the docs with the new migrate call :)

PR #3092, opened 2023-04-19
Generate H0443 | psp-5711

FuriousLlama:


source/backend/api/Services/PropertyService.cs
```diff
@@ -29,16 +31,31 @@ public PimsProperty GetById(long id)
...
+        {
+            _logger.LogInformation("Getting multiple properies by id");
+            _user.ThrowIfNotAuthorized(Permissions.PropertyView);
+
+            List<PimsProperty> properties = _propertyRepository.GetAllByIds(ids);
+            foreach (PimsProperty property in properties)
+            {
+                if (property?.Location != null)
+                {
+                    property.Location = TransformCoordiates(property.Location);
```
devinleighsmith, line 52: nit: coordinates

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -147,7 +147,7 @@ public PimsProperty GetById(long id)
         /// </summary>
         /// <param name="ids"></param>
         /// <returns></returns>
-        public PimsProperty GetAllByIds(IEnumerable<long> ids)
```
devinleighsmith, line 150: Changing the return type makes sense, but why the input type? are we saying only lists are valid inputs but not arrays, sets, etc?

source/frontend/src/constants/formDocumentTypes.ts
```diff
@@ -0,0 +1,16 @@
+export enum FormDocumentType {
```
devinleighsmith, line 1: nit: enum should match file name.

source/backend/api/Constants/FormDocumentType.cs
```diff
@@ -0,0 +1,33 @@
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace Pims.Api.Constants
```
devinleighsmith, line 4: thanks for this.

source/backend/api/Services/PropertyService.cs
```diff
@@ -72,5 +89,12 @@ public PimsProperty Update(PimsProperty property)
 
             return GetById(newProperty.Internal_Id);
         }
+
+        private Point TransformCoordiates(Geometry location)
+        {
+            // return property spatial location in lat/long (4326)
+            var newCoords = _coordinateService.TransformCoordinates(SpatialReference.BCALBERS, SpatialReference.WGS84, location.Coordinate);
```
devinleighsmith, line 96: Yeah, I think this is fair for now that we are always going to be going from albers to wgs84?

source/backend/dal/Constants/FormTypeCode.cs
```diff
@@ -1,6 +1,6 @@
 namespace Pims.Dal.Constants
 {
-    public static class DocumentFormType
+    public static class FormTypeCode
```
devinleighsmith, line 3: Is it our normal pattern to have 2 enums for a single type on the backend? one for the dal and one for the api?

source/frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -16,9 +16,9 @@ import PropertyListView from './PropertyListView';
 jest.mock('@react-keycloak/web');
 jest.mock('hooks/pims-api');
 
-const mockApiGetPropertiesPaged = jest.fn<Promise<AxiosResponse<IPagedItems<IProperty>>>, any>();
+const mockApiGetPropertiesPagedApi = jest.fn<Promise<AxiosResponse<IPagedItems<IProperty>>>, any>();
```
devinleighsmith, line 19: nit: mockApiGetPropertiesPagedApi thats a lot of "Api"

source/frontend/src/features/properties/map/acquisition/AcquisitionContainer.test.tsx
```diff
@@ -261,16 +261,4 @@ describe('AcquisitionContainer component', () => {
 
     await waitFor(async () => viewProps.formikRef.current?.submitCount === 1);
   });
-
```
devinleighsmith, line 264: ?

source/frontend/src/features/properties/map/acquisition/common/AcquisitionMenu.tsx
```diff
@@ -64,21 +66,7 @@ const AcquisitionMenu: React.FunctionComponent<
...
-            className="no-gutters"
-            onClick={() => {
-              props.onGenerateLetter();
-            }}
-          >
-            <Col>Generate Letter</Col>
-          </StyledRow>
-        </StyledMenuGenerateWrapper>
-      )}
+      <GenerateFormContainer acquisitionFileId={props.acquisitionFileId} View={GenerateFormView} />
```
devinleighsmith, line 69: Oh, I see. Sure this can be in a subcontainer.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+    getAcquisitionFile: { execute: getAcquisitionFile },
+  } = useAcquisitionProvider();
+
+  const {
+    getMultiplePropertiesById: { execute: getMultipleProperties },
+  } = useProperties();
+
+  const { generateDocumentDownloadWrappedRequest: generate } = useDocumentGenerationRepository();
+
+  const generateLetter = async (acquisitionFileId: number) => {
```
devinleighsmith, line 25: generateLetter -> generateH0443

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+      });
+      generatedFile?.status === ExternalResultStatus.Success!! &&
+        generatedFile?.payload &&
+        showFile(generatedFile?.payload);
+    }
+  };
+  return generateLetter;
+};
+
+function getOwnerName(owner: Api_AcquisitionFileOwner): string {
```
devinleighsmith, line 88: I would like to standardize this if we can. You and I should discuss this. I think we'll have a bunch of headaches if we have multiple slightly different implementations of this kicking around.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+  } else {
+    return {
+      last_name: owner.lastNameAndCorpName || '',
+      first_name: owner.givenName || '',
+      phone: owner.contactPhoneNum || '',
+    };
+  }
+}
+
+function getPropertyAddress(address: Api_Address): Address {
```
devinleighsmith, line 119: same here, lets standardize this. I don't care if where this lives but I don't see a good reason why we would re-implement this every time we need an address.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+    municipality: address.municipality || '',
+    postal_code: address.postal || '',
+    province: address.province?.description || '',
+    street_address_1: address.streetAddress1 || '',
+    street_address_2: address.streetAddress2 || '',
+    street_address_3: address.streetAddress3 || '',
+  };
+}
+
+function getProperty(property: Api_Property): Property {
```
devinleighsmith, line 130: same here as well I think.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+  return {
+    address: property.address !== undefined ? getPropertyAddress(property.address) : null,
+    legal_description: property.landLegalDescription || '',
+    pid: property.pid?.toString() || '',
+  };
+}
+
+interface Person {
```
devinleighsmith, line 138: nit: co-locating this with the hook that uses this makes sense, but maybe not in the same file.

devinleighsmith (changes requested):
I'd like us to talk about standardizing some common attributes, or at least making a decision not to standardize. I also want to come up with a solution for error handling during document generation.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
```diff
@@ -0,0 +1,172 @@
...
+      last_name: owner.lastNameAndCorpName || '',
+      first_name: owner.givenName || '',
+      phone: owner.contactPhoneNum || '',
+    };
+  }
+}
+
+function getPropertyAddress(address: Api_Address): Address {
+  return {
+    municipality: address.municipality || '',
```
devinleighsmith, line 121: Also, in my story at least all of the addresses were defined as needing to hide empty lines when a field doesn't exist: ie: don't display blank lines for streetAddress2 if not defined. This is obnoxious in carbone. Not sure if your story has that same requirement.

I went with a single string into the template and then convCLRF to add the newlines.

devinleighsmith:
Is the template for this one on the jira? just in case I want to check this out to test generation.

source/backend/dal/Constants/FormTypeCode.cs
FuriousLlama: Well, one is not an enum. but we do have this sometimes

FuriousLlama:
> Is the template for this one on the jira? just in case I want to check this out to test generation.

Yes it is, I uploaded it to the ticket

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateH0443.ts
FuriousLlama: ah, that is a good point. I dont remember being explicitly stated, but it would make sense

source/backend/dal/Repositories/PropertyRepository.cs
FuriousLlama: We can change it if the need arise. As it is it, only passes things from the controller, which only takes a list

devinleighsmith (approved):

PR #3095, opened 2023-04-20
Wired consultations other description | psp-5927

FuriousLlama:


devinleighsmith:
For the life of me I can't seem to find requirements for how this should display but it seems to work given some manual testing: 
image

devinleighsmith (approved):

PR #3097, opened 2023-04-20
Added note logic when updating research status | PSP-5874

FuriousLlama:
image


eddherrera (approved):
Maybe add a unit test for the note addition?

FuriousLlama:
> Maybe add a unit test for the note addition?

Good idea. Added a couple

PR #3103, opened 2023-04-24
Psp 5743 - add address to initial property when adding via map popup flyout.

devinleighsmith:


source/frontend/src/components/maps/providers/MapStateContext.tsx
```diff
@@ -123,9 +123,6 @@ export const MapStateContextProvider: React.FC<
           return produce(prevState, draft => {
             draft.loading = action.loading;
             if (action.loading === true) {
-              draft.selectedInventoryProperty = null;
```
devinleighsmith, line 126: this should not be reset during loading, only the currently selected feature.

asanchezr (approved):

PR #3110, opened 2023-04-26
Psp 5707 - agreements

devinleighsmith:


source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
```diff
@@ -0,0 +1,89 @@
...
+      ? getPersonConcept(negotiatingAgent?.personId)
+      : Promise.resolve(null);
+    const provincialSolicitorConcept = provincialSolicitor?.personId
+      ? getPersonConcept(provincialSolicitor?.personId)
+      : Promise.resolve(null);
+    const ownerSolicitorConcept = ownerSolicitor?.personId
+      ? getPersonConcept(ownerSolicitor?.personId)
+      : Promise.resolve(null);
+
+    const persons = await Promise.all([
```
FuriousLlama, line 46: nice use of asynchronous calls

source/frontend/src/features/properties/map/agreement/detail/AgreementContainer.tsx
```diff
@@ -5,6 +5,7 @@ import { useEffect } from 'react';
 import { useCallback } from 'react';
 import { useContext } from 'react';
 
+import { useGenerateAgreement } from '../../acquisition/common/GenerateForm/hooks/useGenerateAgreement';
```
FuriousLlama, line 8: nit: this looks like a smell that we should move the hooks to somewhere more common

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
```diff
@@ -0,0 +1,89 @@
...
+import { AgreementTypes, Api_Agreement } from 'models/api/Agreement';
+import { ExternalResultStatus } from 'models/api/ExternalResult';
+import { Api_GenerateAgreement } from 'models/generate/GenerateAgreement';
+import { Api_GenerateFile } from 'models/generate/GenerateFile';
+
+export const useGenerateAgreement = () => {
+  const { getPersonConcept } = useApiContacts();
+  const { getAcquisitionFile, getAcquisitionFileProperties } = useApiAcquisitionFile();
+  const { generateDocumentDownloadWrappedRequest: generate } = useDocumentGenerationRepository();
+  const generateLetter = async (agreement: Api_Agreement) => {
```
FuriousLlama, line 15: nit: having this function retrieve the agreement instead of expecting it to be passed would make it more reusable

source/frontend/src/models/api/AcquisitionFile.ts
```diff
@@ -45,6 +45,14 @@ export interface Api_AcquisitionFilePerson extends Api_ConcurrentVersion, Api_Au
   isDisabled?: boolean;
 }
 
+export interface Api_AcquisitionFileOwnerSolicitor extends Api_ConcurrentVersion, Api_AuditFields {
+  id?: number;
+  acquisitionFileId?: number;
+  personId?: number;
+  person?: Api_Person;
+  isDisabled?: boolean;
```
FuriousLlama, line 53: Should these be null instead?

source/frontend/src/models/generate/GenerateAgreement.ts
```diff
@@ -0,0 +1,32 @@
...
+  completion_date: string;
+  termination_date: string;
+  commencement_date: string;
+  status: string;
+  deposit_amount: string;
+  purchase_price: string;
+  survey_plan_number: string;
+  no_later_than_days: string;
+
+  constructor(agreement: Api_Agreement, file: Api_GenerateFile) {
```
FuriousLlama, line 19: nit: file-> generateFile (imo file implies that this is the api_acqfile vs the generate kind)

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
devinleighsmith: That is actually a bit of a sticky implementation detail that we should probably think through. with this implementation we generate based on the agreement that is currently visible. If we swapped this to request on click in theory that could result in pulling someone else's concurrent changes. That could either be a good or a bad thing. Might need to talk to the BA's

source/frontend/src/models/generate/GenerateFile.ts
```diff
@@ -7,19 +7,45 @@ import { GenerateProperty } from './GenerateProperty';
...
   owners: GenerateOwner[];
+  person_owners: GenerateOwner[];
+  organization_owners: GenerateOwner[];
   file_number: string;
   file_name: string;
   project_number: string;
   project_name: string;
+  prov_solicitor?: GeneratePerson;
+  owner_solicitor?: GeneratePerson;
+  neg_agent?: GeneratePerson;
```
FuriousLlama, line 20: I think we are moving away from the implied undefined to null.

source/frontend/src/models/api/AcquisitionFile.ts
devinleighsmith: yes, will update. still following the old pattern brainlessly

source/frontend/src/models/generate/GenerateFile.ts
```diff
@@ -7,19 +7,45 @@ import { GenerateProperty } from './GenerateProperty';
 export class Api_GenerateFile {
```
FuriousLlama, line 7: Is the Api_GenerateFile class a generic data generation class?

FuriousLlama (approved):

source/frontend/src/models/generate/GeneratePerson.ts
```diff
@@ -22,5 +27,11 @@ export class GeneratePerson {
         : '';
     this.organizations =
       person?.personOrganizations?.map(o => o.organization?.name).join(', ') ?? '';
+    this.address = person
+      ? new GenerateAddress(getApiPersonOrOrgMailingAddress(person) ?? null)
+      : null;
+    this.full_name_string = [this.given_name, this.middle_names, this.last_name]
```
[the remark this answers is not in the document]
devinleighsmith: will use that instead.

source/frontend/src/features/properties/map/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
FuriousLlama: that is fair, however, the other data (what is being retrieved on the hook) does not follow that pattern. As it is it expects the caller to have retrieved only the agreement data previously.

source/frontend/src/models/generate/GenerateFile.ts
devinleighsmith: its a mapping between our api model and a model used for generation.
The idea was to use a similar pattern for how to map between our api models and frontend.

PR #3123, opened 2023-05-02
Ran scaffold and clenead up unused models. Fixed formating

FuriousLlama:


source/backend/entities/ef/BcaAreaAmendment.cs
FuriousLlama, line 1: cleanup of old entity models

devinleighsmith (approved):

PR #3126, opened 2023-05-02
PSP-5887 Manage users - Identify a user as a staff/contractor

asanchezr:
user-profile-edit

image


devinleighsmith:
This looks fine. Seeing what seems to be an unrelated issue when trying to accept an access request however. May have been introduced by the previous contractor PR.

devinleighsmith (approved):

asanchezr:
Yeah, I had to unflatten the usertype on a model that is shared by both Access Request and User profile form - that's why the "unrelated" changes. What issue do you see with access request?

PR #3127, opened 2023-05-03
psp-5858 create new roles, role_claims. Migrate existing roles to new…

devinleighsmith:
… equivalents. Remove old roles.


source/database/mssql/scripts/dbscripts/PSP_PIMS_S52_00/Build/075_DML_PIMS_ROLE_CLAIM.sql
```diff
@@ -58,11 +58,11 @@ select @rolePimsR = CLAIM_ID FROM PIMS_CLAIM where NAME = 'ROLE_PIMS_R';
 DECLARE @adminUsers BIGINT;
 DECLARE @adminProjects BIGINT;
 DECLARE @adminProperties BIGINT;
-DECLARE @systemAdministrator BIGINT;
```
devinleighsmith, line 61: This is an unrelated error in these scripts, there is a system-administrator claim and a System Administrator role. Thankfully this would not have caused adverse changes as both values have the same id (1).

Makefile
```diff
@@ -255,7 +255,7 @@ db-clean: ## create a new, clean database using the script file in the database.
 
 db-seed: ## create a new, database seeded with test data using the script file in the database. defaults to using the folder specified in database/mssql/.env, but can be overriden with n=PSP_PIMS_S15_00.
 	@echo "$(P) Seed the database with test data. n=FOLDER_NAME (PSP_PIMS_S15_00)"
-	TARGET_SPRINT=$(n) SEED=TRUE docker-compose up -d --build database;
+	TARGET_SPRINT=$(n) SEED=TRUE docker-compose up -d --build --force-recreate database;
```
devinleighsmith, line 258: --force-recreate ensures that a pre-existing database can't interfere with a seed.

FuriousLlama (approved):

Makefile
FuriousLlama: good, i've been doing a seed on the previous just to force it

PR #3142, opened 2023-05-09
Limit documents extensions | psp-5855

FuriousLlama:


source/backend/api/Services/DocumentService.cs
```diff
@@ -32,6 +33,28 @@ public class DocumentService : BaseService, IDocumentService
         private readonly IAvService avService;
         private readonly IMapper mapper;
 
+        private static readonly string[] validExtensions = {
```
FuriousLlama, line 36: This could be in the DB. Currently, the schema does not support that so its static

devinleighsmith (approved):

PR #3145, opened 2023-05-09
Fixed leases not being created with new property

FuriousLlama:


source/frontend/src/components/maps/providers/MapStateContext.tsx
```diff
@@ -112,6 +110,7 @@ export const MapStateContextProvider: React.FC<
             draft.selectedFeature = action.selectedFeature;
           });
         case MapStateActionTypes.SELECTED_FILE_FEATURE:
+          console.log('MapStateContextProvider.SELECTED_FILE_FEATURE', action.selectedFileFeature);
```
eddherrera, line 113: is this console.log needed?

FuriousLlama: good catch, it isnt

devinleighsmith (approved):

PR #3146, opened 2023-05-10
PROD HOTFIX: Psp 6146 ensure selected properties match map draft marker numbering

devinleighsmith:


devinleighsmith:
This will be validated in test before it is hotfixed to uat and prod.

PR #3150, opened 2023-05-10
psp-5863 remove popup when expand arrow clicked on layer controls

buddy326:
Disabled map layer popup from displaying when expand arrow gets clicked on the layer controls.

### DIFF
--- a/source/frontend/src/components/maps/leaflet/LayersControl/LayersMenu.tsx
+++ b/source/frontend/src/components/maps/leaflet/LayersControl/LayersMenu.tsx
@@ -203,9 +203,25 @@ const LayersTree: React.FC<React.PropsWithChildren<{ items: TreeMenuItem[] }>> =
           return (
             <ParentNode key={node.key} id={node.key}>
               {node.isOpen ? (
-                <OpenedIcon onClick={node.toggleNode} />
+                <OpenedIcon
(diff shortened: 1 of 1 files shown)


source/frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -290,13 +290,6 @@ const Map: React.FC<React.PropsWithChildren<MapProps>> = ({
                     ))}
                   </LayerGroup>
                 )}
-                {!!layerPopup && (
-                  <LayerPopup
-                    ref={popupRef}
-                    layerPopup={layerPopup}
-                    onViewPropertyInfo={onViewPropertyClick}
-                  />
-                )}
```
asanchezr, line 299: I'm not sure this is the right fix. This would remove the layer popup for **every** map interaction. I think we still want to show popups on map click, right @devinleighsmith ?

devinleighsmith: correct, probably what we need to do is prevent event propagation when clicking on the layer tree.

devinleighsmith: @buddy326 does the above give you enough to try and attempt this fix again?

devinleighsmith (approved):

PR #3171, opened 2023-05-16
PSP-5859 : Restrict contractors to acquisition files where they are on the acq team

eddherrera:


source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Security.Claims;
+using Pims.Dal.Entities;
+using Pims.Dal.Exceptions;
+using Pims.Dal.Repositories;
+
+namespace Pims.Core.Extensions
+{
+    public static class PrincipalExtensions
+    {
+        public static void HasAccessToAcquisitionFile(this ClaimsPrincipal principal, IUserRepository userRepository, PimsAcquisitionFile acquisitionFile)
```
devinleighsmith, line 11: Preference for Throw * naming. In my mind a Has* function would return a boolean.

source/backend/entities/Partials/AcquisitionFile.cs
```diff
@@ -10,6 +11,11 @@ public partial class PimsAcquisitionFile : StandardIdentityBaseAppEntity<long>,
         #region Properties
         [NotMapped]
         public override long Internal_Id { get => this.AcquisitionFileId; set => this.AcquisitionFileId = value; }
+
+        public bool PersonIsAssignedToFile(long personId)
```
devinleighsmith, line 15: hmm, I think I'm a bit uncomfortable with this being on the entity directly, since there is no guarantee that the child entities have been loaded.

I think we need an extension method that:
1) always calls the acquisition repository in a manner that ensures the acqpeople are loaded, probably using the file id.
2) checks if the person id is assigned.

If not we would be relying on the caller to know that they need to pass an acq file with the people loaded, otherwise we would return a false negative.

source/backend/entities/Partials/User.cs
```diff
@@ -18,6 +18,12 @@ public partial class PimsUser : StandardIdentityBaseAppEntity<long>, IDisableBas
         [NotMapped]
         public override long Internal_Id { get => UserId; set => UserId = value; }
 
+        /// <summary>
+        /// get - User is Contractor
+        /// </summary>
+        [NotMapped]
+        public bool IsContractor => UserTypeCode is not null && UserTypeCode == EnumUserTypeCodes.CONTRACT.ToString();
```
devinleighsmith, line 25: unlike the above, I don't think we have an issue with data loading here since we should always have the userTypeCode.

Slight preference for implementing as an extension method since this field is not in the database and may confuse someone.

devinleighsmith (changes requested):
See comments, I feel like the only thing that really needs changing is the function on the acquisition file entity, I feel like that could easily result in false negatives that could easily evade unit/integration testing.

source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
eddherrera: Renamed

source/backend/entities/Partials/AcquisitionFile.cs
eddherrera: moved the check to the extension method.

source/backend/entities/Partials/User.cs
eddherrera: ok

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -281,6 +289,7 @@ public PimsCompensationRequisition AddCompensationRequisition(long acquisitionFi
             _logger.LogInformation("Adding compensation requisition for acquisition file id ...", acquisitionFileId);
 
             _user.ThrowIfNotAuthorized(Permissions.CompensationRequisitionAdd);
+            _user.ThrowInvalidAccessToAcquisitionFile(_userRepository, _acqFileRepository, acquisitionFileId);
```
devinleighsmith, line 292: I think we need this on GetAcquisitionCompensations, UpdateAgreements, GetAgreement, Add, GetChecklistItems, GetOwners, GetProperties

devinleighsmith: @eddherrera with this change I think we are good.

devinleighsmith (approved):

PR #3172, opened 2023-05-16
psp-6119 display error message when required fields missing

buddy326:
If the code and logic looks good, I can implement this change in all places the SideBarFooter component is used and push my branch again so every form with required fields would display an error message if the save button is pressed while those fields are missing.

### DIFF
--- a/source/frontend/src/features/leases/add/AddLeaseContainer.tsx
+++ b/source/frontend/src/features/leases/add/AddLeaseContainer.tsx
@@ -29,8 +29,8 @@ export const AddLeaseContainer: React.FunctionComponent<
   const history = useHistory();
   const formikRef = useRef<FormikProps<LeaseFormModel>>(null);
   const { selectedFileFeature } = React.useContext(MapStateContext);
-
   const { addLease } = useAddLease();
+  const [errorMessage, setErrorMessage] = useState<string | undefined>();
   const [addLeaseParams, setAddLeaseParams] = useState<
     { lease: Api_Lease; userOverride?: string } | undefined
   >();
@@ -71,6 +71,16 @@ export const AddLeaseContainer: React.FunctionComponent<
   };
 
   const handleSave = () => {
+    const isFormValid = formikRef?.current?.isValid;
+
+    if (!isFormValid) {
+      setErrorMessage('Please check the required fields.');
+    }
+
+    if (isFormValid) {
+      setErrorMessage(undefined);
+    }
+
     if (formikRef !== undefined) {
       formikRef.current?.setSubmitting(true);
       formikRef.current?.submitForm();
@@ -90,6 +100,7 @@ export const AddLeaseContainer: React.FunctionComponent<
           isOkDisabled={formikRef.current?.isSubmitting || bcaLoading}
           onSave={handleSave}
           onCancel={handleCancel}
+          errorMessage={errorMessage}
         />
       }
       showCloseButton
--- a/source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
+++ b/source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
@@ -9,6 +9,7 @@ interface ISidebarFooterProps {
   editMode?: boolean;
   onEdit?: (editMode: boolean) => void;
   onCancel: () => void;
+  errorMessage?: string | undefined;
 }
 
 const SidebarFooter: React.FunctionComponent<ISidebarFooterProps> = ({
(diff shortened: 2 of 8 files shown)


source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
```diff
@@ -18,18 +20,31 @@ const SidebarFooter: React.FunctionComponent<ISidebarFooterProps> = ({
   onSave,
   onCancel,
   isOkDisabled,
+  isMissingFields,
 }) => {
+  const [errorMessage, setErrorMessage] = useState('');
```
devinleighsmith, line 25: Whenever implementing changes, we always want to implement the minimum possible added state, as every usage of state complicates the application and generally requires testing.

Additionally, we should avoid using state in non-container components.
files with *container* in the name should be used to hold state and contain state manipulation functions. 
all other components are considered "display" components and should not contain state. If view components require the ability to manipulate state they should do that via props.

source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
```diff
@@ -18,18 +20,31 @@ const SidebarFooter: React.FunctionComponent<ISidebarFooterProps> = ({
...
+    if (isMissingFields) {
+      setErrorMessage('Please check the required fields.');
+    }
+  };
+
   return (
     <SidebarFooterBar className="justify-content-end mt-auto no-gutters">
       {(!showEdit || editMode) && (
         <>
+          <Col xs="auto" className="pr-3" style={{ paddingTop: '0.7rem', color: 'red' }}>
```
devinleighsmith, line 38: this can likely be done with the className: "invalid-feedback"
Also in general the issue with inlining styles like this is that it makes the css very hard to override - the specificity is very high. Also inlined styles aren't reusable in the way that classnames or styled components are.

source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
```diff
@@ -18,18 +20,31 @@ const SidebarFooter: React.FunctionComponent<ISidebarFooterProps> = ({
...
+            {errorMessage}
```
devinleighsmith, line 39: instead of storing the errorMessage in a state variable, how about conditionally rendering a section with a static error message depending on if isMissingFields is true?

source/frontend/src/features/leases/add/AddLeaseContainer.tsx
```diff
@@ -71,6 +71,10 @@ export const AddLeaseContainer: React.FunctionComponent<
   };
 
   const handleSave = () => {
+    if (formikRef?.current?.isValid) {
+      setIsMissingFields(false);
```
devinleighsmith, line 75: this pattern works, but minor nitpick. Validation will fail for a number of reasons, invalid fields, missing fields, fields that are too long, etc. variable could be renamed to something more generic such as isFormInvalid.

source/frontend/src/features/leases/add/AddLeaseContainer.tsx
```diff
@@ -71,6 +71,10 @@ export const AddLeaseContainer: React.FunctionComponent<
   };
 
   const handleSave = () => {
+    if (formikRef?.current?.isValid) {
```
devinleighsmith, line 74: note: in situations where the form is a parent of the buttons, you should be able to use the value of the formik isValid directly instead of needing a separate state variable. In this instance you will need the state variable however since the ref will not trigger an update in the SidebarFooter component.

devinleighsmith (changes requested):
see comments.

source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
buddy326: Makes sense! However, is useEffect okay to use in a "display" component?

devinleighsmith: In general, no. a useEffect should live close to the state that it manipulates.

source/frontend/src/features/properties/map/shared/SidebarFooter.tsx
buddy326: I tried using that class but the message format and placement was a little off so instead I went with a styled component way of styling.

source/frontend/src/features/leases/add/AddLeaseContainer.tsx
```diff
@@ -70,7 +70,17 @@ export const AddLeaseContainer: React.FunctionComponent<
     }
   };
 
-  const handleSave = () => {
+  const handleSave = (error?: string | undefined) => {
```
devinleighsmith, line 73: Can you explain why passing the error in the params is necessary? You should be able to just use the errorMessage from the state above. Actually, not sure you need this check at all, setting the errorMessage to 'Please check the required fields.' when it is already set to that value should not result in a re-render.

buddy326: Your right, passing it in as a param is unnecessary so I will change that along with the check too as I do see now that if the value of errorMessage is not changing then it wont be re-rendered.

devinleighsmith (approved):

PR #3173, opened 2023-05-16
psp-6147 fix alignment of payments table

buddy326:


devinleighsmith (changes requested):
see comments. normally we use styled components and not inlined styles but for a single static width like this I think it is fine.

devinleighsmith:
Have a look here:
https://confluence.th.gov.bc.ca/display/PIMS/Best+Practices

that wiki page may help understand some of the patterns we are using.

devinleighsmith (approved):

PR #3182, opened 2023-05-22
add escapes to keycloak sync to avoid special characters being treate…

devinleighsmith:
…d as part of the URL. Also increased speed of keycloak sync by requesting all roles instead of one role at a time.

### DIFF
--- a/tools/keycloak/sync/SyncFactory.cs
+++ b/tools/keycloak/sync/SyncFactory.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
+using System.Web;
 using Microsoft.CodeAnalysis;
 using Microsoft.Extensions.Logging;
 using Pims.Core.Exceptions;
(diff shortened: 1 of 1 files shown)


tools/keycloak/sync/SyncFactory.cs
```diff
@@ -230,7 +225,7 @@ private async Task RemoveRolesFromPims(IEnumerable<KModel.RoleModel> roles, Stri
                 try
                 {
                     _logger.LogInformation($"Deleting keycloak role '{role.Name}'");
-                    await _keycloakManagementClient.DeleteAsync($"{_keycloakManagementClient.GetIntegrationEnvUri()}/roles/{role.Name}");
+                    await _keycloakManagementClient.DeleteAsync($"{_keycloakManagementClient.GetIntegrationEnvUri()}/roles/{Uri.EscapeDataString(role.Name)}");
```
devinleighsmith, line 228: the api expects % encoding on spaces and / characters, this is the only .net method that does both.

eddherrera (approved):

PR #3185, opened 2023-05-23
psp-6220 fix the order of improvements view form

buddy326:
Currently, the improvements tab under a lease is divided into 3 sections: Commercial, Other and Residential. And the sections are displayed in that order so this was to make sure that Residential comes before Other.

### DIFF
--- a/source/frontend/src/features/leases/detail/LeasePages/improvements/Improvements.tsx
+++ b/source/frontend/src/features/leases/detail/LeasePages/improvements/Improvements.tsx
@@ -1,4 +1,6 @@
+import * as API from 'constants/API';
 import { FieldArray, getIn, useFormikContext } from 'formik';
+import useLookupCodeHelpers from 'hooks/useLookupCodeHelpers';
 import { IFormLease } from 'interfaces';
 import { ILeaseImprovement } from 'interfaces/ILeaseImprovement';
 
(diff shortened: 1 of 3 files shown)


devinleighsmith:
Unfortunately this doesn't quite work. the id being used here is just the id of the improvement in the db. if the user:
Creates an other improvement
Creates a residential improvement
The system will display the output in the view form as: Other, Residential

You'll want to change your implementation to use the propertyImprovemntType which comes from the [PIMS_PROPERTY_IMPROVEMENT_TYPE]

that table includes a DISPLAY_ORDER column which will tell you the order (recommended to use that instead of hardcoding the order).

Those code values can be obtained from this logic:

  const improvementTypeCodes = getByType(API.PROPERTY_IMPROVEMENT_TYPES);

devinleighsmith:
@buddy326 I'd also like to see a unit test for this one. Let me know if you want to discuss that.

devinleighsmith (changes requested):
see comment.

source/frontend/src/features/leases/detail/LeasePages/improvements/Improvements.tsx
```diff
@@ -13,6 +15,21 @@ export const Improvements: React.FunctionComponent<IImprovementsProps> = props =
   const { values } = useFormikContext<IFormLease>();
   const improvements: ILeaseImprovement[] = getIn(values, 'improvements') ?? [];
 
+  const { getByType } = useLookupCodeHelpers();
+  const improvementTypeCodes = getByType(API.PROPERTY_IMPROVEMENT_TYPES);
+
+  improvements.sort((improvementOne: ILeaseImprovement, improvementTwo: ILeaseImprovement) => {
+    const findDisplayOrder = (x: ILeaseImprovement): number => {
+      for (let typeCode of improvementTypeCodes) {
```
devinleighsmith, line 23: this is find but the .find javascript function would provide a more concise solution:
ie:
improvementTypeCodes.find(itc => itc.id === x.propertyImprovementTypeId)

also, be careful naming a variable x. If you start to have multiple variables named like that in scope at the same time reading the code can be very confusing:
bad: x
better: li
best: leaseImprovement

Recommend using the full proper english name unless length is a concern or the logic is very simple. If length is a concern that may be code smell to refactor/simplify the logic you are working on.

devinleighsmith (approved):

PR #3191, opened 2023-05-23
Psp 6223 Hide non functional parts of the application

buddy326:


source/frontend/src/components/common/styles.ts
```diff
@@ -88,7 +88,7 @@ export const Tabs = styled(BsTabs)`
     height: 2.4rem;
   }
   .nav-item {
-    min-width: 7rem;
+    min-width: 5rem;
```
devinleighsmith, line 91: hmm, what was the rationale for this?

devinleighsmith (approved):

PR #3203, opened 2023-05-25
psp-2175 fix frontend tests console output

buddy326:
Addressing the act warnings for when the tests are run. Checked and fixed all the tests in the frontend/src/components/ folder.

### DIFF
--- a/source/frontend/src/components/common/form/NotesModal.test.tsx
+++ b/source/frontend/src/components/common/form/NotesModal.test.tsx
@@ -4,12 +4,11 @@ import MockAdapter from 'axios-mock-adapter';
 import { Formik } from 'formik';
 import { createMemoryHistory } from 'history';
 import { noop } from 'lodash';
-import React from 'react';
 
 import NotesModal, { INotesModalProps } from '@/components/common/form/NotesModal';
 import { mockLookups } from '@/mocks/lookups.mock';
 import { lookupCodesSlice } from '@/store/slices/lookupCodes';
-import { renderAsync, RenderOptions, waitFor } from '@/utils/test-utils';
+import { act, renderAsync, RenderOptions, waitFor } from '@/utils/test-utils';
 
 const history = createMemoryHistory();
 const storeState = {
@@ -51,7 +50,9 @@ describe('NotesModal component', () => {
     const {
       component: { getByTitle, findByTestId },
     } = await setup({ initialValues: { note: 'test note' } });
-    userEvent.click(getByTitle('notes'));
+    act(() => {
+      userEvent.click(getByTitle('notes'));
+    });
     const noteField = await findByTestId('note-field');
     expect(noteField).toHaveValue('test note');
   });
@@ -60,14 +61,23 @@ describe('NotesModal component', () => {
     const {
       component: { getByTitle, getByText, findByTestId },
     } = await setup({});
-    userEvent.click(getByTitle('notes'));
+    act(() => {
+      userEvent.click(getByTitle('notes'));
+    });
 
     const noteField = await findByTestId('note-field');
-    userEvent.type(noteField, 'test note');
+    act(() => {
+      userEvent.type(noteField, 'test note');
+    });
     const saveButton = getByText('Save');
-    userEvent.click(saveButton);
+    act(() => {
+      userEvent.click(saveButton);
+    });
 
-    userEvent.click(getByTitle('notes'));
+    await act(async () => {
+      userEvent.click(getByTitle('notes'));
+    });
+
     expect(noteField).toHaveValue('test note');
   });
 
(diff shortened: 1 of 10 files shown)


source/frontend/src/components/common/form/NotesModal.test.tsx
```diff
@@ -59,31 +60,51 @@ describe('NotesModal component', () => {
...
+    });
 
-    userEvent.click(getByTitle('notes'));
-    expect(noteField).toHaveValue('test note');
+    act(() => {
+      userEvent.click(getByTitle('notes'));
+    });
+    await act(async () => {
+      expect(noteField).toHaveValue('test note');
+    });
```
asanchezr, line 81: Not sure we need to **act** here. Did you try without it? Do you still get warnings?

source/frontend/src/components/common/form/ParentSelect.test.tsx
```diff
@@ -66,8 +65,10 @@ it('changes to corresponding child value on click', async () => {
     });
   });
   const childElement = getByRole('option');
-  await waitFor(() => {
-    fireEvent.click(childElement);
+  await act(async () => {
+    await waitFor(() => {
+      fireEvent.click(childElement);
+    });
```
asanchezr, line 71: I think the waitfor could be removed here - could you please try if this works?

```tsx
await act(async () => {
    fireEvent.click(childElement);
});
```

source/frontend/src/components/common/form/ProjectSelector/ProjectSelector.test.tsx
```diff
@@ -90,13 +90,13 @@ describe('ProjectSelector component', () => {
...
     await waitFor(() => expect(handleTypeaheadSearch).toHaveBeenCalled());
 
     const items = await findItems();
     expect(items).toHaveLength(1);
-    items[0].focus();
-    await act(() => userEvent.keyboard('{Enter}'));
-    expect(onChange).toHaveBeenCalled();
+    act(() => items[0].focus());
+    act(() => userEvent.keyboard('{Enter}'));
+    act(() => expect(onChange).toHaveBeenCalled());
```
asanchezr, line 100: Usually we don't need to act on "expect" statements - could you try removing the act and see if you get no warnings?

asanchezr (approved):
Approved but see my comments

source/frontend/src/components/common/form/NotesModal.test.tsx
buddy326: I tried without it and I still get warnings. But as soon as I wrap it in an act(), the warnings go away.

source/frontend/src/components/common/form/ProjectSelector/ProjectSelector.test.tsx
buddy326: When I remove it, it gives the warnings.

source/frontend/src/components/common/form/ParentSelect.test.tsx
devinleighsmith: @buddy326 did you try this?

source/frontend/src/components/common/form/ProjectSelector/ProjectSelector.test.tsx
devinleighsmith: are you referring to removing the acts on 98? 99? 100? a combination?

source/frontend/src/components/common/form/NotesModal.test.tsx
devinleighsmith: would you mind changing the act on 76 to be an await act(async and then removing the act around the expect?

source/frontend/src/components/common/form/ProjectSelector/ProjectSelector.test.tsx
buddy326: I only removed the act on line 100 which resulted in warnings.

source/frontend/src/components/common/form/ParentSelect.test.tsx
buddy326: I did and it still works so that change has been made.

source/frontend/src/components/common/form/NotesModal.test.tsx
buddy326: That worked so change has been made.

source/frontend/src/components/common/form/ProjectSelector/ProjectSelector.test.tsx
buddy326: However I did just try replacing the act on line 99 to be an await act(async () => ...) which fixed the warnings so its okay now.

PR #3210, opened 2023-05-31
PSP-6031 ACQ file: Add Owner Representative

asanchezr:
file-details

file-update


devinleighsmith (approved):
@asanchezr please see my comment on the story regarding whether or not this should allow an organization to be selected for this field

asanchezr:
@devinleighsmith see updated comments on the story. Owner Reps are **always** individuals

PR #3213, opened 2023-05-31
psp-6106 modify land act description field in takes to be select

buddy326:
The select field was moved above the area and end date fields per dev's request.

### DIFF
--- a/source/backend/entities/Partials/LandActType.cs
+++ b/source/backend/entities/Partials/LandActType.cs
@@ -1,0 +1,38 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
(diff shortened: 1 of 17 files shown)


source/backend/entities/Partials/LandActType.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string Id { get => this.LandActTypeCode; set => this.LandActTypeCode = value; }
+
+        [NotMapped]
+        public bool? IsDisabled { get; set; }
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Create a new instance of a PimsTakeStatusType class.
```
devinleighsmith, line 28: nit: copy paste - should be PimsLandActType

devinleighsmith (approved):

PR #3217, opened 2023-06-02
PSP-6023 : Add/Edit OWNER Payee to Create/Edit H120 (Compensation Requisition Input)

eddherrera:
image


eddherrera:
- Pending Unit Tests will next PR.

devinleighsmith (changes requested):
@eddherrera unfortunately it is not safe to return the financial codes as lookup codes as those are loaded once during application start and cached. Financial values can and will change during a single user session.

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -50,7 +50,15 @@ public PimsCompensationRequisition GetById(long compensationRequisitionId)
         {
             User.ThrowIfNotAuthorized(Permissions.CompensationRequisitionView);
             var entity = Context.PimsCompensationRequisitions
+                .Include(x => x.YearlyFinancial)
+                .Include(x => x.ChartOfAccounts)
+                .Include(x => x.Responsibility)
                 .Include(c => c.PimsCompReqH120s)
+                    .ThenInclude(y => y.FinancialActivityCode)
+                .Include(x => x.PimsAcquisitionPayees)
```
devinleighsmith, line 58: @FuriousLlama you already have a separate require for getting payees for a comp req right? does that mean we can take this out and better stick to our limitation of nested entities?

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -64,6 +72,8 @@ public PimsCompensationRequisition Update(PimsCompensationRequisition compensati
                 .FirstOrDefault(x => x.CompensationRequisitionId.Equals(compensationRequisition.Internal_Id)) ?? throw new KeyNotFoundException();
 
             Context.Entry(existingCompensationRequisition).CurrentValues.SetValues(compensationRequisition);
+            Context.UpdateChild<PimsCompensationRequisition, long, PimsCompReqH120, long>(p => p.PimsCompReqH120s, compensationRequisition.Internal_Id, compensationRequisition.PimsCompReqH120s.ToArray());
+            //Context.UpdateGrandchild<PimsCompensationRequisition, long, PimsAcquisitionPayee>(p => p.PimsAcquisitionPayees, ch => ch.PimsAcqPayeeCheques, compensationRequisition.Internal_Id, compensationRequisition.PimsAcquisitionPayees.ToArray());
```
devinleighsmith, line 76: likely can just remove this, don't think we need it.

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -52,6 +76,31 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
     }
   };
 
+  const onFinancialActivitiesUpdated = (values: CompensationRequisitionFormModel) => {
```
devinleighsmith, line 79: Let me check with Priyanka on this, but I believe that the ETLs are not entering data into the PIMS_PAYEE_CHEQUE, they are expecting that the system will just use the summed values from the financials. I'm partial to that as well since it reduces the risk of the calculations and the value in PIMS_PAYEE_CHEQUE gets out of sync.

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -1,39 +1,63 @@
...
+  yearlyFinancialOptions,
   onSave,
   onCancel,
 }) => {
+  const fiscalYearOptions = generateFiscalYearOptions();
   const { setModalContent, setDisplayModal } = useModalContext();
-  let fiscalYearOptions = generateFiscalYearOptions();
+  const formikRef = useRef<FormikProps<CompensationRequisitionFormModel>>(null);
+  const dummyOptions: SelectOption[] = [];
+  const [activitiesUpdated, setActivitiesUpdated] = useState<boolean>(false);
```
devinleighsmith, line 60: lets stick with our patterns and not track state within non-container components.

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+  };
+
+  return (
+    <FieldArray
+      name="financials"
+      render={arrayHelpers => {
+        return (
+          <>
+            {values.financials.map((financial, index) => (
+              <div key={`financial-act-${index}`}>
```
devinleighsmith, line 71: in general we want to avoid using the index to form the key, as per: 

You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.

Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.

Note that your components won’t receive key as a prop. It’s only used as a hint by React itself. If your component needs an ID, you have to pass it as a separate prop: <Profile key={id} userId={id} />.

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
devinleighsmith: Ok, I'm wrong, your pattern here should work fine.

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+                      field={`financials[${index}].totalAmount`}
+                      disabled
+                    />
+                  </SectionField>
+                </>
+                {index !== values.financials.length - 1 && <StyledSpacer className="my-3" />}
+              </div>
+            ))}
+            <LinkButton
+              data-testid="add=financial-activity"
```
devinleighsmith, line 147: add= ? maybe add- ?

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+              onClick={() => {
+                const activity = new FinacialActivityFormModel(null, compensationRequisitionId);
```
devinleighsmith, line 149: nit: FinacialActivityFormModel -> FinancialActivityFormModel

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+          </>
+        );
+      }}
+    ></FieldArray>
+  );
+};
+
+export default FinancialActivitiesSubForm;
+
+const getCurrenclyCleanValue = (stringValue: string): number => {
```
devinleighsmith, line 182: nit: currencly => currency

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -62,83 +111,156 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
+                    field={withNameSpace('payees.0', 'cheques[0].pretaxAmount')}
+                    formikProps={formikProps}
+                    disabled
+                  />
+                </SectionField>
+                <SectionField label="GST amount">
+                  <FastCurrencyInput
+                    field={withNameSpace('payees.0', 'cheques[0].taxAmount')}
+                    formikProps={formikProps}
+                    disabled
```
devinleighsmith, line 219: despite this being disabled in the mockups, we need to allow these to be user-overridable. This allows the user to manually work around situations where the gst constant has been changed but they still need to use the old value.

This should follow the existing patterns, where the user overriden value may be replaced whenever the user changes any values that are required for the calculation.

devinleighsmith:
gst number doesn't appear to be saving at the moment.

devinleighsmith:
get this error when removing all financial activities from a comp req:
**Reduce of empty array with no initial value**

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
eddherrera: would rather still keep it for the details view.

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
eddherrera: still needed to update the cheque data

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
eddherrera: is the same patter as above? need this to keep track of activity updates.

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
eddherrera: updated

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+                        onPretaxAmountUpdated(index, e.target.value);
+                      }}
+                      onBlurChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                        activitiesUpdated();
+                      }}
+                    />
+                  </SectionField>
+
+                  <SectionField label="GST applicable?" required>
+                    <Select
```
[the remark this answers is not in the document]
eddherrera: will need to replace later on, since I would need to make additional changes to work with onBlur.

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
eddherrera: upadted

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
eddherrera: updated

source/frontend/src/features/properties/map/compensation/update/financials/FinancialActivitiesSubForm.tsx
eddherrera: updated

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -62,83 +111,156 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
+                <SectionField label="Expropriation vesting date" labelWidth="4" contentWidth="4">
+                  <FastDatePicker field="expropriationVestingDateTime" formikProps={formikProps} />
+                </SectionField>
+                <SectionField label="Special instructions" labelWidth="12">
+                  <MediumTextArea field="specialInstruction" />
+                </SectionField>
+              </Section>
+
+              <Section header="Financial coding">
+                <SectionField label="Product" labelWidth={'4'}>
```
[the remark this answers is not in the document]
eddherrera: updated

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
eddherrera: enabled

eddherrera:
> get this error when removing all financial activities from a comp req: **Reduce of empty array with no initial value**

fixed

devinleighsmith (approved):

PR #3219, opened 2023-06-02
Added payee logic and models

FuriousLlama:


source/backend/api/Models/Concepts/CompensationRequisition/CompensationPayeeMap.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            config.NewConfig<Entity.PimsAcquisitionPayee, CompensationPayeeModel>()
+                .PreserveReference(true)
+                .Map(dest => dest.AcquisitionPayeeId, src => src.AcquisitionPayeeId)
+                .Map(dest => dest.CompensationRequisitionId, src => src.CompensationRequisitionId)
+                .Map(dest => dest.AcquisitionOwnerId, src => src.AcquisitionOwnerId)
+                .Map(dest => dest.InterestHolderId, src => src.InterestHolderId)
+                .Map(dest => dest.OwnerRepresentativeId, src => src.OwnerRepresentativeId)
+                .Map(dest => dest.OwnerSolicitorId, src => src.OwnerSolicitorId)
+                .Map(dest => dest.MotiSolicitorId, src => src.AcquisitionFilePersonId)
```
devinleighsmith, line 18: Hmm, I guess how this works is that the payee only references acquisitionfilePersonId values with type Moti_Solicitor?

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -30,18 +30,28 @@ public PimsCompensationRequisition GetById(long compensationRequisitionId)
...
-            if(compensationRequisitionId != compensationRequisition.CompensationRequisitionId)
+            _logger.LogInformation($"Updating Compensation Requisition with id ${compensationRequisition.CompensationRequisitionId}");
+
+            var currentCompensation = _compensationRequisitionRepository.GetById(compensationRequisition.CompensationRequisitionId);
+            var currentPayee = currentCompensation.PimsAcquisitionPayees.FirstOrDefault();
+            var updatedPayee = compensationRequisition.PimsAcquisitionPayees.FirstOrDefault();
+            if (currentPayee != null && updatedPayee != null)
             {
-                throw new BadRequestException("Invalid compensationRequisitionId.");
+                if (currentPayee.InterestHolderId == updatedPayee.InterestHolderId &&
```
devinleighsmith, line 45: Hey, I don't really understand what the purpose of this is, am I being thick or does this need a comment?

source/backend/dal/Helpers/Extensions/EntityExtensions.cs
```diff
@@ -135,8 +137,11 @@ public static T ThrowIfNotAllowedToEdit<T>(this T entity, string paramName, Clai
 
             foreach (var oldItem in dbItemsMap.Values)
             {
-                accessor.Remove(dbEntity, oldItem);
-                context.Remove(oldItem);
+                if (children.FirstOrDefault(x => x.Internal_Id.Equals(oldItem.Internal_Id)) == null)
```
devinleighsmith, line 140: Right, I guess we either need this or line 133 should get moved out of the if on 130.

source/frontend/src/features/mapSideBar/tabs/SectionStyles.tsx
```diff
@@ -68,6 +68,7 @@ export const ArrowDropUpIcon = styled(MdArrowDropUp)`
 
 export const StyledSummarySection = styled.div`
   background-color: ${props => props.theme.css.filterBackgroundColor};
+  padding-top: 1rem;
```
devinleighsmith, line 71: This affects ~20 pages in the application, is this something that should be on all of those forms that consume this?

source/frontend/src/features/properties/map/compensation/models.ts
```diff
@@ -53,9 +69,151 @@ export class CompensationRequisitionFormModel {
...
+
+  private constructor(api_id: number, text: string, value: string, payeeType: PayeeType) {
+    this.api_id = api_id;
+    this.text = text;
+    this.value = value;
+    this.payeeType = payeeType;
+  }
+
+  public static fromApi(payeeModels: Api_CompensationPayee[]): string {
+    if (payeeModels.length === 1) {
```
devinleighsmith, line 101: should we throw something here?

source/frontend/src/features/properties/map/compensation/models.ts
```diff
@@ -53,9 +69,151 @@ export class CompensationRequisitionFormModel {
...
+    }
+
+    const payeeOption = options.find(x => x.value === value);
+
+    if (payeeOption === undefined) {
+      return null;
+    }
+
+    const payee: Api_CompensationPayee = {
+      id: 0,
```
devinleighsmith, line 138: oh, I understand your service logic now. you aren't passing the compensation payee id up so you need to manually check if the subentity has changed.

Hmm... personally I think I'd rather do some work on the frontend to maintain the rowversion/id rather then the backend. What was your thinking for implementing this way?

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -67,76 +69,93 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
+                </SectionField>
+                <SectionField label="Cost type" labelWidth="7">
+                  {acquisitionFile.project?.costTypeCode?.code ?? ''}
+                </SectionField>
+                <SectionField label="Fiscal year" labelWidth="7" contentWidth="4" required>
+                  <Select field="fiscalYear" options={fiscalYearOptions} placeholder="Select..." />
+                </SectionField>
+              </Section>
+
+              <Section header="Payment">
```
devinleighsmith, line 129: The only problem I see with this approach is that payee is essentially untracked by formik, so non of the unsaved changes logic will work. That may result in this failing FT. 
This could be implemented using formik could it not? and then all of the normal form logic would work properly?

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -67,76 +69,93 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
             />
-            <SidebarFooter
-              onSave={() => formikProps.submitForm()}
-              isOkDisabled={formikProps.isSubmitting || !formikProps.dirty}
-              onCancel={() => cancelFunc(formikProps.resetForm, formikProps.dirty)}
-            />
+            <StyledFooter>
+              <SidebarFooter
+                onSave={() => formikProps.submitForm()}
+                isOkDisabled={formikProps.isSubmitting || !formikProps.dirty}
```
devinleighsmith, line 155: Right, like if you just change the payee option can you even save since the formik form won't be considered dirty since it won't have registered any changes?

source/backend/api/Services/CompensationRequisitionService.cs
FuriousLlama: In a nutshell, the front end might override the id of the payee or completely remove it. However, that does not necessarily mean that it was updated. The code here makes sure that an update is actually needed by adding the missing payee id if necessary

source/backend/dal/Helpers/Extensions/EntityExtensions.cs
FuriousLlama: yes, before this change the method was removing the entity even if was not intended due to the new flag

source/frontend/src/features/mapSideBar/tabs/SectionStyles.tsx
FuriousLlama: I believe so, I did a naive test of the pages and it looked correct to me.

source/frontend/src/features/properties/map/compensation/models.ts
FuriousLlama: The problem is that the combobox that holds the payee could be selected and unselected (or the other way around) causing the system to remove the id. I will put extra logic to pass the payee id, but to maintain the pattern we have that I think the backend should not trust the id coming from the frontend anyway

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
FuriousLlama: not sure what you mean. Formik-wise, it only cares about the payee id+type combination. If that changes it will detect if the form needs updating

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
FuriousLlama: it will as I mentioned in the previous comment. Let me know to jump on call to chat about this further

devinleighsmith (changes requested):
Confluence lists payee as a required field -> don't see that on this PR.

source/frontend/src/features/properties/map/compensation/update/UpdateCompensationRequisitionForm.tsx
devinleighsmith: nevermind, not sure what I was on about, you are right.

devinleighsmith:
I also get a cannot read properties of undefined error from:

image

when opening an acq file with no payees. Even though this can't happen via system input I wonder if this can happen via etl? Probably better to not throw an error in that case but just return a formik error.

source/backend/api/Models/Concepts/CompensationRequisition/CompensationPayeeMap.cs
FuriousLlama: Correct, that is validated on the front end. The backend only knows that is part of the acq team.

source/frontend/src/features/properties/map/compensation/models.ts
FuriousLlama: I've spent some time on this and I don't think is a good idea to pass the id of the payee if it's going to be validated on the backend again. Let me know if you want to chat about this.

devinleighsmith (approved):

PR #3221, opened 2023-06-03
Psp 6084 - interest holders within stakeholder tab

devinleighsmith:


source/backend/api/Areas/Acquisition/Controllers/InterestHolderController.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            var interestHolders = _acquisitionService.GetInterestHolders(id);
+            return new JsonResult(_mapper.Map<List<Api.Models.Concepts.InterestHolderModel>>(interestHolders));
+        }
+
+        /// <summary>
+        /// Update the acquisition file InterestHolders.
+        /// </summary>
+        /// <returns>The updated interest holder items.</returns>
+        [HttpPost("{id:long}/interestholders")]
```
asanchezr, line 67: Shouldn't this be a PUT operation (instead of POST) given that it's an update?

source/backend/tests/unit/api/Services/AcquisitionFileServiceTest.cs
```diff
@@ -1219,8 +1219,126 @@ public void AddCompensationsRequisitions_Success()
...
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+        }
+
+        [Fact]
+        public void GetInterestHolders_NotAuthorized_Contractor()
+        {
+            // Arrange
+            var service = CreateAcquisitionServiceWithPermissions(Permissions.AcquisitionFileView, Permissions.CompensationRequisitionView);
```
asanchezr, line 1243: nit: don't think this test needs the CompensationRequisitionView permission so would be inclined to remove from test setup

source/backend/tests/unit/api/Services/AcquisitionFileServiceTest.cs
```diff
@@ -1219,8 +1219,126 @@ public void AddCompensationsRequisitions_Success()
...
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+        }
+
+        [Fact]
+        public void GetInterestHolders_Success()
+        {
+            // Arrange
+            var service = CreateAcquisitionServiceWithPermissions(Permissions.AcquisitionFileView, Permissions.CompensationRequisitionView);
```
asanchezr, line 1263: nit: please remove unrelated Compensation claim from these tests

source/backend/tests/unit/api/Services/AcquisitionFileServiceTest.cs
```diff
@@ -1219,8 +1219,126 @@ public void AddCompensationsRequisitions_Success()
...
+
+            // Assert
+            act.Should().Throw<NotAuthorizedException>();
+        }
 
+        [Fact]
+        public void UpdateInterestHolders_NotAuthorized_Contractor()
+        {
+            // Arrange
+            var service = CreateAcquisitionServiceWithPermissions(Permissions.CompensationRequisitionAdd);
```
asanchezr, line 1302: please change this permission to AcquisitionFileEdit so the contractor path is actually tested. As it is right now, the check for AcquisitionFileEdit claim runs first and throws the exception before checking for contractor status

source/backend/tests/unit/dal/Repositories/InterestHolderRepositoryTest.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        [Fact]
+        public void Get_Empty()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission(Permissions.AcquisitionFileEdit);
+
+            var context = helper.CreatePimsContext(user, true);
+            var repository = helper.CreateRepository<InterestHolderRepository>(user);
```
asanchezr, line 38: nit: these 4 lines are repeated in every test. would be inclined to move them to the test class constructor (like the controller test pattern) to avoid boilerplate

source/frontend/src/features/properties/map/acquisition/AcquisitionView.tsx
```diff
@@ -158,6 +158,8 @@ const getEditTitle = (editFormName: EditFormType) => {
       return 'Updating Acquisition Properties';
     case EditFormType.AGREEMENTS:
       return 'Updating Agreements';
+    case EditFormType.STAKEHOLDERS:
+      return 'Updating Stakeholders';
```
asanchezr, line 162: nit: I think the pattern for update forms was "Update [XYZ]" - not "Updating [XYZ]"

I checked the uxpin but it doesn't reflect the title change when switching to edit mode so not sure which one should be. I see a mix of "Update..." and "Updating..." in this file

source/frontend/src/features/properties/map/acquisition/detail/AcquisitionFileTabs.tsx
```diff
@@ -119,6 +121,26 @@ export const AcquisitionFileTabs: React.FC<IAcquisitionFileTabsProps> = ({
     });
   }
 
+  if (acquisitionFile?.id && hasClaim(Claims.AGREEMENT_VIEW)) {
+    tabViews.push({
```
asanchezr, line 125: double-checking that AgreementView is the right claim here (seems unrelated)

source/frontend/src/features/properties/map/acquisition/update/stakeholders/InterestHolderPropertiesTable.tsx
```diff
@@ -0,0 +1,51 @@
...
+  {
+    Header: 'Identifier',
+    accessor: 'property',
+    align: 'left',
+    minWidth: 60,
+    maxWidth: 60,
+    Cell: (cellProps: CellProps<Api_PropertyFile, Api_Property | undefined>) => {
+      const propertyName = getFilePropertyName(cellProps.row.original);
+      return <OverflowTip fullText={`${propertyName.label}: ${propertyName.value}`} />;
+    },
```
asanchezr, line 26: question - did you test properties with "friendly" names; eg "Corner lot of land"?
This code seems to indicate that those properties would display as
```
[checkbox] "Descriptive Name: Corner Lot"
```

instead of 
```
[checkbox] "Corner Lot"
```

which is how it shows in UXpin

source/frontend/src/features/properties/map/acquisition/update/stakeholders/UpdateStakeHolderYupSchema.ts
```diff
@@ -0,0 +1,17 @@
...
+      legalSurveyPlanNum: Yup.string().max(
+        250,
+        'Legal survey plan must be less than 250 characters',
+      ),
+      interestTypeCode: Yup.string().required('Interest type is required').nullable(),
+      impactedProperties: Yup.array().min(1, 'At lease one impacted property is required'),
+      contact: Yup.object().required('Interest holder is required').nullable(),
+    }),
+  ),
+  nonInterestPayees: Yup.array().min(0),
```
asanchezr, line 16: I'm a bit confused about this form validation schema: 

- I couldn't find the field "legalSurveyPlanNum" in the update stakeholder form
- Not sure what making a field required and then nullable would accomplish...
  - e.g. `Yup.string().required('Interest type is required').nullable(),`
- I would assume "nonInterestPayees" would validate "impactedProperties" (in the same manner it gets validated for "interestHolders"). I couldn't find in Confluence anything saying nonInterestPayees could be added without impacted properties

source/frontend/src/features/properties/map/acquisition/update/stakeholders/models.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    const apiInterestHolders: Api_InterestHolder = {
+      ...emptyApiInterestHolder,
+      interestHolderId: 1,
+      interestHolderProperties: [
+        { ...emptyInterestHolderProperty, interestTypeCode: { id: 'PI' }, interestHolderId: 1 },
+        { ...emptyInterestHolderProperty, interestTypeCode: { id: 'IP' }, interestHolderId: 1 },
+      ],
+    };
+    const stakeholderModel = StakeHolderForm.fromApi([apiInterestHolders]);
+    console.log(stakeholderModel);
```
asanchezr, line 36: console.log

source/frontend/src/features/properties/map/acquisition/update/stakeholders/models.ts
```diff
@@ -0,0 +1,219 @@
...
+  id: number | null = null;
+  identifier: string = '';
+  groupedPropertyInterests: InterestHolderViewRow[] = [];
+
+  static fromApi(apiModel: Api_InterestHolderProperty) {
+    const interestHolderViewForm = new InterestHolderViewForm();
+    interestHolderViewForm.id = apiModel.acquisitionFileProperty?.id ?? null;
+    interestHolderViewForm.identifier = `${
+      getFilePropertyName(apiModel.acquisitionFileProperty ?? undefined).label
+    }: ${getFilePropertyName(apiModel.acquisitionFileProperty ?? undefined).value}`;
```
asanchezr, line 163: I think this also needs to be tested with friendly property names

source/frontend/src/hooks/pims-api/useApiInterestHolders.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export const useApiInterestHolders = () => {
+  const api = useAxiosApi();
+
+  return React.useMemo(
+    () => ({
+      getAcquisitionInterestHolderApi: (acqFileId: number) =>
+        api.get<Api_InterestHolder[]>(`/acquisitionfiles/${acqFileId}/interestholders`),
+      postAcquisitionholderApi: (acqFileId: number, agreements: Api_InterestHolder[]) =>
+        api.post<Api_InterestHolder[]>(
```
asanchezr, line 18: As my comment in the backend route -  think this should be a PUT instead of POST because we are always updating, not just adding interest holders

asanchezr (approved):

source/frontend/src/features/properties/map/acquisition/update/stakeholders/UpdateStakeHolderYupSchema.ts
devinleighsmith: thanks, corrected.

source/frontend/src/features/properties/map/acquisition/update/stakeholders/InterestHolderPropertiesTable.tsx
devinleighsmith: I'll check in with Bobbi on this one - was hoping for consistency with our other logic.

PR #3232, opened 2023-06-07
PSP-5952 : View H120 Details (Compensation Requisition)

eddherrera:
image

image

### DIFF
--- a/source/backend/api/Services/CompensationRequisitionService.cs
+++ b/source/backend/api/Services/CompensationRequisitionService.cs
@@ -92,5 +92,13 @@ namespace Pims.Api.Services
             _compensationRequisitionRepository.CommitTransaction();
             return fileFormToDelete;
         }
+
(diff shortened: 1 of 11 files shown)


source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -92,5 +92,13 @@ public bool DeleteCompensation(long compensationId)
             _compensationRequisitionRepository.CommitTransaction();
             return fileFormToDelete;
         }
+
+        public PimsAcquisitionPayee GetPayee(long compensationRequisitionId)
```
devinleighsmith, line 96: you and I are going to have some conflicts with payee logic, I also implemented this method but it looks different (using payeeId instead of compreqId, also uses a separate repo in my case).

devinleighsmith (approved):

PR #3233, opened 2023-06-07
IS-56.00 Database Schema

dfilteau:
PSP_PIMS S56.00 Sprint 55 2023-Jun-07
- Dropped tables:
  - PIMS_ACQ_PAYEE_CHEQUE
  - PIMS_ACQ_PAYEE_CHEQUE_HIST


devinleighsmith (changes requested):
Unfortunately this is not what we agreed on. As per the meeting notes from 6/6/2023 we need to:
is_payment_in_trust and gst_number need to be moved up from PIMS_ACQ_PAYEE_CHEQUE to PIMS_ACQUISITION_PAYEE. That will allow us to remove PIMS_ACQ_PAYEE_CHEQUE

Just to be clear, that would require:
1) create new is_payment_in_trust and gst_number fields in PIMS_ACQUISITION_PAYEE table
2) migrate existing data from PIMS_ACQ_PAYEE_CHEQUE
3) drop PIMS_ACQUISITION_PAYEE table

FuriousLlama (approved):

devinleighsmith (approved):

PR #3249, opened 2023-06-13
Re-labeled source root alias

FuriousLlama:
Imports are not separated between external (node modules), absolute (common components) , relative (i.e. feature components)
image


source/frontend/.dependency-cruiser.js
```diff
@@ -0,0 +1,437 @@
+/** @type {import('dependency-cruiser').IConfiguration} */
+module.exports = {
+  forbidden: [
+    /* rules from the 'recommended' preset: */
+    {
+      name: 'no-circular',
+      severity: 'warn',
+      comment:
+        'This dependency is part of a circular relationship. You might want to revise ' +
```
asanchezr, line 9: given that this pr alters the whole codebase, I recommend to take out this change. That way we can assess the impact of adding this new tool (dependency cruiser) in a separate pr

FuriousLlama: sure thing

asanchezr (approved):

PR #3254, opened 2023-06-15
psp-6312 correct typings of string values in report.

devinleighsmith:


source/backend/api/Models/Concepts/Lease/LeasePaymentReportModel.cs
```diff
@@ -20,15 +20,15 @@ public class LeasePaymentReportModel
         ///
         [DisplayName("L-File Number")]
         [CsvHelper.Configuration.Attributes.Name("L-File Number")]
-        public long LFileNumber { get; set; }
```
devinleighsmith, line 23: while these fields have "number" in the name they are actually alphanumeric

FuriousLlama (approved):

PR #3258, opened 2023-06-16
Psp 6404, psp-6152

devinleighsmith:
the two are related, as changes to how the cql build works is required to fix the issue with bc assessment lookups.

### DIFF
--- a/source/frontend/src/components/common/List/SelectInput.tsx
+++ b/source/frontend/src/components/common/List/SelectInput.tsx
@@ -19,6 +19,7 @@ interface ISelectInputProps<SelectTypes> {
     label: string;
     placeholder: string;
   }[];
+  autoSetting?: string;
   /** Callback to notify when the selected option in the drop-down changes */
   onSelectItemChange?: (e: React.ChangeEvent<HTMLSelectElement>) => void;
 }
--- a/source/frontend/src/components/maps/hooks/useMapSearch.tsx
+++ b/source/frontend/src/components/maps/hooks/useMapSearch.tsx
@@ -17,9 +17,9 @@ export const useMapSearch = () => {
   const {
     loadProperties: { execute: loadProperties, loading, response },
   } = useMapProperties();
-  const { parcelsLayerUrl, propertiesUrl } = useTenant();
+  const { parcelsLayerUrl, boundaryLayerUrl } = useTenant();
   const parcelsService = useLayerQuery(parcelsLayerUrl);
-  const pimsService = useLayerQuery(propertiesUrl, true);
+  const pimsBoundaryService = useLayerQuery(boundaryLayerUrl, true);
   const { setModalContent, setDisplayModal } = useModalContext();
   const keycloak = useKeycloakWrapper();
   const logout = keycloak.obj.logout;
@@ -38,7 +38,7 @@ export const useMapSearch = () => {
           lat: +filter.latitude,
           lng: +filter.longitude,
         });
-        const task2 = pimsService.findOneWhereContains(
+        const task2 = pimsBoundaryService.findOneWhereContains(
           {
             lat: +filter.latitude,
             lng: +filter.longitude,
--- a/source/frontend/src/components/maps/leaflet/Map.tsx
+++ b/source/frontend/src/components/maps/leaflet/Map.tsx
@@ -86,7 +86,6 @@ const getQueryParams = (filter: IPropertyFilter): IGeoSearchParams => {
     STREET_ADDRESS_1: filter.address,
     latitude: filter.latitude,
     longitude: filter.longitude,
-    forceExactMatch: true,
   };
 };
 
--- a/source/frontend/src/components/maps/leaflet/mapUtils.tsx
+++ b/source/frontend/src/components/maps/leaflet/mapUtils.tsx
@@ -1,4 +1,5 @@
 import { DraftCircleNumber } from 'components/propertySelector/selectedPropertyList/DraftCircleNumber';
+import { IWfsCqlFlags } from 'hooks/pims-api';
 import { IProperty } from 'interfaces';
 import L, { DivIcon, GeoJSON, LatLngExpression, Layer, Map, Marker } from 'leaflet';
 import React from 'react';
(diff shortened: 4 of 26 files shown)


source/frontend/src/components/common/List/SelectInput.tsx
```diff
@@ -19,6 +19,7 @@ interface ISelectInputProps<SelectTypes> {
     label: string;
     placeholder: string;
   }[];
+  autoSetting?: string;
```
devinleighsmith, line 22: allows setting an attribute that should prevent browsers from trying to autofill. This makes sense for fields where we are already doing autocomplete like functionality, as sometimes the browser will draw its suggestions over our suggestions.

source/frontend/src/components/maps/hooks/useMapSearch.tsx
```diff
@@ -38,7 +38,7 @@ export const useMapSearch = () => {
           lat: +filter.latitude,
           lng: +filter.longitude,
         });
-        const task2 = pimsService.findOneWhereContains(
```
devinleighsmith, line 41: Really this should be against the boundary as that is much more likely to return a result. In the future when we have a method of populating the boundary as well as the location this will be even more likely to return results (currently this will only return etl'd results).

source/frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -86,7 +86,6 @@ const getQueryParams = (filter: IPropertyFilter): IGeoSearchParams => {
     STREET_ADDRESS_1: filter.address,
     latitude: filter.latitude,
     longitude: filter.longitude,
-    forceExactMatch: true,
```
devinleighsmith, line 89: forcing an exact match for map searches is not desired as per psp-6152. For example, searching for a pid with 01 in it should return all pids that contain 01. Using exact match only a pid of 000-000-001 or a pin of 1 would match.

source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -242,17 +243,16 @@ export const asProperty = (point: PointFeature): IProperty => {
 
 /**
  * Convert any object to a cql filter string, assuming the object's keys should be used as CQL filter properties.
- * AND all object keys together within the generated cql filter string.
```
devinleighsmith, line 245: This comment was wrong, if multiple search fields are provided (aka: pid, pin) we should OR them not AND them.

source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -261,28 +261,20 @@ export const toCqlFilter = (
...
   const cql: string[] = [];
   Object.keys(object).forEach((key: string) => {
     if (object[key]) {
       if (
         ((key === 'PID' || key === 'PID_PADDED') && object[key]?.length === 9) ||
-        forceExactMatch
+        flags?.forceExactMatch
       ) {
         cql.push(`${key} = '${object[key]}'`);
-      } else if ((key === 'PID' || key === 'PID_PADDED') && object[key] && !forceSimplePid) {
```
devinleighsmith, line 277: if code needs to filter by pid and pid_padded and pin it should do that by providing those three keys in the object passed to this function not by falling into this specific case.

source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -261,28 +261,20 @@ export const toCqlFilter = (
  * NOTE: The resulting string value is left as-is. It will not be URL-encoded
  * @param object an object to convert to a cql filter string.
  */
-export const toCqlFilterValue = (
-  object: Record<string, string>,
-  forceSimplePid?: boolean,
-  forceExactMatch?: boolean,
-) => {
+export const toCqlFilterValue = (object: Record<string, string>, flags?: IWfsCqlFlags) => {
```
devinleighsmith, line 264: standardized the cql flags in an interface that limits the risk of flags being sent in the wrong order.

source/frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -261,28 +261,20 @@ export const toCqlFilter = (
...
-          `PIN ilike '%${object[key]}%' OR PID ilike '%${object[key]}%' OR PID_PADDED ilike '%${object[key]}%'`,
-        );
       } else {
         cql.push(`${key} ilike '%${object[key]}%'`);
       }
     }
   });
 
-  return cql.length ? (forceExactMatch ? cql.join(' OR ') : cql.join(' AND ')) : '';
+  return cql.length ? (flags?.useCqlOr ? cql.join(' OR ') : cql.join(' AND ')) : '';
```
devinleighsmith, line 279: allow the or or and of cql concatenation to be determined independently of the forceExactMatch which is also used to control whether ilike or = is used.

source/frontend/src/features/properties/map/hooks/useMapProperties.tsx
```diff
@@ -15,11 +15,11 @@ export const useMapProperties = () => {
     (params?: IGeoSearchParams) => {
       const geoserver_params = {
         STREET_ADDRESS_1: params?.STREET_ADDRESS_1,
-        PID: params?.PID,
+        PID_PADDED: params?.PID,
```
devinleighsmith, line 18: according to psp-6152 searching for 01 should match 236-626-010 as well as 010-463-251. The only way for that to be possible is to search against the padded pid instead of the pid (which is a number).

source/frontend/src/hooks/pims-api/useGeoServer.ts
```diff
@@ -42,17 +42,5 @@ export const useGeoServer = () => {
     onError: useAxiosErrorHandler('Failed to retrieve property information from BC Data Warehouse'),
   });
 
-  const getPropertyWithPidWfs = useCallback(
```
devinleighsmith, line 45: this is an unused function.

source/frontend/src/hooks/useBcAssessmentLayer.ts
```diff
@@ -100,6 +100,7 @@ export const useBcAssessmentLayer = (
       }
       let folioId = legalDescriptionResponse?.features[0]?.properties?.FOLIO_ID;
       let rollNumber = legalDescriptionResponse?.features[0]?.properties?.ROLL_NUMBER;
+      let jurisdictionCode = legalDescriptionResponse?.features[0]?.properties?.JURISDICTION_CODE;
```
devinleighsmith, line 103: this is the fix for psp-6404, which combines ensuring the CQL is being generated with an AND instead of an OR and also including a restriction by jurisidiction when looking for matches.

source/frontend/src/components/maps/leaflet/mapUtils.tsx
asanchezr: Thanks for fixing this logic. I never liked this built-in edge case

asanchezr (approved):

PR #3271, opened 2023-06-20
Techdebt/refactor map split

FuriousLlama:


source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -0,0 +1,250 @@
+import axios from 'axios';
```
devinleighsmith, line 1: ok, so looks like you are trying to split map specific functionality into this file -> map vs sidebar and layout functionality to mapview, and then cleaning up the container to not contain logic better suited to being in these two files.

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/hooks/usePropertyDetails.ts
```diff
@@ -13,9 +13,12 @@ import {
 
 export function usePropertyDetails(property?: Api_Property): IPropertyDetailsForm | undefined {
   const isMounted = useIsMounted();
-  const electoralService = useAdminBoundaryMapLayer();
-  const legalAdminService = useLegalAdminBoundariesMapLayer();
-  const firstNationsService = useIndianReserveBandMapLayer();
+  const electoralService_ = useAdminBoundaryMapLayer();
```
devinleighsmith, line 16: nit: could make this a single line right? this electoralservice line isn't used elsewhere.

source/frontend/src/components/maps/MapLeafletView.tsx
FuriousLlama: Thats correct. The map page (homepage) should as a container for the layout with the different components. The map itself is just a wrapper for the leaflet with the buttons.

source/frontend/src/components/common/List/SelectInput.test.tsx
```diff
@@ -9,9 +9,7 @@ import { SelectInput } from './SelectInput';
 jest.mock('formik');
 
 (useFormikContext as jest.Mock).mockReturnValue({
-  values: {
-    includeAllProperties: false,
```
FuriousLlama, line 13: includeAllProperties was not being used at all

source/frontend/src/features/mapSideBar/hooks/usePropertyDetails.ts
FuriousLlama: yes it could. I will revisit this again to make sure the memoing is actually working as I expected

PR #3272, opened 2023-06-20
psp-6227: fix leases filtering for tenant names

buddy326:


source/backend/dal/Helpers/Extensions/LeaseExtensions.cs
```diff
@@ -152,8 +152,8 @@ private static IQueryable<Entities.PimsLease> GenerateCommonLeaseQuery(this IQue
 
             if (!string.IsNullOrWhiteSpace(filter.TenantName))
             {
-                query = query.Where(l => l.PimsLeaseTenants.Any(tenant => tenant.Person != null && EF.Functions.Like(tenant.Person.Surname + ", " + tenant.Person.FirstName + ", " + tenant.Person.MiddleNames, $"%{filter.TenantName}%"))
-                || l.PimsLeaseTenants.Any(tenant => tenant.Organization != null && EF.Functions.Like(tenant.Organization.OrganizationName, $"%{filter.TenantName}%")));
+                query = query.Where(l => l.PimsLeaseTenants.Any(tenant => tenant.Person != null && EF.Functions.Like(tenant.Person.FirstName + " " + tenant.Person.MiddleNames + " " + tenant.Person.Surname, $"%{filter.TenantName}%"))
```
devinleighsmith, line 155: The original code made this mistake as well, so I see why this is still here, but we'll need to filter out any of the three name components if they are empty, otherwise the only way to search for Devin Smith is by typing Devin(two spaces)Smith

my recommendation would be to put each name component in an array, filter out the empty array elements, and then use string.join to concatenate them.

devinleighsmith (approved):

PR #3273, opened 2023-06-20
psp-6222 make canada as default selection for address component

buddy326:


source/frontend/src/features/contacts/contact/create/components/address/Address.tsx
```diff
@@ -46,6 +46,13 @@ export const Address: React.FunctionComponent<React.PropsWithChildren<IAddressPr
     setSelectedCountryId(countryId);
   }, [countryId, namespace, setFieldValue, setSelectedCountryId]);
 
+  useEffect(() => {
+    if (!countryId) {
+      setFieldValue(withNameSpace(namespace, 'countryId'), '1');
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
devinleighsmith, line 54: @buddy326 is there any reason why the dependencies aren't being set here?

buddy326: @devinleighsmith Because we want Canada to be selected only on the first render, and if we had countryId as a dependency then every time it would get changed, the hook would run and set it back to Canada so thats why I did not have any, but I may be mistaking how useEffect works so my assumptions could be wrong.

devinleighsmith: @buddy326 every time your hook would run would it not see that the countryId had been set and therefore your if statement would prevent it from being set again?

buddy326: @devinleighsmith shoot I forgot about the if statement being there so your right I'll include the dependencies.

asanchezr (approved):

devinleighsmith (approved):

devinleighsmith (approved):

PR #3274, opened 2023-06-20
PSP-6422 : Not possible to save compensation requisition that has a p…

eddherrera:
…ayee with no payee_cheque

### DIFF
--- a/source/backend/api/Services/CompensationRequisitionService.cs
+++ b/source/backend/api/Services/CompensationRequisitionService.cs
@@ -99,9 +99,14 @@ namespace Pims.Api.Services
                 }
             }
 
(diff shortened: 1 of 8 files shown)


source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -99,10 +99,15 @@ public PimsCompensationRequisition Update(PimsCompensationRequisition compensati
                 }
             }
 
-            if (payeeCheque is not null)
+            if (payeeCheque is not null && payeeCheque.AcqPayeeChequeId > 0)
```
devinleighsmith, line 102: Should this not be >=0?

PR #3281, opened 2023-06-21
PSP-6286 : Restrict changing final compensation requisition status back to draft

eddherrera:


source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using DocumentFormat.OpenXml.Office2010.Excel;
```
devinleighsmith, line 5: this can likely be removed.

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailView.tsx
```diff
@@ -120,18 +119,28 @@ export const CompensationRequisitionDetailView: React.FunctionComponent<
...
       <RightFlexDiv>
-        {setEditMode !== undefined && hasClaim(Claims.COMPENSATION_REQUISITION_EDIT) && (
-          <EditButton
-            title="Edit compensation requisition"
-            onClick={() => {
-              setEditMode(true);
-            }}
-          />
-        )}
+        {setEditMode !== undefined &&
```
devinleighsmith, line 134: could we have this refactored so that there is a single block that conditionally renders the edit button? we should never have a case where multiple edit buttons are returned, so I'd rather the render logic more clearly reflected that.

devinleighsmith: maybe turn these if statements into named functions, so we can do something like: 
if(isDraftAndCanEdit || isNotDraftAndAdmin) or similar

devinleighsmith (approved):

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -128,5 +125,15 @@ private void AddNoteIfStatusChanged(long compensationRequisitionId, long acquisi
...
+
+        private void CheckDraftStatusUpdateAuthorized((bool? currentStatus, bool? newStatus) statusComparable)
+        {
+            if (statusComparable.currentStatus.HasValue && statusComparable.currentStatus.Value.Equals(false)
+                && ((statusComparable.newStatus.HasValue && statusComparable.newStatus.Value.Equals(true)) || !statusComparable.newStatus.HasValue)
+                && !_user.HasPermission(Permissions.AdminUsers))
+            {
+                throw new NotAuthorizedException();
+            }
+        }
```
asanchezr, line 137: I see here that we are checking for this permission: **AdminUsers** - which according to the code comments "Can administer user accounts." 

This is not the same as an ADMIN user which would have the `Permissions.SystemAdmin` claim - wouldn't it make sense to use SystemAdmin here instead?

asanchezr (approved):

source/backend/api/Services/CompensationRequisitionService.cs
eddherrera: that naming convention for claims is certainly ambiguous. If that's the one needed I'll change it.

asanchezr: Double check with @devinleighsmith but I believe SystemAdmin is the correct one...

source/backend/api/Services/CompensationRequisitionService.cs
eddherrera: removed

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailView.tsx
eddherrera: updated to use bool variables.

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailView.tsx
```diff
@@ -144,29 +153,41 @@ export const CompensationRequisitionDetailView: React.FunctionComponent<
...
+            >
               {formatMoney(payeeDetails?.totalAmount ?? 0)}
             </HeaderField>
           </Col>
         </StyledRow>
       </Section>
 
       <Section header="Requisition Details">
-        <SectionField label="Status" labelWidth="4">
+        <SectionField label="Status" labelWidth="4" data-testid={'compensation-status'}>
```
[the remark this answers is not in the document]
eddherrera: updated

source/backend/api/Services/CompensationRequisitionService.cs
eddherrera: updated

PR #3288, opened 2023-06-23
PSP-5951 refactor left hand file navigation - Acquisition ONLY

asanchezr:
Left hand sidebar navigation refactored to use routing instead of internal state. Only applies to acquisition file forms (for now)

### File routing
image

### File Property routing
image


devinleighsmith:
something in this change has appeared to cause a regression in the tab view: 
image

source/frontend/src/features/mapSideBar/acquisition/FilePropertyRouter.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const getAcquisitionFileProperty = (
+  acquisitionFile: Api_AcquisitionFile,
+  selectedMenuIndex: number,
+) => {
+  const properties = acquisitionFile?.fileProperties || [];
+  const selectedPropertyIndex = selectedMenuIndex - 1;
+  const acquisitionFileProperty = properties[selectedPropertyIndex];
+  if (!!acquisitionFileProperty.file) {
```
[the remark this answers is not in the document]
asanchezr: I see. Didn't consider that edge case. Will update the error message.

devinleighsmith:
our default handling of when a file fails to load is not very good. Predates this change so I'll log tech debt separately.

source/frontend/src/features/mapSideBar/acquisition/AcquisitionRouter.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export const AcquisitionRouter: React.FC<IAcquisitionRouterProps> = props => {
+  const { path, url } = useRouteMatch();
+
+  if (props.acquisitionFile === undefined || props.acquisitionFile === null) {
+    return null;
+  }
+
+  // render edit forms
+  if (props.isEditing) {
```
devinleighsmith, line 40: what should we do if we are editing and our path doesn't match one of these? currently we display an empty edit screen.

source/frontend/src/features/mapSideBar/acquisition/FilePropertyRouter.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const { path, url } = useRouteMatch();
+
+  if (props.acquisitionFile === undefined || props.acquisitionFile === null) {
+    return null;
+  }
+
+  const fileProperty = getAcquisitionFileProperty(props.acquisitionFile, props.selectedMenuIndex);
+
+  // render edit forms
+  if (props.isEditing) {
```
devinleighsmith, line 38: same question here.

source/frontend/src/features/mapSideBar/acquisition/AcquisitionRouter.tsx
asanchezr: The old code in ViewSelector throws an error - `throw Error('Active edit form not defined');`
I can create a default route that does the same

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/acquisition/FilePropertyRouter.tsx
asanchezr: I can throw an error like the old code used to do or redirect to file/property details if route doesn't match. Not sure what's better

source/frontend/src/features/mapSideBar/acquisition/FilePropertyRouter.tsx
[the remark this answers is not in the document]
asanchezr: Fixed. We show a toast now and also redirect to File Details screen when accessing a non-existing property index.

source/frontend/src/features/mapSideBar/acquisition/AcquisitionRouter.tsx
asanchezr: Fixed. We redirect to the default File tab (file details)

source/frontend/src/features/mapSideBar/acquisition/FilePropertyRouter.tsx
asanchezr: Fixed. We redirect to the default property tab (property details)

PR #3294, opened 2023-06-27
psp-6246 increase keycloak sync resiliancy via polly, also update dep…

devinleighsmith:
…recated calls from keycloak api.

### DIFF
--- a/tools/keycloak/sync/SyncFactory.cs
+++ b/tools/keycloak/sync/SyncFactory.cs
@@ -5,7 +5,6 @@ using System.Net;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
-using System.Web;
 using Microsoft.CodeAnalysis;
 using Microsoft.Extensions.Logging;
 using Pims.Core.Exceptions;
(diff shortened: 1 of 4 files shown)


tools/keycloak/sync/SyncFactory.cs
```diff
@@ -384,24 +383,34 @@ private async Task SyncUsersAsync(StringBuilder log, StringBuilder errorLog)
...
+                }
+                LogInfo(log, $"Executed operation 'add' on roles '{string.Join(',', userRolesToAdd.Select(r => r.Name))}' to user '{username}'");
+            }
+
+            foreach (RoleModel userRoleToRemove in userRolesToRemove)
+            {
+                _logger.LogInformation($"Executing operation 'delete' on role '{userRoleToRemove.Name}' to user '{username}'");
 
-                var response = await _keycloakManagementClient.SendJsonAsync($"{_keycloakManagementClient.GetIntegrationEnvUri()}/user-role-mappings", HttpMethod.Post, operation);
+                var response = await _keycloakManagementClient.SendAsync($"{_keycloakManagementClient.GetIntegrationEnvUri()}/users/{Uri.EscapeDataString(username)}/roles/{Uri.EscapeDataString(userRoleToRemove.Name)}", HttpMethod.Delete);
```
asanchezr, line 408: nit: there seems to be a specialized "DeleteAsync" method - wouldn't it make sense to use that over the generic SendAsync?

asanchezr (approved):

PR #3304, opened 2023-07-05
psp-6428 fix layers colors

buddy326:


asanchezr (approved):

source/frontend/src/components/maps/leaflet/Control/LayersControl/data.ts
```diff
@@ -101,7 +101,7 @@ export const layersTree: ILayerItem[] = [
         format: 'image/png',
         zIndex: 10,
         id: 'firstNationsReserves',
-        color: '#ebe0d7',
+        color: '#DCBDAD',
         opacity: 0.9,
```
asanchezr, line 105: would be nice to see a screenshot of the new colors in the layers menu. We usually attach screenshots to PR that involve user interface changes

buddy326: Screenshot (9)
Screenshot (10)
Screenshot (11)
Screenshot (12)

PR #3310, opened 2023-07-06
HOTFIX: psp-6495 Correct issue with object matching not working as expected.

devinleighsmith:


asanchezr (approved):
Question: are we increasing the HOTFIX # with each PR or will that be done separately?

devinleighsmith:
> Question: are we increasing the HOTFIX # with each PR or will that be done separately?

no, I think we will increment once per actual release to uat, so before the meeting on wednesday we will need an increment.

PR #3314, opened 2023-07-06
PSP-6324 : Select multiple contacts (Owners) for generated Owner letter

eddherrera:


source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,155 @@ export interface IGenerateFormContainerProps {
...
+          intHolders.map(async holder => {
+            if (holder.personId) {
+              const person = !!holder?.personId
+                ? (await getPersonConcept(holder?.personId))?.data
+                : null;
+
+              if (person) {
+                generateRecipientsList.push(
+                  new LetterRecipientModel(
+                    i++,
```
devinleighsmith, line 77: If I'm understanding what this is I'm not a big fan, as this seems like a way of trying to create a unique key for each of these different entities. I think I'd rather have a function that could return a unique key from a LetterRecipientModel that uses the fields of that object to build a unique key. Maybe I'm not understanding something however.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,155 @@ export interface IGenerateFormContainerProps {
...
+                    'HLDR',
+                    Api_GenerateOwner.fromApiPerson(person),
+                    holder,
+                  ),
+                );
+              }
+            } else if (holder.organizationId) {
+              const org = !!holder?.organizationId
```
devinleighsmith, line 85: this seems redundant, with the if above there is no way this can't be set.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,155 @@ export interface IGenerateFormContainerProps {
...
+    if (file.acquisitionFileOwnerSolicitors && file.acquisitionFileOwnerSolicitors.length > 0) {
+      await Promise.all(
+        file.acquisitionFileOwnerSolicitors.map(async solicitor => {
+          if (solicitor.personId) {
+            const person = !!solicitor?.personId
+              ? (await getPersonConcept(solicitor?.personId))?.data
+              : null;
+
+            if (person) {
+              const personSolicitorModel = Api_GenerateOwner.fromApiPerson(person);
```
devinleighsmith, line 115: seems like maybe this and the org version of this function could be extracted into a common function?

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: updated

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: updated

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: I can't see any other scenario where this is used outside of the GenerateLetter context, and I think it's rather handy to have it next to the model.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.test.tsx
```diff
@@ -89,23 +89,32 @@ describe('GenerateFormContainer component', () => {
...
       expect(generateLetterFn).toHaveBeenCalledTimes(0);
-      expect(generateH0443Fn).toHaveBeenCalledTimes(1);
+      expect(generateH0443Fn).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  it('generates the documment letter on confirmation', async () => {
+    await act(async () => viewProps.onGenerateLetterOk([]));
+    await waitFor(async () => {
+      // expect(handleGenerateLetterOk).toHaveBeenCalledTimes(1);
```
asanchezr, line 115: please remove dead code

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,143 @@ export interface IGenerateFormContainerProps {
 const GenerateFormContainer: React.FunctionComponent<
   React.PropsWithChildren<IGenerateFormContainerProps>
 > = ({ acquisitionFileId, View }) => {
-  const generateLetter = useGenerateLetter();
+  const [fullRecipientsList, setFullRecipientsList] = useState<LetterRecipientModel[]>([]);
+  const { getPersonConcept, getOrganization } = useApiContacts();
```
asanchezr, line 26: In my current PR I added a "GetOrganizationConcept" that gets the Api_Organization concept (instead of the IEditableOrganization that getOrganization returns) 

getOrganization is old code from when each form had unique models. Now we have moved towards sharing "concept" models.

Ideally we should be able to sync my changes into your PR

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,143 @@ export interface IGenerateFormContainerProps {
...
+      return;
+    }
 
-  const generateH120 = useGenerateH120();
+    const fileOwnersFetchCall = retrieveAcquisitionFileOwners(acquisitionFileId);
+    const interestHoldersFetchCall = fetchInterestHolders(acquisitionFileId);
+
+    // Add owners and interest holders
+    await Promise.all([fileOwnersFetchCall, interestHoldersFetchCall]).then(
+      ([fileOwners, intHolders]) => {
```
asanchezr, line 57: This works as-is but could be rewritten to drop the promise.then()

Something like
```tsx
const [fileOwners, intHolders] = await Promise.all([...]);
```

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
asanchezr: Owner solicitor is also used in the Generate Expropriation Forms. I talked to Dev and we agreed to keep using GeneratePerson for owner solicitor in those forms but there's a JIRA to use organizations as solicitors as well so this code can be reused

I think we would need to create a "GenerateContact" that can be either a GeneratePerson or a GenerateOrganization

No need to change this PR but just an FYI that we will need to make this "person or org" code more shareable

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +22,143 @@ export interface IGenerateFormContainerProps {
...
+      file.acquisitionFileOwnerRepresentatives &&
+      file.acquisitionFileOwnerRepresentatives.length > 0
+    ) {
+      await Promise.all(
+        file.acquisitionFileOwnerRepresentatives.map(async rep => {
+          if (rep.personId) {
+            const person = (await getPersonConcept(rep?.personId))?.data;
+            if (person) {
+              const personSolicitorModel = Api_GenerateOwner.fromApiPerson(person);
+              generateRecipientsList.push(
```
asanchezr, line 134: My main comment here is why are we using GenerateOwner instances here for owner solicitors and owner representatives. We do have GeneratePerson - why not use that instead??

source/frontend/src/interfaces/editable-contact/index.ts
```diff
@@ -46,8 +46,10 @@ export interface IBaseAddress {
   regionId?: number;
   districtId?: number;
   provinceId?: number;
+  province?: string;
   countryId?: number;
   countryOther?: string;
+  country?: string;
   postal?: string;
```
asanchezr, line 53: I guess this is okay but this is an example of why we would benefit from using "concept" models instead. 
Did you check that this change doesn't break current functionality in Contacts? (as those models are used only there?)

source/frontend/src/models/generate/GenerateAddress.ts
```diff
@@ -29,4 +30,29 @@ export class Api_GenerateAddress {
...
   }
+
+  public static fromIEditableOrgAddress(model: IEditableOrganizationAddress): Api_GenerateAddress {
+    let add = new Api_GenerateAddress(null);
+    add.line_1 = model.streetAddress1 ?? '';
+    add.line_2 = model.streetAddress2 ?? '';
+    add.line_3 = model.streetAddress3 ?? '';
+    add.city = model.municipality ?? '';
+    add.province = model.province ?? '';
+    add.postal = model.postal ?? '';
```
asanchezr, line 41: this wouldn't be needed if we used Api_Organization concept instead of the IEditableOrganization...

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: The array passed to the template are not 'Person/Contacts' but GenerateOwners.

asanchezr: I see. We could leave as-is but there's nothing stopping us from altering the template to take a list of owners, and a list of additional "contacts" and then display them one after another...
Like in Expropriation forms we have many different "groups" of people that receive the letter and they are not all made into owners to fit one big list. See below. 

image

Not saying you must do it this way but it would be cleaner to pass in the extra people as contacts than "force" then into GenerateOwner instances when they don't fit naturally, but again a matter of dev preference

devinleighsmith:
@eddherrera fyi you have some additional comments to address on this PR before it can be merged, as well as some conflicts.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: Made changes to use the new InterestHolder schema.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.test.tsx
eddherrera: removed

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: Updated with GetOrganizationConcept method.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: Updated.

source/frontend/src/interfaces/editable-contact/index.ts
eddherrera: removed

source/frontend/src/models/generate/GenerateAddress.ts
eddherrera: removed

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: It turned out to be much simpler with the new interest holder logic. now just mapping an interest holder to a generateOwner ( instead of mapping an Owner to a Contact ). didn't wanted to modify the template because at pointing we decided not to modify the template and I rather keep the template simple since it's managed by users.

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +23,102 @@ export interface IGenerateFormContainerProps {
...
+    getAcquisitionFile: { execute: getAcquisitionFile },
+  } = useAcquisitionProvider();
+  const [isGenerateLetterModalOpened, openGenerateLetterModal, closeGenerateLetterModal] =
+    useModalManagement();
+  const {
+    getAcquisitionOwners: {
+      execute: retrieveAcquisitionFileOwners,
+      loading: loadingAcquisitionFileOwners,
+    },
+  } = useAcquisitionProvider();
```
asanchezr, line 38: nit: you already have a previous call to **useAcquisitionProvider()** - please move the getAcquisitionOwners destructuring to the first call and remove this extra call to the hook

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
```diff
@@ -14,16 +23,102 @@ export interface IGenerateFormContainerProps {
...
+                ),
+              );
+            }
+          }
+        }),
+      );
+    }
+
+    setFullRecipientsList(generateRecipientsList);
+    console.table(generateRecipientsList);
```
asanchezr, line 103: please remove **console.table()**

source/frontend/src/models/generate/GenerateOwner.ts
```diff
@@ -37,4 +42,43 @@ export class Api_GenerateOwner {
...
+
+    generateOwner.given_name = '';
+    generateOwner.last_name_or_corp_name = model.name ?? '';
+    generateOwner.other_name = model.alias ?? '';
+    generateOwner.incorporation_number = model.incorporationNumber ?? '';
+    generateOwner.registration_number = '';
+
+    const mailingAddress = model.organizationAddresses?.find(
+      addr => addr?.addressUsageType?.id === AddressTypes.Mailing,
+    );
```
asanchezr, line 77: nit: I believe **getApiPersonOrOrgMailingAddress()** already supports either a person or an org so why not use that method instead of repeating that logic here?

asanchezr (approved):

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: removed

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/GenerateFormContainer.tsx
eddherrera: updated

PR #3315, opened 2023-07-07
HOTFIX: h0074 standardize agreement owner string other name.

devinleighsmith:


source/frontend/src/models/generate/GenerateOwner.ts
```diff
@@ -31,6 +33,8 @@ export class Api_GenerateOwner {
     this.is_primary_contact = this.is_primary_contact = owner?.isPrimaryContact ?? false;
     this.owner_string = this.is_corporation
```
devinleighsmith, line 34: FYI apparently all instances of this pattern can have the other name in brackets, thus the non-passive change.

eddherrera (approved):

PR #3318, opened 2023-07-09
fix null vs undefined deposit logic.

devinleighsmith:


source/frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
```diff
@@ -40,15 +40,15 @@ export class DepositListEntry {
 }
 
 function renderHolder({ row: { original } }: CellProps<DepositListEntry, Api_Contact | undefined>) {
-  if (original.contactHolder !== undefined) {
+  if (!!original.contactHolder) {
```
FuriousLlama, line 43: nit rather have ``` ... && !==null```

asanchezr:
@devinleighsmith are you sure the PR should come from your **test** branch? Could you re-create it from **dev** instead? Don't want to mess with git history...
image

PR #3320, opened 2023-07-12
Map Finite State Machine refactor

FuriousLlama:


devinleighsmith:
added properties (via research file) does not show the new properties once the sidebar is closed. They show up with a hard refresh

devinleighsmith:
Clicking on a map marker no longer opens the sidebar (even if the sidebar is not currently open), seems like the user has to click the slide out "view more details" to make that work now. That would be a functional change from our current logic.

devinleighsmith:
Showing the active properties works for research and acq files but not leases.

devinleighsmith:
when searching, clicking on the map will show the popup but will not highlight the underlying parcel.

devinleighsmith:
the address search doesn't seem to work anymore, it just returns all properties.

devinleighsmith:
@FuriousLlama how would you like to handle the above? we'll have to fix 1,3, and 5. We could possibly defer 2 and 4 with follow up stories.

FuriousLlama:
> @FuriousLlama how would you like to handle the above? we'll have to fix 1,3, and 5. We could possibly defer 2 and 4 with follow up stories.

Good testing! Will fix it along with the merges from master

source/frontend/package.json
```diff
@@ -7,6 +7,7 @@
     "@react-keycloak/web": "3.4.0",
     "@reduxjs/toolkit": "1.8.6",
     "@types/polylabel": "1.0.5",
+    "@xstate/react": "^3.2.2",
```
devinleighsmith, line 10: @FuriousLlama our best practice for these is to lock the version to avoid inadvertent version bumps affecting application functionality.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+  if (context === undefined) {
+    throw new Error('useMapStateMachine must be used within a MapStateMachineContextProvider');
+  }
+  return context;
+}
+
+export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>> = ({
+  children,
+}) => {
+  const locationLoader = useLoactionFeatureLoader();
```
devinleighsmith, line 67: nit: loaction

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+    },
+    services: {
+      loadLocationData: (context, event: any) => {
+        let latLng: LatLngLiteral = { lat: 0, lng: 0 };
+        if (event.type === 'MAP_CLICK') {
+          latLng = event.latlng;
+        } else if (event.type === 'MAP_MARKER_CLICK') {
+          latLng = event.featureSelected.latlng;
+        }
+        const result = locationLoader.showLocationDetails(latLng, context.isSelecting);
```
devinleighsmith, line 93: should else throw an exception instead of showing the location details at 0,0?

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+        return result;
+      },
+      loadFeatures: (context: any, event: any) => {
+        const geoFilter = getQueryParams(context.mapFilter);
+        if (geoFilter.latitude !== undefined && geoFilter.longitude) {
```
devinleighsmith, line 99: this if seems weird, why explicitly check lat for not undefined but then use a truthy check on long?

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+    >
+      {children}
+    </MapStateMachineContext.Provider>
+  );
+};
+
+const getQueryParams = (filter: IPropertyFilter): IGeoSearchParams => {
+  // The map will search for either identifier.
+  const pinOrPidValue = filter.pinOrPid ? filter.pinOrPid?.replace(/-/g, '') : undefined;
+  debugger;
```
devinleighsmith, line 266: debugger here.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+        }
+        const result = locationLoader.showLocationDetails(latLng, context.isSelecting);
+
+        return result;
+      },
+      loadFeatures: (context: any, event: any) => {
+        const geoFilter = getQueryParams(context.mapFilter);
+        if (geoFilter.latitude !== undefined && geoFilter.longitude) {
+          return mapSearch.searchOneLocation(
+            Number(geoFilter.latitude),
```
devinleighsmith, line 101: number() returns nan if parsing fails, not sure how that would present here.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -0,0 +1,275 @@
...
+      }}
+    >
+      {children}
+    </MapStateMachineContext.Provider>
+  );
+};
+
+const getQueryParams = (filter: IPropertyFilter): IGeoSearchParams => {
+  // The map will search for either identifier.
+  const pinOrPidValue = filter.pinOrPid ? filter.pinOrPid?.replace(/-/g, '') : undefined;
```
devinleighsmith, line 265: this looks like a regression on the previously implemented request to accept spacse as well as hyphens as pid delimiters.

source/frontend/src/features/leases/detail/LeasePages/details/PropertiesInformation.tsx
```diff
@@ -21,13 +22,38 @@ export const PropertiesInformation: React.FunctionComponent<
   React.PropsWithChildren<IPropertiesInformationProps>
 > = ({ nameSpace, disabled, hideAddress }) => {
   const { values } = useFormikContext<LeaseFormModel>();
-  const properties: IProperty[] = getIn(values, withNameSpace(nameSpace, 'properties')) ?? [];
+
+  const properties: FormLeaseProperty[] = React.useMemo(() => {
+    return getIn(values, withNameSpace(nameSpace, 'properties')) ?? [];
+  }, [values, nameSpace]);
+  console.log(properties);
```
devinleighsmith, line 29: console.log

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -0,0 +1,299 @@
...
+            requestedFlyTo: (_, event: any) => ({
+              bounds: event.bounds,
+              location: null,
+            }),
+          }),
+          target: 'pendingFlyTo',
+        },
+        REQUEST_FIT_BOUNDS: {
+          actions: assign((context: any) => {
+            if (context.draftLocations.length > 0) {
```
devinleighsmith, line 68: I could be missing something but it seems like "draft" properties is being used for both properties to highlight when selecting and properties to highlight when viewing a file. I'd argue for at least a rename, draft to me implies these properties have not been saved.

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -0,0 +1,299 @@
...
+        src: 'loadLocationData',
+        onDone: {
+          target: 'idle',
+          actions: [
+            assign({
+              isLoading: () => false,
+              showPopup: () => true,
+              mapLocationFeatureDataset: (context, event: any) => event.data,
+            }),
+            // 'navigateToProperty', // TODO:uncomment this to show property information sidebar
```
devinleighsmith, line 153: I see, lets make sure we do something with this otherwise this will just come back as a bug. Do we want to ask the BAs?

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -0,0 +1,299 @@
...
+          actions: assign({
+            sideBarType: (_, event: any) => event.sidebarType,
+          }),
+          target: 'sidebarOpen',
+        },
+      },
+    },
+    sidebarOpen: {
+      entry: assign({
+        sideBarType: (_, event: any) => event.sidebarType,
```
devinleighsmith, line 186: why is this necessary?

source/frontend/src/components/common/mapFSM/machineDefinition/types.ts
```diff
@@ -0,0 +1,59 @@
...
+  mapFilter: IPropertyFilter | null;
+
+  isLoading: boolean;
+  sideBarType: SideBarType;
+  requestedFitBounds: LatLngBounds;
+  requestedFlyTo: RequestedFlyTo;
+  draftLocations: LatLngLiteral[];
+};
+
+// Possible state machine states
```
devinleighsmith, line 38: background on this?

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
```diff
@@ -0,0 +1,101 @@
...
+        districtFeature: null,
+        municipalityFeature: null,
+      };
+      try {
+        // call these APIs in parallel - notice there is no "await"
+        const fullyAttributedTask = fullyAttributedServiceFindOne(latLng);
+        const regionTask = adminBoundaryLayerServiceFindRegion(latLng, 'GEOMETRY');
+        const districtTask = adminBoundaryLayerServiceFindDistrict(latLng, 'GEOMETRY');
+
+        const parcelFeature = await fullyAttributedTask;
```
devinleighsmith, line 55: Have a look at:
https://stackoverflow.com/questions/45285129/any-difference-between-await-promise-all-and-multiple-await/54291660#54291660

and see if you think that is relevant for this case.

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
```diff
@@ -0,0 +1,101 @@
...
+          });
+        }
+
+        // If not selecting and there is no other information availiable
+        if (!isSelecting) {
+          const municipalityFeature = await adminLegalBoundaryLayerServiceFindOneMunicipality(
+            latLng,
+          );
+
+          result.pimsFeature = pimsLocationProperties?.features[0] ?? null;
```
devinleighsmith, line 76: can pimsLocationsProperties?.features never be an empty array?

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
```diff
@@ -0,0 +1,101 @@
...
+          result.parcelFeature = parcelFeature ?? null;
+          result.regionFeature = regionFeature ?? null;
+          result.districtFeature = districtFeature ?? null;
+          result.municipalityFeature = municipalityFeature ?? null;
+        }
+      } finally {
```
devinleighsmith, line 82: ?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+export const useMapSearch = () => {
+  const fullyAttributedService = useFullyAttributedParcelMapLayer();
+  const pimsPropertyLayerService = usePimsPropertyLayer();
+
+  const { setModalContent, setDisplayModal } = useModalContext();
+  const keycloak = useKeycloakWrapper();
+  const logout = keycloak.obj.logout;
+
+  const pimsExecute = pimsPropertyLayerService.loadPropertyLayer.execute;
```
devinleighsmith, line 28: nit: function naming. I hope this doesn't kill the pims application when run.

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+          lng: longitude,
+        });
+        const findOnePropertyTask = pimsPropertyLayerServiceFindOne(
+          {
+            lat: latitude,
+            lng: longitude,
+          },
+          'GEOMETRY',
+        );
+        const parcelFeature = await findOneParcelTask;
```
devinleighsmith, line 51: could be parallelized right?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+        const pimsPropertyFeature = await findOnePropertyTask;
+
+        //if found in inventory return or else non inventory
+        const foundGeometry = pimsPropertyFeature
```
devinleighsmith, line 55: should this be pimsPropertyFeature?.geometry ?? parcelFeature?.geometry

devinleighsmith: actually, what is the point of this foundGeometry thing? don't the existing if/else on 65 and 67 work just fine?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+              features: [parcelFeature],
+            };
+          }
+          toast.info(`Property found`);
+        } else {
+          toast.info('No search results found');
+        }
+      } catch (error) {
+        toast.error((error as Error).message, { autoClose: 7000 });
+      } finally {
```
devinleighsmith, line 79: ?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+      }
+      return result;
+    },
+    [fullyAttributedServiceFindOne, pimsPropertyLayerServiceFindOne],
+  );
+
+  const searchMany = useCallback(
+    async (filter?: IGeoSearchParams) => {
+      let result: MapFeatureData = emptyFeatureData;
+      //TODO: PSP-4390 currently this loads all matching properties, this should be rewritten to use the bbox and make one request per tile.
```
devinleighsmith, line 89: is this comment still relevant in this context?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+          findByPinTask,
+          findByPidTask,
+        ]);
+
+        // If the property was found on the pims inventory, use that.
+        if (pidPinInventoryData?.features && pidPinInventoryData?.features?.length > 0) {
+          const validFeatures = pidPinInventoryData.features.filter(feature => !!feature?.geometry);
+
+          result = {
+            pimsFeatures: pidPinInventoryData,
```
devinleighsmith, line 147: shouldn't this be validFeatures?

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -0,0 +1,204 @@
...
+              ...(pinFullyAttributedData?.features || []),
+              ...(pidFullyAttributedData?.features || []),
+            ],
+            bbox: pinFullyAttributedData?.bbox || pidFullyAttributedData?.bbox,
+          };
+
+          const validFeatures = attributedFeatures.features.filter(feature => !!feature?.geometry);
+          result = {
+            pimsFeatures: emptyPimsFeatureCollection,
+            fullyAttributedFeatures: attributedFeatures,
```
devinleighsmith, line 172: same q here.

source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -1,130 +1,141 @@
...
-  });
-
-  const lastZoom = useAppSelector(state => state.mapViewZoom) ?? zoomProp;
-  const [zoom, setZoom] = useState(lastZoom);
+  const handleMapClickEvent = (latlng: LatLng) => {
+    mapMachine.mapClick(latlng);
+  };
+
+  const [zoom, setZoom] = useState(DEFAULT_MAP_ZOOM);
+  const [isMapReady, setIsMapReady] = useState(false);
```
devinleighsmith, line 69: it feels weird to have this outside the state machine, what do you think?

source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -1,130 +1,141 @@
...
+
+  const mapMachinePendingRefresh = mapMachine.pendingFitBounds;
+  const mapMachineProcessFitBounds = mapMachine.processFitBounds;
+  const mapMachineRequestedFitBounds = mapMachine.requestedFitBounds;
   useEffect(() => {
-    if (lastZoom === DEFAULT_MAP_ZOOM) {
-      dispatch(setMapViewZoom(smallScreen ? 4.9 : 5.5));
-    } else if (lastZoom !== zoom && zoom !== DEFAULT_MAP_ZOOM) {
-      dispatch(setMapViewZoom(zoom));
+    if (isMapReady && mapMachinePendingRefresh && mapRef.current !== null) {
```
devinleighsmith, line 81: oh, ok. So we allow the fitbounds to fire but ignore it if the map isn't ready. hmm.

source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -142,24 +153,19 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
 
   const fitMapBounds = () => {
     if (mapRef.current) {
-      mapRef.current.fitBounds([
-        [60.09114547, -119.49609429],
-        [48.78370426, -139.35937554],
-      ]);
+      // TODO: Is this necessary?
```
devinleighsmith, line 156: @asanchezr see any reason to keep this?

source/frontend/src/components/maps/MapSearch.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { PropertyFilter } from '@/features/properties/filter';
+import {
+  defaultPropertyFilter,
+  IPropertyFilter,
+} from '@/features/properties/filter/IPropertyFilter';
+
+export type MapSearchProps = {};
+
+/**
+ * Creates a Leaflet map and by default includes a number of preconfigured layers.
```
devinleighsmith, line 16: copy pasta?

source/frontend/src/components/maps/MapSearch.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * @param param0
+ */
+const MapSearch: React.FC<React.PropsWithChildren<MapSearchProps>> = () => {
+  const { mapFilter: mapMachineMapFilter, setMapFilter: mapMachineSetMapFilter } =
+    useMapStateMachine();
+
+  const [mapFilter, setMapFilter] = useState<IPropertyFilter | null>(null);
+
+  useEffect(() => {
+    if (mapFilter !== null && !dequal(mapFilter, mapMachineMapFilter)) {
```
devinleighsmith, line 26: can you help me understand why this shouldn't go in the handle function below?

source/frontend/src/components/maps/MapView.test.tsx
```diff
@@ -88,18 +88,8 @@ function createProps(): TestProps {
...
-      showSideBar={false}
-      lat={48.43}
-      lng={-123.37}
-      zoom={zoom}
-      whenReady={done}
-      whenCreated={setMap}
-      {...rest}
-    />
-  );
+  //const { done, setMap, zoom = 6 } = props;
```
devinleighsmith, line 91: seems like this line can be deleted.

source/frontend/src/components/maps/hooks/useMapRefreshEvent.tsx
```diff
@@ -5,9 +5,9 @@ import useEventListener from '@/hooks/useEventListener';
...
  */
 
-export const mapRefreshEventName = 'refreshMap';
-export const mapRefreshEvent = new CustomEvent(mapRefreshEventName);
-export const fireMapRefreshEvent = () => window.dispatchEvent(mapRefreshEvent);
+const mapRefreshEventName = 'refreshMap';
+//const mapRefreshEvent = new CustomEvent(mapRefreshEventName);
+//const fireMapRefreshEvent = () => window.dispatchEvent(mapRefreshEvent);
 
 export const useMapRefreshEvent = (callback: (event: Event) => void) => {
```
devinleighsmith, line 12: this looks like dead code.

source/frontend/src/components/maps/leaflet/Control/ZoomOut/ZoomOutButton.tsx
```diff
@@ -18,20 +18,17 @@ const ZoomButton = styled(Button)`
   }
 `;
 
-export type ZoomOutProps = {
-  /** the default bounds of the map to zoom out to */
-  bounds: LatLngBounds;
-};
+export type ZoomOutProps = {};
```
devinleighsmith, line 21: ? probably can delete.

source/frontend/src/components/maps/leaflet/LayerPopup/LayerPopupContainer.tsx
```diff
@@ -31,27 +33,80 @@ export type LayerPopupInformation = {
...
 >((props, ref) => {
+  const mapMachine = useMapStateMachine();
+
+  const [layerPopup, setLayerPopup] = useState<LayerPopupInformation>(emptyLayerPopupInformation);
+
+  useEffect(() => {
+    if (mapMachine.mapLocationFeatureDataset) {
+      const featureSet = mapMachine.mapLocationFeatureDataset;
+
+      let title__ = 'Location Information';
```
devinleighsmith, line 59: what's the deal with this variable naming?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -31,115 +28,87 @@ export type PointClustererProps = {
...
   const filterState = useFilterContext();
-  const mapStateContext = React.useContext(MapStateContext);
-  const { selectedInventoryProperty: selected } = mapStateContext;
 
-  const [currentSelected, setCurrentSelected] = useState(selected);
-  const [currentCluster, setCurrentCluster] = useState<
-    ICluster<any, Supercluster.AnyProps> | undefined
-  >(undefined);
+  // TODO: Figure out if the currentCluster is needed
+  const [, setCurrentCluster] = useState<ICluster<any, Supercluster.AnyProps> | undefined>(
```
devinleighsmith, line 56: @asanchezr a little help here?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -31,115 +28,87 @@ export type PointClustererProps = {
...
+    undefined,
+  );
+
+  const mapMachine = useMapStateMachine();
+
+  const selectedFeature = useMemo(() => {
```
devinleighsmith, line 62: why is a memo needed here? this is just assigning a variable.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -31,115 +28,87 @@ export type PointClustererProps = {
...
-  minZoom = minZoom ?? 0;
-  maxZoom = maxZoom ?? 18;
+  const minZoom = minZoomProps ?? 0;
+  const maxZoom = maxZoomProps ?? 18;
+
+  const [spider, setSpider] = useState<
+    SpiderSet<PIMS_Property_Location_View | PMBC_FullyAttributed_Feature_Properties>
+  >({});
+
+  const draftPoints = useMemo(() => {
```
devinleighsmith, line 78: is this the right place to do this?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -224,31 +183,29 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
 
       if (groupBounds.isValid()) {
```
devinleighsmith, line 184: does this if block have a purpose if fitbounds is removed?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -224,31 +183,29 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
         filterState.setChanged(false);
-        mapInstance.fitBounds(groupBounds, { maxZoom: zoom > MAX_ZOOM ? zoom : MAX_ZOOM });
       }
     }
   }, [draftFeatureGroupRef, mapInstance, draftPoints]);
 
   /**
-   * Update the map bounds and zoom to make all property clusters visible.
+   * Updates the state of the cluster if the feature group has been updated.
+   * TODO: Figure out if this is still necessary
```
devinleighsmith, line 192: @asanchezr ?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -258,148 +215,142 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
-                    type: MapStateActionTypes.SELECTED_INVENTORY_PROPERTY,
-                    selectedInventoryProperty: convertedProperty,
-                  });
-                },
-              }}
-            />
-          );
+            const isSelected =
+              selectedFeature !== null ? clusterFeature.id === selectedFeature.clusterId : false;
+            // TODO: There might be a cleaner way to know if a marker is from a different source.
```
devinleighsmith, line 252: wait, is it even possible to have clustered markers that aren't from PIMS?

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -258,148 +215,142 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
 
 export default PointClusterer;
+
+/**
+ * @param feature the feature to obtain lat/lng coordinates for.
+ * @returns [lat, lng]
+ */
+const getLatLng = <P,>(feature: Feature<Geometry, P>) => {
+  if (feature.geometry.type === 'Polygon' || feature.geometry.type === 'MultiPolygon') {
+    const latLng = geoJSON(feature.geometry).getBounds().getCenter();
```
devinleighsmith, line 324: I think getCenter is bad times because it can't guarantee that the returned lat/lng is actually within the boundary (think of a property with a boundary shaped like the letter C). I think we want to user polylabel for this.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -258,148 +215,142 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
+    return [latLng.lng, latLng.lat];
+  } else if ('coordinates' in feature.geometry) {
+    // TODO: This is only needed to satisfy the types. Fix this.
```
devinleighsmith, line 327: what is the plan for all of these todos? log tech debt? resolve before merging? trust that we'll fix these at some point?

source/frontend/src/components/maps/leaflet/Layers/Spiderfier.ts
```diff
@@ -78,7 +94,9 @@ export class Spiderfier {
     }
     this.unspiderfy();
     this.cluster = cluster;
-    const centerLatlng = GeoJSON.coordsToLatLng(cluster?.geometry?.coordinates as [number, number]);
+    const clusterX = cluster?.geometry?.coordinates[0];
```
devinleighsmith, line 97: not sure if this is equivalent. The old method would assume an array of size 2 and operate on it, but this specifically tries to dereference index 0 and 1. The old way might not crash if the array doesn't have at least two values but this will.

source/frontend/src/components/maps/leaflet/Layers/util.tsx
```diff
@@ -125,47 +128,64 @@ export const createPoints = (properties: IProperty[], type: string = 'Point') =>
...
-  if (!!isCluster) {
-    return createClusterMarker(feature, latlng);
+export function pointToLayer<
+  P extends PIMS_Property_Location_View | PMBC_FullyAttributed_Feature_Properties,
+  C extends Supercluster.ClusterProperties,
+>(
+  point: Supercluster.ClusterFeature<C> | Supercluster.PointFeature<P>,
+  latlng: LatLngExpression,
+): Layer {
+  console.log('point', point);
```
devinleighsmith, line 138: these probably need to come out.

source/frontend/src/components/maps/leaflet/Layers/util.tsx
```diff
@@ -183,9 +203,20 @@ export const getDraftIcon = (text: string) => {
...
  * @param latlng the point position
  */
-export const createSingleMarker = (feature: ICluster, latlng: LatLngExpression): Layer => {
-  const icon = getMarkerIcon(feature);
-  return new Marker(latlng, { icon });
+export const createSingleMarker = <P,>(
+  feature: Supercluster.PointFeature<P>,
+  latlng: LatLngExpression,
+): Layer => {
+  const isOwned = feature.id?.toString().startsWith('PIMS_PROPERTY_LOCATION_VW') ?? false;
```
devinleighsmith, line 210: might be nice to have this as a function so if we come up with a better way of doing this we can fix all of these at once.

source/frontend/src/components/maps/leaflet/Markers/SingleMarker.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Marker } from 'react-leaflet';
+import { PointFeature } from 'supercluster';
+
+import { useMapStateMachine } from '@/components/common/mapFSM/MapStateMachineContext';
+import { PMBC_FullyAttributed_Feature_Properties } from '@/models/layers/parcelMapBC';
+import { PIMS_Property_Location_View } from '@/models/layers/pimsPropertyLocationView';
+
+import { getMarkerIcon, getNotOwnerMarkerIcon } from '../Layers/util';
+
+interface SinlePropertyMarkerProps {
```
devinleighsmith, line 11: nit: sinle

source/frontend/src/components/maps/leaflet/Markers/SingleMarker.tsx
```diff
@@ -0,0 +1,74 @@
...
+  pointFeature,
+  markerPosition,
+  isSelected,
+  isOwned,
+}) => {
+  const mapMachine = useMapStateMachine();
+
+  const getIcon = () => {
+    if (isOwned) {
+      const pimsFeature = pointFeature as PointFeature<PIMS_Property_Location_View>;
```
devinleighsmith, line 28: I wonder if we should ever do runtime validation on any of these to catch if we make changes to the layer but forget to update the interface. What thinkest thou?

source/frontend/src/components/maps/leaflet/styles_nope.ts
devinleighsmith, line 1: what is with the naming of this file?

source/frontend/src/components/propertySelector/MapClickMonitor.tsx
```diff
@@ -1,36 +1,36 @@
...
 import useDeepCompareEffect from '@/hooks/util/useDeepCompareEffect';
-import { mapFeatureToProperty } from '@/utils/mapPropertyUtils';
+import { featuresetToMapProperty } from '@/utils/mapPropertyUtils';
 
 import { IMapProperty } from './models';
 
 interface IMapClickMonitorProps {
   addProperty: (property: IMapProperty) => void;
-  modifiedProperties: IMapProperty[];
+  modifiedProperties: IMapProperty[]; // TODO: this should be just a list of lat longs
```
devinleighsmith, line 13: should it? doesn't this feed into the actually selected properties list?

source/frontend/src/features/mapSideBar/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -52,6 +52,8 @@ export const AddAcquisitionForm = React.forwardRef<
   const [projectProducts, setProjectProducts] = React.useState<Api_Product[] | undefined>(
     undefined,
   );
+
+  console.log('AddAcquisitionForm', initialValues);
```
devinleighsmith, line 56: remove.

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/update/UpdateStakeHolderForm.tsx
```diff
@@ -87,6 +87,7 @@ export const UpdateStakeHolderForm: React.FunctionComponent<IUpdateStakeHolderFo
                                 View={ContactInputView}
                               ></ContactInputContainer>
                             </Col>
+                            {interestHolder.contact?.organization?.organizationPersons?.length}
```
devinleighsmith, line 90: this looks like that number I mentioned the other day.

source/frontend/src/features/mapSideBar/project/add/AddProjectContainer.test.tsx
```diff
@@ -67,9 +67,9 @@ describe('AddProjectContainer component', () => {
     selectedProperty: Feature<Geometry, GeoJsonProperties> | null = null,
   ) => {
     const utils = render(
-      <MapStateContextProvider values={{ selectedFileFeature: selectedProperty }}>
+      <MapStateMachineProvider /*values={{ selectedFileFeature: selectedProperty }}*/>
```
devinleighsmith, line 70: would prefer this was removed instead of commented out without context.

source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.test.tsx
```diff
@@ -13,7 +13,7 @@ import MotiInventoryContainer, { IMotiInventoryContainerProps } from './MotiInve
 jest.mock('@react-keycloak/web');
 
 const onClose = jest.fn();
-const onZoom = jest.fn();
+//const onZoom = jest.fn();
```
devinleighsmith, line 16: just delete these.

source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.tsx
```diff
@@ -79,13 +81,19 @@ export const MotiInventoryContainer: React.FunctionComponent<
     setShowCancelConfirmModal(false);
   };
 
+  const handleZoom = (apiProperty?: Api_Property | undefined) => {
+    if (apiProperty !== undefined && apiProperty?.longitude && apiProperty?.latitude) {
```
devinleighsmith, line 85: seems like there is some redundancy here.

source/frontend/src/features/properties/map/MapContainer.tsx
```diff
@@ -1,84 +1,48 @@
...
-        MAP_MAX_ZOOM,
-        {
-          animate: false,
-        },
-      ),
-    [mapInstance],
-  );
+  const { iSelecting } = useMapStateMachine();
+
+  const cursorClass = useMemo(() => {
```
devinleighsmith, line 28: I don't think this needs to be memoized.

source/frontend/src/features/properties/map/MapContainer.tsx
```diff
@@ -1,84 +1,48 @@
...
-      mapInstance?.flyTo(
-        { lat: apiProperty?.latitude, lng: apiProperty?.longitude },
-        MAP_MAX_ZOOM,
-        {
-          animate: false,
-        },
-      ),
-    [mapInstance],
-  );
+  const { iSelecting } = useMapStateMachine();
```
devinleighsmith, line 26: iSelecting?

source/frontend/src/hooks/repositories/useComposedProperties.ts
```diff
@@ -157,31 +157,46 @@ export const useComposedProperties = ({
...
-    bcAssessmentWrapper: getSummaryWrapper,
-    composedLoading:
-      getLtsaWrapper?.loading ||
-      getPropertyWrapper?.loading ||
-      getPropertyAssociationsWrapper?.loading ||
-      findByWrapper?.loading ||
-      getPropertyWfsWrapper?.loading ||
-      getSummaryWrapper?.loading,
-  };
+  return useMemo(
```
devinleighsmith, line 168: interesting , the memo isn't required in the other two locations but is here? any idea why?

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
...
+    district: isNumber(districtFeature?.properties.DISTRICT_NUMBER)
+      ? districtFeature?.properties.DISTRICT_NUMBER
+      : DistrictCodes.Unknown,
+    districtName: districtFeature?.properties.DISTRICT_NAME ?? 'Cannot determine',
+    name: pimsFeature?.properties?.NAME ?? undefined,
+  };
+}
+
+export function pidFromFeatureSet(featureset: LocationFeatureDataset): string | null {
+  if (featureset.pimsFeature !== null) {
```
devinleighsmith, line 185: can featureset ever be null or undefined?

source/frontend/src/store/rootReducer.ts
```diff
@@ -20,6 +19,5 @@ export const reducer = {
   [jwtSlice.name]: jwtSlice.reducer,
   [filterSlice.name]: filterSlice.reducer,
   [keycloakReadySlice.name]: keycloakReadySlice.reducer,
-  [mapViewZoomSlice.name]: mapViewZoomSlice.reducer,
```
devinleighsmith, line 23: I'm guessing leafletmouseslice is dead too.

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
     name: feature?.properties?.NAME,
   };
 }
+
+export function featuresetToMapProperty(
+  featureSet: LocationFeatureDataset,
+  address: string = 'unknown',
+): IMapProperty {
+  const pimsFeature = featureSet.pimsFeature;
```
devinleighsmith, line 155: can featureset be null/undefined?

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
...
+  const parcelFeature = featureSet.parcelFeature;
+  const regionFeature = featureSet.regionFeature;
+  const districtFeature = featureSet.districtFeature;
+
+  const propertyId = pimsFeature?.properties.PROPERTY_ID;
```
devinleighsmith, line 160: i don't know enough to know if properties can be null/undefined.

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
...
+  const pid = pidFromFeatureSet(featureSet);
+  const pin = pinFromFeatureSet(featureSet);
+  return {
+    propertyId: propertyId ? Number.parseInt(propertyId?.toString()) : undefined,
```
devinleighsmith, line 164: number.parseInt vs Number()?

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
...
+    pid: pid ?? undefined,
+    pin: pin ?? undefined,
+    latitude: featureSet.location.lat,
```
devinleighsmith, line 167: all features have locations?

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -144,3 +147,59 @@ function toMapProperty(
...
+    longitude: featureSet.location.lng,
+    planNumber: parcelFeature?.properties.PLAN_NUMBER?.toString() ?? undefined,
```
devinleighsmith, line 169: inconsistent usage of ?. on properties.

devinleighsmith:
@FuriousLlama lol, my laptop's chrome could barely handle opening something of this size in chrome. Great work, I doubt I could have come up with something this elegant to replace what we had.

devinleighsmith:
I'll approve once you have a chance to have a pass on my comments.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
FuriousLlama: I think this will be solved once the state machine gets strongly typed.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
FuriousLlama: good catch. Both need to be checked

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
FuriousLlama: It would pass those to the searches as nan and would get nothing. Note that this workflow only happens when the address is selected on the search bar and the autocomplete gets an address + lat long. I think it would be fine to leave it as is

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
FuriousLlama: Right now is the same method that is being used in dev. https://github.com/bcgov/PSP/blob/335915188e90ec530cfb91749a9ecc5b0d8bfefa/source/frontend/src/components/maps/MapView.tsx#L44

The loadPropertiesRequest does have the extra regex parameters as in dev.

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: My intention was to get feature parity, so having it named the same would help match the functionality. I do believe that renaming is better.

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: Yes, I will get Dev's input once this gets merged. There are side effects to having it always open as it is now

source/frontend/src/components/common/mapFSM/machineDefinition/types.ts
FuriousLlama: This will be needed when the state machine becomes strictly typed.

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
FuriousLlama: That's a good link, thanks for sharing it! I have updated this one and when I see future ones like these to follow that style as it looks cleaner (even we don't actually handle errors in this case)

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
FuriousLlama: good catch. I;ve wrapped it in a check to make sure it does not cause issues

source/frontend/src/components/common/mapFSM/useLoactionFeatureLoader.tsx
FuriousLlama: A leftover of the legacy try above. Right now I rather leave it to maintain functional parity. Once we tested it more thoroughly we can remove it.

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
FuriousLlama: I was matching the functionality as it was before. cleaner now

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
FuriousLlama: I believe it is given the search is still looking for the entire map instead of using the bounds.

source/frontend/src/components/maps/MapLeafletView.tsx
FuriousLlama: This is necessary given the component might be ready but not the map. Having an effect with the isMapReady allows the map requests to be processed.
From the map fsm point of view, all map requests are async so it does not need to know if the map is ready.

source/frontend/src/components/maps/MapLeafletView.tsx
FuriousLlama: right, ideally the ```mapRef.current !== null``` part would be reactive, but given it isnt...

source/frontend/src/components/maps/MapSearch.tsx
FuriousLlama: It is a problem of initialization. This component gets initialized whenever the application starts and when the sidebar closes. On the second case the map fsm knows the filter already, while on the first case it does not and expects the url parser on the PropertyFilter to parse the url and update the state.
In summary, the handleMapFilterChange gets called on initialization always, but is not always necessary to set the map filter on the map fsm

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: fair point

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: I believe it is given is only for rendering purpouses

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: It also updates the filterContext. I dont think is doing much right now. But that would be another task to remove the filtercontext imo.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: Yes, the clusters are coming from the feature points passed to the supercluster. So in theory it could have a mixture of both types of features. In practice it only has one type at a time

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: That would be a functional change. Lets create a task for that

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
FuriousLlama: I wrote most of them as markers so we can have a conversation (like this one) about the functionality. I will create tasks for the rest of them

source/frontend/src/components/propertySelector/MapClickMonitor.tsx
FuriousLlama: The modifiedProperties is only used to synchronize the markers on the map. useDraftMarkerSynchronizer only uses the locations from the passed properties.

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/update/UpdateStakeHolderForm.tsx
FuriousLlama: weird, no idea how that got in there.

source/frontend/src/features/properties/map/MapContainer.tsx
FuriousLlama: Saving one character at a time

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: Not at this point.

source/frontend/src/store/rootReducer.ts
FuriousLlama: Not sure what that one does, so it should be removed in another task

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: Not at this point. The method signature guarantees is an object

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: I dont think it can

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: Looks like number() is less lenient on the parameter and can result in a floating, while parse int only return integers

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: Yes, given that they are based on a lat lng search. The type enforces it

FuriousLlama:
> @FuriousLlama lol, my laptop's chrome could barely handle opening something of this size in chrome. Great work, I doubt I could have come up with something this elegant to replace what we had.

Thank you. It took a lot of work.

source/frontend/src/store/rootReducer.ts
devinleighsmith: I will log tech debt.

devinleighsmith (approved):

PR #3324, opened 2023-07-14
PSP-5116 Generate Expropriation Form 1

asanchezr:
image


tools/mayan-edms/docker-compose.yml
```diff
@@ -146,8 +146,7 @@ services:
     command:
       - run_worker
       - worker_a
-      - "--prefetch-multiplier=1"
-      - "--concurrency=10 --max-memory-per-child=1000 -E"
+      - "--prefetch-multiplier=1 --concurrency=10 --max-memory-per-child=1000 -E"
```
asanchezr, line 149: This is the fix for mayan worker containers being restarted continuously

source/frontend/src/models/generate/GenerateOwner.ts
```diff
@@ -3,6 +3,7 @@ import { phoneFormatter } from '@/utils/formUtils';
 
 import { Api_GenerateAddress } from './GenerateAddress';
 export class Api_GenerateOwner {
+  owner_row: number;
```
eddherrera, line 6: what this used for?

eddherrera (approved):

source/backend/api/Areas/Acquisition/Controllers/InterestHolderController.cs
```diff
@@ -55,9 +55,9 @@ public InterestHolderController(IAcquisitionFileService acquisitionService, IMap
         [ProducesResponseType(typeof(IEnumerable<InterestHolderModel>), 200)]
         [SwaggerOperation(Tags = new[] { "acquisitionfile" })]
         [TypeFilter(typeof(NullJsonResultFilter))]
-        public IActionResult GetAcquisitionFileInterestHolders([FromRoute] long id)
+        public IActionResult GetAcquisitionFileInterestHolders([FromRoute] long id, bool? fullDetails)
         {
-            var interestHolders = _acquisitionService.GetInterestHolders(id);
+            var interestHolders = _acquisitionService.GetInterestHolders(id, fullDetails ?? false);
```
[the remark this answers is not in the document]
asanchezr: Yes, I thought of adding the extra joined tables to the original query but this would bring two more related tables into the query "interest holder" -> person -> address (new) -> country (new). 
This is only needed on the generation side as we want to display full addresses including country/province/etc but for the original stakeholder tabs where contacts are added/removed as interest holders on an Acquisition File, the addresses are not needed.

I will remove the extra logic and let's discuss a pattern moving forward when Devin comes back. We have talked in the past of having two ways of getting information (the lightweight model vs the fat model with many sub-entities loaded)

source/frontend/src/models/generate/GenerateOwner.ts
asanchezr: Sadly, this is used to be able to conditionally render "TO:" John Doe on the first owner but use "AND:" on the rest. Here is the logic in the template. I guess I could rename to owner_index if it makes it more clear...

image

This is done to comply with the following acceptance criteria on the template:
"// List all the owners with first owner name Prefixed with ‘TO:’ and subsequent with ‘AND:’, ...."

asanchezr: I tried using Carbone logic for this but could not get it working - that is determining the current index of the array of owners being rendered by carbone

FuriousLlama: on carbone you can refer to the index as well

asanchezr: I tried using this `{d.file.owners[i]:count():ifEQ(1)}` but that didn't work. The count "prints" the row number but it doesn't pass it to the next formatter ifEQ()

asanchezr: I'm modifying GenerateOwner to instead take an "owner_prefix" (empty by default) that way I can set the prefix just for expropriation forms. Don't like the index much

asanchezr: Okay - found a way to do it in carbone. Will remove the owner_row

PR #3325, opened 2023-07-14
PSP-6563 : HOTFIX: in trust is missing on H120

eddherrera:


devinleighsmith:
confirming that:
red text should no longer be used for injected values.
the owners should not be displayed on the property service line.
the letter 's' should be appended to the project name.
@eddherrera

FuriousLlama (approved):

eddherrera:
> confirming that: red text should no longer be used for injected values. the owners should not be displayed on the property service line. the letter 's' should be appended to the project name. @eddherrera

@devinleighsmith this seems to be correct. The template was only modified with the 'in trust' section.

PR #3333, opened 2023-07-17
fix trufflehog install from failing

devinleighsmith:


devinleighsmith:
originally caused by https://github.com/yaml/pyyaml/issues/724

devinleighsmith:
Also logged https://github.com/feeltheajf/trufflehog3/issues/50 with the trufflehog people so we can remove the hardcoded dependencies.

PR #3337, opened 2023-07-18
Frontend alignment with interest holder updates

FuriousLlama:
Updated front end models and fixed some issues related to interest holders


source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
```diff
@@ -31,9 +33,11 @@ export const useGenerateAgreement = () => {
...
       team => team.personProfileTypeCode === 'MOTILAWYER',
     );
-    const ownerSolicitor = file.acquisitionFileOwnerSolicitors?.length
-      ? file.acquisitionFileOwnerSolicitors[0]
-      : undefined;
+    const solicitors = file.acquisitionFileInterestHolders
+      ?.filter(x => x.interestHolderType?.id === InterestHolderType.OWNER_SOLICITOR)
+      .filter((x): x is Api_InterestHolder => !!x);
+    const ownerSolicitor =
+      solicitors !== undefined && solicitors.length > 0 ? solicitors[0] : undefined;
```
asanchezr, line 40: in **useGenerateExpropriationForm1** the following pattern was used. Would lean towards using the same here to keep consistency:

```tsx
    const ownerSolicitor = file.acquisitionFileInterestHolders?.find(
      x => x.interestHolderType?.id === InterestHolderType.OWNER_SOLICITOR,
    );
```

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/update/models.test.ts
```diff
@@ -1,21 +1,44 @@
...
-  StakeHolderForm,
-} from './models';
+import { InterestHolderForm, StakeHolderForm } from './models';
+
+const emptyInterestHolderForm: InterestHolderForm = {
+  interestHolderId: null,
+  //person: null,
+  //personId: '',
+  //organization: null,
+  //organizationId: '',
```
asanchezr, line 12: please remove dead code (or take out the comments)

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/update/models.ts
```diff
@@ -13,112 +15,127 @@ export class StakeHolderForm {
...
+        return apiInterestHolder;
       })
-      .value();
+      .value()
+      .filter((x): x is Api_InterestHolder => x !== null);
+  }
+}
+
+export class InterestHolderPropertyFormModel {
+  //acquisitionFileProperty = {};
```
asanchezr, line 85: please remove comment or re-enable the property

asanchezr (approved):

devinleighsmith:
@FuriousLlama there is a 1 showing up here for some reason: 
image

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
FuriousLlama: is doing the same, the difference is that the second one has potentially many interest holders

FuriousLlama:
> //acquisitionFileProperty = {};

Checked and my local does not have that on the stakeholder screen. That might be something on your side

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.ts
asanchezr: The 2nd one still only pick the first one it finds- ie "solicitors[0]" so to me that is doing the same as "find" which returns the first it finds. But not a big deal. We can align that later

PR #3344, opened 2023-07-19
psp-3242: add plan number search for map and list views

buddy326:


source/frontend/src/components/maps/hooks/useMapSearch.tsx
```diff
@@ -78,19 +78,26 @@ export const useMapSearch = () => {
...
+
       loadPropertiesTask = pimsPropertyLayerService.loadPropertyLayer.execute(filter);
       if (filter?.PIN) {
         findByPinTask = fullyAttributedService.findByPin(filter?.PIN);
       }
       if (filter?.PID) {
         findByPidTask = fullyAttributedService.findByPid(filter?.PID);
       }
+      if (filter?.SURVEY_PLAN_NUMBER) {
+        findByPlanNumberTask = fullyAttributedService.findByPlanNumber(filter?.SURVEY_PLAN_NUMBER);
```
FuriousLlama, line 93: this should be its own method to make sure the other filter parameters  (pid,pin)are not trumping this. Also, I dont think it would be possible to do a pid,pin + survey plan search

devinleighsmith: agreed pid/pin is a simultaneous search. Ie. a single search field that can be interpreted two ways. not the case here.

buddy326: Fixed along with some minor refactoring of the searchByMany method code due to a lot of repetition.

devinleighsmith:
@buddy326 see if you can get through resolving the above conflicts.

buddy326:
Working on this, merging results in all my code breaking and changes getting deleted so just trying to understand how all the new map stuff works and refactoring my changes.

buddy326:
Conflicts Resolved.

devinleighsmith:
@buddy326 plan number needs to be an exact match search, currently the search is using ILIKE in all searches.

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -77,6 +77,10 @@ private static IQueryable<Entity.PimsProperty> GenerateCommonPropertyQuery(this
             {
                 query = query.Where(p => EF.Functions.Like(p.Address.StreetAddress1, $"%{filter.Address}%") || EF.Functions.Like(p.Address.MunicipalityName, $"%{filter.Address}%"));
             }
+            if (!string.IsNullOrWhiteSpace(filter.PlanNumber))
```
devinleighsmith, line 80: same here, plan number should be an exact match unlike pid/pin. This is because plan numbers do not have a fixed length so someone looking for plan VIP8 will find thousands of matches in the province for plans VIP8X, VIP8XX, and VIP8XXX etc.

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -28,10 +28,48 @@ export const useMapSearch = () => {
   const loadPimsProperties = pimsPropertyLayerService.loadPropertyLayer.execute;
   const fullyAttributedServiceFindByPin = fullyAttributedService.findByPin;
   const fullyAttributedServiceFindByPid = fullyAttributedService.findByPid;
+  const fullyAttributedServiceFindByPLanNumber = fullyAttributedService.findByPlanNumber;
```
devinleighsmith, line 31: within useFullyAttributedParcelMapLayer.tsx, pass the forceExactMatch param as true to resolve this issue.

devinleighsmith (changes requested):
see comments.

buddy326:
Fixed.

devinleighsmith (approved):

devinleighsmith:
@buddy326 you have a conflict, but approved.

PR #3350, opened 2023-07-21
psp-6306: add primary contact selector for owner solicitor

buddy326:
There's a bug when the selected organization has no contacts associated with it - in that case I just display "No contacts available" but because of that, the primaryContactId field does not get set to null and therefore the organization may get updated but the primaryContactId associated with it stays the same as it was for the previous selected organization that DID have a primary contact selected. 

Steps to reproduce: 

1) Select an organization with one or more contacts associated with it.
2) Select a primary contact for it and save.
3) Go into edit mode and change the organization to one with 0 contacts associated with it.
4) Save and look at the view. The contact selected for the first organization gets displayed.

Would appreciate some feedback on how to set the primaryContactId as null when no selection is made from the user.


source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/detail/AcquisitionSummaryView.tsx
```diff
@@ -44,6 +45,10 @@ const AcquisitionSummaryView: React.FC<IAcquisitionSummaryViewProps> = ({
     x => x.interestHolderType?.id === InterestHolderType.OWNER_SOLICITOR,
   );
 
+  const ownerSolicitorPrimaryContact = usePersonDetail(
+    ownerSolicitor?.primaryContactId ?? -9,
```
devinleighsmith, line 49: what is this magic number of -9?

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -107,6 +111,8 @@ const AcquisitionDetailSubForm: React.FC<{
     [retrieveProjectProducts],
   );
 
+  console.log(ownerSolicitorContact);
```
devinleighsmith, line 114: remember to remove console.logs before committing.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -83,13 +85,15 @@ const AcquisitionDetailSubForm: React.FC<{
   const [projectProducts, setProjectProducts] = React.useState<Api_Product[] | undefined>(
     undefined,
   );
+  const [isPersonSelected, setIsPersonSelected] = React.useState(true);
```
devinleighsmith, line 88: Remember, we do not use state on non-containers.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -266,6 +296,19 @@ const AcquisitionDetailSubForm: React.FC<{
             View={ContactInputView}
           ></ContactInputContainer>
         </SectionField>
+        {!isPersonSelected && (
```
devinleighsmith, line 299: why not just:
ownerSolicitorContact?.organizationId && !ownerSolicitorContact?.personId 
or 
ownerSolicitorContact?.organizationId && !ownerSolicitorContact?.organizationId
instead of a state variable and a useffect?

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -120,6 +126,30 @@ const AcquisitionDetailSubForm: React.FC<{
     }
   }, [fileStatusTypeCode, setFieldValue]);
 
+  React.useEffect(() => {
```
devinleighsmith, line 129: you could keep this useeffect but instead of setting a variable use this to clear the primary contact when the ownerSolicitorContact changes.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -120,6 +126,30 @@ const AcquisitionDetailSubForm: React.FC<{
...
+
+  const orgPersons = organization?.persons;
+
+  const primaryContacts: SelectOption[] =
+    orgPersons?.map((person: Partial<IEditablePerson>) => {
+      const { firstName, middleNames, surname, id } = person;
+
+      return {
+        label: `${firstName} ${middleNames} ${surname}`,
+        value: id ? (id as number) : ' ',
```
devinleighsmith, line 149: why is this cast necessary?

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/detail/AcquisitionSummaryView.tsx
```diff
@@ -137,6 +142,22 @@ const AcquisitionSummaryView: React.FC<IAcquisitionSummaryViewProps> = ({
             </StyledLink>
           </SectionField>
         )}
+        {ownerSolicitor?.organization && (
```
devinleighsmith, line 145: Looks like you are missing a requirement. If there is only one primary contact, it should be automatically selected. as per: 

2. Screenshot 2: User cannot select different primary contact, if there is only ONE attached to linked organization

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -83,13 +85,15 @@ const AcquisitionDetailSubForm: React.FC<{
...
   );
+  const [isPersonSelected, setIsPersonSelected] = React.useState(true);
   const { retrieveProjectProducts } = useProjectProvider();
   const { getOptionsByType } = useLookupCodeHelpers();
   const regionTypes = getOptionsByType(API.REGION_TYPES);
   const acquisitionTypes = getOptionsByType(API.ACQUISITION_TYPES);
   const acquisitionPhysFileTypes = getOptionsByType(API.ACQUISITION_PHYSICAL_FILE_STATUS_TYPES);
   const fileStatusTypeCodes = getOptionsByType(API.ACQUISITION_FILE_STATUS_TYPES);
   const acquisitionFundingTypes = getOptionsByType(API.ACQUISITION_FUNDING_TYPES);
+  const ownerSolicitorContact = formikProps.values.ownerSolicitor.contact;
```
devinleighsmith, line 96: can ownerSolicitor be null?

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -120,6 +126,30 @@ const AcquisitionDetailSubForm: React.FC<{
...
+    if (!ownerSolicitorContact || ownerSolicitorContact?.personId) {
+      setIsPersonSelected(true);
+    }
+
+    if (ownerSolicitorContact?.organizationId && !ownerSolicitorContact?.personId) {
+      setIsPersonSelected(false);
+    }
+  }, [ownerSolicitorContact]);
+
+  const { organization } = useOrganizationDetail(ownerSolicitorContact?.organizationId ?? 1);
```
devinleighsmith, line 139: this seems undesireable, why make a request with id 1 when no organization is set? better to only make requests when the id is set.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -266,6 +296,19 @@ const AcquisitionDetailSubForm: React.FC<{
             View={ContactInputView}
           ></ContactInputContainer>
         </SectionField>
+        {!isPersonSelected && (
+          <SectionField label="Primary contact" className="mt-4">
+            {primaryContacts.length > 0 ? (
```
devinleighsmith, line 301: Looks like you are missing a requirement. If there is only one primary contact, it should be automatically selected. as per:

Screenshot 2: User cannot select different primary contact, if there is only ONE attached to linked organization

devinleighsmith (changes requested):
see comments.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
buddy326: react doesnt allow conditional usage of hooks so one way to do that would be to wrap the call to useOrganizationDetails in wrapper component and communicate the organization info back to the parent component via state.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
buddy326: yes it is at the start when no contacts are selected

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -120,6 +125,27 @@ const AcquisitionDetailSubForm: React.FC<{
...
+  const { organization } = useOrganizationDetail(ownerSolicitorContact?.organizationId);
+
+  const orgPersons = organization?.persons;
+
+  React.useEffect(() => {
+    if (orgPersons?.length === 0) {
+      setFieldValue('ownerSolicitor.primaryContactId', null);
+    } else if (orgPersons?.length === 1) {
+      setFieldValue('ownerSolicitor.primaryContactId', orgPersons[0].id);
+    } else {
```
devinleighsmith, line 137: @buddy326 do you need this else?

buddy326: your right, its redundant so removed.

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
```diff
@@ -120,6 +125,27 @@ const AcquisitionDetailSubForm: React.FC<{
     }
   }, [fileStatusTypeCode, setFieldValue]);
 
+  const { organization } = useOrganizationDetail(ownerSolicitorContact?.organizationId);
```
devinleighsmith, line 128: @buddy326 
Ok, couple of things here. It looks like with the current design of the contact selector this may be redundant as the contact selector search returns organizations with a nested list of organization persons.

For example, this is the value of ownerSolicitorContact after I select an organization using the contact picker:

{
    "id": "O3",
    "organizationId": 3,
    "organization": {
        "id": 3,
        "isDisabled": false,
        "name": "Dairy Queen Forever! Property Management",
        "organizationPersons": [
            {
                "personId": 3,
                "organizationId": 3,
                "isDisabled": false,
                "rowVersion": 1
            }
        ],
        "organizationAddresses": [],
        "contactMethods": [],
        "rowVersion": 1
    },
    "rowVersion": 0,
    "summary": "Dairy Queen Forever! Property Management",
    "organizationName": "Dairy Queen Forever! Property Management",
    "isDisabled": false
}

As a result, I think you can remove this query entirely, which should also resolve an issue where an error toast would display when clearing the owner solicitor.

devinleighsmith (changes requested):
see latest comment

source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/UpdateAcquisitionForm.tsx
buddy326: If I make this change, then the organization persons are each of type Api_OrganizationPerson which does not have the persons details such as name so that would require me to make a call to usePersonDetail() hook for each personId which would be tricky \as react does not like calls to hooks within conditionals like loops so I figured getting the organization details along with all the persons details with just one call is simpler to do.

devinleighsmith: If that is the case you'll need to change your use of useOrganizationDetail.
1) that hook does not get the "concept" form of the model. Going forward we only use models of the "concept" style. Only legacy code is allowed to use non-concept models.
2) that hook also displays a toast error when the org is cleared which doesn't work for this screen.

buddy326: So in this case, what do you mean when you say it does not get the "concept" form of the model? It returns <IEditableOrganization> which contains all the fields we need. I get the second point and agree its not good to have that error pop up to the user's screen but just a little confused about the first point. Also, when you said I'll need to change my use, does that entail changing the implementation of the hook or rather coming up with a workaround in how I call it. Thanks!

devinleighsmith: Yep, so that hook returns a IEditableOrganization which is a legacy code only model. All new code must use the concept models stored in /models/api on the frontend or /concepts on the backend.

In order to do that, you will either need to find an existing hook that returns the organization concept model, or create a new hook that returns the organization concept model.

You should leave the original hook as-is otherwise you would need to rewrite large portions of the contact pages that use hook.

buddy326: ahh I see what you mean, in that case the new hook I create will be returning Api_Organization but another difference between this hypothetical new hook and useOrganizationDetails is that it would not be displaying a toast error to the screen correct? That would resolve the nasty red message from popping up everytime the contact gets cleared. Just want to make sure Im picking up what your putting down haha? Thanks for the thorough replies so far!

devinleighsmith: yes, the red toast is not desirable. Also the new hook (if one is required, please check that one doesn't already exist) should use the new format of using useApiRequestWrapper.

buddy326: ok sweet, I'll get going on it if theres no existing one. Thank you!

source/frontend/src/features/contacts/hooks/useConceptPersonDetail.ts
```diff
@@ -0,0 +1,57 @@
...
+import { useApiRequestWrapper } from '@/hooks/util/useApiRequestWrapper';
+import { IApiError } from '@/interfaces/IApiError';
+import { Api_Person } from '@/models/api/Person';
+import { logError } from '@/store/slices/network/networkSlice';
+
+/**
+ * hook that fetches the person given the person's id.
+ * @param personId
+ */
+export const useConceptPersonDetail = (personId?: number | null) => {
```
devinleighsmith, line 16: @buddy326 getting closer, but this is more or less the same format as the existing usePersonDetail, which is a really old pattern. Please update this to be more like our *Repository hooks. Ie. useTakesRepository.

devinleighsmith:
@buddy326 if you haven't already take a look at: https://confluence.th.gov.bc.ca/display/PIMS/Best+Practices

source/frontend/src/features/contacts/hooks/useConceptPersonDetail.ts
buddy326: Updated! I will remove the concept hooks as soon as the new repository hooks get approved.

devinleighsmith (approved):

PR #3359, opened 2023-07-25
Fixed loading of properties. Added sidebar of properties info from fullscreen

FuriousLlama:


source/frontend/src/components/common/mapFSM/useLocationFeatureLoader.tsx
```diff
@@ -70,23 +70,19 @@ const useLocationFeatureLoader = () => {
           });
         }
 
-        // If not selecting and there is no other information availiable
-        if (!isSelecting) {
-          const municipalityFeature = await adminLegalBoundaryLayerServiceFindOneMunicipality(
```
devinleighsmith, line 75: Hmm, so now when we are querying for a feature we will make a request to the municipality layer as well? is that required?

FuriousLlama: We should figure out what information is required when processing a draft drop. Right now, the information is loaded twice. Once on click and another on the selected properties view

devinleighsmith: Please log some tech debt for that so we can keep track of these map related issues.

devinleighsmith (approved):

source/frontend/src/components/common/mapFSM/useLocationFeatureLoader.tsx
FuriousLlama: I did. Once we remove IMapProperty we will have to addresss it

PR #3361, opened 2023-07-27
PSP-6608 Advanced Filter Bar

asanchezr:
ezgif com-optimize


FuriousLlama:
@asanchezr IMO the use of 'sidebar' is creating confusion by conflating the left pane, what we currently call sidebar, and the filter/layer. I think we should just name it AdvancedFilter or something like that to avoid confusion. Thoughts?

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -223,6 +240,7 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
           { mapVisible: { sideBar: 'sidebarOpen' } },
           { mapVisible: { sideBar: 'selecting' } },
         ].some(state.matches),
+        sideBarType: state.context.sideBarType,
```
FuriousLlama, line 243: The current sidebar is the left pane of the map. This would be mixing the filters with that conceptually

asanchezr:
> @asanchezr IMO the use of 'sidebar' is creating confusion by conflating the left pane, what we currently call sidebar, and the filter/layer. I think we should just name it AdvancedFilter or something like that to avoid confusion. Thoughts?

@FuriousLlama I'm fine renaming it - but the user story actually says Map Filter Sidebar, and it is indeed a sidebar (just to the right of the map instead of the left). But I'm open to hearing better names

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -224,6 +224,27 @@ const sideBarStates = {
   },
 };
 
+const advancedFilterSideBarStates = {
+  initial: 'fullScreen',
```
FuriousLlama, line 228: If this is the state of the filter, should it be initially closed?

asanchezr: I modelled this after the "regular" sidebar - which starts in the "fullScreen" mode == closed

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
asanchezr: This is here because we need to expose the type of "thing" being displayed in the "regular" sidebar. That is to comply with the following AC from my story:

- Given advanced search is open, when property marker is clicked, property details left hand sidebar should display and advanced search should stay open
- Given advanced search is open, when user navigates to a file, the advanced search should be closed.

asanchezr: So there's logic in MapContainer that listens for what kind of sidebar we have in order to conditionally close the Advance Filter sidebar

source/frontend/src/components/maps/MapView.tsx
```diff
@@ -20,18 +20,22 @@ export type MapViewProps = {};
...
 
   return (
     <VisibilitySensor partialVisibility={true}>
       {({ isVisible }: { isVisible: boolean }) => (
         <Styled.MapGrid
           ref={resizeRef}
-          className={classNames('px-0', 'map', { sidebar: showSideBar })}
+          className={classNames('px-0', 'map', {
+            sidebar: isSidebarOpen || isAdvancedFilterSidebarOpen,
+          })}
```
FuriousLlama, line 35: It would be better if the map FSM controls the visibility completely and the interaction between the different components.

source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -218,10 +219,21 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
 
         <LegendControl />
         <ZoomOutButton />
+        <AdvancedMapFilterButton
+          open={mapMachine.isAdvancedFilterSidebarOpen}
+          onOpen={() => {
+            setLayersOpen(false);
+            mapMachine.openAdvancedFilterSidebar();
+          }}
+          onClose={() => mapMachine.closeAdvancedFilterSidebar()}
```
FuriousLlama, line 228: Let's move this logic to the map FSM given it is coordinating between different states

asanchezr: SetLayersOpen is **not** in the FSM - it is on local state. I'd rather do that on a separate story so we can have the filter sidebar merged

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: I actually like the idea of having the controls of the map as a state given only one can be valid at a time. Having the states being 
none,
advancedFilter,
layers.

source/frontend/src/components/maps/MapLeafletView.tsx
FuriousLlama: Take a look at the PR I posted, it controls the visibility directly on the fsm

devinleighsmith: My two cents, prefer having the FSM manage the state of the advanced search/sidebar. I wouldn't say that because the layer is currently controlled by local state that it should continue to be, I'd rather have a consistent implementation.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
devinleighsmith: I like the idea of keeping the current sidebar and the advanced search/layers separate.

FuriousLlama (approved):

PR #3363, opened 2023-07-27
psp-6651 ensure that searches for pids with leading 0s are supported.

devinleighsmith:


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using System.Text.RegularExpressions;
+using ClosedXML.Report.Utils;
```
asanchezr, line 6: Do we need this XML import here? Doesn't look related to this PR...

asanchezr (approved):

PR #3382, opened 2023-08-02
PSP-6642 - Comp Req Draft to Final Message should appear on save

eddherrera:
image


source/frontend/src/mocks/compensations.mock.ts
```diff
@@ -93,7 +93,7 @@ export const getMockApiCompensation = (): Api_CompensationRequisition => ({
       compensationRequisition: null,
       isPaymentInTrust: true,
       gstNumber: '9999',
-      acquisitionOwnerId: null,
+      acquisitionOwnerId: 1,
```
asanchezr, line 96: Is this relevant to this PR? I didn't see this property used in this PR

asanchezr (approved):

source/frontend/src/mocks/compensations.mock.ts
eddherrera: I wouldn't let me submit the form otherwise. It needed at least one payee in the form.

PR #3389, opened 2023-08-04
Scaffols is 60

FuriousLlama:
Note: 
Payee, landtype will not be functional without front end fixes


source/backend/api/Areas/CompensationRequisition/Controllers/CompensationRequisitionController.cs
```diff
@@ -124,9 +124,12 @@ public IActionResult GetCompensationRequisitionPayee([FromRoute] long id)
...
             _logger.LogInformation("Dispatching to service: {Service}", _compensationRequisitionService.GetType());
 
-            var compensationPayee = _compensationRequisitionService.GetPayeeByCompensationId(id);
+            // TODO:
+            //var compensationPayee = _compensationRequisitionService.GetPayeeByCompensationId(id);
+            var compensationPayee = string.Empty;
 
-            return new JsonResult(_mapper.Map<CompensationPayeeModel>(compensationPayee));
+            //return new JsonResult(_mapper.Map<CompensationPayeeModel>(compensationPayee));
+            return new JsonResult(string.Empty);
```
asanchezr, line 132: I don't know where this is used in the FE but I'm assuming it will get fixed in another PR??

FuriousLlama: Yes, as noted on the pr description, the payee functionality requires additional fixes

source/backend/api/Models/Concepts/CompensationRequisition/CompensationRequisitionMap.cs
```diff
@@ -23,11 +23,9 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.ExpropriationNoticeServedDate, src => src.ExpropNoticeServedDt)
                 .Map(dest => dest.ExpropriationVestingDate, src => src.ExpropVestingDt)
                 .Map(dest => dest.SpecialInstruction, src => src.SpecialInstruction)
-                .Map(dest => dest.Payees, src => src.PimsAcquisitionPayees)
```
asanchezr, line 26: It's okay to remove the compensation payee mapping here but I was expecting to see the payee fields (that were moved into compensation requisition) mapped here as well . For example:

- AcquisitionOwnerId
- InterestHolderId
- IsPaymentInTrust
- etc

devinleighsmith (approved):

source/backend/api/Models/Concepts/CompensationRequisition/CompensationRequisitionMap.cs
FuriousLlama: Same as above, this PR is about getting the scaffold updated and the backend compiling. Another PR will need to take a holistic approach to the compensation requisition functionality alignment.

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -49,11 +48,9 @@ public PimsCompensationRequisition GetById(long compensationRequisitionId)
                 .Include(x => x.YearlyFinancial)
                 .Include(x => x.ChartOfAccounts)
                 .Include(x => x.Responsibility)
-                .Include(c => c.PimsCompReqH120s)
+                .Include(c => c.PimsCompReqFinancials)
                     .ThenInclude(y => y.FinancialActivityCode)
-                .Include(x => x.PimsAcquisitionPayees)
-                    .ThenInclude(y => y.AcquisitionOwner)
-                .Include(x => x.PimsAcquisitionPayees)
+                .Include(x => x.AcquisitionOwner)
```
asanchezr, line 53: We need to include here the other potential payee types - InterestHolder, TeamPersonID, etc

Could you please put a TODO here so we don't forget?

FuriousLlama: Not sure is necessary. At this point, the payee side on the front end will simply not work for owners. If the changes require changes in this section of the code they will have to do so

asanchezr (approved):

PR #3390, opened 2023-08-04
psp-6674 fix lease notes issues.

devinleighsmith:


asanchezr (approved):
This is an ETL story so I'll approve as-is

FuriousLlama (approved):

PR #3391, opened 2023-08-07
PSP-6659 When creating a file or project after certain time, the created stamp date is a day after

asanchezr:
The affected date format function **prettyFormatDate** was not assuming that the dates we get from our API are UTC-based. 

Fixed that, cleaned up the code and updated failing tests.

### DIFF
--- a/source/frontend/src/features/leases/detail/LeasePages/payment/table/terms/TermsForm.tsx
+++ b/source/frontend/src/features/leases/detail/LeasePages/payment/table/terms/TermsForm.tsx
@@ -50,7 +50,7 @@ export const TermsForm: React.FunctionComponent<React.PropsWithChildren<ITermsFo
     'receivedDate',
     'desc',
   );
-  const lastPaymentDate = allPayments.length ? prettyFormatDate(allPayments[0]?.receivedDate) : '';
+  const lastPaymentDate = allPayments.length ? allPayments[0]?.receivedDate : '';
 
   /** This is the payments subtable displayed for each term row. */
   const renderPayments = useDeepCompareMemo(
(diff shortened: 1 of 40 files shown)


source/frontend/src/utils/dateUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+export const prettyFormatDateTime = (date?: string | Date | Moment | null) => {
+  return formatApiDateTime(date, 'MMM D, YYYY hh:mm a');
+};
+
+/**
+ * Format the passed string date assuming the date was recorded in UTC (as is the case with the pims API server).
+ * Returns a date formatted for display in the current time zone of the user.
+ * @param date utc date/time string.
+ */
+export const formatApiDateTime = (
```
FuriousLlama, line 24: nit: this might not be necessarily from the API. Instead a UTC date. so how about something like prettyUTCDateTime

source/frontend/src/features/leases/detail/LeasePages/payment/table/terms/TermsForm.tsx
```diff
@@ -50,7 +50,7 @@ export const TermsForm: React.FunctionComponent<React.PropsWithChildren<ITermsFo
     'receivedDate',
     'desc',
   );
-  const lastPaymentDate = allPayments.length ? prettyFormatDate(allPayments[0]?.receivedDate) : '';
```
FuriousLlama, line 53: good catch!

FuriousLlama (approved):

source/frontend/src/utils/dateUtils.ts
asanchezr: Sure - I can rename the function method but it was like that before. I agree the new name is better so will change

PR #3392, opened 2023-08-08
psp-6543: add missing fields error for acquisition file screens

buddy326:


devinleighsmith (changes requested):
All of these changes are against formik forms, so it seems odd that we would need to add state to track errors when formik contains an errors object that can be used to store/display errors. 

Can you please update this implementation to set a key/value pair in formik instead of in external state? likely against a common key at the root level of the errors object, maybe something like "global". That would dramatically simplify this change and reduce the amount of state required.

buddy326:
Ready for one final review and then good to be merged in.

devinleighsmith (approved):

PR #3394, opened 2023-08-08
PSP-6561 : Add "Final Date" to Compensation Requisition Input

eddherrera:
image


source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -73,12 +73,28 @@ public PimsCompensationRequisition Update(PimsCompensationRequisition compensati
...
+                    return currentValue;
+                }
+
+                if(statusComparable.newStatusIsDraft.HasValue)
+                {
+                    return statusComparable.newStatusIsDraft.Value ? null : DateTime.UtcNow;
+                }
+
+                return null;
+            }
```
asanchezr, line 97: question: is this a function within another function body? if so, I don't think we follow that pattern.

source/backend/dal/Helpers/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -72,10 +72,11 @@ public static IServiceCollection AddPimsDalRepositories(this IServiceCollection
             repositories.AddScoped<Repositories.IAgreementRepository, Repositories.AgreementRepository>();
             repositories.AddScoped<Repositories.ICompensationRequisitionRepository, Repositories.CompensationRequisitionRepository>();
             repositories.AddScoped<Repositories.IH120CategoryRepository, Repositories.H120CategoryRepository>();
-            repositories.AddScoped<Repositories.ICompReqFinancialRepository, Repositories.ICompReqFinancialRepository>();
+            repositories.AddScoped<Repositories.ICompReqFinancialRepository, Repositories.CompReqFinancialRepository>();
```
asanchezr, line 75: good catch!

asanchezr (approved):

source/backend/api/Services/CompensationRequisitionService.cs
eddherrera: yes, it's a local function. I like it because it's only useful to the method that contains it.

PR #3395, opened 2023-08-08
Fix dependency injection config

asanchezr:
image


asanchezr:
Fix is included in @eddherrera latest PR. Closing

PR #3396, opened 2023-08-09
Payee updates | psp-6686

FuriousLlama:


source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailView.test.tsx
```diff
@@ -124,29 +143,33 @@ describe('Compensation Detail View Component', () => {
   });
 
   it('User does not have the option to Edit Compensation when is in "FINAL" status', async () => {
-    const mockFinalCompensation = getMockApiFinalCompensation();
+    const mockFinalCompensation = getMockApiDefaultCompensation();
     const { queryByTitle } = await setup({
       claims: [Claims.COMPENSATION_REQUISITION_EDIT],
-      props: { compensation: mockFinalCompensation },
+      props: { compensation: { ...mockFinalCompensation, isDraft: false } },
```
FuriousLlama, line 149: test should be explicit about what conditions are being tested.

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -69,28 +52,27 @@ public PimsCompensationRequisition Update(PimsCompensationRequisition compensati
...
 
-            CheckDraftStatusUpdateAuthorized(compReqStatusComparable);
+            CheckDraftStatusUpdateAuthorized(currentCompensation.IsDraft, compensationRequisition.IsDraft);
             CheckTotalAllowableCompensation(compensationRequisition.AcquisitionFileId, compensationRequisition);
-            compensationRequisition.FinalizedDate = CheckFinalizedDate(compReqStatusComparable, currentCompensation.FinalizedDate);
+            compensationRequisition.FinalizedDate = CheckFinalizedDate(currentCompensation.IsDraft, compensationRequisition.IsDraft, currentCompensation.FinalizedDate);
 
             PimsCompensationRequisition updatedEntity = _compensationRequisitionRepository.Update(compensationRequisition);
-            AddNoteIfStatusChanged(compensationRequisition.Internal_Id, compensationRequisition.AcquisitionFileId, compReqStatusComparable);
+            AddNoteIfStatusChanged(compensationRequisition.Internal_Id, compensationRequisition.AcquisitionFileId, currentCompensation.IsDraft, compensationRequisition.IsDraft);
```
asanchezr, line 61: I think this is where @eddherrera mentioned that using that tuple would capture the value of "IsDraft" 4 lines above here so we don't end up using updated values. Notice this check is done after `_compensationRequisitionRepository.Update(compensationRequisition);` so there's a chance of the IsDraft being changed by the call to Update and then the note would not be added. Perhaps @eddherrera can explain better.

If you really don't like the tuple then we would need to capture the value of IsDraft in separate values to preserve their original values before the Update - e.g.

```csharp
  bool = currentIsDraft = currentCompensation.IsDraft;
  bool = newIsDraft = compensationRequisition.IsDraft;
  ...
  _compensationRequisitionRepository.Update(compensationRequisition);

  // after update
  AddNoteIfStatusChanged(..., ..., currentIsDraft, newIsDraft);
  ...
```

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -51,6 +51,8 @@ public PimsCompensationRequisition GetById(long compensationRequisitionId)
                 .Include(c => c.PimsCompReqFinancials)
                     .ThenInclude(y => y.FinancialActivityCode)
                 .Include(x => x.AcquisitionOwner)
+                .Include(x => x.AcquisitionFilePerson)
+                .Include(x => x.InterestHolder)
```
asanchezr, line 55: Github doesn't let me add comments to code that has not been changed so I will put it here but note I am talking about the next method in this file - **Update(...)**

- Shouldn't this PR also fix updating payee information for an existing compensation requisition?
- I see the TODO: fix this is still there....

source/frontend/src/components/maps/leaflet/Control/AdvancedFilter/FilterContentContainer.tsx
```diff
@@ -38,8 +38,6 @@ export const FilterContentContainer: React.FC<
   const onChange = useCallback(
     (model: PropertyFilterFormModel) => {
       filterProperties(model.toApi());
-      console.log('here', model);
-      console.log('here api', model.toApi());
```
asanchezr, line 42: good catch!

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailView.tsx
```diff
@@ -53,31 +56,37 @@ export const CompensationRequisitionDetailView: React.FunctionComponent<
...
-      payeeDetail.displayName = formatApiPersonNames(compensationPayee.motiSolicitor);
-      payeeDetail.personId = compensationPayee.motiSolicitor?.id!;
+    } else if (compensation.interestHolderId) {
+      if (compensationContactPerson) {
+        payeeDetail.displayName = formatApiPersonNames(compensationContactPerson);
+        payeeDetail.contactString = 's' + compensationContactPerson?.id;
+      } else if (compensationContactOrganization) {
+        payeeDetail.displayName = compensationContactOrganization?.name ?? '';
+        payeeDetail.contactString = 'p' + compensationContactOrganization.id;
+      }
```
asanchezr, line 86: If I understood correctly, here contactString can take the following values: "s1" or "p1" (assuming Id = 1)
But if I recall correctly, the convention for contact-based URLs are "P1" or "O1"  (for person or organization).

Could you please comment where the "s" and "p" are coming from?

asanchezr (approved):

PR #3397, opened 2023-08-09
Added Lease filters and anomalies to advanced property | psp-4845 & psp-6230

FuriousLlama:
image


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -381,6 +381,7 @@ public HashSet<long> GetMatchingIds(PropertyFilterCriteria filter)
         {
             var query = Context.PimsProperties.AsNoTracking();
```
devinleighsmith, line 382: @FuriousLlama I feel like I'm missing something obvious, how could this possibly be working if there are no includes on the PimsProperties query?

devinleighsmith:
minor display issue, could fix this as a bug due to time constraints: 
image

devinleighsmith:
Huh, we probably should have implemented a refresh button or clear button for the whole filter but I don't see that in the requirements or mockup so should be fine for this initial release.

devinleighsmith:
well, despite the missing includes everything works, so will approve. I must just be missing something.

devinleighsmith (approved):

source/backend/dal/Repositories/PropertyRepository.cs
FuriousLlama: Afaik this means, loading all properties. The individual blocks below make sure to add conditionals

PR #3399, opened 2023-08-09
Regenerated scaffold s_60

FuriousLlama:


source/backend/entities/ef/TmpPimsPropPropAdjacentLandType.cs
```diff
@@ -0,0 +1,21 @@
...
+    [Keyless]
+    [Table("TMP_PIMS_PROP_PROP_ADJACENT_LAND_TYPE")]
+    public partial class TmpPimsPropPropAdjacentLandType
+    {
+        [Column("PROPERTY_ID")]
+        public long? PropertyId { get; set; }
+        [Column("PROPERTY_ADJACENT_LAND_TYPE_CODE")]
+        [StringLength(50)]
+        public string PropertyAdjacentLandTypeCode { get; set; }
+    }
```
asanchezr, line 20: Just double-checking that we are keeping this temp table now in EF @FuriousLlama ??

asanchezr (approved):

source/backend/entities/ef/TmpPimsPropPropAdjacentLandType.cs
FuriousLlama: for now, yes, Next migration would be removing this and the corresponding table

PR #3407, opened 2023-08-11
PSP-6738 : Allow Contractor to Create Files

eddherrera:


source/backend/api/Areas/Acquisition/Controllers/AcquisitionFileController.cs
```diff
@@ -103,10 +104,17 @@ public IActionResult AddAcquisitionFile([FromBody] AcquisitionFileModel model, [
...
-            var acquisitionFile = _acquisitionService.Add(acqFileEntity, userOverrideCodes.Select(oc => UserOverrideCode.Parse(oc)));
+            try
+            {
+                var acqFileEntity = _mapper.Map<Dal.Entities.PimsAcquisitionFile>(model);
+                var acquisitionFile = _acquisitionService.Add(acqFileEntity, userOverrideCodes.Select(oc => UserOverrideCode.Parse(oc)));
 
-            return new JsonResult(_mapper.Map<AcquisitionFileModel>(acquisitionFile));
+                return new JsonResult(_mapper.Map<AcquisitionFileModel>(acquisitionFile));
+            }
+            catch(BusinessRuleViolationException ex)
```
devinleighsmith, line 114: @eddherrera this scares me a bit since normally business rule validation exceptions return a 400. I think this is a bit risky, as now other business rule violations within the service layer will get interpretated as a 409. Also the frontend will start displaying any response with a 409 in a toast where it may not be doing that currently.

Could we do something a bit safer here? maybe something like how the user override code gets returned from the backend?
`
else if (ex is UserOverrideException)
            {
                var exception = ex as UserOverrideException;
                code = HttpStatusCode.Conflict;
                message = exception.Message;
                errorCode = (ex as UserOverrideException).UserOverride?.Code;

                _logger.LogError(ex, "User override required to complete this action.");
            }
            `
or perhaps we just handle this issue on the frontend entirely? We should already know if a user is a contractor, should be easy to check if they are on the acq team of the file they are creating as well?

devinleighsmith (approved):

PR #3410, opened 2023-08-14
PSP-6475 : Generate Expropriation Form - Form 8

eddherrera:
image

image


source/frontend/src/models/generate/GenerateAddress.ts
```diff
@@ -43,8 +43,9 @@ export class Api_GenerateAddress {
       .filter(a => !!a)
       .join(', ');
 
-    this.address_String_multiline_compressed =
```
devinleighsmith, line 46: @eddherrera is this used in any other templates? If so does this require us to update those templates?

eddherrera: no, this is the only template so far that uses this format. I created this variation of the address so it's new.

devinleighsmith:
This should have been part of the previous pr that introduced this but it also affects this pr.
If an owner_string is an org, I'm guessing that Inc. No should only be included if inc. no is non-null:

Current logic:
    generateOwner.owner_string = `${generateOwner.last_name_or_corp_name}, Inc. No. ${generateOwner.incorporation_number}`;

@eddherrera

devinleighsmith (approved):

devinleighsmith:
This is fine actually, you can fix this with your next pr.

PR #3412, opened 2023-08-14
PSP-6739 : FT:Reg: User with limited access to region is not filterin…

eddherrera:
…g as expected in the L/L view screen


devinleighsmith (approved):

devinleighsmith:
@eddherrera there is a test failure here.

devinleighsmith (changes requested):
lets hold off on merging this, the fact that we had a unit test for this makes me nervous.

eddherrera:
> lets hold off on merging this, the fact that we had a unit test for this makes me nervous.

@devinleighsmith I think we can close this PR.

PR #3424, opened 2023-08-21
psp-6784: fix missing fields required message displays

buddy326:


buddy326:
A form component has state because me and Manuel both agreed that it was better than the solution of passing the save function as well as the isValid as props.

devinleighsmith (changes requested):
needs to wait for release.

source/frontend/src/features/mapSideBar/acquisition/add/AddAcquisitionForm.tsx
```diff
@@ -66,6 +66,7 @@ export const AddAcquisitionForm = React.forwardRef<
   };
 
   const handleSubmit = (values: AcquisitionForm, formikHelpers: FormikHelpers<AcquisitionForm>) => {
+    debugger;
```
devinleighsmith, line 69: @buddy326 please remove this.

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -60,14 +58,13 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
   onCancel,
-  onFooterSave,
-  missingFieldsError,
 }) => {
   const fiscalYearOptions = generateFiscalYearOptions();
   const { setModalContent, setDisplayModal } = useModalContext();
   const formikRef = useRef<FormikProps<CompensationRequisitionFormModel>>(null);
   const [activitiesUpdated, setActivitiesUpdated] = useState<boolean>(false);
   const [showModal, setShowModal] = useState(false);
+  const [isValid, setIsValid] = useState<boolean>(true);
```
devinleighsmith, line 67: not sure how we still end up with state outside of containers despite my best efforts in code review.

buddy326: As I said in my comment above, this was previously in the container but Manuel suggested that in this particular case it was less messy/convoluted to have the isValid variable declared here which I agreed with as well. Before, we were passing isValid and setIsValid as props and it made things less readable. Regardless, I will make the change :)

devinleighsmith:
@buddy326 is this a R4.0 story/bug? if so you'll need to update the PR to target test instead of dev.

buddy326:
@devinleighsmith this is a bug so I will update the PR

devinleighsmith (approved):

devinleighsmith (approved):

PR #3426, opened 2023-08-21
Activity cleanup

FuriousLlama:


source/backend/api/Services/LeaseService.cs
```diff
@@ -32,7 +32,7 @@ public class LeaseService : BaseService, ILeaseService
 
         public LeaseService(
             ClaimsPrincipal user,
-            ILogger<ActivityService> logger,
+            ILogger<LeaseService> logger,
```
asanchezr, line 35: good catch!

asanchezr (approved):

asanchezr:
@FuriousLlama approved but unit tests need to be fixed
image

PR #3430, opened 2023-08-22
Added tenure filters | psp-4804

FuriousLlama:
image


devinleighsmith:
Hmm, are we not unit testing any of this functionality?

devinleighsmith:
highway/road details filter does not appear to work.

devinleighsmith (changes requested):
highway/road details filter doesn't seem to do anything.

source/frontend/src/components/maps/leaflet/Control/AdvancedFilter/FilterContentForm.tsx
```diff
@@ -49,10 +49,28 @@ export const FilterContentForm: React.FC<React.PropsWithChildren<IFilterContentF
   }, [initialFilter, onChange]);
 
   const { getByType } = useLookupCodeHelpers();
+
+  // Propery options
```
asanchezr, line 53: typo: Propery -> Property

FuriousLlama:
> Hmm, are we not unit testing any of this functionality?

The filter is already being tested. I could add a test for creating the snap, but I am not sure how much value that would bring

PR #3433, opened 2023-08-23
PSP-6737 : FT-REG: Lease & Licenses - new date field looks misaligned…

eddherrera:
… on Deposit Returns modal


devinleighsmith (changes requested):
this is a r4.0 bug and should be fixed in test.

PR #3436, opened 2023-08-23
Generated scaffold from s61

FuriousLlama:


source/backend/entities/PimsBaseContext.cs
```diff
@@ -2088,6 +2086,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
                 entity.Property(e => e.AgreementDt).HasComment("Agreement date.");
 
+                entity.Property(e => e.AlternateProjectId).HasComment("Link a file to an \"Alternate Project\", so the user can make alternate payments that may be due after the original file's project closes.");
+
```
asanchezr, line 2090: nit: being picky here but it should say "Link a compensation requisition to an Alternate Project..." - probably copy/paste from file description...

asanchezr (approved):

source/backend/entities/PimsBaseContext.cs
FuriousLlama: probably

PR #3439, opened 2023-08-23
psp-6610: add alternate project field for comp reqs

buddy326:
Having a test failure in the new unit test having to do with the save button not accepting click events? Would appreciate some feedback there. Thanks!


buddy326:
Screenshot (28)
Screenshot (29)

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -53,6 +53,7 @@ public PimsCompensationRequisition GetById(long compensationRequisitionId)
                 .Include(x => x.AcquisitionOwner)
                 .Include(x => x.AcquisitionFilePerson)
                 .Include(x => x.InterestHolder)
+                .Include(x => x.AlternateProject)
```
FuriousLlama, line 56: this might not be necessary given that the front end can query the project by id.

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -321,6 +334,24 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
...
+            {formikRef.current?.values.alternateProject?.id === acquisitionFile.projectId &&
+              showError() && (
+                <GenericModal
+                  display={showAltProjectError}
+                  className="projectError"
+                  title="Alternate Project Error"
+                  message={[
+                    <strong>Error: </strong>,
+                    `You have selected an alternate project that is the same as the file project, please select a different project`,
+                  ]}
```
FuriousLlama, line 347: @devinleighsmith I thought we talked about this with dev and it was not necessary anymore. Can you confirm?

devinleighsmith: what we were talking about the other day was the converse - the user changes the primary project to be the same as the alternate project. I think there is still some merit to this case, although having an entire popup for it is probably overkill. However that is what the approved story states so fine for this PR.

source/frontend/src/models/api/CompensationRequisition.ts
```diff
@@ -8,11 +8,14 @@ import { Api_CompensationFinancial } from './CompensationFinancial';
 import { Api_ConcurrentVersion } from './ConcurrentVersion';
 import { Api_FinancialCode } from './FinancialCode';
 import { Api_InterestHolder } from './InterestHolder';
+import { Api_Project } from './Project';
 
 export interface Api_CompensationRequisition extends Api_ConcurrentVersion, Api_AuditFields {
   id: number | null;
   acquisitionFileId: number;
   acquisitionFile: Api_AcquisitionFile | null;
+  alternateProject?: Api_Project;
```
devinleighsmith, line 17: @buddy326 I am a little concerned that this PR seems to primarily interact with undefined instead of null, but then the model here is using a mixture of null and undefined. Can you please confirm if the api is sending undefined or null when the alternate project is not set and update the typing here and all of the checks as appropriate.

devinleighsmith: In general new controllers should be returning null instead of undefined - not sure where the compensation controllers fall on that issue.

buddy326: Ok I will take a look, thank you!

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/update/UpdateCompensationRequisitionForm.tsx
buddy326: The controller is returning nulls and I've adjusted the models accordingly but for some reason, the alternate project attribute is still being set as undefined?

devinleighsmith:
@buddy326 there appears to be a snapshot test failure now

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
eddherrera: It's necessary in order to display the Alternate Project information in the View for the comp req.

devinleighsmith (approved):

PR #3441, opened 2023-08-24
PSP-6494 : Acquisition List View Export (Excel)

eddherrera:
image

image


source/backend/api/Areas/Acquisition/Controllers/ReportController.cs
```diff
@@ -0,0 +1,71 @@
...
+    [Area("acquisitionfiles")]
+    [Route("v{version:apiVersion}/[area]/export")]
+    [Route("[area]/export")]
+    public class ReportController : ControllerBase
+    {
+        private readonly IAcquisitionFileService _acquisitionService;
+        private readonly ILogger _logger;
+
+        public ReportController(IAcquisitionFileService acquisitionService, ILogger<ReportController> logger)
+        {
```
asanchezr, line 32: @eddherrera - we have a separate folder for "reports" - we put them under `api/Areas/Reports/Controllers/AcquisitionController.cs`. Please consider moving it there (or merging with the one controller from Devin's PR once that gets merged

eddherrera: Ok, I'll merge after his.

eddherrera: relocated

devinleighsmith:
@eddherrera filename does not match AC
actual: pims-acquisition-files.xlsx
expected: Acquisition_File_Export.xlsx

devinleighsmith:
AlejandroSanchez, BobBillySmith

devinleighsmith:
Worksheet name does not match AC:
actual: PIMS_Acquisition_Files
expected: Acquisition File Export

devinleighsmith:
acquisition file number appears to be the id of the acquisition file, which is incorrect. Should be the file number aka: 
image

source/backend/api/Helpers/Middleware/ErrorHandlingMiddleware.cs
```diff
@@ -178,6 +178,15 @@ private async Task HandleExceptionAsync(HttpContext context, Exception ex)
 
                 _logger.LogError(ex, "Contractor User missing as a team member on the creation of Acquisition File");
             }
+            else if (ex is ExportHasNoDataException)
+            {
+                var exception = ex as ExportHasNoDataException;
+                code = HttpStatusCode.Conflict;
```
devinleighsmith, line 184: would a 404 make more sense? that is what we throw when a user requests an id that does not exist.

source/backend/dal/Helpers/Extensions/AddressExtensions.cs
```diff
@@ -17,5 +19,48 @@ public static string FormatAddress(this PimsAddress address, bool includeMunicip
...
+            if (!string.IsNullOrEmpty(address.StreetAddress2))
+            {
+                stringBuilder.Append(" " + address.StreetAddress2);
+            }
+            if (!string.IsNullOrEmpty(address.StreetAddress3))
+            {
+                stringBuilder.Append(" " + address.StreetAddress3);
+            }
+
+            stringBuilder.Append(" " + address.MunicipalityName);
```
devinleighsmith, line 46: why no if statement check here? is this a required field in the DB?

source/backend/dal/Helpers/Extensions/PersonExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public static string FormatName(this PimsPerson pimsPerson)
+        {
+            StringBuilder stringBuilder = new();
+
+            stringBuilder.Append(pimsPerson.FirstName);
+
+            if (!string.IsNullOrEmpty(pimsPerson.MiddleNames))
+            {
+                stringBuilder.Append(pimsPerson.MiddleNames);
```
devinleighsmith, line 16: see my other comment, seems to be missing spaces.

source/backend/dal/Models/AcquisitionFileExportDto.cs
```diff
@@ -0,0 +1,61 @@
...
+        [DisplayName("General Location")]
+        public string GeneralLocation { get; set; }
+
+        [DisplayName("PID")]
+        public string Pid { get; set; }
+
+        [DisplayName("PIN")]
+        public string Pin { get; set; }
+
+        [DisplayName("Status(File)")]
```
devinleighsmith, line 34: nit (spacing)

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -62,6 +64,122 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter, HashSet<shor
...
         }
 
+        /// <summary>
+        /// Get an Acquisition Files Export based on Filter.
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <param name="regions"></param>
+        /// <param name="filterPersonId"></param>
+        /// <returns>List of AcquisitionFileExportDto.</returns>
+        public List<AcquisitionFileExportDto> GetAcquisitionFileExport(AcquisitionFilter filter, HashSet<short> regions, long? filterPersonId = null)
```
devinleighsmith, line 74: This seems dangerous, as we now have two filters, one that returns data for the acquisition file list view, and this one that returns data for the export. there is no guarantee that the two will return the same data, or remain in sync in the case of future updates. 

Can you explain why we need this function instead of just re-using the existing acquisition search that populates the list view? even if we don't reuse that entire repository method could we at least extract out the database query part to make sure they are the same?

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
```diff
@@ -0,0 +1,50 @@
...
+      try {
+        const { data, status } = await apiExportAcquisitionFiles(filter, outputFormat);
+        dispatch(logSuccess({ name: requestId, status }));
+        dispatch(hideLoading());
+        // trigger file download in client browser
+        fileDownload(data, fileName);
+      } catch (e) {
+        if (axios.isAxiosError(e)) {
+          const axiosError = e as AxiosError<IApiError>;
+          console.log(axiosError);
```
devinleighsmith, line 37: console.log

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
```diff
@@ -0,0 +1,50 @@
...
+import { logRequest, logSuccess } from '@/store/slices/network/networkSlice';
+
+export const useAcquisitionFileExport = () => {
+  const { exportAcquisitionFiles: apiExportAcquisitionFiles } = useApiAcquisitionFile();
+  const dispatch = useDispatch();
+
+  const exportAcquisitionFiles = useCallback(
+    async (
+      filter: IPaginateAcquisition,
+      outputFormat: 'csv' | 'excel' = 'excel',
```
devinleighsmith, line 22: I don't think that csv is supported here so can probably simplify this.

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
```diff
@@ -0,0 +1,50 @@
...
+        fileDownload(data, fileName);
+      } catch (e) {
+        if (axios.isAxiosError(e)) {
+          const axiosError = e as AxiosError<IApiError>;
+          console.log(axiosError);
+          if (axiosError?.response?.status === 409) {
+            toast.error('Export contains no data');
+          }
+          dispatch(hideLoading());
+          //catchAxiosError(e, dispatch, actionTypes.DELETE_PARCEL);
```
devinleighsmith, line 42: ?

source/frontend/src/hooks/pims-api/useApiAcquisitionFile.ts
```diff
@@ -42,6 +42,19 @@ export const useApiAcquisitionFile = () => {
             Accept: 'application/vnd.ms-excel',
           },
         }),
+      exportAcquisitionFiles: (
+        filter: IPaginateAcquisition,
+        outputFormat: 'csv' | 'excel' = 'excel',
```
devinleighsmith, line 47: same comment here regarding csv, pretty sure this will throw an error from the backend if you request csv.

source/backend/dal/Helpers/Extensions/PersonExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using System.Text;
+using Pims.Dal.Entities;
+
+namespace Pims.Dal.Helpers.Extensions
+{
+    public static class PersonExtensions
+    {
+        public static string FormatName(this PimsPerson pimsPerson)
```
devinleighsmith, line 8: is this necessary give that GetFullName extension method already exists?

source/backend/dal/Helpers/Extensions/AcquisitionOwnerExtensions.cs
```diff
@@ -0,0 +1,27 @@
+using System.Text;
+using Pims.Dal.Entities;
+
+namespace Pims.Dal.Helpers.Extensions
+{
+    public static class AcquisitionOwnerExtensions
+    {
+        public static string FormatOwnerName(this PimsAcquisitionOwner pimsOwner)
+        {
+            StringBuilder stringBuilder = new();
```
devinleighsmith, line 10: Personally I prefer this syntax for conciseness/readibility, what do you think?

string[] names = { person.FirstName, person.MiddleNames, person.Surname };
return string.Join(" ", names.Where(n => n != null && n.Trim() != string.Empty));

devinleighsmith (changes requested):
I'm mostly concerned about not re-using the acquisition list view filter method, the two filters should return the same results.

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
eddherrera: removed

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
eddherrera: simplified,

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
eddherrera: fixed

source/backend/dal/Helpers/Extensions/PersonExtensions.cs
eddherrera: the existing extension method doesn't include Suffix

eddherrera:
> I'm mostly concerned about not re-using the acquisition list view filter method, the two filters should return the same results.

Updated with a new private static method in the repository. also other comments addressed.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -52,14 +54,98 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter, HashSet<shor
...
+                              from fp in filePropGJ.DefaultIfEmpty()
+                              join property in Context.PimsProperties
+                                .Include(a => a.Address)
+                                    .ThenInclude(pr => pr.ProvinceState)
+                                .Include(a => a.Address)
+                                    .ThenInclude(pr => pr.Country)
+                                on fp.PropertyId equals property.PropertyId into propGJ
+                              from prop in propGJ.DefaultIfEmpty()
+                              orderby acqFile.AcquisitionFileId, prop.PropertyId
+                              select new AcquisitionFileExportDto()
```
devinleighsmith, line 116: I don't know what others think @asanchezr or @FuriousLlama but I can't get behind this syntax. This also seems to be conflating getting data into entities with transforming entities into a display format all within a repository which seems like a violation of separation of concerns.
I would prefer to see:
1) controller handles formatting models, validating requests and returning http responses
2) service handles composing repo calls and responses, business logic, and authorization.
3) repo handles getting data from database or external services in whatever format that db/service returns that data in.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
asanchezr: I agree that we should try to use the existing "GetPage" functionality of acquisition list view. If we wanted to get all records instead of pages we could send `PageSize = Int.MaxValue` (or whatever C# constant is for max integer). 

I don't see the need for a separate function and as Devin mentions I would be concerned about the two functions not returning the same data

source/backend/dal/Repositories/AcquisitionFileRepository.cs
asanchezr: Agreed. I would prefer that the Entity -> DTO conversion is done at the Controller level. The repo should always return DB entities to keep separation of concerns. 
Also a minor note on the export DTO name - we usually do not use "DTO" in their names (even if they are DTOs in purpose). A more fitting name would be **AcquisitionFileExportModel**

asanchezr: Forgot to mention that I don't think it would be necessary to create a Mapster mapping for this export model conversion. If you look at Devin's PR you can pass the entity in the constructor and create an export model from it. I am using that pattern as well in my export H120 story. 

It would make sense to use mappings for things that are meant to be shared/composed. These "export" models are one-offs so no need to create mappings for them (in my opinion)

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
```diff
@@ -0,0 +1,52 @@
...
+      } catch (e) {
+        if (axios.isAxiosError(e)) {
+          const axiosError = e as AxiosError<IApiError>;
+          if (axiosError?.response?.status === 409) {
+            toast.error('Export contains no data');
+          } else {
+            catchAxiosError(e, dispatch, actionTypes.DELETE_PARCEL);
+          }
+          dispatch(hideLoading());
+        }
```
asanchezr, line 45: A couple issues I see here:
- I would question why are we triggering the download in the "use..." hook. I'd rather just return the raw blob data and let the consumer of the hook decide what to do with it. Please see Devin's PR for an example. I'm doing the same in my export PR. This hook should be concerned with calling the API and returning the response - not with also triggering the file save popup.
- I see this on line 42 `catchAxiosError(e, dispatch, actionTypes.DELETE_PARCEL);` - is this copy/paste? This hook is not related to parcels at all...

asanchezr: One more comment. I also don't think this new functionality warrants a separate "use..." hook. The method **exportAcquisitionFiles** could be added to the **useAcquisitionProvide** hook. The functionality is sufficiently relevant to that provider that it should go there instead of a separate hook with just one function

source/backend/api/Helpers/Middleware/ErrorHandlingMiddleware.cs
asanchezr: I also question the need for another exception. To me 404 would be sufficient. Or another option I am pondering on my PR is to use a non-error status code **204 No Content** - because in theory the API shouldn't be concerned that the frontend wants to display a warning if no results are found. 
In my opinion, the API should not return an error code because no error happened, it just didn't find any matching records. 

What do you think of using 204 instead of 400/404 here @eddherrera and @devinleighsmith ?

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
eddherrera: This is the patter used for exporting leases. Not the first time I get comments for using patterns used in other sections of the code.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
eddherrera: Should I move the query to the properties repository since the query is returning one row per property in the Acquisition file?

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
asanchezr: Sorry if my comments came in as a bit harsh - totally not my intention. As I see it, PR comments are more "suggestions" than anything (unless it is some glaring error in the code). In my view, separating the download from the useXXX hook makes it more re-usable. Also, leases is the oldest portion of the codebase so the patterns used there might have been refined lately

source/backend/api/Areas/Reports/Controllers/AcquisitionController.cs
```diff
@@ -74,6 +83,44 @@ public IActionResult ExportLeases([FromBody] AcquisitionReportFilterModel filter
...
+            }
+
+            var acceptHeader = (string)Request.Headers["Accept"];
+            if (acceptHeader != ContentTypes.CONTENTTYPEEXCEL && acceptHeader != ContentTypes.CONTENTTYPEEXCELX)
+            {
+                throw new BadRequestException($"Invalid HTTP request header 'Accept:{acceptHeader}'.");
+            }
+
+            var acquisitionFileData = _acquisitionFileService.GetAcquisitionFileExport((AcquisitionFilter)filter);
+            if (acquisitionFileData.Count.Equals(0))
```
asanchezr, line 117: nit: not sure if we would need to check for acquisitionFileData to not be null?

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -87,6 +87,44 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter)
             return _acqFileRepository.GetPage(filter, userRegions, personId);
         }
 
+        public List<AcquisitionFileExportModel> GetAcquisitionFileExport(AcquisitionFilter filter)
+        {
+            _logger.LogInformation("Searching for acquisition files...");
+            _logger.LogDebug("Acquisition file search with filter", filter);
```
asanchezr, line 93: nit: I think you need to include a placeholder in the log statement - i.e.

```csharp
_logger.LogDebug("Acquisition file search with filter: {filter}", filter);
```

eddherrera: Agreements doesn't has one...

image

Now that I think about it, shouldn't it be ??

`_logger.LogDebug("Acquisition file search with filter: {0}", filter);  `

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -52,14 +55,34 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter, HashSet<shor
                 throw new ArgumentException("Argument must have a valid filter", nameof(filter));
             }
 
+            IQueryable<PimsAcquisitionFile> query = GetCommonAquisitionFileQuery(filter, regions, filterPersonId);
+
+            var queryResult = query.ToList();
```
asanchezr, line 60: nit: do we want to call ToList() here? That would bring the whole list into memory (instead of one page of results)...

The old code worked by doing the Skip, Take on the original query, and then calling ToList() - could we do the same here?

source/backend/api/Areas/Reports/Controllers/AcquisitionController.cs
eddherrera: it doesn't seem so. It just returns an empty list instead of a null.

source/backend/api/Services/AcquisitionFileService.cs
eddherrera: updated

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -1,9 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.Drawing;
```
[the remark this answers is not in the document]
eddherrera: removed

source/backend/dal/Repositories/AcquisitionFileRepository.cs
eddherrera: updated

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -87,6 +87,42 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter)
...
+                                    FileName = fileProperty.file.FileName ?? string.Empty,
+                                    MotiRegion = fileProperty.file.RegionCodeNavigation?.Description ?? string.Empty,
+                                    MinistryProject = fileProperty.file.Project != null ? $"{fileProperty.file.Project.Code} {fileProperty.file.Project.Description}" : string.Empty,
+                                    CivicAddress = (fileProperty.fp?.Property != null && fileProperty.fp.Property.Address != null) ? fileProperty.fp.Property.Address.FormatFullAddressString() : string.Empty,
+                                    GeneralLocation = (fileProperty.fp?.Property != null) ? fileProperty.fp.Property.GeneralLocation : string.Empty,
+                                    Pid = fileProperty.fp is not null && fileProperty.fp.Property.Pid.HasValue ? fileProperty.fp.Property.Pid.ToString() : string.Empty,
+                                    Pin = fileProperty.fp is not null && fileProperty.fp.Property.Pin.HasValue ? fileProperty.fp.Property.Pin.ToString() : string.Empty,
```
asanchezr, line 113: **nit:** I see mixed usage of `!= null` and `xyz is not null` - would appreciate sticking to one pattern

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -87,6 +87,42 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter)
...
+                                    FileDeliveryDate = fileProperty.file.DeliveryDate.HasValue ? fileProperty.file.DeliveryDate.Value.ToString("dd-MMM-yyyy") : string.Empty,
+                                    FileAcquisitionCompleted = fileProperty.file.CompletionDate.HasValue ? fileProperty.file.CompletionDate.Value.ToString("dd-MMM-yyyy") : string.Empty,
+                                    FilePhysicalStatus = fileProperty.file.AcqPhysFileStatusTypeCodeNavigation != null ? fileProperty.file.AcqPhysFileStatusTypeCodeNavigation.Description : string.Empty,
+                                    FileAcquisitionType = fileProperty.file.AcquisitionTypeCodeNavigation != null ? fileProperty.file.AcquisitionTypeCodeNavigation.Description : string.Empty,
+                                    FileAcquisitionTeam = string.Join(", ", fileProperty.file.PimsAcquisitionFilePeople.Select(x => x.Person.GetFullName(true))),
+                                    FileAcquisitionOwners = string.Join(", ", fileProperty.file.PimsAcquisitionOwners.Select(x => x.FormatOwnerName())),
```
asanchezr, line 122: I think there's potential for NPE errors by not checking for nulls on a few properties here:

- fileProperty.file.AcquisitionFileStatusTypeCodeNavigation.Description
- fileProperty.file.PimsAcquisitionFilePeople.Select(...)
- fileProperty.file.PimsAcquisitionOwners.Select(...)

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
```diff
@@ -0,0 +1,51 @@
...
+          );
+        } else {
+          dispatch(logSuccess({ name: requestId, status }));
+          fileDownload(data, fileName);
+          dispatch(hideLoading());
+        }
+        // trigger file download in client browser
+      } catch (axiosError) {
+        if (axios.isAxiosError(axiosError)) {
+          catchAxiosError(axiosError, dispatch, actionTypes.DELETE_PARCEL);
```
asanchezr, line 43: not sure DELETE_PARCEL action type is relevant here - what action should we use instead?

source/frontend/src/features/acquisition/list/AcquisitionListView.tsx
```diff
@@ -48,6 +54,23 @@ export const AcquisitionListView: React.FunctionComponent<
     'No matching results can be found. Try widening your search criteria.',
   );
 
+  const { exportAcquisitionFiles } = useAcquisitionFileExport();
+
+  /**
+   * @param {'csv' | 'excel'} accept Whether the fetch is for type of CSV or EXCEL
+   */
```
asanchezr, line 61: I think there was a previous comment about removing all references to CSV

source/frontend/src/hooks/pims-api/useApiAcquisitionFile.ts
```diff
@@ -42,6 +42,19 @@ export const useApiAcquisitionFile = () => {
...
+      ) =>
+        api.get<Blob>(
+          `/reports/acquisition?${filter ? queryString.stringify({ ...filter, all: true }) : ''}`,
+          {
+            responseType: 'blob',
+            headers: {
+              Accept: outputFormat === 'csv' ? 'text/csv' : 'application/vnd.ms-excel',
+            },
```
asanchezr, line 55: csv format not supported by our export endpoints (currently)

asanchezr (approved):

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -87,6 +87,42 @@ public Paged<PimsAcquisitionFile> GetPage(AcquisitionFilter filter)
...
+            var acqFiles = _acqFileRepository.GetAcquisitionFileExport(filter, userRegions, personId);
+
+            return acqFiles.SelectMany(file => file.PimsPropertyAcquisitionFiles.Where(fp => fp.AcquisitionFileId.Equals(file.AcquisitionFileId)).DefaultIfEmpty(), (file, fp) => (file, fp))
+                                .Select(fileProperty => new AcquisitionFileExportModel
+                                {
+                                    FileNumber = fileProperty.file.FileNumber ?? string.Empty,
+                                    LegacyFileNumber = fileProperty.file.LegacyFileNumber ?? string.Empty,
+                                    FileName = fileProperty.file.FileName ?? string.Empty,
+                                    MotiRegion = fileProperty.file.RegionCodeNavigation?.Description ?? string.Empty,
+                                    MinistryProject = fileProperty.file.Project != null ? $"{fileProperty.file.Project.Code} {fileProperty.file.Project.Description}" : string.Empty,
```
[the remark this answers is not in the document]
eddherrera: question raised to Julian. Changes should be handled with new story if needed.

source/backend/api/Services/AcquisitionFileService.cs
eddherrera: updated

source/backend/api/Services/AcquisitionFileService.cs
eddherrera: Added check for status. each item is checked and collection are initialized as empty list from the parent constructor.

source/frontend/src/features/acquisition/hooks/useAcquisitionFileExport.ts
eddherrera: changed to log error.

source/frontend/src/features/acquisition/list/AcquisitionListView.tsx
eddherrera: removed

source/frontend/src/hooks/pims-api/useApiAcquisitionFile.ts
eddherrera: removed

devinleighsmith (approved):

PR #3445, opened 2023-08-25
psp-6803 ensure that nip/ip interests are separated correctly.

devinleighsmith:


asanchezr (approved):

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderContainer.tsx
```diff
@@ -36,12 +36,14 @@ export const StakeHolderContainer: React.FunctionComponent<IStakeHolderContainer
...
     apiInterestHolders?.flatMap(interestHolder => interestHolder.interestHolderProperties) ?? [];
-  const interestProperties = allInterestProperties.filter(ip =>
-    ip.propertyInterestTypes.some(pit => pit?.id !== 'NIP'),
-  );
-  const nonInterestProperties = allInterestProperties.filter(ip =>
-    ip.propertyInterestTypes.some(pit => pit?.id === 'NIP'),
-  );
+  const interestProperties = allInterestProperties.map(ip => {
+    const filteredTypes = ip.propertyInterestTypes.filter(pit => pit?.id !== 'NIP');
+    return { ...ip, propertyInterestTypes: filteredTypes };
```
FuriousLlama, line 41: would this not give all properties but some with propertyInterestType set to undefined?

devinleighsmith: mmm, I think you are correct actually, works in the case of a property with both types, but not with just one, need to combine this with previous implementation.

PR #3449, opened 2023-08-29
4.0 acquisition, codes and pims etl corrections.

devinleighsmith:
ensure comp req codes are mapped correctly
    ensure that '55' is respected uniformly on code values
    interest holders and subtables now mapped
    pims land parcels all imported now
    owner reps and solicitors corrected.


devinleighsmith:
based on final, validated fixes performed by Sue.

asanchezr (approved):

PR #3456, opened 2023-08-31
psp-6615: add alternate project to generated comp req

buddy326:


source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/update/models.ts
```diff
@@ -79,6 +81,13 @@ export class CompensationRequisitionFormModel {
     compensation.serviceLine = apiModel.chartOfAccountsId?.toString() || '';
     compensation.chartOfAccounts = apiModel.chartOfAccounts;
     compensation.responsibilityCentre = apiModel.responsibilityId?.toString() || '';
+    compensation.alternateProject =
+      apiModel.alternateProject !== null
+        ? {
+            id: apiModel.alternateProject?.id || 0,
```
devinleighsmith, line 87: this should probably be changed, if the alternate project id is not set we should not be setting the alternate project.

devinleighsmith (changes requested):
see comment.

devinleighsmith:
Also has a conflict, You'll likely need to open a new PR with these changes @asanchezr

PR #3472, opened 2023-09-12
psp-6836 ensure boundary added when property location is added.

devinleighsmith:


source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -488,20 +488,29 @@ private static void ValidateStaff(PimsAcquisitionFile pimsAcquisitionFile)
...
                     if (geom.SRID != SpatialReference.BCALBERS)
                     {
                         var newCoords = _coordinateService.TransformCoordinates(geom.SRID, SpatialReference.BCALBERS, geom.Coordinate);
                         propertyToUpdate.Location = GeometryHelper.CreatePoint(newCoords, SpatialReference.BCALBERS);
                         _propertyRepository.Update(propertyToUpdate, overrideLocation: true);
                     }
+
+                    // apply similar logic to the boundary
+                    var boundaryGeom = acquisitionProperty.Boundary;
+                    if (boundaryGeom != null && boundaryGeom.SRID != SpatialReference.BCALBERS)
```
devinleighsmith, line 508: at some point it would be nice to cut down on the code duplication between these services - it seems that research, lease and acquisition have essentially copied the same property manipulation methods.

devinleighsmith:
Tested by temporarily displaying the boundaries in the PSP database as a layer. Was able to see the newly added property's boundary on the map. Indicating that the boundary was added with the correct srid and there was no loss of precision.

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -123,7 +123,6 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
   };
 
   const onMinistryProjectSelected = async (param: IAutocompletePrediction[]) => {
-    debugger;
```
asanchezr, line 126: thanks for catching this

asanchezr (approved):

PR #3487, opened 2023-09-20
Property Contacts | psp-6825 & PSP-5745

FuriousLlama:
image

image


devinleighsmith:
the property contact section should be collapseable, will address this as a follow up bug.

devinleighsmith:
The primary contact logic is not working properly, the primary contact should 
1) auto select the contact if there is only one available person for an org.
2) not display primary contact at all if the contact is a person (currently it displays if you select a person with an org).

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/PropertyContactEditForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export interface IPropertyContactEditFormProps {
+  isLoading: boolean;
+  propertyContact: Api_PropertyContact;
+  onSave: (apiModel: Api_PropertyContact) => void;
+}
+
+export const PropertyContactEditForm = React.forwardRef<
+  FormikProps<any>,
+  IPropertyContactEditFormProps
```
devinleighsmith, line 29: I think this needs a yup schema otherwise:
1) the purpose is supposed to be required but that isn't being enforced
2) the purpose has a character limit that should also be enforced.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/PropertyContactEditForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const [contact, setContact] = useState<IContactSearchResult | null>(null);
+
+  React.useEffect(() => {
+    if (contact?.organizationId) {
+      fetchOrganization(contact.organizationId);
+    }
+  }, [contact, fetchOrganization]);
+
+  const initialForm = useMemo(() => {
+    const initialmodel = PropertyContactFormModel.fromApi(props.propertyContact);
```
devinleighsmith, line 48: nit: initialmodel

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/PropertyContactEditForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+>((props, ref) => {
+  const {
+    getOrganizationDetail: {
+      execute: fetchOrganization,
+      response: organization,
+      loading: getOrganizationLoading,
+    },
+  } = useOrganizationRepository();
+
+  const [contact, setContact] = useState<IContactSearchResult | null>(null);
```
devinleighsmith, line 39: I'm not a big fan of this and the repo call being in this form component. Container components should be responsible for state and data fetching as per our best practices and for ease of testing.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/models.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  static fromApi(model: Api_PropertyContact | null): PropertyContactFormModel {
+    const newFormModel = new PropertyContactFormModel();
+    newFormModel.id = model?.id || 0;
+    newFormModel.propertyId = model?.propertyId || 0;
+    newFormModel.isDisabled = model?.isDisabled || false;
+    newFormModel.rowVersion = model?.rowVersion || 0;
+    if (model?.organization) {
+      newFormModel.contact = fromApiOrganization(model.organization);
+    } else if (model?.person) {
```
devinleighsmith, line 38: I'm not sure if this is safe, right now it seems the api is returning the organization of a person whenever a person is selected.

source/frontend/src/utils/personUtils.ts
```diff
@@ -15,3 +16,16 @@ export function formatApiPersonNames(person?: Api_Person | null): string {
 export function formatNames(nameParts: Array<string | undefined | null>): string {
   return nameParts.filter(n => n !== null && n !== undefined && n.trim() !== '').join(' ');
 }
+
+export function formatContactSearchResult(
```
devinleighsmith, line 20: perhaps better to live in contactUtils? this seems non person specific.

devinleighsmith:
saving returns the user to the details screen instead of the management screen.

devinleighsmith:
This PR doesn't contain any unit tests, will split those out into separate stories. We at least need a minimal set of unit tests to meet the outline in our best practices.

devinleighsmith (approved):
granting conditional approval, this will require some fixes early next week due to a number of known issues.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/PropertyContactEditForm.tsx
FuriousLlama: I agree. However we need a organization primary contact component that takes care of that abstraction similar to the contact picker

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/update/models.ts
FuriousLlama: fair point

FuriousLlama: I checked and this should be fine given the model is a property contact, and that can only be either a Person or organization but not both.

FuriousLlama:
> formatContactSearchResult

This is something that could wait until the refactor of the property tab routing. At this point, the property container is defaulting to a tab

FuriousLlama:
> The primary contact logic is not working properly, the primary contact should
> 
> 1. auto select the contact if there is only one available person for an org.
> 2. not display primary contact at all if the contact is a person (currently it displays if you select a person with an org).

That is not on the acceptance criteria or in the mapping. Is that the intention for the primary contacts?

PR #3488, opened 2023-09-21
PSP-6865 property management models

asanchezr:
First PR with basic backend+ frontend models. Actual edit functionality to come in a separate PR


source/backend/api/Areas/Property/Controllers/PropertyController.cs
```diff
@@ -62,6 +65,23 @@ public IActionResult GetPropertyAssociationsWithId(long id)
...
+        /// Get all associated leases for the specified unique property 'id'.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("{id}/leases")]
+        [HasPermission(Permissions.PropertyView)]
+        [HasPermission(Permissions.LeaseView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(IEnumerable<Api.Models.Concepts.PropertyLeaseModel>), 200)]
+        [SwaggerOperation(Tags = new[] { "property" })]
+        public IActionResult GetPropertyLeases(long propertyId)
```
devinleighsmith, line 79: I think we need to confirm that this approach is appropriate given that is requires LeaseView. That means that a user that has management view will get an error unless they also have lease-view when viewing the management summary.

We may need to return just the boolean field against the property calculated on the backend with no lease restrictions instead.

devinleighsmith: @asanchezr

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/detail/ManagementSummaryView.tsx
```diff
@@ -0,0 +1,36 @@
...
+  isLoading,
+  property,
+  propertyLeases,
+  setEditMode,
+}) => {
+  const { hasClaim } = useKeycloakWrapper();
+  return (
+    <Section header="Summary">
+      <StyledEditWrapper className="mr-3 my-1">
+        {/** TODO: Use MANAGEMENT CLAIMS when available */}
```
devinleighsmith, line 26: @asanchezr the management claims are available, can you add those please?

devinleighsmith:
The summary section should be collapseable.

source/backend/api/Areas/Property/Controllers/PropertyController.cs
asanchezr: I see, as said in the PR description this is the basic model stuff. I can change that in my next PR

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/detail/ManagementSummaryView.tsx
asanchezr: They are available now - not when I sent the PR originally. Did you add them to the frontend enum as well or do I need to do that?

asanchezr:
> The summary section should be collapseable.

will fix in next pr

devinleighsmith (approved):

devinleighsmith:
@asanchezr feel free to merge this as long as you can finish the other PR before EOD. If not both stories will need to move to IS64

asanchezr:
> @asanchezr feel free to merge this as long as you can finish the other PR before EOD. If not both stories will need to move to IS64

Oh - I just merged it. Working on the rest. Given the hours left not sure if I can finish by EOD. If not done I will have to revert this PR

asanchezr:
> @asanchezr feel free to merge this as long as you can finish the other PR before EOD. If not both stories will need to move to IS64

Also, not sure why we would need to revert the PR as it doesn't contain anything breaking but can revert if needed

PR #3491, opened 2023-09-22
introduce keycloak to dev test ci/cd pipelines as the action appears …

devinleighsmith:
…stable. Will need alternate action for uat prod that runs in openshift.


.github/workflows/ci-cd-pims-dev.yml
```diff
@@ -98,10 +102,55 @@ jobs:
           ./openshift/4.0/player.sh deploy api $DESTINATION -apply
           ./openshift/4.0/player.sh deploy app $DESTINATION -apply
 
+  # the command:
+  # 1) creates an openshift job with generated name to avoid name conflict, substituting the variables in the template.
+  # 2) greps the generated name from the previous step.
+  # 3) waits for the job to complete using the generated name.
+  database-upgrade:
```
devinleighsmith, line 109: Note that while I believe that we will eventually want to roll this out to our other environments, I'd be more comfortable testing the stability of this in the dev environment first given how high traffic it is. If this and the keycloak sync are stable in dev, they should be stable in our manual-only environments with less frequent deployments.

asanchezr (approved):

PR #3492, opened 2023-09-25
psp-6864 missing fields required message.

devinleighsmith:


source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.tsx
```diff
@@ -57,7 +58,8 @@ export const MotiInventoryContainer: React.FunctionComponent<
   const handleSaveClick = async () => {
     if (formikRef !== undefined) {
       formikRef.current?.setSubmitting(true);
-      formikRef.current?.submitForm();
+      await formikRef.current?.submitForm();
+      setIsValid(formikRef.current?.isValid ?? false);
```
asanchezr, line 62: I see other places doing: 
```tsx
var result = await (formikRef.current?.submitForm() ?? Promise.resolve());
setIsValid(...)
```
why not use the same pattern here (awaiting the promise)?

source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
```diff
@@ -117,6 +118,7 @@ export const ResearchContainer: React.FunctionComponent<
     if (formikRef !== undefined) {
       formikRef.current?.setSubmitting(true);
       formikRef.current?.submitForm();
+      setIsValid(formikRef.current?.isValid || false);
```
asanchezr, line 121: Other code awaits the result of `submitForm()` but this one doesn't. Would it make sense to apply the same pattern here?

asanchezr (approved):

PR #3501, opened 2023-10-04
PSP-6945 : Access to PIMS Overview (Help File) - Header

eddherrera:
image


devinleighsmith:
Hey Eduardo, I think you may have missed the mockup for psp-6945:

image

source/frontend/src/features/help/components/HelpModal.tsx
```diff
@@ -57,31 +58,44 @@ const HelpModal: FunctionComponent<React.PropsWithChildren<IModalProps>> = ({
...
-
         <Modal.Body>
+          <H3Styled>Get started at PIMS</H3Styled>
+          <p>
+            This overview has useful tools that will support you to start using the application. You
+            can also watch the video demos.
+          </p>
+          <LinkStyled
+            target="_blank"
+            href="https://sp.th.gov.bc.ca/sites/PropertiesServices/Project%20Management/03.%20Execution%20and%20Control/PIMS%20Training"
```
devinleighsmith, line 73: I think we discussed this, but I assume this will be replaced with the actual tenant/openshift configmap based url so that we can modify this after deployment without needing a code change?

source/frontend/src/features/help/containers/HelpModalContentContainer.tsx
```diff
@@ -1,73 +1,44 @@
...
-      <hr></hr>
-      <h5>Submit</h5>
+      <H3Styled>Contact us:</H3Styled>
       <HelpSubmitBox
         user={displayName ?? ''}
         email={email ?? ''}
-        activeTicketType={activeTicketType}
-        setActiveTicketType={setActiveTicketType}
-        page={helpPage?.name ?? location.pathname}
+        page={location.pathname}
```
devinleighsmith, line 31: do we still need this? I feel like help is no longer context sensitive.

devinleighsmith:
@eddherrera submitting the help form seems bugged, subject and body both coming through as undefined now.

devinleighsmith (changes requested):
mailto action is not working anymore, this is a regression.

devinleighsmith:
@eddherrera we have another story to update modal styling but not one to update header styling. I'll leave it up to you if you'd rather
1) update the header to closer match the mockup: (divider line looks too thick, icon style and size do not match)
-OR-
2) log a story to handle the above.

devinleighsmith:
@eddherrera also, I'm still seeing undefined as the subject and body when clicking yes on the help modal, that needs to be fixed.

source/frontend/src/features/help/components/HelpModal.tsx
eddherrera: Updated

source/frontend/src/features/help/containers/HelpModalContentContainer.tsx
eddherrera: I am passing the path as part of the email subject. should I remove it alltogether?

devinleighsmith (approved):

PR #3503, opened 2023-10-05
Research header last-updated-by | PSP-6930

FuriousLlama:


devinleighsmith:
Hmm, adding a note doesn't seem to cause an update on the research file. Trying other entity types

devinleighsmith:
Same with documents, the original last update time is still being returned. let me make sure I have the code and everything.

devinleighsmith:
Property doesn't seem to update the time either, looking a the code.

devinleighsmith:
@FuriousLlama nevermind, was looking in the wrong place.

source/backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -138,6 +139,120 @@ public PimsResearchFile Update(PimsResearchFile researchFile)
...
+                    AppLastUpdateUserGuid = rd.Document.AppLastUpdateUserGuid,
+                    AppLastUpdateTimestamp = rd.Document.AppLastUpdateTimestamp,
+                })
+                .OrderByDescending(lu => lu.AppLastUpdateTimestamp)
+                .Take(1)
+                .ToList();
+            lastUpdatedByAggregate.AddRange(documentsLastUpdatedBy);
+
+            // This is needed to get the document last-updated-by from the document that where deleted
+            var documentsHistoryLastUpdatedBy = this.Context.PimsResearchFileDocumentHists.AsNoTracking()
```
devinleighsmith, line 180: From practical testing this doesn't seem to work, not sure why yet.

devinleighsmith: (tried deleting a document, last update date/time does not change).

devinleighsmith: Ok, think I've got it figured out. Backend logic works fine, frontend logic does not. Looks like the frontend is re-querying the update info before the actual delete is fully processed? if I do the delete the request doesn't show the correct time, but if I then refresh the page everything works.

source/backend/dal/Repositories/ResearchFileRepository.cs
```diff
@@ -138,6 +139,120 @@ public PimsResearchFile Update(PimsResearchFile researchFile)
...
+                .Where(rdh => rdh.ResearchFileId == id)
+                .Select(rdh => new LastUpdatedByModel()
+                {
+                    ParentId = id,
+                    AppLastUpdateUserid = rdh.AppLastUpdateUserid, // TODO: Update this once the DB tracks the user
```
devinleighsmith, line 185: is this specific to the document history? only see this comment here.

source/frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -48,6 +49,7 @@ const DocumentListContainer: React.FunctionComponent<
   const onDelete = async (
     documentRelationship: Api_DocumentRelationship,
   ): Promise<boolean | undefined> => {
+    props.onSuccess?.();
```
devinleighsmith, line 52: yeah, this might be the issue, this seems to run before the actual delete request.

source/backend/dal/Repositories/ResearchFileRepository.cs
FuriousLlama: uhm, thats odd. Does the same happen with notes?

source/frontend/src/features/documents/list/DocumentListContainer.tsx
```diff
@@ -66,6 +68,7 @@ const DocumentListContainer: React.FunctionComponent<
   };
 
   const onSuccess = async () => {
+    props.onSuccess?.();
```
devinleighsmith, line 71: maybe the same issue here actually.

source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
```diff
@@ -77,17 +92,42 @@ export const ResearchContainer: React.FunctionComponent<
...
-    setResearchFile(retrieved);
     setFile({ ...retrieved, fileType: FileTypes.Research });
-    setStaleFile(false);
-  }, [getResearchFile, getResearchFileProperties, props.researchFileId, setFile, setStaleFile]);
+  }, [getResearchFile, getResearchFileProperties, props.researchFileId, setFile]);
+
+  const fetchLastUpdatedBy = React.useCallback(async () => {
+    var retrieved = await getLastUpdatedBy(props.researchFileId);
+    if (retrieved !== undefined) {
+      setLastUpdatedBy(retrieved);
```
devinleighsmith, line 101: strictly speaking if you aren't using the data object returned by the wrapper (which automatically does an ismounted check) I believe you need to do one here as well.

devinleighsmith (approved):
everything seems to work except for what I noted about documents.

source/frontend/src/features/documents/list/DocumentListContainer.tsx
FuriousLlama: good catch. I updated the notes one and this one slipped by

source/backend/dal/Repositories/ResearchFileRepository.cs
FuriousLlama: Is something that was missed during analysis. The deleted stamp is only retrieved from history but the triggers do not a user that deleted. This is something we might want everywhere.

PR #3504, opened 2023-10-05
PSP-6824 Property management view

asanchezr:
image


devinleighsmith:
This is minor, but the spacing seems a bit off the mockup. Can we move the field values a bit more to the left or was this done to be consistent with another screen?

devinleighsmith:
On my screen I'm getting two scroll bars, are you seeing that @asanchezr 

image

asanchezr:
> On my screen I'm getting two scroll bars, are you seeing that @asanchezr
> 
> image

I did not see this but good catch. This must be a bug within the TAB container not this story but will fix here

asanchezr:
> This is minor, but the spacing seems a bit off the mockup. Can we move the field values a bit more to the left or was this done to be consistent with another screen?

I have used the default spacing that comes with the `<SectionField />` component - like other forms/views in the app. I would lean towards leaving as is for consistency within the app. Most likely UXpin spacing is not consistent with other forms

devinleighsmith (approved):

FuriousLlama (approved):

PR #3507, opened 2023-10-06
Acquisition File Header | psp-6936

FuriousLlama:


devinleighsmith:
mild issue, but worth mentioning, when I modify properties on an acq file I see two requests to updateinfo.

devinleighsmith:
doesn't seem to affect all entities, checklist updates only fire one updateinfo request.

devinleighsmith:
The AC specifies that form8 add update/delete should trigger this but I'm not seeing the request fire.

devinleighsmith:
@FuriousLlama same thing with takes, doesn't seem to be wired up.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -124,6 +124,438 @@ public PimsAcquisitionFile GetById(long id)
...
+                AppLastUpdateUserid = aeph.AppLastUpdateUserid, // TODO: Update this once the DB tracks the user
+                AppLastUpdateUserGuid = aeph.AppLastUpdateUserGuid, // TODO: Update this once the DB tracks the user
+                AppLastUpdateTimestamp = aeph.EndDateHist ?? DateTime.UnixEpoch,
+            })
+            .OrderByDescending(lu => lu.AppLastUpdateTimestamp)
+            .Take(1)
+            .ToList();
+            lastUpdatedByAggregate.AddRange(expPaymentsHistoryLastUpdatedBy);
+
+            // Acquisition Compensation Requsition Financials
```
devinleighsmith, line 488: nit: comment.

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -124,6 +124,438 @@ public PimsAcquisitionFile GetById(long id)
...
+                    AppLastUpdateUserid = aepi.AppLastUpdateUserid,
+                    AppLastUpdateUserGuid = aepi.AppLastUpdateUserGuid,
+                    AppLastUpdateTimestamp = aepi.AppLastUpdateTimestamp,
+                })
+                .OrderByDescending(lu => lu.AppLastUpdateTimestamp)
+                .Take(1)
+                .ToList();
+            lastUpdatedByAggregate.AddRange(expPaymentsItemsLastUpdatedBy);
+
+            // Acquisition Deleted Compensation Requsition Financials
```
devinleighsmith, line 504: nit: comment

source/backend/dal/Repositories/AcquisitionFileRepository.cs
```diff
@@ -124,6 +124,438 @@ public PimsAcquisitionFile GetById(long id)
                 .FirstOrDefault(x => x.AcquisitionFileId == id) ?? throw new KeyNotFoundException();
         }
 
+        /// <summary>
+        /// Retrieves the acquisition file with the specified id last update information.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        public LastUpdatedByModel GetLastUpdateBy(long id)
```
devinleighsmith, line 132: The size of this is a bit of a code smell, but I feel like refactoring this to use some kind of generic method might also be messy. Thoughts?

source/frontend/src/features/mapSideBar/property/tabs/takes/update/models.ts
```diff
@@ -10,7 +10,7 @@ export const TakesYupSchema = Yup.object().shape({
   takes: Yup.array().of(
     Yup.object().shape({
       description: Yup.string().max(4000, 'Description must be at most ${max} characters'),
-      takeTypeCode: Yup.string().required('Take type is required'),
+      takeTypeCode: Yup.string().required('Take type is required').nullable(),
```
devinleighsmith, line 13: hmm?

devinleighsmith (approved):
just one weird line there in they yup schema that I wasn't expecting, and then those two sub-entities that I mentioned.

source/frontend/src/features/mapSideBar/property/tabs/takes/update/models.ts
FuriousLlama: This fixes an issue with the take validation where it was showing an error instead of the message

source/backend/dal/Repositories/AcquisitionFileRepository.cs
FuriousLlama: I agree. For now I rather leave it as is and create a functor later.

PR #3515, opened 2023-10-13
Regenerated scaffold and cleaned up backend and frontend

FuriousLlama:


source/backend/api/Models/Concepts/CompensationRequisition/CompensationRequisitionModel.cs
```diff
@@ -49,9 +49,9 @@ public class CompensationRequisitionModel : BaseAppModel
 
         public InterestHolderModel InterestHolder { get; set; }
 
-        public long? AcquisitionFilePersonId { get; set; }
+        public long? AcquisitionFileTeamId { get; set; }
 
-        public AcquisitionFilePersonModel AcquisitionFilePerson { get; set; }
+        public AcquisitionFileModel AcquisitionFileTeam { get; set; }
```
asanchezr, line 54: I think it would be AcquisitionFileTeamModel here - right?

source/backend/entities/Partials/AcquisitionTeamPerson.cs
```diff
@@ -0,0 +1,15 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsAcquisitionFileTeam class, provides an entity for the datamodel to manage the relationship between the Acquisition Files and it's team.
+    /// </summary>
+    public partial class PimsAcquisitionFileTeam : StandardIdentityBaseAppEntity<long>, IBaseAppEntity
```
asanchezr, line 8: nit: filename should match class name - e,g. **AcquisitionFileTeam.cs**

source/frontend/dependency-graph.svg
asanchezr, line 1: Not sure we want to include this file in the PR

source/frontend/high-level-dependency-graph.svg
asanchezr, line 1: Not sure we want to include this file in the PR

asanchezr (approved):

PR #3516, opened 2023-10-16
psp-6887 add date and time aliases to frontend. Add model typing to backend.

devinleighsmith:
time aliases for frontend/backend models.


devinleighsmith:
converting to draft due to stability risk for a release sprint.

source/frontend/src/features/acquisition/list/AcquisitionSearchResults/models.ts
```diff
@@ -6,6 +6,7 @@ import {
 import { Api_CompensationRequisition } from '@/models/api/CompensationRequisition';
 import { Api_Project } from '@/models/api/Project';
 import Api_TypeCode from '@/models/api/TypeCode';
+import { UtcIsoDateTime } from '@/models/api/UtcIsoDateTime';
```
FuriousLlama, line 9: Note that the frontend nor the backend can guarantee that the date is in UTC. For that reason how about isoDateTime?

FuriousLlama: and I mean guarantee as a type.

devinleighsmith: by that logic there is no guarantee that the string is an iso formatted string either. I lean more towards leaving utc in as it describes the intended contents - in the same was as IsoDateTime does.

devinleighsmith: @FuriousLlama any thoughts?

FuriousLlama: We do control that the backend will format the DateTime type to an iso string, we can't enforce that the date time being encoded will be in utc. In a nutshell, there is no DateTimeUTC on the backend we can rely on so we cannot make that assumption on the frontend either. Although I apreciate the intent of the utc on the name, it might lead to bugs or mistakes where the developer presumes that using that type guarantees utc-ness

devinleighsmith: Thinking about this more, I feel like we also want to keep utc in the name since we want to explicitly only send utc strings to the backend. In my mind, there is never a time we would want to send or receive a date that is not storing a utc date.

devinleighsmith: lets discuss in dev meeting tmrw.

PR #3525, opened 2023-10-19
allow markers with no pids to be opened/viewed.

devinleighsmith:


source/frontend/src/hooks/repositories/useMapProperties.tsx
```diff
@@ -20,6 +20,7 @@ export const useMapProperties = () => {
         STREET_ADDRESS_1: params?.STREET_ADDRESS_1,
         PID_PADDED: params?.PID?.replace(/[-\s]/g, ''),
         PIN: params?.PIN,
+        PROPERTY_ID: params?.PROPERTY_ID,
```
FuriousLlama, line 23: nit: PIMS_PROPERTY_ID would make it explicit that we are loading something from our side.

source/frontend/src/components/common/mapFSM/useLocationFeatureLoader.tsx
```diff
@@ -63,6 +68,11 @@ const useLocationFeatureLoader = () => {
           | undefined = undefined;
 
         // Load PimsProperties
+        if (selectedFeature?.pimsLocationFeature?.PROPERTY_ID) {
+          pimsLocationProperties = await loadProperties({
+            PROPERTY_ID: selectedFeature.pimsLocationFeature.PROPERTY_ID,
+          });
+        }
         if (parcelFeature !== undefined) {
```
FuriousLlama, line 76: this call might need to be in the else clause of the previous if, otherwise it risks re-assigning the pimsLocationProperties value

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -101,7 +101,7 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
         } else if (event.type === 'MAP_MARKER_CLICK') {
           latLng = event.featureSelected.latlng;
         }
-        const result = locationLoader.loadLocationDetails(latLng);
+        const result = locationLoader.loadLocationDetails(latLng, event.featureSelected);
```
FuriousLlama, line 104: the loadLocationData presumes we are trying to load based on a click of a marker (lat-lng). If the problem is that the pims property is using the pid-pin to find it after searching for the parcel, then instead it should query the pims layer by the lat-lng instead of attempting to use the pims id

source/frontend/src/components/common/mapFSM/useLocationFeatureLoader.tsx
```diff
@@ -63,6 +68,11 @@ const useLocationFeatureLoader = () => {
           | undefined = undefined;
 
         // Load PimsProperties
+        if (selectedFeature?.pimsLocationFeature?.PROPERTY_ID) {
+          pimsLocationProperties = await loadProperties({
```
FuriousLlama, line 72: should it instead try to get the pims property by it's lat-lng?

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
devinleighsmith: I'll try that, was concerned about performance since the lat/lng is not indexed. Will do some testing in one of our more marker heavy envs to check performance.

FuriousLlama: well, it would be doing that search on the other layers too, so assuming those are equal performant it should be fine. Also, as it is now, the pims call waits on the external layer to resolve the pid. If they are doing the call independently they could be parallelized.

FuriousLlama (approved):

PR #3539, opened 2023-10-24
PSP-7032 : FT:HelpDesk:Header UI(Font and Icon size) is not displaying accurately for Help and User text

eddherrera:
PSP-7057 : FT:HelpDesk: map icon is not close PIMS naming on the Header


devinleighsmith:
Well, the help icon doesn't match still, is that a problem? (see figma)

devinleighsmith (approved):

PR #3548, opened 2023-10-26
psp-7099 prevent infinite refresh loop of acq files when user is not …

devinleighsmith:
…authorized or file does not exist.


source/frontend/src/features/mapSideBar/acquisition/AcquisitionView.tsx
```diff
@@ -158,6 +162,12 @@ export const AcquisitionView: React.FunctionComponent<IAcquisitionViewProps> = (
             }
             bodyComponent={
               <StyledFormWrapper>
+                {error && (
+                  <b>
+                    Failed to load Acquisition File. Check the detailed error in the top right for
+                    more details.
+                  </b>
+                )}
```
asanchezr, line 170: nit: please add unit tests to cover this new logic
image

asanchezr (approved):

PR #3564, opened 2023-11-02
Psp 6988 - update property sidebar to use router instead of state.

devinleighsmith:


source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.tsx
```diff
@@ -120,10 +138,8 @@ export const MotiInventoryContainer: React.FunctionComponent<
       onClose={props.onClose}
     >
       <>
-        <PropertyViewSelector
+        <PropertyRouter
           composedPropertyState={composedPropertyState}
```
FuriousLlama, line 142: this composed property state might not be necessary anymore if the router can handle the expected state

devinleighsmith: composedPropertyState is maybe a misnomer. This is the object that contains all of the ltsa/property/bc assessment information, as well as the loading states for each. Still required.

FuriousLlama (approved):

PR #3568, opened 2023-11-03
PSP-7075 : Show ACQ team (organization) in project exports

eddherrera:


source/frontend/src/features/projects/reports/models.ts
```diff
@@ -19,3 +22,22 @@ export enum ProjectExportTypes {
   COMPENSATION = 'Compensation Requisition Export',
   AGREEMENT = 'Agreement Export',
 }
+
+const getParameterIdFromOptions = (
```
devinleighsmith, line 26: would you mind moving this out to a shared method (since I think your last pr uses this as well). maybe in contactutils?

devinleighsmith (approved):

source/frontend/src/features/projects/reports/models.ts
eddherrera: different logic, this one returns an array, the other one returns a string. I thought about it.

PR #3574, opened 2023-11-06
Takes - Acquire property toggle

FuriousLlama:


source/backend/tests/unit/api/Services/AcquisitionFileServiceTest.cs
```diff
@@ -43,27 +43,6 @@ private AcquisitionFileService CreateAcquisitionServiceWithPermissions(params Pe
             return this._helper.Create<AcquisitionFileService>(user);
         }
 
-        public static object[] takes = new List<object[]>() {
```
FuriousLlama, line 46: moved this section to be where the parametrized test is.

devinleighsmith:
minor placement issue, still may result in a bug.

image

devinleighsmith (approved):

PR #3578, opened 2023-11-08
Allow Products with multiple projects

FuriousLlama:


devinleighsmith:
tested h179, h120, and h0774 and h0443 form templates to ensure project printed correctly, appears to be working.

devinleighsmith:
Also tested project export, looks good.

devinleighsmith:
also tested alternate project and acquisition list view project search, that is working.

devinleighsmith:
project advanced search also working.

source/backend/api/Models/Concepts/Project/ProjectProductMap.cs
```diff
@@ -0,0 +1,25 @@
...
+            config.NewConfig<Entity.PimsProjectProduct, ProjectProductModel>()
+                .Map(dest => dest.Id, src => src.ProjectProductId)
+                .Map(dest => dest.ProjectId, src => src.ProjectId)
+                .Map(dest => dest.Product, src => src.Product)
+                .Map(dest => dest.ProductId, src => src.ProductId)
+                .Map(dest => dest.Project, src => src.Project)
+                .Inherits<Entity.IBaseEntity, BaseModel>();
+
+            config.NewConfig<ProjectProductModel, Entity.PimsProjectProduct>()
+                .Map(dest => dest.ProjectProductId, src => src.Id)
```
devinleighsmith, line 19: seems like an unusual mixture of fields here (one id and then one entity), is there a reason why that is necessary?

source/backend/api/Areas/Projects/Controllers/ProjectController.cs
```diff
@@ -113,11 +115,11 @@ public IActionResult GetAll()
         [ProducesResponseType(typeof(ProjectModel), 200)]
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "project" })]
-        public IActionResult AddProject(ProjectModel projectModel)
+        public IActionResult AddProject(ProjectModel projectModel, [FromQuery] string[] userOverrideCodes)
```
devinleighsmith, line 118: what happens if the frontend doesn't provide this query param?

source/backend/api/Services/ProjectService.cs
```diff
@@ -134,44 +136,107 @@ public PimsProject Add(PimsProject project)
...
 
             return updatedProject;
         }
 
-        /*private void CheckForDuplicateProducts(IEnumerable<PimsProduct> products, long projectId)
+        /*
+        Updates the passed project with the matched products in place.
+        Note: The return list contains the external products matched
+        */
+        private List<PimsProduct> MatchProducts(PimsProject project)
```
devinleighsmith, line 177: This function does seem like one we should have good test coverage on (via the public method).

devinleighsmith:
@FuriousLlama the estimate dates on products to not appear to work.

devinleighsmith:
always gets reset to empty string:
image

PR #3580, opened 2023-11-08
psp-7209 do not display acquisition file until all loading state is c…

devinleighsmith:
…omplete and context is initialized.

### DIFF
--- a/source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
+++ b/source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
@@ -275,7 +275,12 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
   };
 
   // UI components
-  if (loadingAcquisitionFile || (loadingAcquisitionFileProperties && !isPropertySelector)) {
+  if (
+    loadingAcquisitionFile ||
+    (loadingAcquisitionFileProperties && !isPropertySelector) ||
+    file?.fileType !== FileTypes.Acquisition ||
+    file?.id !== acquisitionFileId
+  ) {
     return <LoadingBackdrop show={true} parentScreen={true}></LoadingBackdrop>;
   }
 
--- a/source/frontend/src/features/mapSideBar/acquisition/AcquisitionView.tsx
+++ b/source/frontend/src/features/mapSideBar/acquisition/AcquisitionView.tsx
@@ -76,9 +76,6 @@ export const AcquisitionView: React.FunctionComponent<IAcquisitionViewProps> = (
   const history = useHistory();
   const match = useRouteMatch();
   const { file, lastUpdatedBy } = useContext(SideBarContext);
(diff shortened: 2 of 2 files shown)


FuriousLlama (approved):

source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
```diff
@@ -275,7 +275,12 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
   };
 
   // UI components
-  if (loadingAcquisitionFile || (loadingAcquisitionFileProperties && !isPropertySelector)) {
+  if (
+    loadingAcquisitionFile ||
+    (loadingAcquisitionFileProperties && !isPropertySelector) ||
+    file?.fileType !== FileTypes.Acquisition ||
```
FuriousLlama, line 281: this is odd. Do you know when this would be true?

devinleighsmith: Yes, so essentially there is a race condition where when the file is loading the previous file is still in context. The state transition to set loading to false occurs before the new file is set into the context, and so this occurs.

Another solution would be to set the file in the context to null immediately during load.

asanchezr: This bug happens when you load a Research file into the sidebar (which populates the sidebar context), then without closing the sidebar go to an ACQ file (via the PIMS files tab of a property of the file). Then the container renders what is already in context thinking it is an ACQ file when it is not.

source/frontend/src/features/mapSideBar/acquisition/AcquisitionView.tsx
```diff
@@ -76,9 +76,6 @@ export const AcquisitionView: React.FunctionComponent<IAcquisitionViewProps> = (
   const history = useHistory();
   const match = useRouteMatch();
   const { file, lastUpdatedBy } = useContext(SideBarContext);
-  if (!!file && file?.fileType !== FileTypes.Acquisition) {
-    throw Error('Context file is not an acquisition file');
-  }
```
asanchezr, line 81: I actually think this is a good error check. Without it we would have never found this bug/race condition. Just my opinion, but I would actually like to see this restored and a similar check added to ResearchView - but we can discuss on our Friday meeting

asanchezr (approved):

source/frontend/src/features/mapSideBar/acquisition/AcquisitionView.tsx
devinleighsmith: If we leave it in, it would be dead code, because the parent makes the same check. This if statement will never run with the recommended changes.

asanchezr: well, in theory this view could be embedded "anywhere" by another container but as I said would like us to discuss this use case for other views that make use of a context that might not have the information the view is expecting

PR #3595, opened 2023-11-15
PSP-6778 : Update dialog boxes

eddherrera:
image

image

image

### DIFF
--- a/source/frontend/src/assets/scss/App.scss
+++ b/source/frontend/src/assets/scss/App.scss
@@ -2,7 +2,7 @@
 
 .modal-dialog {
   div.modal-header {
-    height: 38px;
+    height: 48px;
     padding: 0 16px;
     display: flex;
     flex-direction: row;
@@ -12,7 +12,7 @@
 
     .modal-title {
       font-family: BcSans-Bold;
-      font-size: 18px;
+      font-size: 22px;
       height: 2.75rem;
       height: auto;
     }
@@ -28,8 +28,8 @@
   }
 
   div.modal-body {
-    padding: 24px 16px;
-    text-align: justify;
+    padding: 24px 18px;
+    font-size: 18px;
   }
 
   div.modal-footer {
@@ -41,8 +41,8 @@
 
     .button-wrap {
       display: inline-flex;
-      margin-top: 36px;
-      margin-bottom: 36px;
+      margin-top: 24px;
+      margin-bottom: 24px;
     }
 
     button {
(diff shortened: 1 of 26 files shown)


source/frontend/src/assets/scss/App.scss
```diff
@@ -1,19 +1,112 @@
 @import './styles.scss';
 
-.modal-dialog.modal-xl {
-  max-width: 100rem;
-}
+.modal-dialog {
+  div.modal-header {
+    height: 48px;
```
devinleighsmith, line 5: all of our sizing should be done in rems if possible.

so 4.8rem
0 1.6rem
etc

source/frontend/src/components/common/GenericModal.tsx
```diff
@@ -134,29 +139,65 @@ export const GenericModal = (props: Omit<BsModalProps, 'onHide'> & ModalProps) =
...
+      {...rest}
+      show={showState}
+      modalSize={modalSize}
+      onHide={noop}
+      className={className}
+    >
       <Modal.Header closeButton={closeButton} onHide={close}>
-        <Modal.Title>{title}</Modal.Title>
+        <Modal.Title>
+          {getHeaderIcon() && <div className="header-icon">{getHeaderIcon()}</div>}
```
devinleighsmith, line 171: nit: call this once and store the result somewhere, instead of calling this twice.

source/frontend/src/components/common/GenericModal.tsx
```diff
@@ -134,29 +139,65 @@ export const GenericModal = (props: Omit<BsModalProps, 'onHide'> & ModalProps) =
...
   return (
-    <ModalContainer {...rest} show={showState} modalSize={modalSize} onHide={noop}>
+    <ModalContainer
+      {...rest}
+      show={showState}
+      modalSize={modalSize}
+      onHide={noop}
+      className={className}
+    >
       <Modal.Header closeButton={closeButton} onHide={close}>
```
devinleighsmith, line 169: the flex is on this div, which means the direct children are the title and the close button. That means the header icon and the text end up being grandchildren and end up in a display: block element within the title. That probably means you are going to end up with some wonky behaviour if the text in the title overlaps the length of the modal.

devinleighsmith (approved):
@eddherrera at the very least please convert px values to rem.

source/frontend/src/assets/scss/App.scss
eddherrera: Updated

source/frontend/src/components/common/GenericModal.tsx
eddherrera: Updated

PR #3597, opened 2023-11-16
Hotfixes

devinleighsmith:
psp-7241
psp-7227
psp-7231

### DIFF
--- a/source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.test.tsx
+++ b/source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.test.tsx
@@ -88,7 +88,30 @@ describe('useGenerateAgreement functions', () => {
     await act(async () => {
       await generate(mockAgreementsResponse()[1]);
       expect(generateFn).toHaveBeenCalled();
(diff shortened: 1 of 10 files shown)


source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateAgreement.test.tsx
```diff
@@ -111,7 +134,30 @@ describe('useGenerateAgreement functions', () => {
...
+        },
+      ],
+    };
+    const generate = setup({ acquisitionResponse: responseWithTeam });
+
+    await act(async () => {
+      await generate(mockAgreementsResponse()[1]);
+      expect(generateFn).toHaveBeenCalled();
+      expect(getPersonConceptFn).toHaveBeenCalledWith(1);
       expect(getAcquisitionFileProperties).toHaveBeenCalled();
```
asanchezr, line 161: not sure I follow this test. We already have a test for negotiating agent being a person so I would expect this test for an org - right?
If it is for an org, then the test description should be changed to "... negotiating agent with org" and also the test assertions should change to `expect(getOrganizationConceptFn).toHaveBeenCalledWith(1);` instead of `expect(getPersonConceptFn).toHaveBeenCalledWith(1);`

asanchezr (approved):

PR #3602, opened 2023-11-21
Added check status logic when updating acq and related entities | psp-7006

FuriousLlama:


devinleighsmith:
@FuriousLlama I checked this out, but I don't see this anywhere:

Edit Button Is Hidden

 GIVEN
User does not have permissions to edit a tab
WHEN
User views restricted tab
THEN
System will show a tooltip that indicates why editing is not available.
"The file you are viewing is in a non-editable state. Change the file status to active or draft to allow editing."

devinleighsmith:
Hmm, the story does not go into how to control the behaviour of making an agreement editable/non-editable when it is swapped from draft to final in a file that is already in the final status. I do think it is a bit jarring to disable the fields the instant the status change happens (before the agreement is saved). Could that be changed to only make non-editable when the agreement status is final AND there is no id (indicating an unsaved agreement)?

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -240,6 +244,12 @@ public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile, IEnumerab
             ValidateVersion(acquisitionFile.Internal_Id, acquisitionFile.ConcurrencyControlNumber);
             ValidateDrafts(acquisitionFile.Internal_Id);
 
+            AcqusitionStatusTypes? currentAcquisitionStatus = GetCurrentAcquisitionStatus(acquisitionFile.Internal_Id);
```
devinleighsmith, line 247: nit: acqusition -> acquisition

source/backend/api/Constants/AcqusitionStatusTypes.cs
```diff
@@ -0,0 +1,31 @@
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace Pims.Api.Constants
+{
+    [JsonConverter(typeof(JsonStringEnumMemberConverter))]
+    public enum AcqusitionStatusTypes
```
devinleighsmith, line 7: nit: acqusition

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -393,6 +409,31 @@ public IEnumerable<PimsAgreement> UpdateAgreements(long acquisitionFileId, List<
...
             _user.ThrowInvalidAccessToAcquisitionFile(_userRepository, _acqFileRepository, acquisitionFileId);
 
+            var currentAcquisitionStatus = GetCurrentAcquisitionStatus(acquisitionFileId);
+
+            var currentAgreements = _agreementRepository.GetAgreementsByAcquisitionFile(acquisitionFileId);
+
+            var toBeUpdated = currentAgreements.Where(ca => agreements.Any(na => ca.AgreementId == na.AgreementId && !ca.IsEqual(na)));
+            var toBeDeleted = currentAgreements.Where(ca => !agreements.Any(na => ca.AgreementId == na.AgreementId));
+
+            foreach (var agreement in toBeUpdated)
```
devinleighsmith, line 419: is there a reason that agreements being edited and agreements being deleted are being split out into their own loops? Seems like the business logic that is being executed is identical, sort of adds an implication that updates should not be handles the same as deletes.

source/backend/api/Services/AcquisitionFileService.cs
devinleighsmith, line 1: It looks like you added an explicit check/throw for stakeholders and checklists even though those errors will not get thrown in current state. I'm fine with that as it provides a consistent pattern for update logic, but doesn't that mean you should add a check for updateProperties as well?

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -59,8 +62,15 @@ public PimsCompensationRequisition Update(PimsCompensationRequisition compensati
 
             var currentCompensation = _compensationRequisitionRepository.GetById(compensationRequisition.CompensationRequisitionId);
 
-            CheckDraftStatusUpdateAuthorized(currentCompensation.IsDraft, compensationRequisition.IsDraft);
-            CheckTotalAllowableCompensation(compensationRequisition.AcquisitionFileId, compensationRequisition);
+            var currentAcquisitionFile = _acqFileRepository.GetById(currentCompensation.AcquisitionFileId);
+            var currentAcquisitionStatus = Enum.Parse<AcqusitionStatusTypes>(currentAcquisitionFile.AcquisitionFileStatusTypeCode);
+
+            if (!_statusSolver.CanEditOrDeleteCompensation(currentAcquisitionStatus, currentCompensation.IsDraft) && !_user.HasPermission(Permissions.SystemAdmin))
```
devinleighsmith, line 68: I actually don't see a requirement for an admin to be able to override this in the story, can you point that out to me?

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -90,6 +100,15 @@ public bool DeleteCompensation(long compensationId)
             _logger.LogInformation("Deleting compensation with id ...", compensationId);
             _user.ThrowIfNotAuthorized(Permissions.CompensationRequisitionDelete, Permissions.AcquisitionFileEdit);
 
+            var currentCompensation = _compensationRequisitionRepository.GetById(compensationId);
+
+            var currentAcqusitionStatus = GetCurrentAcqusitionStatus(currentCompensation.AcquisitionFileId);
```
devinleighsmith, line 105: nit: getcurrent acqusition

source/backend/api/Services/AcquisitionFileService.cs
FuriousLlama: From what I gather on the confluence page the properties are allowed to be updated at all times (all statues)

source/backend/api/Services/CompensationRequisitionService.cs
FuriousLlama: It's from a different story. I figured those where additive instead of replaced.

source/backend/api/Services/AcquisitionFileService.cs
devinleighsmith: right, but the same applies to stakeholders and checklists, why have explicit logic for those and not for properties?

source/backend/entities/Partials/Agreement.cs
```diff
@@ -11,5 +11,28 @@ public partial class PimsAgreement : StandardIdentityBaseAppEntity<long>, IBaseA
         [NotMapped]
         public override long Internal_Id { get => this.AgreementId; set => this.AgreementId = value; }
         #endregion
+
+        public bool IsEqual(PimsAgreement other)
```
devinleighsmith, line 15: In my experience isEqual methods generally handle being passed a null other value.

source/backend/tests/core/TestHelper.cs
```diff
@@ -11,8 +11,10 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Moq;
+using Pims.Api.Services;
 using Pims.Dal;
 using Pims.Dal.Configuration.Generators;
+using Pims.Dal.Entities;
```
devinleighsmith, line 17: this seems odd, looks like you are adding imports but then not using them anywhere?

source/frontend/src/features/mapSideBar/acquisition/tabs/checklist/detail/AcquisitionChecklistView.tsx
```diff
@@ -38,10 +39,12 @@ export const AcquisitionChecklistView: React.FC<IAcquisitionChecklistViewProps>
   const checklist = acquisitionFile?.acquisitionFileChecklist || [];
   const lastUpdated = lastModifiedBy(checklist);
 
+  const statusSolver = new StatusUpdateSolver(acquisitionFile);
+
   return (
     <StyledSummarySection>
       <StyledEditWrapper className="mr-3 my-1">
-        {keycloak.hasClaim(Claims.ACQUISITION_EDIT) && acquisitionFile !== undefined ? (
```
devinleighsmith, line 44: are we ok removing that undefined acquisitionFile check?

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailContainer.tsx
```diff
@@ -91,11 +91,10 @@ export const CompensationRequisitionDetailContainer: React.FunctionComponent<
   return compensation ? (
     <View
       loading={loading || loadingPerson || loadingOrganization}
+      acquisitionFile={acquisitionFile}
       compensation={compensation}
       compensationContactPerson={payeePerson}
       compensationContactOrganization={payeeOrganization}
-      acqFileProject={acquisitionFile?.project}
```
devinleighsmith, line 97: huh, I wonder why these were split out like this?

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderContainer.test.tsx
```diff
@@ -83,101 +78,4 @@ describe('StakeHolderContainer component', () => {
     const { asFragment } = setup({});
     expect(asFragment()).toMatchSnapshot();
   });
-
-  it('groups multiple interest holder properties by acquisition file id', async () => {
```
devinleighsmith, line 87: ?

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderView.test.tsx
```diff
@@ -128,11 +150,13 @@ describe('StakeHolderView component', () => {
...
     );
 
+    organizerMock.getInterestProperties.mockReturnValue([model]);
+
     model.identifier = 'PID: 025-196-375';
     const { getByText } = setup({
       props: {
-        groupedInterestProperties: [],
-        groupedNonInterestProperties: [model],
+        //groupedInterestProperties: [],
```
devinleighsmith, line 158: ? if this is a wip, please fix. if not, can you remove these comments?

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderContainer.tsx
```diff
@@ -34,75 +31,14 @@ export const StakeHolderContainer: React.FunctionComponent<IStakeHolderContainer
...
-      )}
-      legacyStakeHolders={acquisitionFile.legacyStakeholders ?? []}
+      acquisitionFile={acquisitionFile}
+      interestHolders={apiInterestHolders}
       onEdit={onEdit}
     />
   );
 };
 
-const getGroupedInterestProperties = (
```
devinleighsmith, line 71: Is this related to this PR? If not, I'd really prefer we separate these kind of changes into their own PR.

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/stakeholderOrganizer.ts
```diff
@@ -0,0 +1,84 @@
+import { Api_AcquisitionFile } from '@/models/api/AcquisitionFile';
```
devinleighsmith, line 1: This looks a bit like unrelated tech-debt. I prefer this, but feel like this should have been done in a separate PR at least.

devinleighsmith:
I'm seeing a lot of line coverage warnings for these changes. Can you confirm that this PR conforms to our minimum test coverage requirements:
1. one test per business logic.
2. each layer should be mocked, validate that mocked layers are called in the expected manner.

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderView.test.tsx
FuriousLlama: ups. forgot to cleanup this tests

source/frontend/src/features/mapSideBar/acquisition/tabs/checklist/detail/AcquisitionChecklistView.tsx
FuriousLlama: the status solver does take into account the acq file being null/undefined

source/frontend/src/features/mapSideBar/acquisition/tabs/compensation/detail/CompensationRequisitionDetailContainer.tsx
FuriousLlama: note sure, but now is more cohesive

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderContainer.test.tsx
FuriousLlama: All the tests for the grouping logic was movied into its own test file that only deals with that.

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/StakeHolderContainer.tsx
FuriousLlama: Thats fair. I rather keep it as is now and in the future create new PRs

source/frontend/src/features/mapSideBar/acquisition/tabs/stakeholders/detail/stakeholderOrganizer.ts
FuriousLlama: True, I refactored to leverage similar props being passed. It also contains the grouping logic instead of being part of the view

FuriousLlama:
> I'm seeing a lot of line coverage warnings for these changes. Can you confirm that this PR conforms to our minimum test coverage requirements:
> 
> 1. one test per business logic.
> 2. each layer should be mocked, validate that mocked layers are called in the expected manner.

The backend logic has very thorough tests. I can add some to the front end too for the solver, but otherwise, it should have the same amount of tests.

source/backend/entities/Partials/Agreement.cs
FuriousLlama: That's a good point, however, this one is not a Comparable but rather a helper method. I can make it more standard too but I am not sure we gain much from it.

source/backend/api/Services/AcquisitionFileService.cs
FuriousLlama: Although true, doing the loop on each would be the same as doing the loop on an aggregated collection.

source/backend/api/Services/AcquisitionFileService.cs
FuriousLlama: Ah, I understand what you are saying now. Initially I had it like that, but it felt like adding those checks was adding unnecessary complexity. If you feel strongly about it I can add them back.

devinleighsmith (approved):
Approved, but you are going to get a test reject from not displaying the warning tooltip in place of the edit icon.

FuriousLlama:
> Approved, but you are going to get a test reject from not displaying the warning tooltip in place of the edit icon.

Tooltip is there :)

PR #3606, opened 2023-11-21
PSP-7267 : FT: Leases & Licenses: System shows "Leases & Licenses" wh…

eddherrera:
…en you do mouse hover on the icon


asanchezr (approved):

devinleighsmith:
@eddherrera what is the plan for these?

PR #3617, opened 2023-11-24
Psp 7281

devinleighsmith:


source/backend/api/Models/Concepts/DispositionFile/DispositionFileMap.cs
```diff
@@ -0,0 +1,51 @@
...
+                .Map(dest => dest.FileNumber, src => src.FileNumber)
+                .Map(dest => dest.FileName, src => src.FileName)
+                .Map(dest => dest.FileReference, src => src.ReferenceNumber)
+                .Map(dest => dest.AssignedDt, src => src.AssignedDate)
+                .Map(dest => dest.InitiatingDocumentDt, src => src.InitiatingDocumentDate)
+                .Map(dest => dest.CompletedDt, src => src.CompletionDate)
+                .Map(dest => dest.DispositionFundingTypeCodeNavigation, src => src.FundingTypeCode)
+                .Map(dest => dest.DispositionFileStatusTypeCodeNavigation, src => src.FileStatusTypeCode)
+                .Map(dest => dest.DspPhysFileStatusTypeCodeNavigation, src => src.PhysicalFileStatusTypeCode)
+                .Map(dest => dest.DispositionTypeCodeNavigation, src => src.DispositionTypeCode)
```
asanchezr, line 41: just checking that we have a mapping that goes from model to the typecode navigation...
I think we usually set the string on the way in, not the navigation property itself

source/backend/api/Models/Concepts/DispositionFile/DispositionFileModel.cs
```diff
@@ -0,0 +1,90 @@
+using System;
+using System.Collections.Generic;
+
+/*
+* Frontend model
+* LINK @frontend/src\models\api\DispositionFile.ts:14
```
asanchezr, line 6: nit and question: are we forced to use a mix of forward slashes and backward slashes here or can we use forward slashes only?

source/backend/api/Models/Concepts/DispositionFile/DispositionFileTeamMap.cs
```diff
@@ -0,0 +1,33 @@
...
+                .Map(dest => dest.Id, src => src.DispositionFileTeamId)
+                .Map(dest => dest.DispositionFileId, src => src.DispositionFileId)
+                .Map(dest => dest.Person, src => src.Person)
+                .Map(dest => dest.PersonId, src => src.PersonId)
+                .Map(dest => dest.Organization, src => src.Organization)
+                .Map(dest => dest.OrganizationId, src => src.OrganizationId)
+                .Map(dest => dest.PrimaryContact, src => src.PrimaryContact)
+                .Map(dest => dest.PrimaryContactId, src => src.PrimaryContactId)
+                .Map(dest => dest.TeamProfileType, src => src.DspFlTeamProfileTypeCodeNavigation)
+                .Map(dest => dest.TeamProfileTypeCode, src => src.DspFlTeamProfileTypeCode)
```
asanchezr, line 20: TeamProfile and TeamProfileTypeCode seems to be duplicated (one is the string code, the other the typemodel representation of it). Perhaps the string one could be removed?

source/backend/api/Models/Concepts/DispositionFile/DispositionFileTeamModel.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// get/set - The Team's profile type code.
+        /// </summary>
+        public string TeamProfileTypeCode { get; set; }
+
+        /// <summary>
+        /// get/set - The Team's profile type code.
+        /// </summary>
+        public TypeModel<string> TeamProfileType { get; set; }
```
asanchezr, line 60: is there a need to keep the duplicate `TeamProfileTypeCode` in the model when we already have the TypeModel for it?

source/backend/dal/Repositories/DispositionFileRepository.cs
```diff
@@ -0,0 +1,167 @@
...
+                .Include(d => d.DspPhysFileStatusTypeCodeNavigation)
+                .Include(d => d.PimsDispositionFileTeams)
+                    .ThenInclude(d => d.Organization)
+                .Include(d => d.PimsDispositionFileTeams)
+                    .ThenInclude(d => d.Person)
+                .Include(d => d.PimsDispositionFileTeams)
+                    .ThenInclude(d => d.PrimaryContact)
+                .Include(d => d.PimsDispositionFileTeams)
+                    .ThenInclude(d => d.DspFlTeamProfileTypeCodeNavigation)
+                .FirstOrDefault(d => d.DispositionFileId == id);
```
asanchezr, line 60: want to confirm this follows existing patterns - I believe we throw a keynotfound exception if no entity is found by id?

source/backend/dal/Repositories/Interfaces/ILookupRepository.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using DocumentFormat.OpenXml.InkML;
```
asanchezr, line 2: is this import needed?

source/backend/api/Models/Concepts/DispositionFile/DispositionFileModel.cs
devinleighsmith: That would be a better question for @FuriousLlama, as currently he is the only one with this library functional. This conforms to the pattern that I've seen on other backend model files.

source/frontend/src/features/mapSideBar/disposition/DispositionView.tsx
```diff
@@ -0,0 +1,235 @@
...
+                  onChange={onMenuChange}
+                  onShowPropertySelector={onShowPropertySelector}
+                />
+              </>
+            }
+            bodyComponent={
+              <StyledFormWrapper>
+                {error && (
+                  <b>
+                    Failed to load Diposition File. Check the detailed error in the top right for
```
asanchezr, line 157: nit: type "Diposition" -> Disposition

source/frontend/src/models/api/DispositionFile.ts
```diff
@@ -0,0 +1,52 @@
...
+  id?: number;
+  dispositionFileId: number;
+  personId?: number;
+  person?: Api_Person;
+  organizationId?: number;
+  organization?: Api_Organization;
+  primaryContactId?: number;
+  primaryContact?: Api_Person;
+  teamProfileTypeCode?: string;
+  teamProfileType?: Api_TypeCode<string>;
```
asanchezr, line 51: like the backend models I think the type model is enough - no need to keep the string code as well

asanchezr (approved):

PR #3618, opened 2023-11-27
Pims api to Frontend ts generator and generated files

FuriousLlama:


devinleighsmith:
I'm not going to review all 460 files, are there any files that I should take a look at or are we confident based on your presentation from Friday that this is good to ship as-is?

FuriousLlama:
> I'm not going to review all 460 files, are there any files that I should take a look at or are we confident based on your presentation from Friday that this is good to ship as-is?
> 
> @FuriousLlama

I am pretty confident things will work. If you have spare cycles having a second set of eyes on the changes to the controllers would be good given those where more prone to error.

devinleighsmith (changes requested):
@FuriousLlama I checked this out and this appears to break the keycloak-sync. Can you open that solution and fix any errors please?

devinleighsmith:
@FuriousLlama keycloak-sync isn't compiling still. Does this PR depend on something else maybe?

FuriousLlama:
> @FuriousLlama keycloak-sync isn't compiling still. Does this PR depend on something else maybe?

should be compiling now!

FuriousLlama:
@devinleighsmith Ready for re-review.

devinleighsmith (approved):

PR #3622, opened 2023-11-29
psp-7346 correct bracket usage, ensure logic limited to file close only.

devinleighsmith:


source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -242,7 +242,7 @@ public PimsAcquisitionFile Update(PimsAcquisitionFile acquisitionFile, IEnumerab
 
             _user.ThrowInvalidAccessToAcquisitionFile(_userRepository, _acqFileRepository, acquisitionFile.Internal_Id);
             ValidateVersion(acquisitionFile.Internal_Id, acquisitionFile.ConcurrencyControlNumber);
-            ValidateDrafts(acquisitionFile.Internal_Id);
+            ValidateDrafts(acquisitionFile);
```
FuriousLlama, line 245: kinda of a nit: The current acq file is being retrieved multiple times for each validation function, ideally we would retrieve it once and pass it to the validation logic

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -948,7 +949,7 @@ private void ValidateInterestHoldersDependency(long acquisitionFileId, List<Pims
             var currentAcquisitionFile = _acqFileRepository.GetById(acquisitionFileId);
             AcquisitionStatusTypes currentAcquisitionStatus;
 
-            if (Enum.TryParse(currentAcquisitionFile.AcquisitionFileStatusTypeCode, out currentAcquisitionStatus))
+            if (Enum.TryParse(currentAcquisitionFile?.AcquisitionFileStatusTypeCode, out currentAcquisitionStatus))
```
FuriousLlama, line 952: the acq file should never be null here.

FuriousLlama: imo, having the code throw here if null is a good idea

source/backend/api/Services/AcquisitionFileService.cs
devinleighsmith: probably not part of this PR, but easy enough to fix so why not?

devinleighsmith: Is this true? I had a look and while I see acquisition sub-entities being loaded in the various Validate* functions I don't see the acquisition file being retrieved multiple times.

devinleighsmith: Hmm, I only see one request for the acquisition file in any validate functions within Update:
ValidatePayeeDependency
the only other call is at the end of the update function, but that is after the transaction is committed so I think that makes sense.

FuriousLlama: for example teh  GetCurrentAcquisitionStatus will retrieve the latest acq file for the given id

FuriousLlama (approved):

PR #3634, opened 2023-12-06
PSP-6952 tests

asanchezr:


source/backend/api/Services/DispositionFileService.cs
```diff
@@ -75,7 +75,8 @@ public IEnumerable<PimsPropertyDispositionFile> GetProperties(long id)
         public IEnumerable<PimsDispositionFileTeam> GetTeamMembers()
         {
             _logger.LogInformation("Getting disposition team members");
-            _user.ThrowIfNotAuthorized(Permissions.DispositionView, Permissions.ContactView);
+            _user.ThrowIfNotAuthorized(Permissions.DispositionView);
```
devinleighsmith, line 78: interesting changing, does this result in different behaviour?

devinleighsmith (approved):

source/backend/api/Services/DispositionFileService.cs
asanchezr: Yes, it is! It caught me by surprise by passing a test it should have failed due to lack of claims...
It is very subtle - passing a list of permissions will allow it if the user has ANY of the claims (instead of ALL).
To check for all permissions we should call to **ThrowIfNotAllAuthorized** as shown below, which calls `user.HasPermissions(...)` (plural) vs `user.HasPermission(...)` (singular)

The fact that `user.HasPermission()` can take a list but only enforces that you have at least one is misleading and easy to confuse with `HasPermissions()` . IMO, we should file a tech debt task to improve naming and make logic more explicit...

image

PR #3659, opened 2023-12-14
psp-7388 correct table behaviour such that pagination works as expected.

devinleighsmith:


source/frontend/src/components/Table/Table.tsx
```diff
@@ -295,7 +295,7 @@ export const Table = <T extends IIdentifiedObject, TFilter extends object = {}>(
   const manualSortBy = !!externalSort || props.manualSortBy;
   const totalItems = externalTotalItems ?? data?.length;
   const pageCount =
-    externalPageCount ?? internalPageSize > 0 ? Math.ceil(totalItems / internalPageSize) : 0;
+    externalPageCount ?? (internalPageSize > 0 ? Math.ceil(totalItems / internalPageSize) : 0);
```
devinleighsmith, line 298: these brackets are actually required, otherwise the pageCount does not update properly when the page size changes.

source/frontend/src/components/contact/ContactManagerView/ContactManagerView.tsx
```diff
@@ -83,9 +83,8 @@ const ContactManagerView = ({
   const changeFilter = useCallback(
     (filter: IContactFilter) => {
       setFilter(filter);
-      setCurrentPage(0);
```
devinleighsmith, line 86: due to my updates to setfilter, the page is always set to 0 internally in the useSearch hook anyways. This no longer fires two search requests.

source/frontend/src/hooks/useSearch.ts
```diff
@@ -85,56 +102,65 @@ export function useSearch<ISearchResult extends object, IFilter extends object>(
...
+    }
+    callApi(currentPage, filter);
+  }, [currentPage, filter, callApi]);
 
   // allow manual re-triggers of the API calls
-  const execute = () => {
+  const executeFn = useCallback(() => {
     setRefreshIndex(r => r + 1);
-  };
+  }, []);
```
asanchezr, line 143: **nit:** I don't think we ever use manual re-triggers of the search call - I think we could try to remove this. Also if you notice from the old code, the `refreshIndex` was in the dependency array of **useDeepEffect** to force the effect to re-run when the execute function was called. I don't see a similar logic with the updated code (one more reason to remove it)

asanchezr (approved):
Thanks for fixing this!

PR #3661, opened 2023-12-14
PSP-7393 : FT:Reg:Lease/License Files:Map Filter: Lease file with lease expiry date or lease term expiry date is Blank and the property associated with this file should pop up on the lease filter

eddherrera:


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -406,68 +407,79 @@ public PimsProperty TransferFileProperty(PimsProperty property, bool isOwned, bo
...
+                                        || (!pl.Lease.PimsLeaseTerms.Any() && !pl.Lease.OrigExpiryDate.HasValue) // no terms and orig is null
+                                        || (!pl.Lease.PimsLeaseTerms.Any() && pl.Lease.OrigExpiryDate.HasValue && pl.Lease.OrigExpiryDate.Value.Date >= DateTime.Now.Date) // no terms and orig not expired
+                                        || (pl.Lease.OrigExpiryDate.HasValue // has terms and term expired is bigger than Lease expiry is not expired
+                                            && pl.Lease.PimsLeaseTerms.Any(lt => lt.TermExpiryDate.HasValue && lt.TermExpiryDate.Value.Date > pl.Lease.OrigExpiryDate.Value.Date && lt.TermExpiryDate.Value.Date >= DateTime.Now.Date))
+                                        || (pl.Lease.OrigExpiryDate.HasValue // has terms and term expired is lower than Lease expiry
+                                            && (!pl.Lease.PimsLeaseTerms.Any(lt => lt.TermExpiryDate.HasValue && lt.TermExpiryDate.Value.Date > pl.Lease.OrigExpiryDate.Value.Date) && pl.Lease.OrigExpiryDate.Value.Date >= DateTime.Now.Date)))));
```
asanchezr, line 469: I see the old logic is using **DateTime.Now.Date** but I wonder if we should be using **DateTime.UtcNow.Date** instead given that we store date-time in UTC in the database. Otherwise this code might return different results based on server timezone (PST vs UTC)

What do you think @devinleighsmith ?

asanchezr (approved):

PR #3663, opened 2023-12-14
Added improved error handling of the useParcelMapLayer service | psp-7416

FuriousLlama:


source/frontend/src/components/common/mapFSM/useLocationFeatureLoader.tsx
```diff
@@ -48,52 +48,49 @@ const useLocationFeatureLoader = () => {
...
+      }
+
+      const municipalityFeature = await adminLegalBoundaryLayerServiceFindOneMunicipality(latLng);
 
-        result.parcelFeature = parcelFeature ?? null;
-        result.regionFeature = regionFeature ?? null;
-        result.districtFeature = districtFeature ?? null;
-        result.municipalityFeature = municipalityFeature ?? null;
-      } finally {
-        // TODO: Remove once the try above is deemed no longer necessary.
```
FuriousLlama, line 94: this is the only real change, the rest is the try wrapping removal being marked as a change

source/frontend/src/hooks/layer-api/useWfsLayer.ts
```diff
@@ -67,6 +67,7 @@ export const useWfsLayer = (
       },
       [layerOptions, url],
     ),
+    throwError: true,
```
FuriousLlama, line 70: the usewfs layer will bubble up the errors now

source/frontend/src/hooks/repositories/mapLayer/useParcelMapLayer.ts
```diff
@@ -24,61 +25,85 @@ export const useParcelMapLayer = () => {
 
   const { execute: getAllFeatures, loading: getAllFeaturesLoading } = getAllFeaturesWrapper;
 
+  const handleError = useCallback((error: unknown) => {
+    toast.error('Unable to contact Parcel Map');
+  }, []);
```
FuriousLlama, line 30: at some point in the future we should improve the network slice to maintain the state of the requests and use a component that monitors it to show or not show errors. For now this is just using the regular toasts as the other services

source/frontend/src/hooks/layer-api/useWfsLayer.ts
asanchezr: Did you check all usages of `useWfsLayer` to make sure they handle errors correctly?

asanchezr (approved):

PR #3694, opened 2024-01-03
Psp 6953 disposition checklist and generic checklist components.

devinleighsmith:


source/backend/dal/Repositories/Interfaces/ILookupRepository.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using DocumentFormat.OpenXml.InkML;
```
eddherrera, line 2: What is this for?

eddherrera (approved):

PR #3708, opened 2024-01-11
psp-7323 Contact ambiguous

devinleighsmith:


source/frontend/src/features/mapSideBar/research/tabs/fileDetails/update/UpdateSummaryForm.tsx
```diff
@@ -162,11 +162,13 @@ const UpdateSummaryForm: React.FunctionComponent<IUpdateSummaryFormProps> = prop
...
           />
         </SectionField>
-        {values.requestor?.id.startsWith('P') && (
-          <SectionField label="Organization" className="pb-4">
-            {values.requestor.organizationName ?? 'none'}
-          </SectionField>
-        )}
+        {values.requestor &&
+          isPersonResult(values.requestor) &&
+          values.requestor.person?.personOrganizations?.length !== undefined && (
```
FuriousLlama, line 167: not sure this guarantees there is at least one personorganization given 0 is different than undefined

source/frontend/src/interfaces/IContactSearchResult.ts
```diff
@@ -27,59 +19,107 @@ export interface IContactSearchResult {
   tenantType?: string;
```
FuriousLlama, line 19: no need to change now, but these should not be optional, but rather allow for undefined or null instead

source/frontend/src/components/contact/ContactManagerView/ContactManagerView.test.tsx
```diff
@@ -111,7 +110,7 @@ describe('ContactManagerView', () => {
   });
 
   it('searches by city/municipality', async () => {
-    setupMockSearch([{ ...defaultSearchResult, municipalityName: 'victoria' }]);
+    setupMockSearch([{ ...defaultPersonSearchResult, municipalityName: 'victoria' }]);
```
FuriousLlama, line 113: we need to move away from static moc objects given some other section of the code could change it. instead, use a factory approach to have a method construct new objects on each call

FuriousLlama (approved):

source/frontend/src/components/contact/ContactManagerView/ContactManagerView.test.tsx
devinleighsmith: true, this isn't really part of this PR (pre-existing mock objects). This was just a re-name. I don't mind correcting this however.

devinleighsmith:
had another look at this. There are too many issues with integrating this without the full swapover to the APIGEN models. will retarget this to 5.1 and put into draft to indicate the dependency.

PR #3713, opened 2024-01-13
Dead code/lint update - read description

devinleighsmith:
recommend ignoring commit shas:
7f8da1c0b0e54223e8b4484cbc3912282a451adc
56aac960df1ea0bb30d54f30281819a30635068b
16427689e9e588c69a169153c2a5dd6c0dd99aad

as the first is automated changes and the second is hand-picked changes that are non-functional, the third contains minor lint test corrections (that result in test passes) and dead code removal as identified by a tool.


FuriousLlama (approved):

source/frontend/src/components/common/ModalContainer.tsx
```diff
@@ -5,11 +5,7 @@ import { ModalContext } from '@/contexts/modalContext';
 
 import GenericModal from './GenericModal';
 
-interface IModalContainerProps {}
-
-export const ModalContainer: React.FunctionComponent<
-  React.PropsWithChildren<IModalContainerProps>
-> = props => {
+export const ModalContainer: React.FunctionComponent<React.PropsWithChildren<unknown>> = () => {
```
FuriousLlama, line 8: I am guessing there is a non prop version?

devinleighsmith: this is just cleaning up unused variables - in this case props is never used.

FuriousLlama (changes requested):
Most of the changes look fine, however, the ones with the double exclamation mark removal could be incorrect, Where those removed because the evaluation of undefined, null or zero will be converted to boolean under the hood?
i.e !!required

FuriousLlama:
@devinleighsmith thanks for the reply. I will test it locally to be 100% I am not missing anything

FuriousLlama (approved):
Checked and the removing the double bang keeps the logic as you mentioned.

PR #3720, opened 2024-01-17
Siteminder Refresh

devinleighsmith:
POC possible implementation of siteminder token refresh.

### DIFF
--- a/source/frontend/src/hooks/useRefreshSiteminder.ts
+++ b/source/frontend/src/hooks/useRefreshSiteminder.ts
@@ -1,0 +1,22 @@
+import axios from 'axios';
+import { useCallback, useMemo } from 'react';
+
+import { useTenant } from '@/tenants';
+
+export const useRefreshSiteminder = () => {
+  const { parcelsLayerUrl } = useTenant();
+
+  const refresh = useCallback(() => {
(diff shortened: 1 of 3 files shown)


devinleighsmith:
Converting back to draft, this should not be merged until 5.1

source/frontend/src/hooks/useRefreshSiteminder.ts
```diff
@@ -0,0 +1,19 @@
...
+import { useTenant } from '@/tenants';
+
+export const useRefreshSiteminder = () => {
+  const { parcelsLayerUrl, siteminderRefreshUrl } = useTenant();
+
+  const refresh = useCallback(() => {
+    axios.get(parcelsLayerUrl + siteminderRefreshUrl);
+  }, [parcelsLayerUrl, siteminderRefreshUrl]);
+
+  return useMemo(
```
FuriousLlama, line 13: this looks kind odd, but should it be a state rather than a computed variable that only depends on itself for updating?

devinleighsmith: uh, yeah, this is from a pattern I use where multiple functions are being returned from a hook. Not necessary in this case.

PR #3731, opened 2024-01-19
PSP-7296 : Create/Edit Sales Details (Disposition)

eddherrera:
image

image


source/backend/api/Areas/Disposition/Controllers/DispositionFileController.cs
```diff
@@ -346,6 +346,59 @@ public IActionResult GetDispositionFileSales([FromRoute]long id)
...
+                DateTime.Now);
+
+            _logger.LogInformation("Dispatching to service: {Service}", _dispositionService.GetType());
+
+            var dispositionSaleEntity = _mapper.Map<Dal.Entities.PimsDispositionSale>(dispositionFileSale);
+            var updatedSale = _dispositionService.UpdateDispositionFileSale(id, saleId, dispositionSaleEntity);
+
+            return new JsonResult(_mapper.Map<DispositionFileSaleModel>(updatedSale));
+        }
+
```
devinleighsmith, line 401: Adding unit tests for controllers is pretty quick, can you please add the minimal amount to allow for test coverage?

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionFileSaleModel.cs
```diff
@@ -87,13 +87,13 @@ public class DispositionFileSaleModel : BaseConcurrentModel
...
         /// <summary>
-        /// get/set - A list of disposition Sale Purchaser(s) Agents.
+        /// get/set - Disposition Sale Purchaser(s)'s Agents.
         /// </summary>
-        public IList<DispositionSalePurchaserAgentModel> DispositionPurchaserAgents { get; set; }
+        public DispositionSalePurchaserAgentModel DispositionPurchaserAgent { get; set; }
 
         /// <summary>
-        /// get/set - A list of disposition Sale Purchaser(s) Solicitors.
+        /// get/set - Disposition Sale Purchaser(s)'s Solicitor.
```
devinleighsmith, line 95: Purchaser(s)'s?

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
```diff
@@ -1,35 +1,49 @@
...
+    readonly rowVersion: number | null = null,
+  ) {
     this.id = id;
     this.dispositionFileId = dispositionFileId;
+    this.rowVersion = rowVersion;
   }
 
-  static fromApi(entity: Api_DispositionFileSale) {
-    const model = new DispositionSaleFormModel(entity.id, entity.dispositionFileId);
+  static fromApi(entity: Api_DispositionFileSale): DispositionSaleFormModel {
```
devinleighsmith, line 34: This is a pretty complex model, it looks like a good candidate for model unit tests. See some of our other model unit tests for example (they are quite simple).

devinleighsmith:
setting isGstRequired to true sends the string value "true" to the backend which causes a 400 error.

image

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleFormYupSchema.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-template-curly-in-string */
+import * as yup from 'yup';
+
+export const DispositionSaleFormYupSchema = yup.object().shape({
```
devinleighsmith, line 4: these are all strings, but in actuality a lot of these values are numbers or dates, please fix.
Make sure to enforce number size limits.
Also, it probably makes sense to limit the fiscal year to 4 digits.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSalePurchasersSubForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { setModalContent, setDisplayModal } = useModalContext();
+
+  return (
+    <FieldArray
+      name="dispositionPurchasers"
+      render={arrayHelpers => (
+        <>
+          {values.dispositionPurchasers.map(
+            (purchaser: DispositionSaleContactModel, index: number) => (
+              <React.Fragment key={`purchaser-${index}`}>
```
devinleighsmith, line 34: Please don't do this unless there is no readily available unique identifier. See React docs for rationale:

image

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    setGSTDerivedAmountFields(isGstRequired, cleanValue);
+  };
+
+  const onUpdateGstApplicable = (gstOption: string): void => {
+    const isGstRequired = stringToBoolean(gstOption);
+    const taxCollectedAmountValue = getIn(formikProps.values, 'gstCollectedAmount');
+
+    if (!isGstRequired) {
+      setModalContent({
```
devinleighsmith, line 52: this looks like it isn't covered in tests, but is one of the more complex workflows in this logic, probably should be covered.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/hooks/useCalculateNetProceedsBeforeSPP.tsx
```diff
@@ -0,0 +1,79 @@
...
+    if (
+      (finalSaleAmountTouched ||
+        realtorCommissionTouched ||
+        totalCostOfSaleTouched ||
+        netBookAmountTouched ||
+        gstTouched ||
+        sppTouched) &&
+      !isSubmitting
+    ) {
+      if (gstDecimal && isGstEligible) {
```
devinleighsmith, line 38: this may not be safe, as if gstDecimal is 0 all of this logic will be skipped.

devinleighsmith (changes requested):
see comments.

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
```diff
@@ -47,22 +75,34 @@ export class DispositionSaleFormModel {
       finalConditionRemovalDate: emptyStringtoNullable(this.finalConditionRemovalDate),
       saleCompletionDate: emptyStringtoNullable(this.saleCompletionDate),
       saleFiscalYear: emptyStringtoNullable(this.saleFiscalYear),
-      finalSaleAmount: this.finalSaleAmount,
-      realtorCommissionAmount: this.realtorCommissionAmount,
+      finalSaleAmount: this.finalSaleAmount ? parseFloat(this.finalSaleAmount.toString()) : null,
```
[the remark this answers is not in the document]
eddherrera: When a value it's removed from the Input, it's being passed as an Empty String, (although it should have been parsed as a null, it doesn't, so a check is needed.)

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
eddherrera: updated

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionFileSaleModel.cs
eddherrera: updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleFormYupSchema.ts
eddherrera: updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/hooks/useCalculateNetProceedsBeforeSPP.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const sppTouched = getIn(touched, 'sppAmount');
+
+  const finalSaleAmount = getIn(values, 'finalSaleAmount');
+  const gstAmount = getIn(values, 'gstCollectedAmount');
+  const realtorCommissionAmount = getIn(values, 'realtorCommissionAmount');
+  const totalCostOfSale = getIn(values, 'totalCostAmount');
+  const netBookAmount = getIn(values, 'netBookAmount');
+  const sppAmount = getIn(values, 'sppAmount');
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
eddherrera: the amounts are update when any of the fields are touched. not when all are touched.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleForm.tsx
eddherrera: updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/hooks/useCalculateNetProceedsBeforeSPP.tsx
eddherrera: updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/hooks/useCalculateNetProceeds.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const finalSaleAmount = getIn(values, 'finalSaleAmount');
+  const gstAmount = getIn(values, 'gstCollectedAmount');
+  const realtorCommissionAmount = getIn(values, 'realtorCommissionAmount');
+  const totalCostOfSale = getIn(values, 'totalCostAmount');
+  const netBookAmount = getIn(values, 'netBookAmount');
+  const sppAmount = getIn(values, 'sppAmount');
+
+  useEffect(() => {
+    if (
+      (finalSaleAmountTouched ||
```
devinleighsmith, line 30: I'm still not a big fan of these touched checks. If I'm in edit mode, don't I want to see what my current sums are regardless of whether I've touched one of the fields?

devinleighsmith: Regardless of the above, things don't seem to be working correctly. I see this when I first go to a completely filled out sales edit form:

image

devinleighsmith: then, if I change the final sale price it clears the net proceeds before spp cost field:

image

eddherrera: removed the "touch" from the useEffect, but I am not seing the empty fields as you mention.

devinleighsmith: I'll re-test.

devinleighsmith (approved):

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -246,6 +246,45 @@ public PimsDispositionSale GetDispositionFileSale(long dispositionFileId)
...
+        public PimsDispositionSale UpdateDispositionFileSale(long dispositionFileId, long saleId, PimsDispositionSale dispositionSale)
+        {
+            _logger.LogInformation("Updating disposition file Sale with DispositionFileId: {id}", dispositionFileId);
+            _user.ThrowIfNotAuthorized(Permissions.DispositionEdit);
+
+            var dispositionFileParent = _dispositionFileRepository.GetById(dispositionFileId);
+            if (dispositionFileId != dispositionSale.DispositionFileId || dispositionSale.DispositionSaleId != saleId || dispositionFileParent is null)
+            {
+                throw new BadRequestException("Invalid dispositionFileId.");
+            }
```
FuriousLlama, line 258: this validation looks like it should happen on the controller instead of the service

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -246,6 +246,45 @@ public PimsDispositionSale GetDispositionFileSale(long dispositionFileId)
...
+        public PimsDispositionSale AddDispositionFileSale(long dispositionFileId, PimsDispositionSale dispositionSale)
+        {
+            _logger.LogInformation("Adding disposition file Sale to Disposition File with Id: {id}", dispositionFileId);
+            _user.ThrowIfNotAuthorized(Permissions.DispositionEdit);
+
+            var dispositionFileParent = _dispositionFileRepository.GetById(dispositionFileId);
+            if (dispositionFileId != dispositionSale.DispositionFileId || dispositionFileParent is null)
+            {
+                throw new BadRequestException("Invalid dispositionFileId.");
+            }
```
FuriousLlama, line 275: same here. the service does not need to get Ids just to validate them

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionFileSaleMap.cs
```diff
@@ -40,8 +42,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.SppAmt, src => src.SppAmount)
                 .Map(dest => dest.RemediationAmt, src => src.RemediationAmount)
                 .Map(dest => dest.PimsDispositionPurchasers, src => src.DispositionPurchasers)
-                .Map(dest => dest.PimsDspPurchAgents, src => src.DispositionPurchaserAgents)
-                .Map(dest => dest.PimsDspPurchSolicitors, src => src.DispositionPurchaserSolicitors);
+                .Map(dest => dest.PimsDspPurchAgents, src => src.DispositionPurchaserAgent == null ? null : new List<DispositionSalePurchaserAgentModel> { src.DispositionPurchaserAgent })
+                .Map(dest => dest.PimsDspPurchSolicitors, src => src.DispositionPurchaserSolicitor == null ? null : new List<DispositionSalePurchaserSolicitorModel> { src.DispositionPurchaserSolicitor });
```
FuriousLlama, line 46: this looks odd. why would the ef entities be expecting a list but  the model only contain one?

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserAgentModel.cs
```diff
@@ -15,7 +15,7 @@ public class DispositionSalePurchaserAgentModel : BaseConcurrentModel
         /// <summary>
         /// Parent Disposition Sale.
         /// </summary>
-        public long DispositionSaleId { get; set; }
+        public long? DispositionSaleId { get; set; }
```
FuriousLlama, line 18: does this match the ef required field?

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserModel.cs
```diff
@@ -15,7 +15,7 @@ public class DispositionSalePurchaserModel : BaseConcurrentModel
         /// <summary>
         /// Parent Disposition Sale.
         /// </summary>
-        public long DispositionSaleId { get; set; }
+        public long? DispositionSaleId { get; set; }
```
FuriousLlama, line 18: same as last question

source/backend/api/Services/DispositionFileService.cs
eddherrera: Controllers should not do validations, and be as simple as possible.

source/backend/api/Services/DispositionFileService.cs
eddherrera: Controllers should not do validations, and be as simple as possible.

source/backend/dal/Repositories/DispositionFileRepository.cs
```diff
@@ -325,6 +325,26 @@ public PimsDispositionSale GetDispositionFileSale(long dispositionId)
...
         }
 
+        public PimsDispositionSale AddDispositionFileSale(PimsDispositionSale dispositionSale)
+        {
+            Context.PimsDispositionSales.Add(dispositionSale);
+
+            return dispositionSale;
+        }
+
+        public PimsDispositionSale UpdateDispositionFileSale(long saleId, PimsDispositionSale dispositionSale)
```
FuriousLlama, line 335: q: does the dispositionSaleId not valid here? why is the saleId prefered?

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleContactModel.ts
```diff
@@ -0,0 +1,74 @@
...
+    readonly rowVersion: number | null = null,
+    contact: IContactSearchResult | null = null,
+  ) {
+    this.id = id;
+    this.dispositionSaleId = dispositionSaleId;
+    this.contact = contact;
+    this.rowVersion = rowVersion;
+  }
+
+  toApi(): Api_DispositionSaleContact | null {
```
FuriousLlama, line 30: please don't use the old API models anymore. instead the generated ones to avoid type missmatches

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
```diff
@@ -38,6 +52,20 @@ export class DispositionSaleFormModel {
 
     model.netProceedsBeforeSppAmount = calculateNetProceedsBeforeSppAmount(entity);
     model.netProceedsAfterSppAmount = calculateNetProceedsAfterSppAmount(entity);
+
+    model.dispositionPurchasers = entity.dispositionPurchasers?.map(
```
FuriousLlama, line 56: not sure why this is not barking, but the result of this could be an undefined, that does not match the type in the model

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleForm.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <SectionField label="Final sale price ($)" labelWidth="5" contentWidth="5">
+        <FastCurrencyInput
+          formikProps={formikProps}
+          field="finalSaleAmount"
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            onFinalSaleAmountUpdated(e.target.value);
+          }}
+        />
+      </SectionField>
+      <SectionField label=" Realtor commission ($)" labelWidth="5" contentWidth="5">
```
FuriousLlama, line 147: nit space on the label

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
```diff
@@ -38,6 +52,20 @@ export class DispositionSaleFormModel {
...
+      x => DispositionSaleContactModel.fromApi(x) || [],
+    );
+
+    model.dispositionPurchaserAgent = entity.dispositionPurchaserAgent
+      ? DispositionSaleContactModel.fromApi(entity.dispositionPurchaserAgent)
+      : new DispositionSaleContactModel(null, entity.id, 0, null);
```
FuriousLlama, line 62: this constructor is being called with the same hardcoded parameters, that look like a codesmell. instead how about using a static method that builds the object using only the entity id?

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSalePurchasersSubForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+            data-testid="add-purchaser-button"
+            className="mb-3"
+            onClick={() => {
+              const purchaserContact = new DispositionSaleContactModel(
+                null,
+                dispositionSaleId,
+                0,
+                null,
+              );
+              arrayHelpers.push(purchaserContact);
```
FuriousLlama, line 97: dame here

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <StyledFormWrapper>
+      <Formik<DispositionSaleFormModel>
+        enableReinitialize
+        validationSchema={DispositionSaleFormYupSchema}
+        initialValues={initialValues}
+        onSubmit={async (values: DispositionSaleFormModel, formikHelpers) => {
+          try {
+            const sale = await onSave(values.toApi());
+            if (sale && sale.id) {
```
FuriousLlama, line 60: note that the sale id could be zero

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const StyledContent = styled.div`
+  background-color: ${props => props.theme.css.filterBackgroundColor};
+`;
+
+const StyledFooter = styled.div`
+  margin-right: 1rem;
+  padding-bottom: 1rem;
+  z-index: 0;
+`;
```
FuriousLlama, line 121: any chance this is duplicated?

source/backend/api/Services/DispositionFileService.cs
FuriousLlama: that is not true. Please move api validation to the controller

FuriousLlama (changes requested):
based on comments

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionFileSaleMap.cs
eddherrera: Entities per the scaffolding are one to many, (which is not true) the relationship should be 1 to 1.

FuriousLlama: ah, thats not good. If we need db changes does Dough know?

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserAgentModel.cs
eddherrera: For new entities I am passing null in the model and letting EF to resolve the primary key to be generated.

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserModel.cs
eddherrera: For new entities I am passing null in the model and letting EF to resolve the primary key to be generated.

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserAgentModel.cs
FuriousLlama: if the ef model does not allow nulls, then the mapper will assign a zero to it. It is probably better to have that being explicit on the api model too

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSaleForm.tsx
eddherrera: Updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
eddherrera: Very likely, we could create a story to define a new css definition for these kind of components.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
eddherrera: that's ok, when Zero it would mean the Api request call wasn't successful.

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
eddherrera: Updated

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleFormModel.ts
eddherrera: I would rather just simplify the constructor instead of adding another method.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/form/DispositionSalePurchasersSubForm.tsx
eddherrera: Updated

source/frontend/src/features/mapSideBar/disposition/models/DispositionSaleContactModel.ts
eddherrera: Updated to new models

source/backend/api/Areas/Disposition/Controllers/DispositionFileController.cs
eddherrera: updated

source/backend/api/Services/DispositionFileService.cs
devinleighsmith: Right, the service probably only needs to know about the dispositionSale. If the service is always expecting the dispositionSale (which contains the saleId and dispositionFileId) to always match - we shouldn't need to pass those to the service at all.

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
FuriousLlama: under that logic it should check for a zero as well

source/backend/dal/Repositories/DispositionFileRepository.cs
devinleighsmith: In general we want to think of these methods as being independent from their calling services. The way this method is written, there is a hidden requirement that the saleId should match the dispositionSale.Id which is not guaranteed. The hidden requirement is met by the one service calling this method, but future services may want to use it who may not be doing this validation.

Also, In general, reducing a methods parameters to only what is required is a good rule of thumb.

source/backend/api/Services/DispositionFileService.cs
devinleighsmith: Another way to think of this is using Single Responsibility Principle. The only reason we have all of these parameters in the first place is because the controller needs them to be structured in a consistent manner for REST. Therefore, if these parameters are only **necessary** in the controller, they should not be passed onto other areas of the system.

eddherrera: Updated

source/backend/api/Services/DispositionFileService.cs
eddherrera: Updated

source/backend/apimodels/Models/Concepts/DispositionFile/DispositionSalePurchaserAgentModel.cs
eddherrera: updated

source/frontend/src/features/mapSideBar/disposition/tabs/offersAndSale/dispositionSale/update/UpdateDispostionSaleView.tsx
eddherrera: updated

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -246,6 +246,34 @@ public PimsDispositionSale GetDispositionFileSale(long dispositionFileId)
             return _dispositionFileRepository.GetDispositionFileSale(dispositionFileId);
         }
 
+        public PimsDispositionSale UpdateDispositionFileSale(long dispositionFileId, long saleId, PimsDispositionSale dispositionSale)
```
FuriousLlama, line 249: saleId and dispositionfileid are no longer necessary

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -246,6 +246,34 @@ public PimsDispositionSale GetDispositionFileSale(long dispositionFileId)
...
+            _logger.LogInformation("Updating disposition file Sale with DispositionFileId: {id}", dispositionFileId);
+            _user.ThrowIfNotAuthorized(Permissions.DispositionEdit);
+
+            var updatedSale = _dispositionFileRepository.UpdateDispositionFileSale(dispositionSale);
+            _dispositionFileRepository.CommitTransaction();
+
+            return updatedSale;
+        }
+
+        public PimsDispositionSale AddDispositionFileSale(long dispositionFileId, PimsDispositionSale dispositionSale)
```
FuriousLlama, line 260: does the dispositionsale have an invalid dispositionfileId? otherwise the file id does not need to be passed separately into the service

source/backend/api/Services/DispositionFileService.cs
eddherrera: Updated

source/backend/api/Services/DispositionFileService.cs
eddherrera: Updated

FuriousLlama (approved):

PR #3735, opened 2024-01-22
PSP-7509 : Allow Contractor to Create or Edit Disposition File

eddherrera:


source/backend/api/Helpers/Extensions/DispositionFileExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            ArgumentNullException.ThrowIfNull(dispositionFile);
+
+            ArgumentNullException.ThrowIfNull(principal);
+
+            var pimsUser = userRepository.GetUserInfoByKeycloakUserId(principal.GetUserKey());
+
+            if (pimsUser?.IsContractor == true && !dispositionFile.PimsDispositionFileTeams.Any(x => x.PersonId == pimsUser.PersonId))
+            {
+                throw new UserOverrideException(UserOverrideCode.ContractorNotInTeam, "As a Contractor your user contact information should be assigned to the Disposition File's team");
```
devinleighsmith, line 23: This doesn't seem to match the error text provided in the story:
As a contractor, you must add yourself as a team member to the file in order to create or save changes.

Also, the story specifies that this should be thrown as an error. The contract user should not be allowed to override this error - the system should completely disallow saving a file when a contractor is creating it and not on the team.

UserOverrideExceptions should only be used when the user is being shown a warning that can be optionally overriden.

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -52,12 +55,14 @@ public class DispositionFileService : IDispositionFileService
...
             _lookupRepository = lookupRepository;
             _checklistRepository = checklistRepository;
+            _userRepository = userRepository;
         }
 
         public PimsDispositionFile Add(PimsDispositionFile dispositionFile, IEnumerable<UserOverrideCode> userOverrides)
         {
             _logger.LogInformation("Creating Disposition File {dispositionFile}", dispositionFile);
             _user.ThrowIfNotAuthorized(Permissions.DispositionAdd);
+            dispositionFile.ThrowMissingContractorInTeam(_user, _userRepository);
```
devinleighsmith, line 65: is this function logically equivalent to: _user.ThrowInvalidAccessToDispositionFile(_userRepository, _dispositionFileRepository, id);
?
If so I'd rather not have two functions that provide the same functionality. Especially given that this function seems to require that the disposition team be provided with the file which is not a guarantee.

source/frontend/src/models/api/UserOverrideCode.ts
```diff
@@ -3,6 +3,7 @@ export enum UserOverrideCode {
   ADD_LOCATION_TO_PROPERTY = 'ADD_LOCATION_TO_PROPERTY',
   UPDATE_REGION = 'UPDATE_REGION',
   PROPERTY_OF_INTEREST_TO_INVENTORY = 'PROPERTY_OF_INTEREST_TO_INVENTORY',
+  CONTRACTOR_NOT_IN_TEAM = 'CONTRACTOR_NOT_IN_TEAM',
```
devinleighsmith, line 6: Both of the below two override codes seem suspect, as a user should not be able to override not being in the team or removing themselves.

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -88,6 +94,7 @@ public PimsDispositionFile Update(long id, PimsDispositionFile dispositionFile,
 
             _logger.LogInformation("Updating acquisition file with id {id}", id);
             _user.ThrowIfNotAuthorized(Permissions.DispositionEdit);
+            _user.ThrowInvalidAccessToDispositionFile(_userRepository, _dispositionFileRepository, id);
```
devinleighsmith, line 97: Hmm, doesn't this one also need to know if the incoming disposition file is removing the current contractor from the disposition team.

source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
```diff
@@ -11,18 +11,28 @@ public static class PrincipalExtensions
...
+        public static void ThrowInvalidAccessToDispositionFile(this ClaimsPrincipal principal, IUserRepository userRepository, IDispositionFileRepository dispositionFileRepository, long dispostionFileId)
+        {
+            ArgumentNullException.ThrowIfNull(principal);
+
+            var pimsUser = userRepository.GetUserInfoByKeycloakUserId(principal.GetUserKey());
+            PimsDispositionFile dispostionFile = dispositionFileRepository.GetById(dispostionFileId);
+
+            if (pimsUser?.IsContractor == true && !dispostionFile.PimsDispositionFileTeams.Any(x => x.PersonId == pimsUser.PersonId))
+            {
+                throw new NotAuthorizedException("Contractor is not assigned to the Disposition File's team");
```
devinleighsmith, line 34: This seems better, if a contractor does not have access to a file they should see an error not an override.

source/frontend/src/hooks/useApiUserOverride.ts
```diff
@@ -70,59 +70,75 @@ export const useApiUserOverride = <
...
-        title: 'Note',
-        variant: 'info',
-        message: RemoveSelfContractorContent(),
-        handleOk: async () => {
-          setState({
-            previousUserOverrideCodes: [...state.previousUserOverrideCodes],
-            userOverrideCode: null,
-            message: null,
+          break;
+        case UserOverrideCode.CONTRACTOR_SELFREMOVED:
```
devinleighsmith, line 92: Yes, the fact that the userOverrideCode is null here is a red flag. Why are we using the user override logic when the user isn't able to override anything?

source/frontend/src/hooks/useApiUserOverride.ts
```diff
@@ -70,59 +70,75 @@ export const useApiUserOverride = <
...
+        case UserOverrideCode.CONTRACTOR_NOT_IN_TEAM:
+          setModalContent({
+            title: 'Contractor missing',
+            variant: 'warning',
+            message:
+              'As a contractor, you must add yourself as a team member to the file in order to create or save changes.',
+            handleOk: async () => {
+              setState({
+                previousUserOverrideCodes: [...state.previousUserOverrideCodes],
+                userOverrideCode: null,
```
devinleighsmith, line 84: same problem here as below.

source/frontend/src/hooks/useApiUserOverride.ts
devinleighsmith: > Yes, the fact that the userOverrideCode is null here is a red flag. Why are we using the user override logic when the user isn't able to override anything?

This code demonstrates the correct way to handle network errors (non 409 responses) in concert with the UserOverride logic:

`withUserOverride(
          (userOverrideCodes: UserOverrideCode[]) =>
            handleSubmit(values, formikHelpers, userOverrideCodes),
          [],
          (axiosError: AxiosError<IApiError>) => { // this function is executed when a non-409 400 error code is returned.
            setModalContent({
              variant: 'warning',
              title: 'Warning',
              message: axiosError?.response?.data.error,
              okButtonText: 'Close',
            });
            setDisplayModal(true);
          },
        )`

source/backend/api/Services/DispositionFileService.cs
eddherrera: that's another story.

source/backend/api/Helpers/Extensions/DispositionFileExtensions.cs
eddherrera: Updated

source/frontend/src/models/api/UserOverrideCode.ts
eddherrera: Removed

source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
eddherrera: all good

source/frontend/src/hooks/useApiUserOverride.ts
eddherrera: Updated

source/frontend/src/hooks/useApiUserOverride.ts
eddherrera: Updated

source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
```diff
@@ -11,18 +11,28 @@ public static class PrincipalExtensions
...
                 throw new NotAuthorizedException("Contractor is not assigned to the Acquisition File's team");
             }
         }
+
+        public static void ThrowInvalidAccessToDispositionFile(this ClaimsPrincipal principal, IUserRepository userRepository, IDispositionFileRepository dispositionFileRepository, long dispostionFileId)
+        {
+            ArgumentNullException.ThrowIfNull(principal);
+
+            var pimsUser = userRepository.GetUserInfoByKeycloakUserId(principal.GetUserKey());
+            PimsDispositionFile dispostionFile = dispositionFileRepository.GetById(dispostionFileId);
```
devinleighsmith, line 30: nit: dispostion

devinleighsmith (approved):

source/backend/api/Helpers/Extensions/PrincipalExtensions.cs
eddherrera: Fixing on next pr.

PR #3739, opened 2024-01-23
Tech debt| use new api files and null enhancements

FuriousLlama:
Note, the commits were divided into sets that can be skimmed 'minor' or others that require a closer look 'major',

Where possible, the changes were only made across the api boundary and react components where intended to remain intact.


source/backend/apimodels/Models/Concepts/AcquisitionFile/AcquisitionFileModel.cs
```diff
@@ -96,7 +96,7 @@ public class AcquisitionFileModel : FileWithChecklistModel
         /// <summary>
         /// get/set - A list of research property relationships.
         /// </summary>
-        public IList<AcquisitionFilePropertyModel> FileProperties { get; set; }
+        public new IList<AcquisitionFilePropertyModel> FileProperties { get; set; }
```
devinleighsmith, line 99: I'm not familiar with this syntax. Why is it only being applied to this one list?

source/backend/apimodels/Models/Concepts/File/FilePropertyModel.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public PropertyModel Property { get; set; }
+
+        /// <summary>
+        /// get/set - The relationship's property id.
+        /// </summary>
+        public long PropertyId { get; set; }
+
+        /// <summary>
+        /// get/set - The relationship's acquisition file id.
```
devinleighsmith, line 36: nit: comment accuracy.

devinleighsmith:
Not sure where in this PR this was caused, but I'm unable to save an expiry date during a lease update **when changing the expiry date from null to a non-null value** seems to work fine when changing a pre-existing value to another value.

Also, dates are defaulting to Jan 1, 1970 on the lease form. They should not be defaulting at all.

devinleighsmith:
Adding a property to a lease seems to be broken. When creating a lease if I drop a marker on the map it does not show up on the lease list.

devinleighsmith:
I'm getting render infinite loops on some leases, trying to figure out why. This statement is certainly part of it but maybe not the root cause:

React.useEffect(() => {
    setFilePropertyLocations(locations);
  }, [setFilePropertyLocations, locations]);

devinleighsmith:
> I'm getting render infinite loops on some leases, trying to figure out why. This statement is certainly part of it but maybe not the root cause:
> 
> React.useEffect(() => { setFilePropertyLocations(locations); }, [setFilePropertyLocations, locations]);

Actually I figured it out. All string references in PropertiesInformation.tsx of 'properties' should be 'fileProperties'.

devinleighsmith:
existing terms aren't showing up on the front end (ie. a term exists in the db, and is returned by the backend, but the frontend does not display it).

source/frontend/src/models/api/generated/ApiGen_Concepts_Lease.ts
```diff
@@ -30,23 +27,21 @@ export interface ApiGen_Concepts_Lease extends ApiGen_Base_BaseAudit {
...
   categoryType: ApiGen_Base_CodeType<string> | null;
   purposeType: ApiGen_Base_CodeType<string> | null;
-  statusType: ApiGen_Base_CodeType<string> | null;
   region: ApiGen_Base_CodeType<number> | null;
   programType: ApiGen_Base_CodeType<string> | null;
   returnNotes: string | null;
   responsibilityEffectiveDate: string | null;
-  properties: ApiGen_Concepts_PropertyLease[] | null;
-  securityDeposits: ApiGen_Concepts_SecurityDeposit[] | null;
+  fileProperties: ApiGen_Concepts_PropertyLease[] | null;
```
devinleighsmith, line 41: this change seems to be the cause of lease properties not working as well as the hard crashes.

devinleighsmith:
something got borked in the css:

image
vs current dev:
image

source/frontend/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+      lastModified = item;
+    }
+  }
+
+  return lastModified;
+}
+
+export function isDefaultState(checklistItem: ApiGen_Concepts_FileChecklistItem): boolean {
+  return (
+    checklistItem.statusTypeCode?.id === 'INCOMP' &&
```
devinleighsmith, line 36: at some point we'll probably want a story to replace all of these with their generated enum values. Don't get me wrong though, this PR is large enough as it is.

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -57,54 +57,53 @@ export const getPropertyName = (property: IMapProperty): PropertyName => {
...
-  if (
-    fileProperty.propertyName !== undefined &&
-    fileProperty.propertyName !== null &&
-    fileProperty.propertyName !== '' &&
-    skipName === false
-  ) {
+  if (exists(fileProperty.propertyName) && fileProperty.propertyName !== '' && skipName === false) {
     return { label: NameSourceType.NAME, value: fileProperty.propertyName ?? '' };
-  } else if (!!fileProperty.property) {
+  } else if (exists(fileProperty.propertyName)) {
```
devinleighsmith, line 86: this looks like a functional change to me.

devinleighsmith:
Something is wrong with the method that formats property names:

image

the frontend is returning undefined as a property_name but for some reason the frontend is displaying that as an empty string instead of just displaying the PID.

devinleighsmith:
similar problem on interest holder. This property has no name, but has a pid. Displaying as n/a for some reason:

image

devinleighsmith:
Looks like there may be a regression in interest holder as well. Selecting a person belonging to an org seems to create an interest holder person with a primary contact of themselves.

source/frontend/src/features/leases/detail/LeasePages/insurance/details/Insurance.tsx
```diff
@@ -31,18 +31,18 @@ const InsuranceDetailsView: React.FunctionComponent<
     <div data-testid="insurance-section">
       <Section header="Required insurance">
         <InsuranceTypeList>
-          {sortedInsuranceList.map((insurance: Api_Insurance, index: number) => (
+          {sortedInsuranceList.map((insurance: ApiGen_Concepts_Insurance, index: number) => (
             <li key={`insurance-section-${insurance?.id?.toString() ?? index}`}>
-              {insurance.insuranceType.description}
-              {insurance.insuranceType.id === 'OTHER' && insurance.otherInsuranceType
+              {insurance.insuranceType?.description}
```
devinleighsmith, line 36: interesting that this is coming from the generated models? as far as I can tell this is a required, non-nullable field in the DB.

source/frontend/src/features/leases/detail/LeasePages/tenant/PrimaryContactWarningModal.test.tsx
```diff
@@ -69,9 +69,9 @@ describe('PrimaryContactWarningModal component', () => {
         ...getMockApiLease(),
         tenants: [
           {
-            ...getMockApiLease().tenants[0],
-            primaryContactId: undefined,
-            primaryContact: undefined,
+            ...getMockApiLease().tenants![0],
```
devinleighsmith, line 72: is the ! really necessary?

source/frontend/src/features/mapSideBar/property/tabs/propertyDetailsManagement/activity/list/models/PropertyActivityRow.ts
```diff
@@ -1,10 +1,10 @@
-import { Api_PropertyActivity } from '@/models/api/PropertyActivity';
-import Api_TypeCode from '@/models/api/TypeCode';
+import { ApiGen_Base_CodeType } from '@/models/api/generated/ApiGen_Base_CodeType';
+import { ApiGen_Concepts_PropertyActivity } from '@/models/api/generated/ApiGen_Concepts_PropertyActivity';
 
 export class PropertyActivityRow {
-  activityType: Api_TypeCode<string> | null = null;
-  activitySubType: Api_TypeCode<string> | null = null;
-  activityStatusType: Api_TypeCode<string> | null = null;
+  activityType: ApiGen_Base_CodeType<string> | null = null;
```
devinleighsmith, line 5: note: confirm this is an equivalent entity.

source/frontend/src/components/common/form/UserRegionSelect/UserRegionSelectContainer.tsx
```diff
@@ -26,10 +27,10 @@ export const UserRegionSelectContainer: React.FunctionComponent<
   );
   const { retrieveUserInfo, retrieveUserInfoResponse } = useUserInfoRepository();
   const regionTypes = getOptionsByType(API.REGION_TYPES);
-  const userRegionCodes = retrieveUserInfoResponse?.userRegions?.map(ur =>
-    ur.regionCode?.toString(),
-  );
-  const userRegionTypes = regionTypes.filter(r => userRegionCodes?.includes(r.code));
+  const userRegionCodes =
```
devinleighsmith, line 30: hmm, not so sure about this one. doesn't seem like a non-functional change. seems fine overall but the r.code ?? '' seems odd.

source/frontend/src/features/admin/edit-user/EditUserContainer.tsx
```diff
@@ -50,17 +51,17 @@ const EditUserContainer: React.FunctionComponent<IEditUserContainerProps> = ({ u
     regions: [],
     note: '',
     position: '',
-    userTypeCode: { id: UserTypes.Contractor },
+    userTypeCode: toTypeCodeNullable(UserTypes.Contractor) ?? undefined,
```
devinleighsmith, line 54: seems a bit verbose. is there a non-null version of toTypeCode that we could use here instead since we are passing a constant?

source/frontend/src/features/contacts/contact/detail/utils.ts
```diff
@@ -28,10 +28,10 @@ export function getContactInfo(
   // Get only the valid types
   let filteredFields = contactMethods.reduce(
     (accumulator: ContactInfoField[], method: IContactMethod) => {
-      if (Object.keys(validTypes).includes(method.contactMethodType.id)) {
+      if (Object.keys(validTypes).includes(method.contactMethodType?.id ?? '')) {
```
devinleighsmith, line 31: this also seems funky, I don't think we should be trying to match empty string with validtypes if the contactmethodtype or the id is null/undefined.

source/frontend/src/features/contacts/contact/detail/utils.ts
```diff
@@ -28,10 +28,10 @@ export function getContactInfo(
...
         accumulator.push({
           info: method.value,
-          label: validTypes[method.contactMethodType.id],
+          label: validTypes[method.contactMethodType?.id ?? ''],
```
devinleighsmith, line 34: this also doesn't make much sense to me.

source/frontend/src/features/disposition/list/DispositionSearchResults/columns.tsx
```diff
@@ -88,37 +85,33 @@ export const columns: ColumnWithProps<DispositionSearchResultModel>[] = [
...
-            person: x[0].person || {},
-            organization: null,
-            roles: x
-              .map(t => t.teamProfileType)
-              .filter((z): z is Api_TypeCode<string> => z !== undefined)
-              .flatMap(y => y.description || ''),
-          };
-        })
+        .groupBy((groupedTeams: ApiGen_Concepts_DispositionFileTeam) => groupedTeams.personId)
+        .map<MemberRoleGroup>(x => ({
```
devinleighsmith, line 93: oh boy, this looks sketchy, this must have been missed in review. Not needing to be corrected in your PR.

source/frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
```diff
@@ -11,44 +11,47 @@ import TooltipIcon from '@/components/common/TooltipIcon';
...
   public paidDate: string;
-  public contactHolder?: Api_Contact;
+  public contactHolder?: ApiGen_Concepts_Contact;
   public depositReturnCount: number;
 
-  public constructor(baseDeposit: Api_SecurityDeposit) {
+  public constructor(baseDeposit: ApiGen_Concepts_SecurityDeposit) {
     this.id = baseDeposit.id || -1;
-    if (baseDeposit.depositType.id === 'OTHER') {
+    if (baseDeposit.depositType?.id === 'OTHER') {
```
devinleighsmith, line 30: depositType is also a non-nullable field on PIMS_SECURITY_DEPOSIT. Why is this coming from the generated entity?

source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
```diff
@@ -135,12 +136,10 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
     const acquisitionProperties = await acquisitionPropertiesTask;
     const acquisitionChecklist = await acquisitionChecklistTask;
 
-    if (retrieved) {
-      retrieved.fileProperties = acquisitionProperties;
-      retrieved.fileChecklistItems = acquisitionChecklist ?? [];
-      setFile({ ...retrieved, fileType: FileTypes.Acquisition });
-      setStaleFile(false);
-    }
+    retrieved.fileProperties = acquisitionProperties ?? null;
```
devinleighsmith, line 139: my instinct here is ?? []. why ?? null?

source/frontend/src/features/mapSideBar/acquisition/common/GenerateForm/hooks/useGenerateH120.ts
```diff
@@ -83,7 +83,7 @@ export const useGenerateH120 = () => {
     if (!file) {
       throw Error('Acquisition file not found');
     }
-    file.fileProperties = properties;
+    file.fileProperties = properties ?? null;
```
devinleighsmith, line 86: same here.

source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
```diff
@@ -38,7 +38,7 @@ export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderPr
           <Row className="no-gutters">
             <Col xs="8">
               <HeaderField label="Civic Address:" labelWidth={'3'} contentWidth="9">
-                {apiProperty?.address !== undefined ? formatApiAddress(apiProperty?.address) : '-'}
+                {exists(apiProperty?.address) ? formatApiAddress(apiProperty!.address) : '-'}
```
devinleighsmith, line 41: this one probably needs your todo comment.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetails/update/models/UpdatePropertyDetailsFormModel.ts
```diff
@@ -129,99 +133,116 @@ export class UpdatePropertyDetailsFormModel {
...
-      tenures: this.tenures?.map(e => e.toApi()),
-      roadTypes: this.roadTypes?.map(e => e.toApi()),
-      adjacentLands: this.adjacentLands?.map(e => e.toApi()),
+      anomalies: this.anomalies?.map(e => e.toApi()) ?? null,
+      tenures: this.tenures?.map(e => e.toApi()) ?? null,
+      roadTypes: this.roadTypes?.map(e => e.toApi()) ?? null,
+
+      boundary: null,
+      dataSource: null,
+      dataSourceEffectiveDateOnly: EpochISODateTimeString,
```
devinleighsmith: Hmm, I guess this is the equivalent of setting null for two required fields? would we be better storing and passing of the actual value if we have it?

source/frontend/src/features/mapSideBar/property/tabs/propertyResearch/update/models.ts
```diff
@@ -41,65 +44,64 @@ export class UpdatePropertyFormModel {
...
+    model.isLegalOpinionObtained = !exists(base.isLegalOpinionObtained)
+      ? 'unknown'
+      : base.isLegalOpinionObtained
+      ? 'yes'
+      : 'no';
+    model.documentReference = base.documentReference ?? undefined;
+    model.researchSummary = base.researchSummary ?? undefined;
     model.propertyId = base.property?.id;
-    model.researchFileId = base.file?.id;
-    model.researchFileVersion = base.file?.rowVersion;
```
devinleighsmith, line 70: lost this, not sure if that was intentional.

source/frontend/src/features/mapSideBar/property/tabs/propertyResearch/update/models.ts
```diff
@@ -41,65 +44,64 @@ export class UpdatePropertyFormModel {
...
           : false,
-      documentReference: this.documentReference,
-      researchSummary: this.researchSummary,
-      property: { id: this.propertyId },
-      file: { id: this.researchFileId, rowVersion: this.researchFileVersion },
-      purposeTypes: this.purposeTypes?.map(x => x.toApi()),
-      rowVersion: this.rowVersion,
+      documentReference: this.documentReference ?? null,
+      researchSummary: this.researchSummary ?? null,
+      property: null,
```
devinleighsmith, line 100: this doesn't seem passive.

source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
```diff
@@ -285,7 +297,7 @@ export const ResearchContainer: React.FunctionComponent<
                   show
                 />
                 <ResearchView
-                  researchFile={researchFile}
+                  researchFile={researchFile as unknown as ApiGen_Concepts_ResearchFile | undefined} //todo : clean the forced upcasting
```
devinleighsmith, line 300: were you intending to leave this in this PR unresolved?

source/frontend/src/features/mapSideBar/shared/detail/PropertyFileContainer.tsx
```diff
@@ -80,7 +81,10 @@ export const PropertyFileContainer: React.FunctionComponent<
   if (props.fileContext === FileTypes.Research) {
     tabViews.push({
       content: (
-        <PropertyResearchTabView researchFile={props.fileProperty} setEditMode={props.setEditing} />
+        <PropertyResearchTabView
+          researchFile={props.fileProperty as ApiGen_Concepts_ResearchFileProperty}
```
devinleighsmith, line 85: may want a todo to do an is check in the future.

source/frontend/src/features/notes/models.ts
```diff
@@ -12,31 +13,32 @@ export class NoteForm {
...
-      rowVersion: this.rowVersion,
-      appCreateTimestamp: this.appCreateTimestamp,
-      appCreateUserGuid: this.appCreateUserGuid,
-      appCreateUserid: this.appCreateUserid,
-      appLastUpdateTimestamp: this.appLastUpdateTimestamp,
-      appLastUpdateUserGuid: this.appLastUpdateUserGuid,
-      appLastUpdateUserid: this.appLastUpdateUserid,
+      id: this.id ?? 0,
+      note: this.note ?? null,
+      isSystemGenerated: false,
```
devinleighsmith, line 34: should we not pass the value that we have?

source/frontend/src/models/default_initializers.ts
```diff
@@ -0,0 +1,223 @@
+import moment from 'moment';
```
devinleighsmith, line 1: what's with the file name?

devinleighsmith (approved):

devinleighsmith (changes requested):
whoops, wrong button. still reviewing.

source/frontend/src/features/mapSideBar/router/FilePropertyRouter.tsx
```diff
@@ -55,12 +57,12 @@ export const FilePropertyRouter: React.FC<IFilePropertyRouterProps> = props => {
...
         <Route exact path={`${path}/${InventoryTabNames.property}`}>
           {(() => {
-            if (fileProperty?.property?.id === undefined) {
+            if (!isValidId(fileProperty?.property?.id)) {
               throw Error('Cannot edit property without a valid id');
             }
             return (
               <UpdatePropertyDetailsContainer
-                id={fileProperty?.property?.id}
+                id={fileProperty!.property!.id}
```
devinleighsmith, line 65: maybe a typescript todo here as well?

devinleighsmith:
Getting the following when submitting an access request:

image

steps to reproduce:
1) Submit an access request (manually navigate to /access/request)
2) Decline the access request
3) Attempt to re-submit, get the error.

Unable to repro in dev env.

source/frontend/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function lastModifiedBy(
+  array: ApiGen_Base_BaseAudit[] = [],
+): ApiGen_Base_BaseAudit | undefined {
+  let lastModified: ApiGen_Base_BaseAudit | undefined = undefined;
+
+  for (const item of array) {
+    if (moment(item.appLastUpdateTimestamp).isBefore(moment('1900-01-01T00:00:00.000Z'))) {
+      continue;
```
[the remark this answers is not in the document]
FuriousLlama: that would be good. The change just moves the existing logic to this new file

source/frontend/src/utils/fileUtils.ts
FuriousLlama: agreed! note that this change just moves that existing logic to this new file

source/frontend/src/utils/mapPropertyUtils.ts
FuriousLlama: yep, it was a mistake. A later commit fixes it

source/frontend/src/features/leases/detail/LeasePages/tenant/PrimaryContactWarningModal.test.tsx
FuriousLlama: yes. there is a problem with the current version of typescript that we are using. once we move to a newer version these ! wont be necessary

source/frontend/src/features/mapSideBar/property/tabs/propertyDetails/update/models/UpdatePropertyDetailsFormModel.ts
FuriousLlama: that would be the ideal. in this case I chose to preserve the existing functionality as much as posible

source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
FuriousLlama: yes. the previous version was ok with this kind of upcasting but this one requires a forced one. We need to look deeper into how to do the casting more cleanly

devinleighsmith:
It appears now that terms with null end dates are showing up like this:

image

(null end dates should be allowed for lease terms, it means that the term has no expiry).

devinleighsmith:
> Not sure where in this PR this was caused, but I'm unable to save an expiry date during a lease update **when changing the expiry date from null to a non-null value** seems to work fine when changing a pre-existing value to another value.
> 
> Also, dates are defaulting to Jan 1, 1970 on the lease form. They should not be defaulting at all.

The first issue still exists, I'm unable to change a lease expiry that is currently null

devinleighsmith:
sorry, wrong button again.

devinleighsmith (approved):

source/frontend/src/features/leases/detail/LeasePages/insurance/details/Insurance.tsx
FuriousLlama: I think the problem is that string is still an object on the backend and that could be nullable according to our nullable checks

source/frontend/src/components/common/form/UserRegionSelect/UserRegionSelectContainer.tsx
FuriousLlama: I agree. I left it like that to maintain curent functionalty. Otherwise, filtering valid codes could cause a sideeffect

source/frontend/src/features/admin/edit-user/EditUserContainer.tsx
FuriousLlama: good catch

source/frontend/src/features/leases/detail/LeasePages/deposits/components/DepositsReceivedContainer/columns.tsx
FuriousLlama: because the codetype is an object the backend does not guarantee that it wont be null

source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
FuriousLlama: right now we assume that lists could be null. We could change the backend serializer to return empty lists instead

source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
FuriousLlama: I think we can do a search for ".!" and match places where we are ignoring the compiler

source/frontend/src/features/mapSideBar/property/tabs/propertyResearch/update/models.ts
FuriousLlama: this was only used to store the current research file. the object file does contain all that information so the individual members are not needed

source/frontend/src/features/mapSideBar/property/tabs/propertyResearch/update/models.ts
FuriousLlama: the backend only cares about the property id, In that light setting only that member makes more sense than creating a property that only contains that information. The mapper is aligned with that.

source/frontend/src/features/notes/models.ts
FuriousLlama: it should. but at this point, the form does not contain that information.

PR #3740, opened 2024-01-24
psp-7275 complete a disposition file. psp-7785, psp-7322, psp-7568

devinleighsmith:


source/backend/api/Services/DispositionFileService.cs
```diff
@@ -97,16 +100,31 @@ public PimsDispositionFile Update(long id, PimsDispositionFile dispositionFile,
...
-                {
-                    throw new UserOverrideException(UserOverrideCode.DispositionFileFinalStatus, "You are changing this file to a non-editable state. Only system administrators can edit the file when set to Archived, Cancelled or Completed state). Do you wish to continue?");
-                }
+                throw new BusinessRuleViolationException("You have not added a Sales Price. Please add a Sales Price before completion.");
+            }
+            else if (currentDispositionFile.DispositionFileStatusTypeCode != EnumDispositionFileStatusTypeCode.COMPLETE.ToString()
+                && dispositionFile.DispositionFileStatusTypeCode == EnumDispositionFileStatusTypeCode.COMPLETE.ToString()
+                && currentDispositionFile.PimsDispositionFileProperties.Count > 0)
+            {
+                DisposeofProperties(dispositionFile);
```
FuriousLlama, line 123: nit disposeOfProperties

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -97,16 +100,31 @@ public PimsDispositionFile Update(long id, PimsDispositionFile dispositionFile,
...
-                var doNotAddToStatuses = new List<string>() { EnumDispositionFileStatusTypeCode.COMPLETE.ToString(), EnumDispositionFileStatusTypeCode.ARCHIVED.ToString() };
-                if(doNotAddToStatuses.Contains(dispositionFile.DispositionFileStatusTypeCode))
-                {
-                    throw new UserOverrideException(UserOverrideCode.DispositionFileFinalStatus, "You are changing this file to a non-editable state. Only system administrators can edit the file when set to Archived, Cancelled or Completed state). Do you wish to continue?");
-                }
+                throw new BusinessRuleViolationException("You have not added a Sales Price. Please add a Sales Price before completion.");
+            }
+            else if (currentDispositionFile.DispositionFileStatusTypeCode != EnumDispositionFileStatusTypeCode.COMPLETE.ToString()
+                && dispositionFile.DispositionFileStatusTypeCode == EnumDispositionFileStatusTypeCode.COMPLETE.ToString()
+                && currentDispositionFile.PimsDispositionFileProperties.Count > 0)
```
FuriousLlama, line 121: this looks like good logic to be put in a different solver

source/backend/api/Services/DispositionFileService.cs
devinleighsmith: this is a function, why would it be lowercase?

source/backend/api/Services/DispositionFileService.cs
devinleighsmith: To be honest I'd rather we discuss the "solver" pattern. I think that pattern was added to the code base without a group discussion.

source/backend/api/Services/DispositionFileService.cs
FuriousLlama: I meant capitalization of the O in Of

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -395,6 +395,7 @@ public PimsProperty TransferFileProperty(PimsProperty property, bool isOwned, bo
 
             existingProperty.IsPropertyOfInterest = isPropertyOfInterest;
             existingProperty.IsOwned = isOwned;
+            existingProperty.IsDisposed = isDisposed;
```
eddherrera, line 398: Besides from the Disposed Flag, shouldn't the PropertyClassificationTypeCode be updated to "DISPOSED" also?

devinleighsmith: We don't actually use the PropertyClassificationTypeCode, it is a holdover from CITZ pims.

eddherrera: Could we consider the idea to have a new method called "DisposeProperty"?

eddherrera (approved):

source/backend/api/Services/DispositionFileService.cs
eddherrera: Wouldn't have been better creating extension methods for the Disposition File instead of creating the "solver" approach to determine the Disposition file operability?

PR #3746, opened 2024-01-27
PSP-7547 Auto-select region for disposition file

asanchezr:
Also, prevent user from adding properties outside of their assigned regions:

Screenshot 2024-01-25 205406

### DIFF
--- a/source/backend/api/Areas/Disposition/Controllers/DispositionFileController.cs
+++ b/source/backend/api/Areas/Disposition/Controllers/DispositionFileController.cs
@@ -5,10 +5,8 @@ using MapsterMapper;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
-using Pims.Api.Areas.Acquisition.Controllers;
 using Pims.Api.Helpers.Exceptions;
 using Pims.Api.Models.Concepts.AcquisitionFile;
-using Pims.Api.Models.Concepts.DispositionFile;
 using Pims.Api.Models.Concepts.DispositionFile;
 using Pims.Api.Policies;
 using Pims.Api.Services;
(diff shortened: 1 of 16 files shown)


devinleighsmith:
@asanchezr this story will need to be held until the next IS as you are not able to satisfy all requirements:

image

(currently you are only displaying this message during creation but not during update).

asanchezr:
> @asanchezr this story will need to be held until the next IS as you are not able to satisfy all requirements:
> 
> image
> 
> (currently you are only displaying this message during creation but not during update).

I'm fine delaying this but my interpretation was that this was specifically for create (not update). we might need BAs to clarify. This story started with adding the region of the first property upon creation (not update) and then also included the restriction to not add a property outside of allowed region upon create. Also I think updating properties on Disposition Files does not currently work because it was missed requirements and there is no "/disposition/{id}/properties" endpoint at the moment

asanchezr:
As per current backend code - we only have a way to retrieve a disposition file properties (not a way to update them, ie no PUT)
image

devinleighsmith:
> > @asanchezr this story will need to be held until the next IS as you are not able to satisfy all requirements:
> > image
> > (currently you are only displaying this message during creation but not during update).
> 
> I'm fine delaying this but my interpretation was that this was specifically for create (not update). we might need BAs to clarify. This story started with adding the region of the first property upon creation (not update) and then also included the restriction to not add a property outside of allowed region upon create. Also I think updating properties on Disposition Files does not currently work because it was missed requirements and there is no "/disposition/{id}/properties" endpoint at the moment

Yes, exactly my point, there is no ability to edit properties on disposition files at the moment, and therefore this story cannot be completed as written.

Remember - for the sake of the FT group - we need to make sure that any of our assumptions / modifications to a story are reflected on that story. As a result, this story either needs to be updated to indicate what is achievable as restricted by our technical limitations, or we need to indicate implementation is not possible and hold the story (ie, shipping stories with known issues).

In this case the story pretty clear articulates requirements around property update, so we can't just deliver it as-is to the FT group right?

PR #3760, opened 2024-02-01
Vite 

eddherrera:
TODO's

- Fix msw implementation.
- Update images imports for images from assets to:  Legend.tsx and Layers/util.tsx
- Refactor or migrate Jest to Vitest


source/frontend/index.html
FuriousLlama, line 1: is it a requirement that the index file be on the root of the js project or can that be part of the setup?

eddherrera: It may be possible by Specifying Alternative Root

PR #3769, opened 2024-02-03
psp-7515 allow disposition properties to be added/updated.

devinleighsmith:


source/backend/api/Services/DispositionFileService.cs
```diff
@@ -431,6 +431,64 @@ public List<DispositionFileExportModel> GetDispositionFileExport(DispositionFilt
                 }).ToList();
         }
 
+        public PimsDispositionFile UpdateProperties(PimsDispositionFile dispositionFile, IEnumerable<UserOverrideCode> userOverrides)
+        {
+            _logger.LogInformation("Updating disposition file properties...");
+            _user.ThrowIfNotAuthorized(Permissions.DispositionEdit, Permissions.PropertyView, Permissions.PropertyAdd);
```
asanchezr, line 437: this is quite subtle but this extension method will NOT enforce all of those claims. If the user has **any** of those it will let if pass. The one you want is called something like ThrowIfNotAll or something. 
Alternatively, you could list individual claim checks like in 3 lines of ThrowIfNotAuthorized

I think the naming is confusing and we should address with some tech-debt (not this pr of course)

source/backend/tests/unit/dal/Repositories/DispositionFilePropertyRepositoryTest.cs
```diff
@@ -31,9 +31,9 @@ public void GetByDispositionFileId_Success()
             var user = PrincipalHelper.CreateForPermission(Permissions.DispositionView);
 
             var acqFile = EntityHelper.CreateDispositionFile();
-            var pimsPropertyDispositionFile = new PimsDispositionFileProperty() { DispositionFileId = acqFile.DispositionFileId, Property = EntityHelper.CreateProperty(1) };
+            var pimsDispositionFileProperty = new PimsDispositionFileProperty() { DispositionFileId = acqFile.DispositionFileId, Property = EntityHelper.CreateProperty(1) };
```
asanchezr, line 34: nit: acqFile -> dispFile

source/backend/tests/unit/dal/Repositories/DispositionFilePropertyRepositoryTest.cs
```diff
@@ -104,6 +104,126 @@ public void GetDispositionFilePropertyRelatedCount_NotFound()
...
+        }
+        #endregion
+
+        #region Update
+        [Fact]
+        public void Update_Success()
+        {
+            // Arrange
+            var helper = new TestHelper();
+            var user = PrincipalHelper.CreateForPermission(Permissions.DispositionAdd);
```
asanchezr, line 151: nit: Repositories don't check for permissions but for consistency it should be DispositionUpdate here as we are testing the update method

source/frontend/src/features/mapSideBar/disposition/DispositionContainer.tsx
```diff
@@ -25,6 +27,9 @@ export const DispositionContainer: React.FunctionComponent<IDispositionContainer
   const { setLastUpdatedBy, lastUpdatedBy, staleLastUpdatedBy, staleFile } =
     useContext(SideBarContext);
   const [isValid, setIsValid] = useState<boolean>(true);
+  const withUserOverride = useApiUserOverride<
+    (userOverrideCodes: UserOverrideCode[]) => Promise<any | void>
+  >('Failed to update Acquisition File');
```
asanchezr, line 32: 'Failed to update **Disposition File**'

asanchezr (approved):

PR #3776, opened 2024-02-06
psp-7746 do not remove the disposition file from props during loading, only if…

devinleighsmith:
… the file does not match.


source/frontend/src/features/mapSideBar/disposition/DispositionContainer.tsx
```diff
@@ -122,6 +122,7 @@ export const DispositionContainer: React.FunctionComponent<IDispositionContainer
       (!error && dispositionFileId !== dispositionFile?.id && !loadingDispositionFile) ||
       staleFile
     ) {
+      debugger;
```
eddherrera, line 125: remove debugger

eddherrera (approved):

PR #3783, opened 2024-02-09
psp-7726 guarantee same view/edit order by using id instead of timestamp. Ther…

devinleighsmith:
…e is no guarantee on insert order so rename take title during add.


source/frontend/src/features/mapSideBar/property/tabs/takes/detail/TakesDetailContainer.tsx
```diff
@@ -47,7 +47,7 @@ const TakesDetailContainer: React.FunctionComponent<ITakesDetailContainerProps>
   return (
     <View
       onEdit={onEdit}
-      takes={orderBy(takes, t => moment(t.appCreateTimestamp), 'desc') ?? []}
+      takes={orderBy(takes, t => moment(t.id), 'desc') ?? []}
```
eddherrera, line 50: is this right? if you are ordering by ID why does it need to be a Moment?

eddherrera (approved):

PR #3790, opened 2024-02-13
PSP-7681 : FT-REG: Browser alerts should be replaced by established modal dialogs

eddherrera:


source/frontend/src/features/contacts/contact/create/Organization/CreateOrganizationForm.tsx
```diff
@@ -132,9 +131,6 @@ const CreateOrganizationComponent: React.FC<FormikProps<IEditableOrganizationFor
 
   return (
     <>
-      {/* Router-based confirmation popup when user tries to navigate away and form has unsaved changes */}
-      <UnsavedChangesPrompt />
```
devinleighsmith, line 136: This is a tricky one, because our regular modal doesn't pop up when navigating away, only when the user clicks "cancel" can we confirm with Julian/Dev that what they really want is for that modal not to display?

PR #3806, opened 2024-02-21
Fix local development issues.

eddherrera:


asanchezr (approved):

FuriousLlama (approved):

devinleighsmith (approved):

tools/mayan-edms/docker-compose.yml
```diff
@@ -21,7 +21,7 @@ x-mayan-container: &mayan-container
 networks:
   psp:
     external: true
-    name: psp-net
```
devinleighsmith, line 24: this reminds me, we should really roll this docker-compose into our main one, having two shouldn't be necessary.

PR #3812, opened 2024-02-22
PSP-7676 | Updated the calculation of totals for property activity invoices

FuriousLlama:


devinleighsmith:
Oh, interesting, I was under the impression we weren't using those fields. Good catch.

devinleighsmith (approved):

PR #3821, opened 2024-02-28
psp-7622 Testing

FuriousLlama:


source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -313,6 +313,7 @@ export const mapMachine = createMachine<MachineContext>({
   // Machine identifier
   id: 'map',
   initial: 'notMap',
+  predictableActionArguments: true,
```
FuriousLlama, line 316: benign change. The Map FSM should have had this from day 0. With this we wont get any of those warnings on the test side

devinleighsmith (approved):

PR #3836, opened 2024-03-04
psp 7616 | Subdivision Consolidation navigation

FuriousLlama:


devinleighsmith (approved):

source/frontend/src/components/layout/SideNavBar/SubdCons.tsx
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Subdivision & Consolidation Tray page.
+ * Intended for use within the left side tray.
+ */
+export const SubdivisionConsolidationTray = ({ onLinkClick }: ISideTrayPageProps) => {
+  const { hasClaim } = useKeycloakWrapper();
+  return (
+    <>
+      <HalfHeightDiv>
+        <Styled.TrayHeader>Disposition Files</Styled.TrayHeader>
```
asanchezr, line 19: not sure this header is correct? I would expect it to be "Consolidation & Subdivision" - not disposition file,,,

asanchezr (approved):

source/frontend/src/components/layout/SideNavBar/SubdCons.tsx
FuriousLlama: good point. I will update it in the front end pr

PR #3840, opened 2024-03-05
PSP-7869 Indicate "retired" on property information details 

asanchezr:
image

### DIFF
--- a/source/backend/api/Services/PropertyService.cs
+++ b/source/backend/api/Services/PropertyService.cs
@@ -101,14 +101,8 @@ namespace Pims.Api.Services
 
         public PimsProperty Update(PimsProperty property, bool commitTransaction = true)
         {
-            _logger.LogInformation("Updating property...");
+            _logger.LogInformation("Updating property with id {id}", property.Internal_Id);
             _user.ThrowIfNotAuthorized(Permissions.PropertyEdit);
-
-            var existingProperty = GetById(property.Internal_Id);
-            if (existingProperty.IsRetired.HasValue && existingProperty.IsRetired.Value)
-            {
-                throw new BusinessRuleViolationException("Retired records are referenced for historical purposes only and cannot be edited or deleted.");
-            }
 
             // convert spatial location from lat/long (4326) to BC Albers (3005) for database storage
             var geom = property.Location;
--- a/source/backend/api/Services/PropertyService.cs
+++ b/source/backend/api/Services/PropertyService.cs
@@ -9,6 +9,7 @@ using Pims.Api.Constants;
 using Pims.Api.Helpers.Exceptions;
 using Pims.Api.Models.CodeTypes;
 using Pims.Api.Models.Concepts.Property;
+using Pims.Core.Exceptions;
 using Pims.Core.Extensions;
 using Pims.Dal.Entities;
 using Pims.Dal.Exceptions;
@@ -102,6 +103,12 @@ namespace Pims.Api.Services
         {
             _logger.LogInformation("Updating property...");
             _user.ThrowIfNotAuthorized(Permissions.PropertyEdit);
+
+            var existingProperty = GetById(property.Internal_Id);
+            if (existingProperty.IsRetired.HasValue && existingProperty.IsRetired.Value)
+            {
+                throw new BusinessRuleViolationException("Retired records are referenced for historical purposes only and cannot be edited or deleted.");
+            }
 
             // convert spatial location from lat/long (4326) to BC Albers (3005) for database storage
             var geom = property.Location;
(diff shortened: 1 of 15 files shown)


source/backend/api/Services/PropertyService.cs
```diff
@@ -103,6 +104,12 @@ public PimsProperty Update(PimsProperty property, bool commitTransaction = true)
             _logger.LogInformation("Updating property...");
             _user.ThrowIfNotAuthorized(Permissions.PropertyEdit);
 
+            var existingProperty = GetById(property.Internal_Id);
+            if (existingProperty.IsRetired.HasValue && existingProperty.IsRetired.Value)
```
eddherrera, line 108: I had considered putting the validation at the Repository level.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
```diff
@@ -63,19 +64,32 @@ export const PropertyDetailsTabView: React.FunctionComponent<IPropertyDetailsTab
...
+    }
+    return true;
+  };
+
   return (
     <StyledSummarySection>
       <LoadingBackdrop show={loading} parentScreen={true} />
       <StyledEditWrapper className="mr-3 my-1">
-        {hasClaim(Claims.PROPERTY_EDIT) && (
+        {hasClaim(Claims.PROPERTY_EDIT) && canEditDetails(property) ? (
```
eddherrera, line 78: My take is that users without the edit claim should still see nothing rendered and ONLY users with the edit claim would see the tooltip.

source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
```diff
@@ -63,6 +73,17 @@ export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderPr
               </HeaderField>
             </Col>
           </Row>
+          {isExpired && (
+            <Row className="no-gutters">
+              <Col xs="8"></Col>
+              <Col className="d-flex justify-content-end pr-4">
+                <ExpiredWarning>
+                  <AiOutlineExclamationCircle size={16} />
+                  &nbsp; EXPIRED
```
eddherrera, line 82: not a fan of the "&nbsp;". wouldn't it be better to use CSS to fix?

source/backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
```diff
@@ -521,6 +522,47 @@ public void Update_Property_Success()
             updatedProperty.Description.Should().Be("test");
             updatedProperty.Pid.Should().Be(200);
         }
+        [Fact]
```
eddherrera, line 525: Please just add one scenario where the existing property before update has Is_Retired as null. Since this will be the most likely scenario.

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -283,17 +284,24 @@ public PimsProperty GetAllAssociationsById(long id)
...
         {
             property.ThrowIfNull(nameof(property));
 
             var propertyId = property.Internal_Id;
-            var existingProperty = this.Context.PimsProperties
+            var existingProperty = Context.PimsProperties
                 .Include(p => p.Address)
                 .FirstOrDefault(p => p.PropertyId == propertyId) ?? throw new KeyNotFoundException();
 
+            // PSP-7876 prevent editing on retired properties
```
eddherrera, line 299: nit: we probably don't need this comment with the ticket number?

source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
asanchezr: This is how we display on other pages (ie Leases) but will give it a margin instead

source/backend/dal/Repositories/PropertyRepository.cs
asanchezr: We sometimes put the JIRA # to refer back to business logic that was relevant at the time. Will remove but this "retired" thing will become an integral part of the system and wanted to put that # there to reflect the business decision.

source/frontend/src/features/mapSideBar/property/tabs/propertyDetails/detail/PropertyDetailsTabView.tsx
asanchezr: updated

source/backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
asanchezr: updated

eddherrera (approved):

PR #3846, opened 2024-03-06
Psp 7984 node 20 update

devinleighsmith:


source/frontend/public/mockServiceWorker.js
FuriousLlama, line 1: this file keeps fliping on my machine. Any chance we have conflicting formatting rules?

devinleighsmith: potentially, you mentioned that you stopped using vscode, does that mean you aren't using prettier anymore?

asanchezr: this happens to me as well anytime Manuel commits changes. I keep ignoring the changes on my local but we should work with the same set of linting rules

asanchezr: We should exclude this file from linting regardless of local setup

asanchezr (approved):

PR #3848, opened 2024-03-07
PSP-7860, psp-7962, psp-7963, psp-7964 | Added consolidation history view and fixed issues

FuriousLlama:


devinleighsmith:
the subdivision/consolidation icon is 28px wide while the other icons are 24px (on the navigation pane).

devinleighsmith:
The hyperlinked pids appear incorrect. The story says:

**Each PID is hyperlinked to open new tab with property details page.** 

The hyperlinks do not open in a new tab.

devinleighsmith:
The uxpin mockups are pretty different from the implementation and are unlikely to pass FT without corrections. Please ask Ana to:
1) remove the markers on the subdivision/consolidation mockups
2) make the pid values hyperlinks and remove the square hyperlink icon in the header
3) move the m^2 from the header to be in every column

devinleighsmith:
Subdivision/consolidations do not show loading spinners when data is loading - they display the "there is no subdivision/consolidation" message, which isn't in line with our standard practices for loading data.

devinleighsmith:
For me, the consolidation section never loads any data, even when I have consolidations in my db they are ignored.

Also, it seems like the propertyOperationTypeCode field is not being mapped correctly. See screenshot:

image

devinleighsmith:
Note: my comments on ui were based on the mockup linked in the consolidation story. If those are no longer accurate they can be ignored.

devinleighsmith (approved):

PR #3849, opened 2024-03-07
PSP-8003 : FT:UI/UX- Agreements: Second word starting letter doesn't …

eddherrera:
…start with capital letter In the Agreement Details word


source/frontend/src/components/common/Section/Section.tsx
```diff
@@ -80,6 +80,7 @@ const StyledSectionHeader = styled.h2<{ isStyledHeader?: boolean }>`
   color: ${props => props.theme.css.primaryColor};
   border-bottom: 0.2rem ${props => props.theme.css.primaryColor} solid;
   margin-bottom: 2rem;
+  text-transform: capitalize;
```
FuriousLlama, line 83: This change is very broad. I am not convinced this is necessary given the ticket is about a single word in the entire system.

eddherrera: Based on the number of defects we get regarding the same issue I think it's worth considering.

asanchezr: I think it is worth considering but would like to see this change behind an (optional) flag so that the rest of components using **Section** can work as before and you can pass **true** for your bug fix. 

A wide change like that we should discuss I think

devinleighsmith: let's discuss in our dev meeting today.

PR #3853, opened 2024-03-08
PSP-7907 Add representation of RETIRED properties on the map view and advance filter

asanchezr:
retired-properties


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -493,10 +504,29 @@ public HashSet<long> GetMatchingIds(PropertyFilterCriteria filter)
             }
 
             // Property ownership filters
-            predicate.And(p => (p.IsOwned && filter.IsCoreInventory) ||
-                (p.IsPropertyOfInterest && filter.IsPropertyOfInterest) ||
-                (p.IsOtherInterest && filter.IsOtherInterest) ||
-                (p.IsDisposed && filter.IsDisposed));
+            var ownershipBuilder = PredicateBuilder.New<PimsProperty>(p => false);
+            if (filter.IsCoreInventory)
```
FuriousLlama, line 508: I really liked this logic more explicit

source/frontend/.prettierignore
FuriousLlama, line 1: +1

source/frontend/src/components/maps/leaflet/Control/AdvancedFilter/models.ts
```diff
@@ -49,6 +50,7 @@ export class PropertyFilterFormModel {
       isPropertyOfInterest: this.isPropertyOfInterest,
       isOtherInterest: this.isOtherInterest,
       isDisposed: this.isDisposed,
+      isRetired: this.isRetired,
```
FuriousLlama, line 53: Eventually we will probably need a more robust way of passing these flags

FuriousLlama (approved):

PR #3854, opened 2024-03-08
PSP-8004 : FT:UI/UX- Agreements: tooltip icon for Deposit due no late…

eddherrera:
…r than field in the view screen only is displaying in the second line

image


source/frontend/src/components/common/Section/SectionField.tsx
```diff
@@ -66,4 +66,7 @@ export const StyledCol = styled(Col)`
 
 export const StyledFieldLabel = styled.label`
   font-weight: bold;
+  span {
+    position: absolute;
+  }
```
asanchezr, line 71: again - this seems a broad change (like there are 89 changed snapshots)
Similarly to the other PR perhaps we should discuss as a team

eddherrera: sure, the impact it's barely noticeable but I am ok with a quick call.

eddherrera: aligned all the columns a little more to the right, since it's better than the alternatives.

devinleighsmith (approved):

PR #3862, opened 2024-03-13
Fix PSP-7927

FuriousLlama:
Updated timeouts to be 30s instead of 3s when retrieving information for adding properties

### DIFF
--- a/source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
+++ b/source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
@@ -56,7 +56,7 @@ const AddSubdivisionContainer: React.FC<IAddSubdivisionContainerProps> = ({ onCl
 
   const getAddress = useCallback(
     async (pid: string): Promise<AddressForm | undefined> => {
-      const bcaSummary = await getPrimaryAddressByPid(pid);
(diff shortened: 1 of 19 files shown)


source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
```diff
@@ -56,7 +56,7 @@ const AddSubdivisionContainer: React.FC<IAddSubdivisionContainerProps> = ({ onCl
 
   const getAddress = useCallback(
     async (pid: string): Promise<AddressForm | undefined> => {
-      const bcaSummary = await getPrimaryAddressByPid(pid, 3000);
+      const bcaSummary = await getPrimaryAddressByPid(pid);
```
asanchezr, line 59: did you mean to delete the timeout here? the other changes put the timeout at 30000ms but here it gets removed - why?

asanchezr (approved):

source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
FuriousLlama: I aligned it with the rest of the additions, so 30s timeout

PR #3866, opened 2024-03-14
Test corrections - correct all console warnings/errors

devinleighsmith:


source/frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -12,11 +13,11 @@ export const ApiVersionInfo = () => {
   const { getVersion } = useApiHealth();
   const [version, setVersion] = React.useState<IApiVersion>();
 
-  React.useEffect(() => {
+  useDeepCompareEffect(() => {
     let isActive = true;
     const get = async () => {
       const response = await getVersion();
-      if (isActive) {
+      if (isActive && version?.informationalVersion !== response.data?.informationalVersion) {
```
FuriousLlama, line 20: Can you elaborate on why is this change necessary? I am guessing that fixing the tests uncovered some unwanted behavior on the app.

source/frontend/src/features/mapSideBar/acquisition/add/models.ts
```diff
@@ -21,23 +21,23 @@ export class AcquisitionForm implements WithAcquisitionTeam, WithAcquisitionOwne
...
+  region?: string = '';
   properties: PropertyForm[] = [];
   team: AcquisitionTeamFormModel[] = [];
   owners: AcquisitionOwnerFormModel[] = [];
   fileCheckList: ChecklistItemFormModel[] = [];
 
   project?: IAutocompletePrediction;
   product = '';
-  fundingTypeCode?: string;
+  fundingTypeCode?: string = '';
```
FuriousLlama, line 40: kinda nit: If the variables should not be nullable then it would be a good idea to make their types only strings.

FuriousLlama (approved):
Great job. Thank you for taking the time to get the tests cleaner and fix the issues along.

source/frontend/src/features/mapSideBar/property/tabs/takes/update/TakesUpdateContainer.test.tsx
```diff
@@ -80,20 +80,24 @@ describe('TakesUpdateContainer component', () => {
 
   it('renders as expected', async () => {
     const { asFragment } = setup({});
-    await waitForEffects();
+    await act(async () => {});
```
asanchezr, line 83: I believe this is equivalent to `await waitForEffects()` - did you find any differences in behavior in the tests?

source/frontend/src/components/common/ApiVersionInfo.tsx
devinleighsmith: Yes, in some cases the test uncovered situations where the getVersion request would get fired more then once. In that case, we don't need to overwrite the version information unless it has changed.

source/frontend/src/features/mapSideBar/acquisition/add/models.ts
devinleighsmith: Fair-ish. This is our desire to make things cleanly nullable pushing up against Formik's hard requirement to not allow formik values fields to be null.

I tried to fix this by making the formik-enabled component set an empty string instead of null where possible, but also wanted to minimize the scope of my changes.

source/frontend/src/features/mapSideBar/property/tabs/takes/update/TakesUpdateContainer.test.tsx
devinleighsmith: await waitForEffects() does not appear to work in all cases with the newer version of jest that we are currently on. In all cases where I replaced that function, the test threw warnings before the replacement but stopped after the replacement (I did not touch any tests that were not throwing warnings).

asanchezr: agree we could look into modifying the waitForEffects method later

PR #3871, opened 2024-03-15
Psp 7875 only show error modal or bomb - not both.

devinleighsmith:


source/frontend/src/components/propertySelector/search/PropertySelectorPidSearchContainer.tsx
```diff
@@ -22,9 +25,16 @@ export const PropertySelectorPidSearchContainer: React.FunctionComponent<
   const searchFunc = useCallback(
     async (layerSearch: ILayerSearchCriteria) => {
       if (layerSearch?.pid) {
-        const result = await getPropertyByPidWrapper.execute(layerSearch?.pid);
-        if (result) {
-          setSelectProperty(result);
+        try {
+          const result = await getPropertyByPidWrapper.execute(layerSearch?.pid);
```
devinleighsmith, line 29: this is the underlying story that brought up 7875 - when a property is searched for but does not exist - the error reporting is not clear.

source/frontend/src/hooks/util/useApiRequestWrapper.ts
```diff
@@ -98,7 +96,28 @@ export const useApiRequestWrapper = <
           return;
         }
         const axiosError = e as AxiosError<IApiError>;
-        onError && onError(axiosError);
+        if (onError) {
+          onError(axiosError).catch((axiosError: AxiosError) => {
```
devinleighsmith, line 100: This change ensures that if an explicit error handler is set, the bomb icon will not display, unless the error handler re-throws the exception.

source/frontend/src/hooks/util/useApiRequestWrapper.ts
```diff
@@ -98,7 +96,28 @@ export const useApiRequestWrapper = <
...
+              logError({
+                name: requestName,
+                status: axiosError?.response?.status,
+                error: axiosError ?? {},
+              }),
+            );
+          });
+        } else if (!throwError) {
+          // If no error handling is provided, fall back to the default PIMS error handler (bomb icon).
+          dispatch(
```
devinleighsmith, line 112: if there is no error handler, and an error occurs, show the bomb.

source/frontend/src/utils/axiosUtils.ts
```diff
@@ -47,10 +49,13 @@ export function useAxiosErrorHandlerWithAuthorization(
     (axiosError: AxiosError<IApiError>) => {
       if (axiosError?.response?.status === 400) {
         toast.error(axiosError?.response.data.error, { autoClose: 10000 });
```
devinleighsmith, line 51: this is an example, when the error is handled satisfactorily, we resolve the promise. when we get an unexpected error, we reject and fall back on the default error logging behaviour.

FuriousLlama (approved):

PR #3876, opened 2024-03-18
Added logic to make sure the current status is taken into account

FuriousLlama:


source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -776,11 +776,36 @@ private void TransferPropertiesOfInterest(PimsAcquisitionFile acquisitionFile, b
...
+                    isOtherInterest = false;
+                    isPropertyOfInterest = false;
+                }
+                else if(property.IsOtherInterest || isOtherInterest)
+                {
+                    isOwned = false;
+                    isOtherInterest = true;
+                    isPropertyOfInterest = false;
+                }
+                else if(property.IsPropertyOfInterest || isPropertyOfInterest)
```
FuriousLlama, line 792: the OR is not actually necessary, but I think it improves readability.

source/frontend/public/mockServiceWorker.js
```diff
@@ -8,121 +8,121 @@
  * - Please do NOT serve this file on production.
  */
 
-const INTEGRITY_CHECKSUM = '3d6b9f06410d179a7f7404d4bf4c3c70';
-const activeClientIds = new Set();
+const INTEGRITY_CHECKSUM = '3d6b9f06410d179a7f7404d4bf4c3c70'
```
devinleighsmith, line 11: accidental?

devinleighsmith (approved):

PR #3877, opened 2024-03-18
psp-7995 consolidation/subdivision restrictions.

devinleighsmith:


source/frontend/src/components/propertySelector/search/PropertySelectorPidSearchContainer.tsx
```diff
@@ -28,12 +28,19 @@ export const PropertySelectorPidSearchContainer: React.FunctionComponent<
         try {
           const result = await getPropertyByPidWrapper.execute(layerSearch?.pid);
           if (result) {
+            if (!result.isOwned || result.isRetired) {
+              toast.warn(
+                'Only properties that are part of the Core Inventory (owned) can be subdivided/consolidated. This property is not in core inventory within PIMS.',
+              );
+            }
             setSelectProperty(result);
```
asanchezr, line 36: question - this shows a warning message/toast but still sets the selected property via `setSelectProperty(result)`
does that makes sense or should we not set it to selected in cases where is it not owned or it is retired?

asanchezr (approved):

PR #3884, opened 2024-03-20
PSP-7929 : Backend error - For all files do not include a retired property on selection

eddherrera:


source/backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
```diff
@@ -62,6 +63,11 @@ public PimsPropertyAcquisitionFile Add(PimsPropertyAcquisitionFile propertyAcqui
         {
             propertyAcquisitionFile.ThrowIfNull(nameof(propertyAcquisitionFile));
 
+            if (propertyAcquisitionFile.Property.IsRetired.HasValue && propertyAcquisitionFile.Property.IsRetired.Value)
```
devinleighsmith, line 66: hmm, this would appear to not cover acquisition updates, just add. Was that intentional?

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -227,14 +227,19 @@ public PimsAcquisitionFile Add(PimsAcquisitionFile acquisitionFile, IEnumerable<
             ValidateStaff(acquisitionFile);
             ValidateOrganizationStaff(acquisitionFile);
 
-            acquisitionFile.AcquisitionFileStatusTypeCode = "ACTIVE";
+            if (acquisitionFile.PimsPropertyAcquisitionFiles.Any(x => x.Property.IsRetired.HasValue && x.Property.IsRetired.Value))
```
devinleighsmith, line 230: hmm, yes. This appears to only affect add but I imagine it should affect update as well?

source/backend/api/Services/LeaseService.cs
```diff
@@ -184,6 +185,11 @@ public PimsLease Add(PimsLease lease, IEnumerable<UserOverrideCode> userOverride
             var pimsUser = _userRepository.GetByKeycloakUserId(_user.GetUserKey());
             pimsUser.ThrowInvalidAccessToLeaseFile(lease.RegionCode);
 
+            if (lease.PimsPropertyLeases.Any(x => x.Property.IsRetired.HasValue && x.Property.IsRetired.Value))
```
devinleighsmith, line 188: It seems like your pattern here is to add this logic to both the service and also to the repository, but not in the case of leases?

source/backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
devinleighsmith, line 66: I do sort of like having this logic at the repository level, because that means regardless of what service calls this, the user will be unable to add an acq file with a retired property. Is there a good argument for also having this at the service level?

source/backend/api/Services/AcquisitionFileService.cs
devinleighsmith: Unless you were expecting the repository to handle file update adding new properties? in that case the lease property probably needs something similar.

eddherrera: The adding does not call the AcquisitionFilePropertyRepository methods, so the check had to be added in the Service. 
For the update I am taking into consideration the developer note that states: "Although if a property is already on a file, it can be ignored during this process." So it's only when "UpdatingProperties" and the flow is to "Add".

source/backend/api/Services/LeaseService.cs
eddherrera: I don't think you can update the property on a lease?

source/backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
eddherrera: Not particularly besides short-circuiting the method and the fact that the repository Add is only called in this Service so it's more contained. But yeah, I guess it could be done in the repo.

source/backend/api/Services/LeaseService.cs
eddherrera: Updated

source/backend/dal/Repositories/AcquisitionFilePropertyRepository.cs
eddherrera: Updated

devinleighsmith (approved):

PR #3885, opened 2024-03-21
PSP-7922 Completing disposition file - selected properties and errors

asanchezr:
In this PR the file validation logic was refactored/moved into a separate method to encapsulate all checks in one place. Previous implementation had the checks spread out across several code paths.
Having it in a single method also increases maintainability - if the business rules driving those checks ever change.

### DIFF
--- a/source/backend/api/Services/DispositionFileService.cs
+++ b/source/backend/api/Services/DispositionFileService.cs
@@ -117,35 +117,18 @@ namespace Pims.Api.Services
                 throw new BusinessRuleViolationException("The file you are editing is not active or draft, so you cannot save changes. Refresh your browser to see file state.");
             }
 
-            ValidateStaff(dispositionFile);
             ValidateVersion(id, dispositionFile.ConcurrencyControlNumber);
 
-            dispositionFile.ThrowContractorRemovedFromTeam(_user, _userRepository);
-
-            var doNotAddToStatuses = new List<string>() { EnumDispositionFileStatusTypeCode.COMPLETE.ToString(), EnumDispositionFileStatusTypeCode.ARCHIVED.ToString(), EnumDispositionFileStatusTypeCode.CANCELLED.ToString(), };
+            // validate disposition file state before proceeding with any database updates
             var currentDispositionFile = _dispositionFileRepository.GetById(id);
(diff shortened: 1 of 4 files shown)


source/frontend/src/features/mapSideBar/disposition/tabs/fileDetails/detail/update/UpdateDispositionForm.tsx
```diff
@@ -244,10 +244,8 @@ const UpdateDispositionForm: React.FC<IUpdateDispositionFormProps> = ({
 export default UpdateDispositionForm;
 
 const Container = styled.div`
-  .form-section {
-    margin: 0;
-    padding-left: 0;
-  }
+  background-color: ${props => props.theme.css.filterBackgroundColor};
+  padding-top: 1rem;
```
asanchezr, line 248: Fixing the styling of the Update Disposition form. Was missing the grey background and padding

source/backend/api/Services/DispositionFileService.cs
```diff
@@ -117,35 +117,18 @@ public PimsDispositionFile Update(long id, PimsDispositionFile dispositionFile,
...
-                if (currentDispositionFile?.PimsDispositionSales?.FirstOrDefault()?.SaleFinalAmt == null)
-                {
-                    throw new BusinessRuleViolationException("You have not added a Sales Price. Please add a Sales Price before completion.");
-                }
-                else if (currentDispositionFile.PimsDispositionFileProperties.Count > 0)
-                {
-                    DisposeOfProperties(dispositionFile, userOverrides);
-                }
-            }
+            var isFileClosing = currentDispositionFile.DispositionFileStatusTypeCode != EnumDispositionFileStatusTypeCode.COMPLETE.ToString() &&
```
eddherrera, line 126: You could move this assigment bebore calling  private "ValidateFileBeforeUpdate" method, and pass the value to the method as parameter, that way you would make the assigment once.

eddherrera (approved):

source/backend/api/Services/DispositionFileService.cs
asanchezr: fair- still I feel this is a simple check and perhaps should be kept so the "Validate" doesn't have to trust someone to do the check for itself

PR #3889, opened 2024-03-22
PSP-6765 empty file upload error

asanchezr:
image


devinleighsmith:
Hmm, should this throw an error when the user tries to upload an empty document? Based on the tests it seems like no error is thrown, but also the document is not uploaded to Mayan.

asanchezr:
> Hmm, should this throw an error when the user tries to upload an empty document? Based on the tests it seems like no error is thrown, but also the document is not uploaded to Mayan.

I can make it throw an error but the existing code does not throw if no "document" is returned from mayan so I kept it that way. Let's jump on a call to discuss

devinleighsmith (approved):

PR #3893, opened 2024-03-25
PSP-8051 : FT:User is able to add the retired property record to the …

eddherrera:
…Acquisition&lease/license files

### DIFF
--- a/source/backend/api/Services/LeaseService.cs
+++ b/source/backend/api/Services/LeaseService.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Configuration;
 using System.Globalization;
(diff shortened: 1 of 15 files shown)


source/backend/api/Services/LeaseService.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Configuration;
```
asanchezr, line 3: do we need this import?

asanchezr (approved):

source/backend/api/Services/LeaseService.cs
eddherrera: removed

PR #3896, opened 2024-03-25
Tech debt/vite

FuriousLlama:


devinleighsmith:
Before this is merged, we need to set up a vite image within openshift and test it against this branch.
Additionally, the GHA will need to be switched over, that should also be done on this branch.

devinleighsmith:
Given that you are going to be busy with database changes and stories and this is a short sprint, I'll see if I can get the GHA/OpenShift components working today.

PR #3899, opened 2024-03-26
fix an error introduced in the isRetired conditional logic when getti…

devinleighsmith:
…ng properties by pid.

### DIFF
--- a/source/backend/dal/Repositories/PropertyRepository.cs
+++ b/source/backend/dal/Repositories/PropertyRepository.cs
@@ -195,7 +195,7 @@ namespace Pims.Dal.Repositories
 
             if(includeRetired)
             {
-                query = query.Where(r => r.IsRetired.HasValue && r.IsRetired.Value);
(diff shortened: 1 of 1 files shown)


eddherrera (approved):

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -195,7 +195,7 @@ public PimsProperty GetByPid(int pid, bool includeRetired = false)
 
             if(includeRetired)
             {
-                query = query.Where(r => r.IsRetired.HasValue && r.IsRetired.Value);
+                query = query.Where(r => !r.IsRetired.HasValue || (r.IsRetired.HasValue && r.IsRetired.Value));
```
FuriousLlama, line 198: I think this would work, although it does look a tad odd. Maybe using the predicate builder and an Or statement would be more readable

FuriousLlama (approved):

PR #3904, opened 2024-03-26
Fully fix property getByPid, add unit tests.

devinleighsmith:


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -193,11 +193,7 @@ public PimsProperty GetByPid(int pid, bool includeRetired = false)
 
             var query = Context.PimsProperties.AsNoTracking();
 
-            if(includeRetired)
-            {
-                query = query.Where(r => !r.IsRetired.HasValue || (r.IsRetired.HasValue && r.IsRetired.Value));
```
devinleighsmith, line 198: turns out !r.IsRetired.HasValue is equivalent to is not null, which was resulting in 
isRetired is null || (isRetired is not null && isRetired = true) // this fails when isRetired = 0 explicitly, instead of null

what we really want is just to not filter out retired properties when includeRetired is true.

asanchezr (approved):

PR #3905, opened 2024-03-27
psp-8150 fix consolidation/subdivision issues.

devinleighsmith:


source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -164,6 +164,7 @@ public List<PimsProperty> GetAllByIds(List<long> ids)
                 .Include(p => p.Address)
                     .ThenInclude(a => a.Country)
                 .Where(p => ids.Any(s => s == p.PropertyId))
+                .AsNoTracking()
```
devinleighsmith, line 167: I checked this, the only consumers are a controller, which just returns the result of the get, and the consolidation logic. This, and the following in the property service was the cause of this issue:

`public List<PimsProperty> GetMultipleById(List<long> ids)
{
    _logger.LogInformation("Getting multiple properties by id");
    _user.ThrowIfNotAuthorized(Permissions.PropertyView);

    List<PimsProperty> properties = _propertyRepository.GetAllByIds(ids);
    foreach (PimsProperty property in properties)
    {
        if (property?.Location != null)
        {
            property.Location = TransformCoordinates(property.Location);
        }
    }

    return properties;
}`

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -246,6 +247,10 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
     serviceSend({ type: 'FINISH_SELECTION' });
   }, [serviceSend]);
 
+  const changeSidebar = useCallback(() => {
```
devinleighsmith, line 250: unrelated change, but subdivision/consolidations were not clearing the list of draft markers after the subdivision is saved.

source/frontend/src/features/mapSideBar/consolidation/AddConsolidationContainer.tsx
```diff
@@ -76,7 +76,9 @@ const AddConsolidationContainer: React.FC<IAddConsolidationContainerProps> = ({
           featuresetToMapProperty(selectedFeatureDataset),
         );
         if (isValidString(propertyForm.pid)) {
-          propertyForm.address = await getAddress(propertyForm.pid);
```
devinleighsmith, line 79: This was a wasted of time and inaccurate, as we have this property in our system we should not be looking up the bca address for it.

asanchezr (approved):

PR #3923, opened 2024-04-02
PSP-7886 : Prevent users from adding new takes, and/or deleting completed takes from a non-active (e.g. draft or active) acquisition file

eddherrera:


source/backend/api/Services/TakeService.cs
```diff
@@ -61,11 +61,35 @@ public IEnumerable<PimsTake> UpdateAcquisitionPropertyTakes(long acquisitionFile
...
+                {
+                    var currentCompleteTakes = currentFilePropertyTakes
+                        .Where(x => x.TakeStatusTypeCode == AcquisitionTakeStatusTypes.COMPLETE.ToString()).ToList();
+
+                    if (currentCompleteTakes.Count > 0)
+                    {
+                        foreach (var completeTake in currentCompleteTakes)
+                        {
+                            var updatedTake = takes.FirstOrDefault(x => x.TakeId == completeTake.TakeId);
+                            if (updatedTake is null || updatedTake?.TakeStatusTypeCode != completeTake.TakeStatusTypeCode)
```
devinleighsmith, line 86: I'm not sure about this logic, as it would seem to allow a completed take to be updated as long as its status type is not modified.

given that the frontend prevents editing completed takes, I think we can just ignore any take modifications to completed takes here.

devinleighsmith (approved):

source/backend/api/Services/TakeService.cs
eddherrera: Updated frontend

devinleighsmith (approved):

PR #3929, opened 2024-04-05
correct map not loading properties initially.

devinleighsmith:


source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -383,6 +383,7 @@ export const mapMachine = createMachine<MachineContext>({
     },
     mapVisible: {
       type: 'parallel',
+      entry: [send({ type: 'REFRESH_PROPERTIES', searchCriteria: defaultPropertyFilter })],
```
devinleighsmith, line 386: It looks like the map state machine was actually depending on the useRouterFilter component to trigger the initial properties load.

devinleighsmith: Since the router was removed in the 5.2 branch, that led to the properties load not being fired properly.

eddherrera (approved):

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: The initial interaction with the map property filter are messy. If on first load it will load the correct query from the URL this will be fine. Otherwise that component needs to initiate the first load

FuriousLlama (approved):

PR #3930, opened 2024-04-05
Do not return non-polygon boundaries as the mapping is not idempotent.

devinleighsmith:


devinleighsmith:
Note that I've also added the following: https://jira.th.gov.bc.ca/browse/PSP-8194

There is no reason why we can't support multipolygons on our backend.
In fact, currently in our deployed environments we actually have multipolygons in our database. The issue is that when you convert a multipolygon to a coordinate array, and then try to convert a coordinate array back into a polygon, it doesn't work.

It also seems as those there is no way to directly convert a coordinate array into a multi polygon - so we need to change the format of the model we are passing to/from the backend to contain more of the geometry so that mapping to/from is possible.

FuriousLlama (approved):

PR #3931, opened 2024-04-05
PSP-8094, PSP-8095 Warn user that property is part of an existing acquisition/research file

asanchezr:
add-property-warning

### DIFF
--- a/source/frontend/src/features/mapSideBar/acquisition/add/AcquisitionPropertiesSubForm.tsx
+++ b/source/frontend/src/features/mapSideBar/acquisition/add/AcquisitionPropertiesSubForm.tsx
@@ -8,19 +8,23 @@ import { IMapProperty } from '@/components/propertySelector/models';
 import SelectedPropertyHeaderRow from '@/components/propertySelector/selectedPropertyList/SelectedPropertyHeaderRow';
 import SelectedPropertyRow from '@/components/propertySelector/selectedPropertyList/SelectedPropertyRow';
 import { useBcaAddress } from '@/features/properties/map/hooks/useBcaAddress';
+import { useModalContext } from '@/hooks/useModalContext';
 
 import { AddressForm, PropertyForm } from '../../shared/models';
 import { AcquisitionForm } from './models';
 
-interface AcquisitionPropertiesProp {
+export interface IAcquisitionPropertiesProps {
   formikProps: FormikProps<AcquisitionForm>;
+  confirmBeforeAdd: (propertyId: number) => Promise<boolean>;
 }
 
-export const AcquisitionPropertiesSubForm: React.FunctionComponent<
-  React.PropsWithChildren<AcquisitionPropertiesProp>
-> = ({ formikProps }) => {
+export const AcquisitionPropertiesSubForm: React.FunctionComponent<IAcquisitionPropertiesProps> = ({
+  formikProps,
+  confirmBeforeAdd,
+}) => {
   const { values } = formikProps;
   const { getPrimaryAddressByPid, bcaLoading } = useBcaAddress();
+  const { setModalContent, setDisplayModal } = useModalContext();
 
   return (
     <>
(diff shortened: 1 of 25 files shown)


source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
```diff
@@ -18,26 +14,20 @@ import { ApiGen_Concepts_File } from '@/models/api/generated/ApiGen_Concepts_Fil
...
-import ResearchMenu from './common/ResearchMenu';
 import { useGetResearch } from './hooks/useGetResearch';
 import { useUpdateResearchProperties } from './hooks/useUpdateResearchProperties';
-import ResearchView from './ResearchView';
+import { IResearchViewProps } from './ResearchView';
 
 export interface IResearchContainerProps {
   researchFileId: number;
   onClose: () => void;
+  View: React.FunctionComponent<React.PropsWithChildren<IResearchViewProps>>;
```
asanchezr, line 26: I refactored the `ResearchContainer` to have the view injected via dependency injection (like other file containers in the code base). This makes testing easier and improves code maintainability.

source/frontend/src/features/mapSideBar/research/add/AddResearchContainer.tsx
```diff
@@ -55,12 +67,51 @@ export const AddResearchContainer: React.FunctionComponent<
...
-    if (!!initialForm && !!formikRef.current) {
-      formikRef.current.resetForm();
-      formikRef.current?.setFieldValue('properties', initialForm.properties);
-    }
-  }, [initialForm]);
+    const runAsync = async () => {
+      if (exists(initialForm) && exists(formikRef.current) && needsUserConfirmation) {
+        if (initialForm.properties.length > 0) {
+          const formProperty = initialForm.properties[0];
+          if (formProperty.apiId && (await confirmBeforeAdd(formProperty.apiId))) {
```
devinleighsmith, line 77: I have to admit I'm a bit surprised by this implementation. Is there any reason why this was written instead of the standard, api-driven user override pattern we've been using up until now? Perhaps I'm just missing something that precludes the use of that workflow.

devinleighsmith: Also, I think the 
formProperty.apiId
may be incorrect. Since this is an "Add" workflow, the form should not have an apiId yet - that would only be available in update workflows. I think this is why in my testing I was able to save two new research files back-to-back with the same property.

source/frontend/src/features/mapSideBar/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -104,6 +111,60 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> =
...
 
+  const { initialValues } = helper;
+  // Require user confirmation before adding a property to file
+  // This is the flow for Map Marker -> right-click -> create Acquisition File
+  useEffect(() => {
+    const runAsync = async () => {
+      if (exists(initialValues) && exists(formikRef.current) && needsUserConfirmation) {
+        if (initialValues.properties.length > 0) {
+          const formProperty = initialValues.properties[0];
+          if (formProperty.apiId && (await confirmBeforeAdd(formProperty.apiId))) {
```
devinleighsmith, line 122: same issue here.

source/frontend/src/features/mapSideBar/acquisition/add/AcquisitionPropertiesSubForm.tsx
```diff
@@ -46,14 +50,48 @@ export const AcquisitionPropertiesSubForm: React.FunctionComponent<
...
+                                <p>
+                                  This property has already been added to one or more acquisition
+                                  files.
+                                </p>
+                                <p>Do you want to acknowledge and proceed?</p>
+                              </>
+                            ),
+                            okButtonText: 'Yes',
+                            cancelButtonText: 'No',
+                            handleOk: () => {
```
devinleighsmith, line 70: Hmm, I don't see anything in the AC that reflects why this is here, what am I missing?

devinleighsmith:
Oh, I thought we agreed to do this on save, and that was why we went with "User Override Required" as the header. This is unfortunate, as I think this may have just been an issue with the requirements, and now the implementation is significantly more complex.

source/frontend/src/features/mapSideBar/research/add/AddResearchContainer.tsx
asanchezr: let's chat over Teams. Easier than typing here but my interpretation (and we asked this) is that they want the check done "before" saving the file (so we can not do an api-driven user override)

asanchezr:
> Oh, I thought we agreed to do this on save, and that was why we went with "User Override Required" as the header. This is unfortunate, as I think this may have just been an issue with the requirements, and now the implementation is significantly more complex.

I recall that we devs pushed for doing on Save and BAs went to business and said "no we want on add property before save" that is why I believe it was pointed at 5. To accommodate the fact that we would be going away from stablished pattern

source/frontend/src/features/mapSideBar/research/add/AddResearchContainer.tsx
asanchezr: > Also, I think the formProperty.apiId may be incorrect. Since this is an "Add" workflow, the form should not have an apiId yet - that would only be available in update workflows. I think this is why in my testing I was able to save two new research files back-to-back with the same property.

The **formProperty.apiId** will indeed be populated if you create a file from an existing property on the map (which has apiId set) and that also belongs to **another** ACQ file. I can show you the flow in the way I tested - maybe I missed something

devinleighsmith:
> > Oh, I thought we agreed to do this on save, and that was why we went with "User Override Required" as the header. This is unfortunate, as I think this may have just been an issue with the requirements, and now the implementation is significantly more complex.
> 
> I recall that we devs pushed for doing on Save and BAs went to business and said "no we want on add property before save" that is why I believe it was pointed at 5. To accommodate the fact that we would be going away from stablished pattern

regardless of this, when adding a new research file that has the same property as a pre-existing research file, no warning displays either on property add or on save.

PR #3941, opened 2024-04-09
Features/psp 7836 filter

FuriousLlama:


source/backend/apimodels/Models/Concepts/Property/PropertyViewMap.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace Pims.Api.Models.Concepts.Property
+{
+    public class PropertyViewMap : IRegister
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config.NewConfig<Entity.PimsPropertyLocationVw, PropertyViewModel>()
+               .Map(dest => dest.Id, src => src.PropertyId)
+               .Map(dest => dest.Pid, src => src.Pid)
+.Map(dest => dest.PidPadded, src => src.PidPadded)
```
FuriousLlama, line 13: I will run the formater on a separate PR to address this and other issues like these.

source/backend/tests/unit/dal/Repositories/PropertyRepositoryTest.cs
```diff
@@ -93,6 +93,8 @@ public void GetPage_Properties_NotAuthorized()
             act.Should().Throw<NotAuthorizedException>();
         }
 
+        /*
+        // TODO: Figure out how to add DB views to the context
```
FuriousLlama, line 97: Right now the context does not allow to track views, so this test is not working at the moment.

devinleighsmith:
@FuriousLlama when using the advanced filter or list view filter for property of interest, nothing is returned. the POI's show up on the map when the advanced filter is not in use.

devinleighsmith:
> @FuriousLlama when using the advanced filter or list view filter for property of interest, nothing is returned. the POI's show up on the map when the advanced filter is not in use.

sorry, nvm. backend out of sync, disregard.

source/backend/api/Services/PropertyOperationService.cs
```diff
@@ -165,8 +164,22 @@ public IEnumerable<PimsPropertyOperation> ConsolidateProperty(IEnumerable<PimsPr
...
-        private static void CommonPropertyOperationValidation(IEnumerable<PimsPropertyOperation> operations)
+        private static void CommonPropertyOperationValidation(IEnumerable<PimsPropertyOperation> operations, IEnumerable<PimsProperty> dbSourceProperties)
         {
+
+            foreach (var sourceProperty in dbSourceProperties)
+            {
+                var operation = operations.FirstOrDefault(p => p.SourcePropertyId == sourceProperty.PropertyId);
+                if (operation == null)
+                {
+                    throw new BusinessRuleViolationException("All source properties must exist in the system.");
```
devinleighsmith, line 175: All source properties must exist in the system.

Doesn't seem like an appropriate user facing error. Is this listed in the story somewhere?

If not I'd work with Ana or Julian for some text, this seems more like a developer appropriate error.

FuriousLlama: Yep, not in the story. It is not really a error for the user, but in case the backend receives a bad request

source/backend/api/Services/TakeService.cs
```diff
@@ -75,7 +75,7 @@ public IEnumerable<PimsTake> UpdateAcquisitionPropertyTakes(long acquisitionFile
             var currentAcquisitionStatus = Enum.Parse<AcquisitionStatusTypes>(currentAcquistionFile.AcquisitionFileStatusTypeCode);
 
             // No user can update the takes when the File is not Active/Draft
```
devinleighsmith, line 77: nit: the comment seems out of date.

source/backend/api/Services/TakeService.cs
```diff
@@ -75,7 +75,7 @@ public IEnumerable<PimsTake> UpdateAcquisitionPropertyTakes(long acquisitionFile
...
-            if (!_statusSolver.CanEditTakes(currentAcquisitionStatus))
+            if (!_statusSolver.CanEditTakes(currentAcquisitionStatus) && !_user.HasPermission(Permissions.SystemAdmin))
```
devinleighsmith, line 78: actually, this doesn't seem correct. I think even admins can't edit the takes when the file is complete.

source/backend/api/Services/TakeService.cs
```diff
@@ -86,20 +86,18 @@ public IEnumerable<PimsTake> UpdateAcquisitionPropertyTakes(long acquisitionFile
...
+                        if (!_user.HasPermission(Permissions.SystemAdmin) && (updatedTake is null || (updatedTake is not null && updatedTake.TakeStatusTypeCode != completeTake.TakeStatusTypeCode)))
                         {
-                            var updatedTake = takes.FirstOrDefault(x => x.TakeId == completeTake.TakeId);
-                            if (updatedTake is null || updatedTake?.TakeStatusTypeCode != completeTake.TakeStatusTypeCode)
-                            {
-                                throw new BusinessRuleViolationException("Retired records are referenced for historical purposes only and cannot be edited or deleted. If the take has been added in error, contact your system administrator to re-open the file, which will allow take deletion.");
-                            }
+                            throw new BusinessRuleViolationException("Retired records are referenced for historical purposes only and cannot be edited or deleted. If the take has been added in error, contact your system administrator to re-open the file, which will allow take deletion.");
```
devinleighsmith, line 100: This error seems to refer to a take on a retired property, but I don't see anything in the logic above about a retired property. Are you confident this is correct?

source/backend/api/Services/TakeService.cs
```diff
@@ -112,15 +110,15 @@ public IEnumerable<PimsTake> UpdateAcquisitionPropertyTakes(long acquisitionFile
             var currentProperty = _acqFileRepository.GetProperty(acquisitionFilePropertyId);
             var currentTakes = _takeRepository.GetAllByPropertyId(currentProperty.PropertyId);
 
-            var completedTakes = currentTakes.Where(t => t.TakeStatusTypeCode == "COMPLETE").ToList();
+            var completedTakes = currentTakes.Union(takes)
```
devinleighsmith, line 113: I suppose this works for now - I think this will become problematic when we add completed take deletion.

source/backend/apimodels/Models/Concepts/Property/PropertyViewModel.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        public string Description { get; set; }
+
+        public long? AddressId { get; set; }
+
+        public short RegionCode { get; set; }
+
+        public short DistrictCode { get; set; }
+
+        //public Geometry Geometry { get; set; }
```
devinleighsmith, line 60: If this isn't necessary please remove it. The next developer won't know why this is here.

source/backend/api/Services/TakeService.cs
FuriousLlama: Ah good to know. When I merged eduardo's changes I thougt the story was about allowing sys-admins to delete takes. There is a story about it in this sprint given there are rules around disposed properties and takes

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -97,33 +88,44 @@ private static ExpressionStarter<PimsProperty> GenerateCommonPropertyQuery(Claim
 
             var isRetired = filter.Ownership.Contains("isRetired");
 
-            ExpressionStarter<PimsProperty> ownershipBuilder;
+            ExpressionStarter<PimsPropertyLocationVw> ownershipBuilder;
 
             if (filter.Ownership.Count > 0)
             {
-                ownershipBuilder = isRetired ? PredicateBuilder.New<PimsProperty>(p => p.IsRetired == true) : PredicateBuilder.New<PimsProperty>(p => false);
+                var authorizationTypes = new List<string>() {
```
devinleighsmith, line 95: nit, this isn't all the types, just the types that result in interests, could update naming to better reflect that.

Also, do we not have an enum for this?

source/backend/api/Services/TakeService.cs
FuriousLlama: This came from the merge from another story. I agree, that it does not look like it actually validates for retired.

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -97,33 +88,44 @@ private static ExpressionStarter<PimsProperty> GenerateCommonPropertyQuery(Claim
...
                 if (filter.Ownership.Contains("isPropertyOfInterest"))
                 {
-                    ownershipBuilder = ownershipBuilder.Or(p => p.IsPropertyOfInterest && p.IsRetired != true);
+                    ownershipBuilder.Or(p => p.IsRetired != true && p.HasActiveAcquisitionFile.HasValue && p.HasActiveAcquisitionFile.Value);
+                    ownershipBuilder.Or(p => p.IsRetired != true && p.HasActiveResearchFile.HasValue && p.HasActiveResearchFile.Value);
                 }
                 if (filter.Ownership.Contains("isOtherInterest"))
                 {
-                    ownershipBuilder = ownershipBuilder.Or(p => p.IsOtherInterest && p.IsRetired != true);
+                    var today = DateOnly.FromDateTime(DateTime.Now);
```
devinleighsmith, line 117: this looks like an unused variable. What was this supposed to be used for?

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -525,35 +525,56 @@ public HashSet<long> GetMatchingIds(PropertyFilterCriteria filter)
                     p.PimsPropPropAnomalyTypes.Any(at => filter.AnomalyIds.Contains(at.PropertyAnomalyTypeCode)));
             }
 
+
+            var authorizationTypes = new List<string>(){
```
devinleighsmith, line 529: is this being used?

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
devinleighsmith: Is this being used?

source/frontend/src/features/mapSideBar/consolidation/AddConsolidationContainer.tsx
```diff
@@ -145,6 +148,7 @@ const AddConsolidationContainer: React.FC<IAddConsolidationContainerProps> = ({
   ) => {
     try {
       const propertyOperations = values.toApi();
+      console.log(propertyOperations);
```
devinleighsmith, line 151: probably can remove.

source/frontend/src/features/mapSideBar/router/FilePropertyRouter.tsx
```diff
@@ -37,6 +37,7 @@ export const FilePropertyRouter: React.FC<IFilePropertyRouterProps> = props => {
   const onChildSuccess = () => {
     props.setIsEditing(false);
     setStaleLastUpdatedBy(true);
+    props.onSuccess();
```
devinleighsmith, line 40: what is this for?

source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
```diff
@@ -75,9 +75,11 @@ const AddSubdivisionContainer: React.FC<IAddSubdivisionContainerProps> = ({ onCl
           propertyForm.address = selectedFeatureDataset.pimsFeature?.properties
             ? AddressForm.fromPimsView(selectedFeatureDataset.pimsFeature?.properties)
             : undefined;
-          const consolidationFormModel = new SubdivisionFormModel();
-          consolidationFormModel.sourceProperty = propertyForm.toApi();
-          setInitialForm(consolidationFormModel);
+          const subdivisionFormModel = new SubdivisionFormModel();
+          subdivisionFormModel.sourceProperty = propertyForm.toApi();
+          subdivisionFormModel.sourceProperty.isOwned =
```
devinleighsmith, line 80: do we need something similar for retired?

source/backend/api/Services/TakeService.cs
FuriousLlama: Also the wording of this makes me suspect that this is coming from the acquisition file being completed, but the retired part seems off. I would rather open a new ticket and go over the validation to make sure this is handled correclty taken into accounts all stories related to it

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/router/FilePropertyRouter.tsx
FuriousLlama: once the take is updated, the need to re-fetch the properties is bubbled up though the onSuccess. Otherwise the map still shows the old markers

source/frontend/src/features/mapSideBar/subdivision/AddSubdivisionContainer.tsx
FuriousLlama: Unsure, in theory, you cannot subdivide nor consolidate retired properties (not selectable)

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
FuriousLlama: good catch. I was using it before when doing the queries manually instead of using the view. Will remove.

source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
FuriousLlama: Same as below. I switched the implementation of the filters halfway to use the view instead of replicating the query

PR #3948, opened 2024-04-12
PSP-8203 FT: map icon that hovers when you try to choose property on the map is not displayed

asanchezr:
This was a weird asset loading bug caused by migration to **vite**. Inlined SVGs are converted to data URLs when small enough (< 4kb) and those data urls need to be surrounded with quotes - ie in CSS: 

```css
.c0.draft-cursor .leaflet-interactive {
  cursor: url("data:image/svg+xml, ...") 15 45, pointer;
}
```

our code was missing those quotes in the CSS for `url()` and so the icon was not loading on production builds. launching the frontend via "npm start" locally does not show this bug because the asset/svg is loaded via an URL in development mode (vs a data URL).

### DIFF
--- a/source/frontend/src/features/properties/map/MapContainer.tsx
+++ b/source/frontend/src/features/properties/map/MapContainer.tsx
@@ -45,6 +45,7 @@ const MapContainer: React.FC<React.PropsWithChildren<MapContainerProps>> = () =>
   );
 };
 
+// prettier-ignore
 const StyleMapView = styled.div`
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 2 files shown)


source/frontend/src/features/properties/map/MapContainer.tsx
```diff
@@ -45,14 +45,17 @@ const MapContainer: React.FC<React.PropsWithChildren<MapContainerProps>> = () =>
   );
 };
 
+// prettier-ignore
```
asanchezr, line 48: We need to disable prettier here because we really really need double quotes below (single quotes will not do it). Given that our project is setup to have prettier use single quotes instead of double ones then we need to disable prettier for the next line

devinleighsmith (approved):

PR #3962, opened 2024-04-17
Psp 7883 - remove acquisition completion date.

devinleighsmith:


devinleighsmith:
Note, this story will be handled independently of the database change - the field will no longer be accessible but will remain in the db until it can be migrated.

PR #3964, opened 2024-04-18
PSP-8253: Fixed looking for property in the property listview

FuriousLlama:


source/backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -75,15 +75,15 @@ private static ExpressionStarter<PimsPropertyLocationVw> GenerateCommonPropertyQ
...
-                predicateBuilder = predicateBuilder.And(p => p != null && (EF.Functions.Like(p.Pid.ToString(), $"%{formattedPidPin}%") || EF.Functions.Like(p.Pin.ToString(), $"%{formattedPidPin}%")));
+                predicateBuilder = predicateBuilder.And(p => EF.Functions.Like(p.Pid.ToString(), $"%{formattedPidPin}%") || EF.Functions.Like(p.Pin.ToString(), $"%{formattedPidPin}%"));
             }
             if (!string.IsNullOrWhiteSpace(filter.Address))
             {
                 predicateBuilder = predicateBuilder.And(p => EF.Functions.Like(p.StreetAddress1, $"%{filter.Address}%") || EF.Functions.Like(p.MunicipalityName, $"%{filter.Address}%"));
             }
             if (!string.IsNullOrWhiteSpace(filter.PlanNumber))
             {
-                predicateBuilder = predicateBuilder.And(p => p != null && p.SurveyPlanNumber.Equals(filter.PlanNumber));
```
devinleighsmith, line 86: I have to admit that I don't understand why this solves the problem. Doesn't this just cause p.SurveyPlanNumber or p.PID to throw an NPE if p is null?

FuriousLlama: That would be true in other places where there is no null validation too, but I think the context would not provide a null reference to a table that is not being navigated. Regarding why this is being changed, the not operator (!=) could be used on an entity directly when that table does not have a key (in this case a keyless view). removing the null check fixed the issue without any other noticeable effect. I believe that the view would also not have null entries.

asanchezr (approved):

PR #3968, opened 2024-04-23
PSP-8103 Warn user that property is part of any disposition file

asanchezr:


source/frontend/src/features/mapSideBar/acquisition/add/AddAcquisitionContainer.tsx
```diff
@@ -35,17 +36,41 @@ export const AddAcquisitionContainer: React.FC<IAddAcquisitionContainerProps> =
...
 
   const { execute: getPropertyAssociations } = usePropertyAssociations();
+  const {
+    getPropertyByPidWrapper: { execute: getPropertyByPid },
+    getPropertyByPinWrapper: { execute: getPropertyByPin },
+  } = usePimsPropertyRepository();
   const [needsUserConfirmation, setNeedsUserConfirmation] = useState<boolean>(true);
 
   // Warn user that property is part of an existing acquisition file
   const confirmBeforeAdd = useCallback(
```
devinleighsmith, line 46: it seems like this function has more to do with properties then acquisition files, would recommend moving this to a shared space to reduce code duplication.

devinleighsmith: would need to handle the associations for each type, perhaps via a param? not strictly necessary.

asanchezr: Agree with the code duplication - have to to think about that. Given that we are so close to sprint end I would lean towards merging as-is (if not broken) and then refactoring next sprint

devinleighsmith (approved):

PR #3979, opened 2024-04-30
PSP-8022 : Disposition file states - Hold

eddherrera:


source/frontend/package.json
```diff
@@ -100,6 +100,7 @@
     "happy-dom": "14.7.1",
     "husky": "6.0.0",
     "jest-styled-components": "7.2.0",
+    "jsdom": "^24.0.0",
```
devinleighsmith, line 103: please lock dependency versions.

devinleighsmith (approved):

source/frontend/package.json
eddherrera: Updated

PR #3981, opened 2024-05-01
PSP-8324 Update area convertor in Lease and License

asanchezr:
pims-lease-property-area-view

----

pims-lease-property-area-update


source/frontend/src/features/leases/models.ts
```diff
@@ -214,22 +215,17 @@ export class FormLeaseProperty {
...
       file: null,
       property: formLeaseProperty.property?.toApi() ?? null,
       propertyId: formLeaseProperty.property?.id ?? 0,
       propertyName: formLeaseProperty.name ?? null,
-      leaseArea: numberLeaseArea ?? null,
-      areaUnitType:
-        numberLeaseArea !== undefined
-          ? toTypeCodeNullable(formLeaseProperty.areaUnitTypeCode) ?? null
-          : null,
+      leaseArea: isNumber(formLeaseProperty.landArea) ? formLeaseProperty.landArea : 0,
```
eddherrera, line 225: should this be null when 'landArea' is not a number?

source/frontend/src/components/measurements/AreaView.tsx
```diff
@@ -13,37 +13,35 @@ interface IAreaViewProps {
...
-          </StyledGreenGrey>
-        </Col>
-      </Row>
-    </>
+    <Row>
+      <Col>
+        <StyledGreenCol>
+          <Row>
+            <Col className="text-right">{formatNumber(meters, 0, 2)}</Col>
+            <Col>sq. metres</Col>
```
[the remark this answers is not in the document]
asanchezr: this is part of the common area component (which we are not modifying here) and also that is the approved verbiage - ie "metre" is the Canadian/British spelling vs meter the American spelling

eddherrera (approved):

source/frontend/src/features/leases/models.ts
asanchezr: it is typed as a number so I gave the default value of 0

PR #3986, opened 2024-05-01
psp-8352 properly handle null/empty organization (based on error in test).

devinleighsmith:


asanchezr (approved):

source/frontend/src/features/mapSideBar/research/tabs/fileDetails/details/ResearchSummaryView.tsx
```diff
@@ -58,7 +58,7 @@ const ResearchSummaryView: React.FunctionComponent<IResearchSummaryViewProps> =
     const personOrganizations = props.researchFile!.requestorPerson.personOrganizations;
     const organization =
       exists(personOrganizations) && personOrganizations.length > 0
-        ? personOrganizations[0].organization
+        ? personOrganizations[0]?.organization
```
FuriousLlama, line 61: This is odd. is the exist() method not filtering null | undefined correctly?

devinleighsmith: it looks like there is an array entry that contains a null/undefined value. My guess is that this was introduced by the person merge that Arima is working on.

FuriousLlama: even if the result is coming back with those values, the '''exists(personOrganizations) && personOrganizations.length > 0' should filter those out.

PR #3988, opened 2024-05-02
psp-7928 make document upload block - to reduce number of failed uplo…

devinleighsmith:
…ads.

### DIFF
--- a/source/backend/api/Services/DocumentService.cs
+++ b/source/backend/api/Services/DocumentService.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using System.Configuration;
 using System.IO;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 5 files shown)


source/backend/api/Services/DocumentService.cs
```diff
@@ -133,7 +144,20 @@ public async Task<DocumentUploadResponse> UploadDocumentAsync(DocumentUploadRequ
             if (externalResponse.Status == ExternalResponseStatus.Success)
             {
                 var externalDocument = externalResponse.Payload;
+                if (externalDocument.FileLatest == null && _config.UploadRetries > 0)
+                {
+                    var retryPolicy = Policy<ExternalResponse<DocumentDetailModel>>
```
devinleighsmith, line 149: This is based on the mayan audit logs - I looked at a few examples in test where a document was created in PIMS but it was not possible to download the related file. In the cases I looked at, the document was created, but the file failed to upload or be processed. This causes the document also to be deleted due to a rule in mayan to delete document "stubs"

This change essentially makes upload a document a blocking operation, at least as long as the upload retries is non-zero. We need a long-term solution for this, but at least in the short term, users will not need to manually validate that their file uploaded successfully.

devinleighsmith: Also, OCR still appears to be running despite the OCR app being "disabled" I've added a manual step to update the related config map to target another OCR related env var that seems to stop the process.

FuriousLlama: I think this will work for now. We should run some tests on OS to make sure that this wont result into spawning unnecessary pims api pods

asanchezr (approved):

PR #4002, opened 2024-05-08
PSP-8241 | Updated take calculation to use incoming take if being updated

FuriousLlama:


source/backend/api/Services/TakeService.cs
```diff
@@ -223,7 +223,14 @@ private void RecalculateOwnership(long acquisitionFilePropertyId, PimsTake take)
             var currentTakes = _takeRepository.GetAllByPropertyId(currentProperty.PropertyId);
 
             var allTakes = currentTakes;
-            if (take != null)
+
+            // If the take is being modified, update the one in the list to use the incomming one.
```
devinleighsmith, line 227: nit: incoming

source/backend/api/Services/TakeService.cs
```diff
@@ -223,7 +223,14 @@ private void RecalculateOwnership(long acquisitionFilePropertyId, PimsTake take)
             var currentTakes = _takeRepository.GetAllByPropertyId(currentProperty.PropertyId);
```
devinleighsmith, line 223: Hmm, does this take into account expired takes? given our conversation today, we should not be incorporating expired takes into the calculation.

FuriousLlama: Right now all the takes that could result in ownership cannot expire. Just as a reminder, the interest holder calculation is done "on-the-fly" for the views and filters and that one does take into account the expire date.

devinleighsmith (approved):

source/backend/api/Services/TakeService.cs
devinleighsmith: Ah yes, did not consider that, well thats nice, as it simplifies this.

PR #4009, opened 2024-05-10
psp-5511 display mayan document previews.

devinleighsmith:


source/backend/api/Areas/Documents/DocumentController.cs
```diff
@@ -233,6 +234,52 @@ public async Task<IActionResult> GetDocumentMetadata(long mayanDocumentId)
...
+
+            return new JsonResult(result.Payload.Results);
+        }
+
+        /// <summary>
+        /// Downloads the desired page for the file within the target document.
+        /// </summary>
+        [HttpGet("storage/{mayanDocumentId}/file/{documentFileId}/pages/{documentFilePageId}")]
+        [HasPermission(Permissions.DocumentView)]
+        [ProducesResponseType(typeof(FileContentResult), 200)]
```
FuriousLlama, line 261: Kinda nit: the result is a FileStreamResult

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+    setCurrentPage(0);
+    setDocumentPages([]);
+  }, [setPreviewDocumentId, setShowDocumentPreview]);
+
+  const loadPagesForDocumentId = useCallback(async () => {
+    const data = await retrieveDocumentDetail(previewDocumentId);
+    if (
+      data?.payload?.datetime_created &&
+      moment(data.payload.datetime_created).add(1, 'minutes').isAfter(moment())
+    ) {
```
FuriousLlama, line 63: Lets add some documentation about why this is here. It looks to me that we just want to wait at least one minute before allowing the preview

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+      const pages = await getDocumentFilePageList(previewDocumentId, data.payload.file_latest.id);
+
+      //todo: handle various doc failures.
+      if (pages?.length > 0) {
+        const documentPageImage = await loadDocumentPageImage(
+          previewDocumentId,
+          data.payload.file_latest.id,
+          pages[0].id,
+        );
+        const lightboxPages = pages.map(page => ({ ...page } as LoadedPages));
```
FuriousLlama, line 100: q: why is the typecoercion necessary? is the result of the map not necessarily a LoadedPages?

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+        lightboxPages[0].loadedDocumentImageDataUrl = documentPageImage;
```
FuriousLlama, line 101: this is a tad odd. The left hand side implies that it would be an URL, but the right hand side implies it is an actual image. Should one be renamed or maybe some docs are needed?

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+    retrieveDocumentDetailLoading,
+    getDocumentFilePageListLoading,
+  } = useDocumentProvider();
+  const [documentPages, setDocumentPages] = useState<LoadedPages[]>([]);
+  const [currentPage, setCurrentPage] = useState<number>(0);
+
+  const loadDocumentPageImage = useCallback(
+    async (documentId: number, documentFileId: number, documentFilePageId: number) => {
+      //todo: handle image failure
+      return (createFileUrl ? createFileUrl : URL.createObjectURL)(
```
FuriousLlama, line 44: nothing to change, but this dynamic method calling looked very odd at first glance.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
FuriousLlama, line 100: - Using the lightBoxPage here is leaking abstraction from the view given that the View is responsible for rendering the container does not need to prepare the correct view type

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+      setDisplayModal(true);
+    } else if (data?.payload?.file_latest) {
+      const pages = await getDocumentFilePageList(previewDocumentId, data.payload.file_latest.id);
+
+      //todo: handle various doc failures.
+      if (pages?.length > 0) {
+        const documentPageImage = await loadDocumentPageImage(
+          previewDocumentId,
+          data.payload.file_latest.id,
+          pages[0].id,
```
FuriousLlama, line 98: This looks like is only necessary have the first page ready after the pages have been retrieved. Is this necessary because the useEffect at line 133 wont load the first page? If so, please add a comment with a small blurb indicating that. Otherwise, I would expect that loading the first page would work the same as loading all other pages

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+    getDocumentFilePageList,
+    downloadWrappedDocumentFileLatest,
+    retrieveDocumentDetailLoading,
+    getDocumentFilePageListLoading,
+  } = useDocumentProvider();
+  const [documentPages, setDocumentPages] = useState<LoadedPages[]>([]);
+  const [currentPage, setCurrentPage] = useState<number>(0);
+
+  const loadDocumentPageImage = useCallback(
+    async (documentId: number, documentFileId: number, documentFilePageId: number) => {
```
FuriousLlama, line 42: nit: being explicit about the return type of this function would help readability

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { ApiGen_Mayan_FilePage } from '@/models/api/generated/ApiGen_Mayan_FilePage';
+import { ApiGen_System_HttpStatusCode } from '@/models/api/generated/ApiGen_System_HttpStatusCode';
+
+import { DocumentViewerContext } from './context/DocumentViewerContext';
+import { IDocumentPreviewViewProps } from './DocumentPreviewView';
+import { createFileDownload } from './DownloadDocumentButton';
+import { useDocumentProvider } from './hooks/useDocumentProvider';
+
+export interface LoadedPages extends ApiGen_Mayan_FilePage {
+  loadedDocumentImageDataUrl?: string;
```
FuriousLlama, line 15: q: is this the image Blob?

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { ModalContext } from '@/contexts/modalContext';
+import { ApiGen_Mayan_FilePage } from '@/models/api/generated/ApiGen_Mayan_FilePage';
+import { ApiGen_System_HttpStatusCode } from '@/models/api/generated/ApiGen_System_HttpStatusCode';
+
+import { DocumentViewerContext } from './context/DocumentViewerContext';
+import { IDocumentPreviewViewProps } from './DocumentPreviewView';
+import { createFileDownload } from './DownloadDocumentButton';
+import { useDocumentProvider } from './hooks/useDocumentProvider';
+
+export interface LoadedPages extends ApiGen_Mayan_FilePage {
```
FuriousLlama, line 14: - Singular (LoadedPage)
- It would be better if this uses composition instead of inheritance given a loaded page contains both Mayan page information as well as the image retrieved for that page. That would help disambiguate things once that data is used somewherelese

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -0,0 +1,167 @@
...
+  loadedDocumentImageDataUrl?: string;
+  error: string;
+}
+
+export interface IDocumentPreviewContainerProps {
+  View: React.FC<IDocumentPreviewViewProps>;
+  createFileUrl?: (obj: Blob | MediaSource) => string;
```
FuriousLlama, line 21: This looks like is always necessary, why is it marked as optional here?

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: makes sense, I'll add some doc.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: Loaded pages is just an overload of the Page type with an extra field, I want to manually cast a Page to a LoadedPage so that I can have access to the extra field.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: probably renamed, the image is converted to a dataurl which itself contains the image.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: Fair, I'll move this down a level.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: No, this is the data url that was made from the blob.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: I don't mind changing this, but I struggle to see why the above approach is objectively better.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: nevermind, this is just a rename.

devinleighsmith: changed to composition

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
devinleighsmith: testing to see if this is necessary, this may be an artifact of a previous implementation.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
FuriousLlama: I think is a problem of where the implementation of the injection happens. Right now we don't have a good place for that to happen. I think it would be ok if it is done on the component itself as a default vs on the function

devinleighsmith: updated do default within the props.

FuriousLlama (approved):

PR #4013, opened 2024-05-10
PSP-8256 : Remove PS and LIS number from lease/area

eddherrera:


devinleighsmith:
Something going on with this PR, getting errors due to the PimsFileNumber schema not existing, but the version of the code in this PR doesn't have that schema.

devinleighsmith (approved):

eddherrera:
> Something going on with this PR, getting errors due to the PimsFileNumber schema not existing, but the version of the code in this PR doesn't have that schema.

I still need to update the branch with latest.

PR #4019, opened 2024-05-14
correct help href not being set during initial help modal open.

devinleighsmith:


source/frontend/src/features/help/containers/HelpContainer.tsx
```diff
@@ -45,6 +48,18 @@ export function HelpContainer() {
     update();
   }, [getSettings, config]);
 
+  useDeepCompareEffect(() => {
+    if (isMounted()) {
+      setModalContent({
+        ...modalProps,
+        okButtonHref:
```
devinleighsmith, line 55: Due to how the help modal was re-written, during the initial open, the mailto will not be set in the state var before the modal is displayed. This ensures the modal is updated when the mailto is loaded.

eddherrera (approved):

PR #4025, opened 2024-05-16
psp-7417 add health checks for Major pims services.

devinleighsmith:


source/backend/api/Helpers/Healthchecks/PimsCdogsHealthCheck.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _generationRepository = generationRepository;
+        }
+
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+
+                var fileTypes = await _generationRepository.TryGetFileTypesAsync();
```
devinleighsmith, line 23: I'm focusing on real, but low overhead methods for these health checks.

source/backend/api/Helpers/Healthchecks/PimsExternalApiHealthcheck.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Pims.Core.Exceptions;
+
+namespace Pims.Api.Helpers.Healthchecks
+{
+    public class PimsExternalApiHealthcheck : IHealthCheck
```
devinleighsmith, line 12: The idea here is that this is a generic health check that we can use for any external, unauthenticated api service.

source/backend/api/Helpers/Healthchecks/PimsGeoserverHealthcheck.cs
```diff
@@ -0,0 +1,59 @@
...
+            _statusCode = statusCode;
+        }
+
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            using (var client = new HttpClient())
+            {
+                try
+                {
+                    var basicAuth = Convert.ToBase64String(Encoding.GetEncoding("ISO-8859-1").GetBytes($"{_proxyOptions.GetValue<string>("ServiceUser")}:{_proxyOptions.GetValue<string>("ServicePassword")}"));
```
devinleighsmith, line 40: Will log some tech debt for this - currently all of the proxy logic is directly in the controller for geoserver - as a result the auth logic needs to be reproduced here. Would be nice to have that in a dependency injectable service so that the healthcheck can reuse the exact logic used by the endpoint (this puts this implementation at risk of drifting from the functional implementation).

source/backend/api/Helpers/Healthchecks/PimsLtsaHealthcheck.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            _pid = int.Parse(configuration.GetSection("HealthChecks:Ltsa:Pid").Value);
+            _ltsaService = ltsaService;
+        }
+
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                var titleSummary = await _ltsaService.GetTitleSummariesAsync(_pid);
```
devinleighsmith, line 25: This is the lowest impact rest endpoint ltsa has - the only one that doesn't involve generating a document as a result of the request.

source/backend/api/Helpers/Healthchecks/PimsMayanHealthCheck.cs
```diff
@@ -0,0 +1,35 @@
...
+        public PimsMayanHealthcheck(IEdmsDocumentRepository documentRepository)
+        {
+            _documentRepository = documentRepository;
+        }
+
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                var documentTypes = await _documentRepository.TryGetDocumentTypesAsync(string.Empty, 1, 1);
```
devinleighsmith, line 22: This seemed like a realistic choice from the rest spec, if there is a better choice let me know.

source/backend/api/Helpers/Healthchecks/PimsMetricsHealthCheck.cs
```diff
@@ -7,11 +7,11 @@
 
 namespace Pims.Api.Helpers.HealthChecks
 {
-    public class PimsMetricsHealthCheck : IHealthCheck
+    public class PimsMetricsHealthcheck : IHealthCheck
```
devinleighsmith, line 10: just standardizing naming.

FuriousLlama (approved):

PR #4026, opened 2024-05-17
psp-8185 remove closebutton generic modal prop - show custom pims clo…

devinleighsmith:
…se button by default.


source/frontend/src/components/common/GenericModal.tsx
```diff
@@ -211,19 +208,17 @@ export const GenericModal = (props: Omit<BsModalProps, 'onHide'> & ModalProps) =
       onHide={noop}
       className={getModalClass()}
     >
-      <Modal.Header closeButton={closeButton} onHide={close}>
+      <Modal.Header onHide={close}>
```
devinleighsmith, line 211: This closebutton actually causes a few problems.

Clicking this close button does not fire our close event.
the styling is hard to see and does not use one of our approved icons.

asanchezr (approved):

FuriousLlama (approved):

PR #4028, opened 2024-05-17
psp-8228 | Updated header status display

FuriousLlama:


source/frontend/src/components/common/HeaderField/StatusField.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+    return statusDictionary['cancelled'];
+  };
+
+  const statusFound = translateStatusCode(statusCodeType);
+  return (
+    <StyledBottomRow className="no-gutters justify-content-end align-items-end">
+      <Col />
+      <Col xs="auto" className="align-self-end">
+        <RetiredWarning $variant={statusFound.colorVariant}>
```
devinleighsmith, line 77: a prop with a name starting with a dollar sign seems new, what does that naming indicate?

FuriousLlama: I set it like to differentiate it from a react component (that one is a style). I dont feel strongly about it so will change if it helps

source/frontend/src/components/common/HeaderField/StatusField.tsx
```diff
@@ -0,0 +1,138 @@
...
+    `;
+    } else if (props.$variant === 'grey') {
+      return `
+  color: ${props.theme.css.disabledColor};
+  background-color: ${props.theme.css.disabledFieldBackgroundColor};
+    `;
+    } else if (props.$variant === 'blue') {
+      return `
+  color: ${props.theme.css.slideOutBlue};
+  background-color: ${props.theme.css.filterBoxColor};
```
devinleighsmith, line 116: I'm confused about the naming choices here. It seems like you could have named these colors whatever you wanted, but ended up going with some of the previously used colors that have since been removed.

For example, why call the blue slideOutBlue? I recognize that as a color that we used to have, but if this is the only incidence of this color, why not name this something closer to the usage. Ie. 'completedStatusColor' or similar.

source/frontend/src/assets/scss/colors.scss
```diff
@@ -1,23 +1,45 @@
...
+
 $pims-yellow-10: #fef1d8;
 $pims-yellow-30: #fcba19;
 $pims-red-10: #f2dede;
+$pims-red-80: #a12622;
 $pims-white: #f2f2f2;
-$pims-blue-20: #d9eaf7;
 $pims-yellow-100: #6c4a00;
 
+$pims-grey-10: #e8ecef;
```
devinleighsmith, line 15: I'm hesitant to introduce our own greys when grey 10 to 100 exists in the bcgov standard spec already:
https://github.com/bcgov/design-system/blob/main/packages/design-tokens/dist/css/variables.css

source/frontend/src/components/common/HeaderField/StatusField.tsx
FuriousLlama: tbh I had a hard time with the colors and the documentation on figma so I decided to name it exactly like the figma ones . I will chat with ana when she is back to get those names aligned.

source/frontend/src/assets/scss/colors.scss
```diff
@@ -1,23 +1,45 @@
...
+$light-danger-color: $pims-red-10;
 $highlight-background-color: $pims-white;
 $filter-box-color: $pims-blue-20;
 $text-warning-color: $pims-yellow-100;
 $header-text-color: #{'var(--theme-blue-100)'};
 $header-border-color: #{'var(--theme-blue-90)'};
 $link-color: #{'var(--surface-color-primary-button-default)'};
 $link-hover-color: #{'var(--surface-color-primary-button-hover)'};
+
+$completed-color: $pims-green-80;
```
devinleighsmith, line 37: some of the names here don't seem to match up with their usage in the app which I don't understand.

I think the idea behind these mappings are as follows:
1) in the upper section, define a list of colours **that are not already in the bcgov spec**
2) map those colors to subjective names within the pims app that give developers a clue where that colour is being used. ie. if pims has a "silly" button, perhaps we create a color called "silly-button-color" and map one of our non-gov colours to it. That way if we want to tweak the color of all silly buttons in the app we can do so in one place. (as a result, we may map the same colour to multiple concepts within pims).

Anyways, these colour subjective names seem to have been introduced specifically for these new status labels, and the status labels only, but these names don't seem to reflect that.

source/frontend/src/assets/scss/colors.scss
FuriousLlama: That is reasonable. I added it like that because the figma spec used the code instead of a color name. The rest is the same as the response above.

source/frontend/src/assets/scss/colors.scss
FuriousLlama: I dont like the color names at all either. However, that's what they are called on figma, so I went for that for now.

devinleighsmith: If ana went with that as a color name, it would likely be because she misunderstood the documentation I sent her during the color palette rework. She asked me for a list of old colors and new colors, and maybe got confused as to which was which (perhaps thinking that the list with more colors was newer, which is actually not the case). 

This is actually very unfortunate, because she may have chosen a number of colors from the list that we both (Ana and myself) agreed to remove from the application in favour of colors already in the bcgov spec.

We can go with this for now, and I'll log another tech debt item to clean up the colors again with Ana.

devinleighsmith (approved):

devinleighsmith (approved):

PR #4036, opened 2024-05-22
UI/improvements

FuriousLlama:
Reviewing per commit should be easier.


source/frontend/src/features/mapSideBar/acquisition/tabs/agreement/common/UpdateAcquisitionAgreementForm.tsx
```diff
@@ -81,16 +82,6 @@ const UpdateAcquisitionAgreementForm: React.FunctionComponent<
 
 export default UpdateAcquisitionAgreementForm;
 
-const StyledFormWrapper = styled.div`
```
FuriousLlama, line 84: This class was repeated multiple times. We need to keep an eye out for simplification when possible.

devinleighsmith:
What is your suggestion for reviewing this? also what release should this be added to if the next sprint is a release sprint?

FuriousLlama:
> What is your suggestion for reviewing this? also what release should this be added to if the next sprint is a release sprint?

By commits. Most of the changes are pure padding changes, so it should be fairly trivial. The changes in inputs require a bit of attention, but nothing crazy.

FuriousLlama:
> What is your suggestion for reviewing this? also what release should this be added to if the next sprint is a release sprint?

This release would be fine so we get time to see the changes in dev and test. Nothing functional so IMO is pretty safe

source/frontend/src/components/common/form/ContactInput/ContactInputView.tsx
```diff
@@ -109,30 +119,44 @@ const ContactInputView: React.FunctionComponent<IContactInputViewProps> = ({
 
 export default ContactInputView;
 
-const StyledDiv = styled.div`
-  background: none;
+const StyledInputLikeDiv = styled.div<{ $hasContact: boolean }>`
```
devinleighsmith, line 122: I'm not sure I get this name, isn't this just a div with a classname pointing to a styled components class? why is this "like" a div?

FuriousLlama: I think it makes it clear that that div is imitiating the look of an input.

PR #4037, opened 2024-05-22
PSP-8474 | Fixed property updates removing is-owned

FuriousLlama:


devinleighsmith (changes requested):
Please increment the hotfix version @FuriousLlama to v5.2.3-79.18

devinleighsmith:
@FuriousLlama would it make sense to update the property repository to prevent this field from being set at all by the property service? I think only the take service should be allowed to change ownership, correct?

FuriousLlama:
> @FuriousLlama would it make sense to update the property repository to prevent this field from being set at all by the property service? I think only the take service should be allowed to change ownership, correct?

That could work. We could design something more robust that inclues other fields as well like location and prevents updates to those. For now, I think this change is sufficient

devinleighsmith (approved):

PR #4039, opened 2024-05-23
pims document preview - pre-cache.

devinleighsmith:
By default, Mayan will only cache document pages on a schedule, or when they are requested. This adds an async pre-cache task to the upload action to facilitate that - greatly increasing the odds that a page preview will be ready when requested.


devinleighsmith:
Also adds a retry to the image download, and provides a parameter to control the max pages returned by the previewer.

eddherrera (approved):

PR #4044, opened 2024-05-24
psp 7495 | Lease tenant UI update

FuriousLlama:


source/frontend/src/features/leases/detail/LeasePages/tenant/AddLeaseTenantForm.test.tsx
```diff
@@ -91,7 +91,7 @@ describe('AddLeaseTenantForm component', () => {
       component: { getByText },
     } = await setup({});
 
-    const tenantButton = getByText('Select Tenant(s)');
```
devinleighsmith, line 94: This doesn't seem correct, since the user may select multiple tenants on this screen.

devinleighsmith (approved):

source/frontend/src/features/leases/detail/LeasePages/tenant/AddLeaseTenantForm.test.tsx
FuriousLlama: This is just the test matcher for the text. The text itself has not been changed

PR #4046, opened 2024-05-27
psp-8477 - use fully attributed parcel map.

devinleighsmith:


source/frontend/src/components/maps/leaflet/Layers/util.tsx
```diff
@@ -272,10 +272,10 @@ export const isPimsBoundary = (
...
 };
 
-export const isParcelMap = (
+export const isFaParcelMap = (
   feature: Supercluster.PointFeature<MarkerFeature>,
-): feature is Supercluster.PointFeature<PMBC_Feature_Properties> => {
-  return feature.id?.toString().startsWith('WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW') ?? false;
+): feature is Supercluster.PointFeature<PMBC_FullyAttributed_Feature_Properties> => {
+  return feature.id?.toString().startsWith('WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW') ?? false;
 };
```
asanchezr, line 279: Given that locally we don't have fully attributed parcel map this check would fail, right? Like **isFaParcelMap** will never be true in local development. Are there any implications that we should know about?

source/frontend/src/components/maps/leaflet/Markers/SingleMarker.tsx
```diff
@@ -60,23 +64,23 @@ const SinglePropertyMarker: React.FC<React.PropsWithChildren<SinglePropertyMarke
...
       });
-    } else if (isParcelMap(pointFeature)) {
+    } else if (isFaParcelMap(pointFeature)) {
       mapMachine.mapMarkerClick({
         clusterId: clusterId,
         latlng: latlng,
         pimsLocationFeature: null,
         pimsBoundaryFeature: null,
-        pmbcFeature: pointFeature.properties,
+        fullyAttributedFeature: pointFeature.properties,
```
asanchezr, line 83: This code block will never execute on local development - I'd like to know if that would impact what we can click on the map while doing local development

source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.test.tsx
```diff
@@ -71,7 +71,7 @@ describe('MotiInventoryContainer component', () => {
     mockAxios
       .onGet(
         new RegExp(
-          'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW/ows*',
+          'https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW/ows*',
```
asanchezr, line 74: I think in tests we don't use the FA parcel map so leaving the old value should be okay. just checking

asanchezr (approved):

PR #4051, opened 2024-05-29
psp-7494 | Lease depost UI enhancements

FuriousLlama:


devinleighsmith:
I think it is likely we aren't being strict about these icons not matching the mockup, but they should at least be consistent on this page:
image
image

source/frontend/src/components/common/SectionListHeader.tsx
```diff
@@ -12,6 +12,7 @@ export interface ISectionListHeaderProps {
   addButtonIcon?: JSX.Element;
   onAdd?: () => void;
   claims: Claims[];
+  dataTestid?: string;
```
devinleighsmith, line 15: fyi: this is the way to make this consistent with the standard naming of this prop: 'data-testId': string;

PR #4061, opened 2024-05-30
PSP-8194 Add Complete Support for MultiPolygons

asanchezr:
Tested by temporarily displaying the boundaries in the PSP database as a map layer. I was able to see multi-polygon boundaries being added to the database and displayed correctly on the map

https://github.com/bcgov/PSP/assets/24322224/3ae092dd-81bc-40d5-b0c5-ae92a098cb9a


source/backend/api/Services/PropertyService.cs
```diff
@@ -349,36 +332,41 @@ public PimsProperty PopulateNewProperty(PimsProperty property, bool isOwned = fa
             var boundaryGeom = property.Boundary;
             if (boundaryGeom != null && boundaryGeom.SRID != SpatialReference.BCALBERS)
             {
-                var newCoords = property.Boundary.Coordinates.Select(coord => _coordinateService.TransformCoordinates(boundaryGeom.SRID, SpatialReference.BCALBERS, coord));
-                var gf = NetTopologySuite.NtsGeometryServices.Instance.CreateGeometryFactory(SpatialReference.BCALBERS);
-                property.Boundary = gf.CreatePolygon(newCoords.ToArray());
+                _coordinateService.TransformGeometry(boundaryGeom.SRID, SpatialReference.BCALBERS, boundaryGeom);
```
devinleighsmith, line 335: Does it make sense to do this here, or should we just call the new UpdateLocation method?

devinleighsmith:
Given the scope of these changes, I think we will be better served by pushing this to 5.4, what do you think?

asanchezr:
> Given the scope of these changes, I think we will be better served by pushing this to 5.4, what do you think?

I was actually thinking the same, I can change the PR to target the 5.4 branch if you agree. I'd rather see this tested with lots of time ahead for fixes etc

devinleighsmith (approved):
approved, given that tests are coming in a future PR.

devinleighsmith:
yes, lets push it and the story to 5.4, that will give us some increased confidence, given that a lot of our existing integration tests don't test multipolygons.

source/backend/api/Services/PropertyService.cs
asanchezr: The **UpdateLocation** calls the property repository to- well - update the property in the DB. This method is called when we are adding a brand new one as part of a larger parent entity (say an acquisition file that gets a brand new property added). 

Do you have a suggestion on how to reuse the UpdateLocation to not call the DB?

PR #4075, opened 2024-06-04
psp-8585 display either retired or disposed status in property header…

devinleighsmith:
… if relevant


source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
```diff
@@ -43,6 +44,16 @@ export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderPr
...
+  const isDisposed = React.useMemo(() => {
+    if (
+      exists(geoserverMapData?.features[0]) &&
+      geoserverMapData?.features[0]?.properties?.IS_DISPOSED
+    ) {
+      return true;
+    }
+    return false;
+  }, [geoserverMapData?.features]);
+
```
asanchezr, line 56: **nit:** This code assumes that the features array always has elements - perhaps check for null/empty array to avoid NPE?

asanchezr (approved):

PR #4077, opened 2024-06-04
PSP-8586 : FT-REG: Subdivision/ Consolidation - Menu icon enlarges when user clicks on expand the left menu section

eddherrera:


source/frontend/src/components/layout/SideNavBar/styles.ts
```diff
@@ -20,7 +20,7 @@ export const SideNavBar = styled.div`
   align-items: center;
   grid-area: iconbar;
   background-color: ${props => props.theme.tenant.colour};
-  transition: 0.5s width;
+  transition: 0.3s width;
```
devinleighsmith, line 23: what is the motivation for this change?

devinleighsmith (changes requested):
Hmm, I think something is borked in that svg now, prior to this change you can view the svg in the browser as it is considered "valid". see https://github.com/bcgov/PSP/blob/dev/source/frontend/src/assets/images/subdivisionconsolidation.svg

devinleighsmith:
@eddherrera see: https://raw.githubusercontent.com/bcgov/PSP/b42a01ff309fa7952b17ef556975cbede6f64cc6/source/frontend/src/assets/images/subdivisionconsolidation.svg

source/frontend/src/components/layout/SideNavBar/styles.ts
eddherrera: I was trying to reduce the animation time, but I see that it's not needed.

devinleighsmith:
@eddherrera looks close, vertical padding is a bit off:

image

devinleighsmith (approved):

PR #4078, opened 2024-06-05
PSP-8587 table pagination visual bug

asanchezr:
image

### DIFF
--- a/source/frontend/src/components/Table/PageSizeSelector.tsx
+++ b/source/frontend/src/components/Table/PageSizeSelector.tsx
@@ -4,19 +4,14 @@ import styled from 'styled-components';
 
 import { IMenuItemProps, Menu } from '@/components/menu/Menu';
 
-/** align text to middle of size input */
-const StyledText = styled.span`
-  margin-top: 0.3rem;
-`;
-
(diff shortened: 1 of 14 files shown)


source/frontend/src/components/Table/PageSizeSelector.tsx
```diff
@@ -38,17 +33,25 @@ export const TablePageSizeSelector: React.FC<React.PropsWithChildren<IProps>> =
...
   );
 };
+
+/** align text to middle of size input */
+const StyledText = styled.span`
+  margin-top: 0.3rem;
+`;
+
+const StyledInput = styled(Form.Control)`
+  width: 7.5rem;
```
devinleighsmith, line 51: this appears to be 25ps bigger then it previously was.

asanchezr: It is - we need to make it wider to display 2 and 3 digit numbers. The actual bug is that it says it "shows 1 instead of 10".
I found out that the component does "display" 10 by default but the input is too narrow so it only displays the first digit

devinleighsmith (approved):

PR #4099, opened 2024-06-11
Update pims health checks to support configuration based enable/disab…

devinleighsmith:
…le and period.

### DIFF
--- a/source/backend/api/Startup.cs
+++ b/source/backend/api/Startup.cs
@@ -315,7 +315,7 @@ namespace Pims.Api
             {
                 services.AddHealthChecks().Add(new HealthCheckRegistration(
                     "Ltsa",
-                    sp => new PimsLtsaHealthcheck(this.Configuration, sp.GetService<ILtsaService>()),
+                    sp => new PimsLtsaHealthcheck(allHealthCheckOptions.Ltsa, sp.GetService<ILtsaService>()),
                     null,
                     new string[] { "services", "external" })
                 { Period = TimeSpan.FromMinutes(allHealthCheckOptions.Ltsa.Period) });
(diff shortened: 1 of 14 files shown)


source/backend/api/Startup.cs
```diff
@@ -235,62 +237,109 @@ public void ConfigureServices(IServiceCollection services)
...
+            }
 
-            services.AddHealthChecks()
-                .AddCheck(
-                    "api-metrics",
-                    new PimsMetricsHealthcheck(csBuilder.ConnectionString),
+            if (allHealthCheckOptions.ApiMetrics.Enabled)
+            {
+                services.AddHealthChecks().Add(new HealthCheckRegistration(
+                    "ApiMetrics",
```
asanchezr, line 277: I believe we use the string "api-metrics" in our Prometheus/Sysdig dashboards. We will need to update the dashboards to read from the new "ApiMetrics" information or they will show no data...

asanchezr: NVM - I checked Sysdig and we don't use that name at all - disregard previous comment.

source/backend/api/appsettings.json
```diff
@@ -4,20 +4,41 @@
...
       "Url": "/ows?service=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=PIMS_PROPERTY_LOCATION_VW&srsName=EPSG:4326&count=0",
-      "StatusCode": 200
+      "StatusCode": 200,
+      "Enabled": true,
+      "Duration": 1
     },
     "Ltsa": {
-      "Pid": "6209521"
+      "Enabled": true,
+      "Duration": 1
```
asanchezr, line 33: I see a discrepancy here in regards to the duration/period. In other configuration files I see the override values as "Period" vs here they are set originally as "Duration". I think you need to pick one...
image

asanchezr (approved):

PR #4104, opened 2024-06-11
Psp 8635, psp-8636, psp-8638

devinleighsmith:
see commits for story definitions.

### DIFF
--- a/source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
+++ b/source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
@@ -92,8 +92,8 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
     return mapMachine.filePropertyLocations.map(x => {
       // The values on the feature are rounded to the 4th decimal. Do the same to the draft points.
       return {
-        lat: parseFloat(x.lat.toFixed(4)),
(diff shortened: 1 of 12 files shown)


source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -92,8 +92,8 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
     return mapMachine.filePropertyLocations.map(x => {
       // The values on the feature are rounded to the 4th decimal. Do the same to the draft points.
       return {
-        lat: parseFloat(x.lat.toFixed(4)),
```
devinleighsmith, line 95: this was causing some draft markers to be offset from their parent markers at high zoom levels.

tools/geoserver/geoserver_data/psp/settings.xml
```diff
@@ -11,7 +11,7 @@
     <id>contact</id>
   </contact>
   <charset>UTF-8</charset>
-  <numDecimals>4</numDecimals>
+  <numDecimals>8</numDecimals>
```
devinleighsmith, line 14: I believe this is the reason the draft markers were being limited to four decimal places - but this restriction only exists on our local geoserver and not the moti one.

PR #4106, opened 2024-06-12
PSP-8575 : Lease/License – Improvements Tab (Other Improvements) Tool…

eddherrera:
… Tip

### DIFF
--- a/source/frontend/src/features/leases/detail/LeasePages/improvements/AddImprovementsForm.tsx
+++ b/source/frontend/src/features/leases/detail/LeasePages/improvements/AddImprovementsForm.tsx
@@ -1,5 +1,5 @@
 import { FieldArray, Formik, FormikProps } from 'formik';
-import React from 'react';
+import React, { ReactNode } from 'react';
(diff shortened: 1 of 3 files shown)


source/frontend/src/features/leases/detail/LeasePages/improvements/AddImprovementsForm.tsx
```diff
@@ -51,9 +52,29 @@ export const AddImprovementsForm: React.FunctionComponent<
                         const title =
                           sectionTitles.get(improvement.propertyImprovementTypeId) ?? 'N/A';
                         const nameSpace = `improvements.${index}`;
+
+                        const HeaderDisplayNode = (typeId: string): ReactNode => {
```
devinleighsmith, line 56: In this case I think this duplication is fine, but let's try and use reusable components where appropriate to reduce duplication and reduce the risk of view/edit visual differences.

devinleighsmith (approved):

asanchezr (approved):

PR #4109, opened 2024-06-12
PSP-8566 : FT: Lease & Licences - Tenants: Select Tenants button gets…

eddherrera:
… blue after clicking on it.


source/frontend/src/components/common/SectionListHeader.tsx
```diff
@@ -19,7 +19,10 @@ export const SectionListHeader: React.FunctionComponent<
   React.PropsWithChildren<ISectionListHeaderProps>
 > = props => {
   const { hasClaim } = useKeycloakWrapper();
-  const onClick = () => props.onAdd && props.onAdd();
+  const onClick = (e): void => {
```
devinleighsmith, line 22: @eddherrera I think this bug is a styling issue (CSS) not a functionality issue (JS).

try something like this:

`
export const StyledAddButton = styled(Button)`
  &.btn.btn-primary {
    background-color: ${props => props.theme.bcTokens.iconsColorSuccess};
  }
  &.btn.btn-primary:active,
  &.btn.btn-primary:focus {
    background-color: ${props => props.theme.bcTokens.iconsColorSuccess};
    outline-color: ${props => props.theme.css.pimsGreen80};
  }
  &.btn.btn-primary:hover {
    background-color: ${props => props.theme.css.pimsGreen80};
  }
;
`

eddherrera: @devinleighsmith not so sure about this one. The styling definitions are as expected. This is a well-known issue with many workarounds. Either bootstrap or react is not resetting the button state.

devinleighsmith: Hmm, is this an issue with other buttons? I think the problem here is that a green button turns blue, not that it goes into some kind of "clicked" state that isn't reset properly.

eddherrera: Confirmed with reporter the issue is leaving the focus after modal is closed. Which is a bootstrap modal option. So I am turning it off.

devinleighsmith (approved):

PR #4128, opened 2024-06-21
PSP-8409 Takes UI/UX cleanup

asanchezr:
image

### DIFF
--- a/source/frontend/src/features/mapSideBar/property/tabs/takes/detail/TakesDetailView.tsx
+++ b/source/frontend/src/features/mapSideBar/property/tabs/takes/detail/TakesDetailView.tsx
@@ -3,7 +3,6 @@ import { FaPlus, FaTrash } from 'react-icons/fa';
 import styled from 'styled-components';
 
 import { StyledRemoveLinkButton } from '@/components/common/buttons';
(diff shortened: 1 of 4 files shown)


source/frontend/src/features/mapSideBar/property/tabs/takes/detail/TakesDetailView.tsx
```diff
@@ -182,14 +182,10 @@ export const TakesDetailView: React.FunctionComponent<ITakesDetailViewProps> = (
...
-                    labelWidth="8"
+                    labelWidth="9"
                     tooltip="The term new highway dedication includes municipal road or provincial public highway."
                   >
-                    <YesNoButtons
-                      id="newRightOfWayToggle"
-                      disabled
-                      value={take.isNewHighwayDedication ?? undefined}
-                    />
+                    {booleanToYesNoUnknownString(take.isNewHighwayDedication ?? undefined)}
```
eddherrera, line 188: nit: Probably you don't need the '??' operator, since the property is 'bool | null' and the function accepts 'bool | null'.

eddherrera: same for others in this component.

eddherrera (approved):

PR #4147, opened 2024-07-02
PSP-7158 Show markers positioned corresponding to that file 

asanchezr:
https://github.com/bcgov/PSP/assets/24322224/bcbc9321-57d6-4da6-97fe-9be122eae9c4


source/backend/api/Services/PropertyService.cs
```diff
@@ -377,6 +377,40 @@ public void UpdateLocation(PimsProperty incomingProperty, ref PimsProperty prope
...
+
+        public void UpdateFilePropertyLocation<T>(T incomingFileProperty, T filePropertyToUpdate)
+            where T : IWithPropertyEntity
+        {
+            // TODO: Remove this casting when LOCATION gets added to all remaining file-property types (research, disposition, lease)
+            if (incomingFileProperty is PimsPropertyAcquisitionFile incomingAcquisitionProperty
+                && filePropertyToUpdate is PimsPropertyAcquisitionFile acquisitionPropertyToUpdate)
+            {
+                // convert spatial location from lat/long (4326) to BC Albers (3005) for database storage
+                var geom = incomingAcquisitionProperty.Location;
```
devinleighsmith, line 405: does the above casting also check for null?

source/backend/api/Services/PropertyService.cs
```diff
@@ -377,6 +377,40 @@ public void UpdateLocation(PimsProperty incomingProperty, ref PimsProperty prope
             }
         }
 
+        public T PopulateNewFileProperty<T>(T fileProperty)
+        {
+            // TODO: Remove this casting when LOCATION gets added to all remaining file-property types (research, disposition, lease)
+            if (fileProperty is PimsPropertyAcquisitionFile acquisitionFileProperty)
+            {
+                // convert spatial location from lat/long (4326) to BC Albers (3005) for database storage
+                var geom = acquisitionFileProperty.Location;
```
devinleighsmith, line 386: does this casting check for null?

source/frontend/src/features/mapSideBar/disposition/models/DispositionFormModel.ts
```diff
@@ -102,6 +93,20 @@ export class DispositionFormModel implements WithDispositionTeam {
     };
   }
 
+  private toPropertyApi(x: PropertyForm): ApiGen_Concepts_DispositionFileProperty {
```
devinleighsmith, line 96: Would it be worth creating a single function that converts to ApiGen_Concepts_FileProperty instead?

source/backend/api/Services/PropertyService.cs
asanchezr: According to MS docs: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/tutorials/safely-cast-using-pattern-matching-is-and-as-operators
image

source/frontend/src/features/mapSideBar/disposition/models/DispositionFormModel.ts
asanchezr: Yes, we could create that for the 2 files that don't have extra info in there (Disposition and Acquisition). Research and Lease I believe have extra fields but we could call the shared function and then append the extra stuff. Will look into it

devinleighsmith (approved):

source/backend/api/Services/PropertyService.cs
asanchezr: Also - a note that this casting will go away this sprint with the story to use location on other file types

asanchezr:
Merging this PR as-is to get it to FT. I will address PR comments in my next related PR

PR #4157, opened 2024-07-04
psp-8780 scaffold of IS84

devinleighsmith:


source/backend/entities/ef/PimsPropertyActivityHist.cs
```diff
@@ -50,10 +50,6 @@ public partial class PimsPropertyActivityHist
     [Column("COMPLETION_DT")]
     public DateOnly? CompletionDt { get; set; }
 
-    [Column("DESCRIPTION")]
-    [StringLength(2000)]
-    public string Description { get; set; }
-
```
asanchezr, line 56: I know this is auto-generated but wanted to double check that it makes sense to drop this column in the history table while keeping the original in the PimsPropertyActivity table...

asanchezr (approved):

source/backend/entities/ef/PimsPropertyActivityHist.cs
FuriousLlama: Good catch. We can leave things in the hist table that are being removed from the original table but not the other way around

FuriousLlama: Also, if we are removing a column that never saw the light of prod, then removing it from hist makes sense

PR #4160, opened 2024-07-05
PSP-8414 : Select one or multiple properties for H120 Generation

eddherrera:
image

image


eddherrera:
*** Requires template update before merge

devinleighsmith:
Trying to remove a property that is on a compensation throws an unhandled db exception. Probably want to add to this logic: 

`
if (acqFileProperties.PimsTakes.Any() || acqFileProperties.PimsInthldrPropInterests.Any())
{
    throw new BusinessRuleViolationException("You must remove all takes and interest holders from an acquisition file property before removing that property from an acquisition file");
}
`

devinleighsmith:
The implemented view display of selected properties doesn't match figma and will probably be brought up as a bug by the FT team. Is this something you discussed with Ana, and is that documented on the story somewhere?
implementation:
image
mockup:
image

devinleighsmith:
Also this AC is not met:
image

devinleighsmith:
@eddherrera if I select/deselect multiple properties at a time the current implementation only seems to log one of my selections/deselections when I save.

devinleighsmith (changes requested):
some mismatched requirements, functional issues.

devinleighsmith (approved):

PR #4166, opened 2024-07-08
Psp 8368 variable payments, gst fixes, unit tests.

devinleighsmith:


source/backend/tests/unit/api/Services/PaymentServiceTest.cs
```diff
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Azure;
```
eddherrera, line 5: nit

eddherrera (approved):

PR #4168, opened 2024-07-08
psp-8783 ensure that form state is maintained during collapsing of sidebar.

devinleighsmith:


source/frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
```diff
@@ -27,101 +26,88 @@ const MapSideBarLayout: React.FunctionComponent<
...
-                  tooltipId="expand-sidebar-tooltip-collapse"
-                  tooltip={'Collapse Screen'}
-                >
-                  <span>
-                    <ExpandCollapseButton
-                      expanded={!mapSideBarViewState.isCollapsed}
-                      toggleExpanded={toggleSidebarDisplay}
-                    />
-                  </span>
+    <StyledSidebarWrapper className={mapSideBarViewState.isCollapsed ? '' : 'expanded'}>
```
FuriousLlama, line 29: nit: this could also be passed as a parameter to the styled component

FuriousLlama (approved):

asanchezr:
@devinleighsmith please update the snapshots to resolve conflicts

source/frontend/src/features/mapSideBar/layout/MapSideBarLayout.tsx
asanchezr: The other subtle change here is that we removed the **VisibilitySensor** (which we mock in a lot of unit tests). Given that we don't use it anymore we should update the tests to remove the unnecessary mocking.
Not required for this PR - should be a tech-debt task. Just calling out the change

asanchezr (approved):

PR #4170, opened 2024-07-09
psp-8772 lease payment export corrections.

devinleighsmith:


asanchezr (approved):

devinleighsmith:
Also, if we are adding text fields, the yup schema must be updated to enforce the max character limits of that text field.

PR #4172, opened 2024-07-09
PSP-8693: Rename Header Dates (Lease and License)

asanchezr:
image


devinleighsmith (approved):
You may have some issues coming back from FT on the mockup vs implementation
Current expiry(mockup) vs. Expiry(implementation)
<File Status>(implementation) vs. <File Status>(mockup)

asanchezr:
> You may have some issues coming back from FT on the mockup vs implementation Current expiry(mockup) vs. Expiry(implementation) File: (implementation) vs. (mockup)

Appreciate the feedback. I went with what was described in Confluence. We will need to sync up the mockup and Confluence when Ana is back

PR #4183, opened 2024-07-11
psp-8784 do not display cancel modal unless changes have been made to…

devinleighsmith:
… lease screen.


source/frontend/src/features/leases/add/ConsultationSubForm.tsx
```diff
@@ -84,4 +56,44 @@ const ConsultationSubForm: React.FunctionComponent<
...
+        id: 0,
+        parentLeaseId: lease.id || 0,
+        consultationType: {
+          id: consultationType.id.toString(),
+          description: consultationType.name,
+          displayOrder: null,
+          isDisabled: false,
+        },
+        consultationStatusType: {
+          id: 'UNKNOWN',
```
eddherrera, line 74: is there a generate code type that we could use here instead of hardcoding? There has been many codetypes changes around lately.

devinleighsmith: I agree, I think making improvements to that is outside the scope of this bug for now, since I'm just moving some existing code around.

PR #4188, opened 2024-07-11
PSP-7980 : Display L/L Status in Property Information Pane - Management (Summary Section)

eddherrera:
image


source/backend/api/Services/PropertyService.cs
```diff
@@ -209,9 +209,70 @@ public PropertyManagementModel GetPropertyManagement(long propertyId)
...
             return propertyManagement;
         }
 
+        private static void PropertyHasActiveLease(IEnumerable<PimsPropertyLease> propertyLeases, out bool hasActiveLease, out bool hasActiveExpiryDate)
+        {
+            hasActiveLease = false;
+            hasActiveExpiryDate = false;
+
+            List<PimsLease> activeAgreementsList = propertyLeases.Select(x => x.Lease).ToList();
+            var activeLease = activeAgreementsList.FirstOrDefault(x => x.LeaseStatusTypeCode == LeaseStatusTypes.ACTIVE.ToString());
```
devinleighsmith, line 225: My understanding of the requirements, is that if there is **any** lease that is active for a property, that should be displayed on the management screen.

Given that a single property may have multiple leases (ie, a large lot with many buildings). it seems entirely possible that a single property may have multiple active leases, where some of those active leases may have expired or been terminated.

as a result, I think using FirstOrDefault here doesn't work.

source/backend/api/Services/PropertyService.cs
```diff
@@ -209,9 +209,70 @@ public PropertyManagementModel GetPropertyManagement(long propertyId)
...
+            if (activeLease is not null)
+            {
+                // Lease is Active but has termination Date.
+                if (activeLease.TerminationDate.HasValue)
+                {
+                    return;
+                }
+                else
+                {
+                    var latestRemewal = activeLease.PimsLeaseRenewals.Where(x => x.IsExercised.HasValue && x.IsExercised.Value).OrderByDescending(x => x.CommencementDt).FirstOrDefault();
```
devinleighsmith, line 237: nit latestRemewal -> latestRenewal

devinleighsmith: FYI a fast way of checking if a nullable boolean is a specific value is just x.IsExercised == true, since null != true and false != true

source/backend/api/Services/PropertyService.cs
```diff
@@ -209,9 +209,70 @@ public PropertyManagementModel GetPropertyManagement(long propertyId)
...
+                    if (latestRemewal is null) // No Renewal - Check only Lease dates.
+                    {
+                        if(activeLease.OrigExpiryDate.HasValue && activeLease.OrigExpiryDate.Value.Date >= DateTime.UtcNow.Date)
```
devinleighsmith, line 240: I think this is an example of when you don't want to use the UTC date. If you did, and a user used this logic before 7am, the utc date would not match the pst date.

devinleighsmith: Another way to think of this - when we store dates with times, we store them in UTC. When we store dates only, we store them as-is.

source/backend/api/Services/PropertyService.cs
```diff
@@ -209,9 +209,70 @@ public PropertyManagementModel GetPropertyManagement(long propertyId)
...
+                    }
+                    else
+                    {
+                        if(activeLease.OrigExpiryDate.HasValue && latestRemewal.ExpiryDt.HasValue)
+                        {
+                            hasActiveLease = hasActiveExpiryDate = activeLease.OrigExpiryDate.Value.Date >= DateTime.UtcNow.Date || latestRemewal.ExpiryDt.Value.Date >= DateTime.UtcNow.Date;
+                        }
+                        else if(activeLease.OrigExpiryDate.HasValue && !latestRemewal.ExpiryDt.HasValue)
+                        {
+                            hasActiveLease = true;
```
devinleighsmith, line 261: wouldn't this also mean that the lease has an active expiry date?

source/backend/api/Services/PropertyService.cs
```diff
@@ -209,9 +209,70 @@ public PropertyManagementModel GetPropertyManagement(long propertyId)
...
+                        }
+                        else if (!activeLease.OrigExpiryDate.HasValue && latestRemewal.ExpiryDt.HasValue)
+                        {
+                            hasActiveLease = latestRemewal.ExpiryDt.Value.Date >= DateTime.UtcNow.Date;
```
devinleighsmith, line 265: same thing here?

source/backend/api/Services/PropertyService.cs
asanchezr: Please note that this is backend (C#) code so this runs in the timezone of the .NET container - which by default is UTC in docker/openshift. I'm not aware that we set the timezone to PST on our live env containers.
If we want to do this "Date math" in PST we need to do that manually (either subtract 7hrs from UTC or convert the exp date to UTC)

devinleighsmith (approved):

source/backend/api/Services/PropertyService.cs
eddherrera: Updated to check for other leases with a loop.

source/backend/api/Services/PropertyService.cs
eddherrera: Updated

source/backend/api/Services/PropertyService.cs
eddherrera: not quite sue, I am trying to cover for the scenarios not documented in the AC.

PR #4189, opened 2024-07-12
psp-8771 & psp-8683  | Lease expiry using renewal information

FuriousLlama:
Updated lease expiry search to use renewal information. Updated expiry definition for leases

### DIFF
--- a/source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
+++ b/source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
@@ -11,19 +11,19 @@ namespace Pims.Api.Areas.Reports.Mapping.Lease
     {
         public void Register(TypeAdapterConfig config)
         {
-            config.NewConfig<(Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant), Model.LeaseModel>()
+            config.NewConfig<(Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant, DateTime? expiryDate), Model.LeaseModel>()
                 .AfterMapping((src, dest) =>
                 {
                     MapLease(src, dest);
                 });
         }
 
-        private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant) src, Model.LeaseModel dest)
+        private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant, DateTime? expiryDate) src, Model.LeaseModel dest)
         {
             dest.LFileNo = src.lease.LFileNo;
             dest.MotiRegion = src.lease.RegionCodeNavigation?.RegionName;
             dest.StartDate = src.lease.OrigStartDate.FilterSqlMinDate().ToNullableDateOnly();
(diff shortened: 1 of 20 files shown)


source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
```diff
@@ -11,19 +11,19 @@ public class LeaseMap : IRegister
...
         }
 
-        private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant) src, Model.LeaseModel dest)
+        private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease lease, Entity.PimsPropertyLease property, Entity.PimsLeaseTenant tenant, DateTime? expiryDate) src, Model.LeaseModel dest)
         {
             dest.LFileNo = src.lease.LFileNo;
             dest.MotiRegion = src.lease.RegionCodeNavigation?.RegionName;
             dest.StartDate = src.lease.OrigStartDate.FilterSqlMinDate().ToNullableDateOnly();
-            dest.EndDate = src.lease.OrigExpiryDate?.FilterSqlMinDate().ToNullableDateOnly();
+            dest.EndDate = src.expiryDate.ToNullableDateOnly();
```
devinleighsmith, line 26: Do we need to filter the sql min date here? previously this would cause exceptions converting from an sql min date to a date time.

source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
```diff
@@ -36,7 +36,7 @@ private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease le
             dest.InspectionNotes = src.lease.InspectionNotes;
             dest.InspectionDate = src.lease.InspectionDate?.FilterSqlMinDate();
             dest.LeaseNotes = src.lease.LeaseNotes;
-            dest.IsExpired = (src.lease.GetExpiryDate() < DateTime.Now).BoolToYesNo();
+            dest.IsExpired = (src.expiryDate < DateTime.Now).BoolToYesNo();
```
devinleighsmith, line 39: Do we need to include renewals in this calculation?

source/frontend/src/features/leases/leaseUtils.ts
```diff
@@ -16,6 +20,21 @@ export const getAllNames = (tenants: ApiGen_Concepts_LeaseTenant[]): string[] =>
...
 };
 
+export const getCalculatedExpiry = (
+  lease: ApiGen_Concepts_Lease,
+  renewals: ApiGen_Concepts_LeaseRenewal[],
+): string => {
+  const excercisedRenewalDates = renewals.filter(r => r.isExercised).flatMap(fr => fr.expiryDt);
+
+  let calculatedExpiry: string | null = lease?.expiryDate ?? '';
+  for (let i = 0; i < excercisedRenewalDates.length; i++) {
```
devinleighsmith, line 30: could also do with a sort? this seems fine though.

source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
FuriousLlama: Not 100%. I dont think this field is actually used in the frontend. But I will clean it up as tech debt

devinleighsmith:
@FuriousLlama here is the only area I'm seeing a mismatch so far:
image
(the latest renewal is July 31st on this lease, shows up correctly on list view and export but not here).

devinleighsmith (approved):
approved, only found one instance in application where no expiry not working as intended.

PR #4199, opened 2024-07-15
Psp 8855 fix create file from map property - property attribution display

devinleighsmith:


eddherrera (approved):

asanchezr (approved):

asanchezr:
Need to update snapshots here
image

PR #4202, opened 2024-07-16
Psp 8855 correct lease-specific property attribution when adding from the map.

devinleighsmith:


source/frontend/src/features/leases/add/AddLeaseForm.tsx
```diff
@@ -28,15 +28,16 @@ const AddLeaseForm: React.FunctionComponent<React.PropsWithChildren<IAddLeaseFor
...
+
   const { getByType } = useLookupCodeHelpers();
   const consultationTypes = getByType(API.CONSULTATION_TYPES);
-  apiFormLease.consultations = getConsultations(apiFormLease, consultationTypes);
   if (propertyInfo) {
     defaultFormLease.properties = [];
     defaultFormLease.properties.push(FormLeaseProperty.fromMapProperty(propertyInfo));
     defaultFormLease.regionId = propertyInfo.region ? propertyInfo.region.toString() : '';
   }
+  const apiFormLease = LeaseFormModel.toApi(defaultFormLease);
```
devinleighsmith, line 39: doing the conversion to an apiFormLease after the properties are added to the defaultFormLease now.

asanchezr (approved):

PR #4214, opened 2024-07-18
PSP-8911 : HotFix - Update

eddherrera:


asanchezr (approved):
This hotfix doesn't bring any code changes because the changes had been merged prior. But we deployed an earlier version to UAT and TEST.

PR #4224, opened 2024-07-24
Truncate NVARCHAR(max) Columns Prior to Alter Up

dfilteau:


asanchezr (approved):
Looks good to me but need Manuel to review in case I missed something

FuriousLlama (approved):

PR #4227, opened 2024-07-24
PSP-8746 | GST is modifiable for periods

FuriousLlama:


source/frontend/src/features/leases/detail/LeasePages/payment/modal/period/PeriodForm.tsx
```diff
@@ -64,6 +69,27 @@ export const PeriodForm: React.FunctionComponent<React.PropsWithChildren<IPeriod
...
+  const onGstCheckChange = (
+    formikState: FormikProps<FormLeasePeriod>,
+    field: string,
+    values: boolean,
+  ) => {
+    if (values === true) {
+      const gstDecimal = gstConstant !== undefined ? parseFloat(gstConstant.value) : undefined;
+      const calculated = round((formikState.values.paymentAmount as number) * (gstDecimal / 100));
+      formikState.setFieldValue('gstAmount', calculated);
+    } else {
```
asanchezr, line 83: There is a code path here where **gstDecimal** is undefined (when the GST constant is not defined).
I'm not sure what would be the result of line 81 would be in this case; ie: `calculated = round(... * (undefined/100))`

Would it make sense to put a check here for that case?

source/frontend/src/features/leases/detail/LeasePages/payment/modal/period/VariablePeriodSubForm.tsx
```diff
@@ -1,25 +1,55 @@
...
+  const formikProps = useFormikContext<FormLeasePeriod>();
   const lookups = useLookupCodeHelpers();
   const paymentFrequencyOptions = lookups.getOptionsByType(API.LEASE_PAYMENT_FREQUENCY_TYPES);
+
+  const onGstCheckChange = (field: string, values: boolean) => {
+    if (values === true) {
+      const gstDecimal = gstConstant !== undefined ? parseFloat(gstConstant.value) : undefined;
+      const calculated = round(
+        (getIn(formikProps.values, paymentAmountField) as number) * (gstDecimal / 100),
+      );
```
asanchezr, line 40: same comment here re: gstDecimal being undefined

asanchezr (approved):

source/frontend/src/features/leases/detail/LeasePages/payment/modal/period/PeriodForm.tsx
FuriousLlama: fair point. I added a better default value

source/frontend/src/features/leases/detail/LeasePages/payment/modal/period/VariablePeriodSubForm.tsx
FuriousLlama: same as above

PR #4233, opened 2024-07-31
PSP-7158 Allow moving the position of map markers (within the file) 

asanchezr:
https://github.com/user-attachments/assets/b93d0ae3-7a16-4e9c-9102-a86995171d2e


source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -255,6 +264,26 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
     serviceSend({ type: 'FINISH_SELECTION' });
   }, [serviceSend]);
 
+  const startReposition = useCallback(
+    (
+      repositioningFeatureDataset: LocationFeatureDataset | null,
+      index: number | null,
```
FuriousLlama, line 270: Can these be null if they are necessary for the state?

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -40,6 +50,25 @@ const featureViewStates = {
...
       },
     },
+    repositioning: {
+      on: {
+        FINISH_REPOSITION: {
+          target: 'browsing',
+          actions: [
+            assign({
+              repositioningFeatureDataset: () => null,
+              repositioningPropertyIndex: () => null,
```
FuriousLlama, line 60: does selectingComponentId need to be set to null as well?

source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -206,7 +219,7 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
             ))}
           </LayerGroup>
         )}
-        {mapMachine.showPopup && (
+        {mapMachine.showPopup && !mapMachine.isRepositioning && (
```
FuriousLlama, line 222: this is kinda odd. The mapMachine should be the sole authority in deciding if the popup should be shown. Otherwise, we risk having the logic distributed. Would it be possible to have the state so both showPopup and isRepositioning can never happen together?

source/frontend/src/components/propertySelector/MapClickMonitor.tsx
```diff
@@ -7,14 +9,22 @@ import { featuresetToMapProperty } from '@/utils/mapPropertyUtils';
 import { LocationFeatureDataset } from '../common/mapFSM/useLocationFeatureLoader';
 
 interface IMapClickMonitorProps {
-  addProperty: (property: LocationFeatureDataset) => void; // TODO: This should be a featureDataset
+  addProperty: (property: LocationFeatureDataset) => void;
```
FuriousLlama, line 12: looks like the TODO was not addressed. Does it not need to be a featureDataset anymore?

asanchezr: good catch - will remove the comment

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
asanchezr: good catch - will change

source/frontend/src/components/propertySelector/MapClickMonitor.tsx
asanchezr: I misunderstood your comment. According to git history the TODO was there when the property was an **IMapProperty**. Then Devin refactored that callback to correctly use the LocationFeatureSet (which is what the comments is asking for). Then I noticed that the TODO didn't apply anymore and removed it.

source/frontend/src/components/maps/MapLeafletView.tsx
asanchezr: Updated

FuriousLlama (approved):

PR #4255, opened 2024-08-09
psp-7878 | Improvements to zoom handling

FuriousLlama:


source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -279,9 +250,9 @@ const selectedFeatureLoaderStates = {
 };
 
 const sideBarStates = {
-  initial: 'fullScreen',
+  initial: 'close',
```
devinleighsmith, line 253: nit: given that these are states and not actions I'd rather these were 'closed' and 'opened'

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -372,20 +343,50 @@ const sideBarStates = {
...
-      entry: assign({ mapFilter: () => defaultPropertyFilter }),
+    filtering: {
+      entry: [send({ type: 'REFRESH_PROPERTIES', searchCriteria: defaultPropertyFilter })],
+      exit: [send({ type: 'REFRESH_PROPERTIES' })],
       on: {
-        CLOSE_ADVANCED_FILTER_SIDEBAR: {
-          target: 'fullScreen',
+        TOGGLE_FILTER: {
+          target: 'close',
+          actions: [assign({ showDisposed: () => false }), assign({ showRetired: () => false })],
```
devinleighsmith, line 377: I'm on the fence if it makes sense for the state machine to enforce defaults like this, or if it should be up to the component itself.

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -372,20 +343,50 @@ const sideBarStates = {
...
+        },
+        TOGGLE_LAYERS: {
+          target: 'layerControl',
+        },
+        SET_VISIBLE_PROPERTIES: {
+          actions: assign({ activePimsPropertyIds: (_, event: any) => event.propertyIds }),
+        },
+        SET_SHOW_DISPOSED: {
```
devinleighsmith, line 385: Having this as an action still feels fishy to me - given that there are multiple other property states that have selectors, but we only have explicit actions for these two.

devinleighsmith (approved):
Only issue I could find is that if you search for a pid, and then open the advanced filter, and then close the advanced filter, it searches for the pid from the first step again, but there is no pid in the search bar.

would consider a minor issue, but still confusing. I think I would expect either:
1) the pid is populated when the search occurs.
2) the search criteria is cleared when I close the advanced search.

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: I agree, however, this logic was in place before the refactor. I rather leave it like that and figure out later a better approach

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: Also, agree. But this was like that before, rather keep the logic as is for existing filter parameters

source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
FuriousLlama: yeah, I struggled between the names. I will set it to closed'

FuriousLlama:
> Only issue I could find is that if you search for a pid, and then open the advanced filter, and then close the advanced filter, it searches for the pid from the first step again, but there is no pid in the search bar.
> 
> would consider a minor issue, but still confusing. I think I would expect either:
> 
> 1. the pid is populated when the search occurs.
> 2. the search criteria is cleared when I close the advanced search.

Fair, I will check with the BAs and fix it on a separate pr

PR #4280, opened 2024-08-23
PSP-8977 : Add Total Allowable Compensation to Compensation Section

eddherrera:
image


devinleighsmith (changes requested):
This requirement does not appear to be met:
User cannot save a "total allowable compensation" for a FILE, if the value is less than the total of all existing final state compensations. Instead, the user gets a warning.

devinleighsmith:
Actually this one doesn't seem to be met either:

User cannot save a "total compensation" for a FINAL H120 COMPENSATION REQUISITION greater than "total allowable compensation" (i.e. sum of existing final compensation must not exceed file's total allowable compensation)
Warning: "Your compensation requisition cannot be saved in FINAL status, as the compensation amount exceeds total allowable compensation for this file".

devinleighsmith:
Specifically, updating a current compensation requisition to a value greater than the total allowable is what is not working.

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -266,7 +266,10 @@ private PimsCompensationRequisition UpdateLeaseFileCompensation(PimsCompensation
 
             var currentCompensation = _compensationRequisitionRepository.GetById(compensationRequisition.CompensationRequisitionId);
 
+            var currentLeaseFile = _leaseRepository.Get((long)currentCompensation.LeaseId);
+            CheckTotalAllowableCompensation(currentLeaseFile, compensationRequisition);
```
devinleighsmith, line 270: This should probably be on the update logic, as well as the create.

source/backend/dal/Repositories/CompReqFinancialRepository.cs
```diff
@@ -42,9 +42,26 @@ public IEnumerable<PimsCompReqFinancial> GetAllByAcquisitionFileId(long acquisit
             {
                 query = query.Where(c => c.CompensationRequisition.IsDraft == false);
             }
+
             return query.AsNoTracking().ToArray();
         }
 
+        public IEnumerable<PimsCompReqFinancial> GetAllByLeaseFileId(long leaseFileId, bool? finalOnly)
+        {
+            this._user.ThrowIfNotAllAuthorized(Security.Permissions.CompensationRequisitionView);
```
devinleighsmith, line 51: nit: can you just use ThrowIfNotAuthorized here (there is only one permission).

source/frontend/src/features/mapSideBar/compensation/list/CompensationListContainer.tsx
```diff
@@ -72,6 +77,15 @@ export const CompensationListContainer: React.FC<ICompensationListContainerProps
 
               updatedFileResponse = await putAcquisitionFile(updatedAcquisitionFile, []);
             }
+            break;
+          case ApiGen_CodeTypes_FileTypes.Lease:
+            {
+              const updatedLease = file as ApiGen_Concepts_Lease;
+              file.totalAllowableCompensation = totalAllowableCompensation;
```
devinleighsmith, line 84: I think this works because updatedLease and file point to the same reference, but probably would be clearer to say updatedLease.totalAllowableCompensation

devinleighsmith (approved):

FuriousLlama (approved):

source/backend/dal/Repositories/CompReqFinancialRepository.cs
eddherrera: Updated

source/frontend/src/features/mapSideBar/compensation/list/CompensationListContainer.tsx
eddherrera: updated

PR #4282, opened 2024-08-25
psp-8955 add file streaming back to PIMS.

devinleighsmith:


source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -170,26 +172,14 @@ public async Task<ExternalResponse<QueryResponse<DocumentMetadataModel>>> TryGet
...
 
             try
             {
-                Uri endpoint = new($"{this._config.BaseUri}/documents/{documentId}/files/{fileId}/download/");
-                HttpResponseMessage httpResponse = await client.GetAsync(endpoint).ConfigureAwait(true);
-
-                var response = await ProcessDownloadResponse(httpResponse);
-
-                return response;
+                return await ProcessDownloadResponse(await GetFileAsync(documentId, fileId));
```
FuriousLlama, line 182: nit: for debugging purposes, it would be better to have the result on a different variable and use that on the function.
```
var response = await GetFileAsync(documentId, fileId);
return await ProcessDownloadResponse(response);

File: source/backend/api/Repositories/RestCommon/BaseRestRepository.cs [hunk 2]
```diff
@@ -233,6 +234,53 @@ protected async Task<ExternalResponse<FileDownloadResponse>> ProcessDownloadResp
...
+        {
+            ExternalResponse<FileStreamResponse> result = new ExternalResponse<FileStreamResponse>()
+            {
+                Status = ExternalResponseStatus.Error,
+            };
+
+            Stream responsePayload = await response.Content.ReadAsStreamAsync().ConfigureAwait(true);
+            _logger.LogTrace("Response: {response}", response);
+            response.Content.Headers.TryGetValues("Content-Length", out IEnumerable<string> contentLengthHeaders);
+            long contentLength = contentLengthHeaders?.FirstOrDefault() != null ? int.Parse(contentLengthHeaders.FirstOrDefault(), CultureInfo.InvariantCulture) : responsePayload.Length;
```
FuriousLlama, line 247: nit: using an if would be more readable given its length

source/frontend/src/features/documents/hooks/useDocumentProvider.ts
```diff
@@ -208,6 +211,29 @@ export const useDocumentProvider = () => {
...
+    (
+      mayanDocumentId: number,
+      mayanFileId: number,
+    ) => Promise<AxiosResponse<AxiosResponse<File, any>>>
+  >({
+    requestFunction: useCallback(
+      async (mayanDocumentId: number, mayanFileId: number) =>
+        await streamDocumentFileApiCall(mayanDocumentId, mayanFileId),
+      [streamDocumentFileApiCall],
+    ) as unknown as () => Promise<AxiosResponse<AxiosResponse<File, any>>>,
```
FuriousLlama, line 225: why is the unknown necessary?

source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -7,10 +7,12 @@
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using System.Threading.Tasks;
+using Azure;
```
asanchezr, line 10: not sure this import is needed

source/frontend/src/features/documents/hooks/useDocumentProvider.ts
devinleighsmith: Yeah, this is a headache, but currently I haven't found a good way to tell Typescript that when the rawResponse flag is set, the response returns the whole axios response not just the data.

FuriousLlama (approved):

source/frontend/src/features/documents/hooks/useDocumentProvider.ts
FuriousLlama: ah, yes. I saw this before, we need to make our response type more broad or add another result type on the wrapper.

PR #4286, opened 2024-08-26
psp-9080 add missing LEASE_STAKEHOLDER_TYPE codes.

devinleighsmith:


source/frontend/src/components/common/form/Select.tsx
```diff
@@ -100,10 +100,12 @@ export const Select: React.FC<React.PropsWithChildren<SelectProps>> = ({
   };
 
   const renderPlaceholder = () => {
-    if (!placeholder) {
+    const calculatedPlaceholder =
+      !value || options.find(option => option.value === value) ? null : 'N/A'; // Render N/A in the event that the currently selected value is not in the list.
```
devinleighsmith, line 104: Note: this is required, otherwise if a value is currently selected that is not in the options list, the select will default to showing the first value in the options list, even if that is not what the select's current value is. To correct this, the user would have to select the value which currently appears selected, which is not intuitive.

source/database/mssql/scripts/dbscripts/PSP_PIMS_S87_00/Alter Up/154_DML_PIMS_LEASE_STAKEHOLDER_TYPE_Alter_Up.sql
```diff
@@ -0,0 +1,53 @@
+IF NOT EXISTS (
```
devinleighsmith, line 1: It looks like Doug did a migration of the old values in the alter up, but forgot to add these new values. They are present in the BUILD.

eddherrera (approved):

asanchezr (approved):

PR #4288, opened 2024-08-26
psp-9041 fix issues with PIN-only properties.

devinleighsmith:


source/frontend/src/components/propertySelector/MapSelectorContainer.tsx
```diff
@@ -66,7 +66,7 @@ export const MapSelectorContainer: FunctionComponent<IMapSelectorContainerProps>
           return property;
         }
         const queryObject = {};
-        if (pid.length > 0) {
+        if (isValidId(+pid)) {
```
devinleighsmith, line 69: pid may be undefined.

asanchezr (approved):

PR #4292, opened 2024-08-27
PSP-8932 | Added checks to create pages so navigating with data shows a warning

FuriousLlama:


devinleighsmith:
I'll try this one out in the morning, sometimes these forms have weird interactions with save/cancel logic, so I'll check it out and play with it.

devinleighsmith (changes requested):
clicking cancel on the add lease form with changes requires to clicks instead of one to cancel now (probably two modals).

devinleighsmith:
same for acquisition and research.

devinleighsmith:
For subdivision and consolidation, if I add a property, click cancel, and click yes, no navigation happens.

devinleighsmith:
Project seems to be working as expected however.

devinleighsmith (commented):
I don't have time to re-review, but I'll clear the requested changes.

devinleighsmith (approved):
I don't have time to re-review, but I'll clear the requested changes.

PR #4297, opened 2024-08-28
Psp 9058 - gst calculation fixes.

devinleighsmith:


source/frontend/src/features/leases/detail/LeasePages/payment/modal/period/PeriodForm.test.tsx
```diff
@@ -83,6 +84,32 @@ describe('PeriodForm component', () => {
...
+    });
+
+    const variableButton = getByTestId('radio-isvariable-variable');
+    await act(async () => {
+      userEvent.click(variableButton);
+    });
+
+    expect(getByDisplayValue('$100.00')).toBeVisible();
+    expect(getByDisplayValue('$200.00')).toBeVisible();
+    expect(getByDisplayValue('$300.00')).toBeVisible();
```
eddherrera, line 110: Not quite fond of this pattern since it could lead to false positives when any other element displays the same value. Preferably find the element in cuestion and use toHaveValue to assert results explicitly.

eddherrera (approved):

FuriousLlama (approved):

PR #4303, opened 2024-09-03
PSP-9002 : Generate H120 Lease/Licence

eddherrera:
image


source/frontend/src/models/generate/acquisition/GenerateCompensationPayee.ts
```diff
@@ -15,46 +17,54 @@ export class Api_GenerateCompensationPayee {
...
+        this.name = compensation?.acquisitionFileTeam.organization?.name ?? '';
       }
     } else if (compensation?.legacyPayee) {
-      this.name = compensation.legacyPayee ?? '';
+      this.name = compensation?.legacyPayee ?? '';
+    } else if (compensation?.compReqLeaseStakeholder?.length > 0) {
+      const stakeHolder: ApiGen_Concepts_LeaseStakeholder =
+        compensation?.compReqLeaseStakeholder[0].leaseStakeholder;
+      if (stakeHolder.lessorType.id === ApiGen_CodeTypes_LessorTypes.ORG) {
+        this.name = stakeHolder.organization.name ?? '';
```
asanchezr, line 51: Just to be safe - I would apply a nullcheck here for organization like the rest of the code in this file:

```ts
this.name = stakeholder.organization?.name ?? '';
```

source/frontend/src/models/generate/CompensationRequisition/GenerateCompReqFileLease.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ApiGen_Concepts_LeaseStakeholder } from '@/models/api/generated/ApiGen_Concepts_LeaseStakeholder';
+import { ApiGen_Concepts_PropertyLease } from '@/models/api/generated/ApiGen_Concepts_PropertyLease';
+
+import { Api_GenerateH120Property } from '../acquisition/GenerateH120Property';
+import { Api_GenerateOwner } from '../GenerateOwner';
+import { Api_GenerateProduct } from '../GenerateProduct';
+import { Api_GenerateProject } from '../GenerateProject';
+import { ICompensationRequisitionFile } from './ICompensationRequisitionFile';
+
+export class GenerateCompReqFileLease implements ICompensationRequisitionFile {
```
asanchezr, line 13: Please rename this to "Api_Generate..." to be consistent with naming convention of generate classes

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
```diff
@@ -182,5 +182,13 @@ public List<PimsPropertyLease> GetLeaseCompReqPropertiesById(long compensationRe
...
                 .ToList();
         }
+
+        public IEnumerable<PimsCompReqFinancial> GetCompensationRequisitionFinancials(long id)
+        {
+            return Context.PimsCompReqFinancials
+                .Include(y => y.FinancialActivityCode)
+                .Where(x => x.CompensationRequisitionId == id)
+                .ToList();
+        }
```
asanchezr, line 192: I think this method is only ever used to "read" data from the DB so it would make sense to add **.AsNoTracking()** - thoughts?

source/backend/api/Services/CompensationRequisitionService.cs
```diff
@@ -146,6 +146,14 @@ public IEnumerable<PimsCompensationRequisition> GetFileCompensationRequisitions(
             return compReqs;
         }
 
+        public IEnumerable<PimsCompReqFinancial> GetCompensationRequisitionFinancials(long id)
+        {
+            _logger.LogInformation("Getting compensations financials for id: {acquisitionFileId}", id);
+            _user.ThrowIfNotAuthorized(Permissions.CompensationRequisitionView, Permissions.CompensationRequisitionView);
```
asanchezr, line 152: not sure why we would have the same permission twice here - please remove duplicate

source/backend/api/Areas/CompensationRequisition/Controllers/CompensationRequisitionController.cs
```diff
@@ -204,5 +204,26 @@ public IActionResult GetFileCompensations([FromRoute]FileTypes fileType, [FromRo
...
+                "Request received by Controller: {Controller}, Action: {ControllerAction}, User: {User}, DateTime: {DateTime}",
+                nameof(CompensationRequisitionController),
+                nameof(GetCompensationRequisitionFinancials),
+                User.GetUsername(),
+                DateTime.Now);
+            _logger.LogInformation("Dispatching to service: {Service}", _compensationRequisitionService.GetType());
+
+            var compReqFinancials = _compensationRequisitionService.GetCompensationRequisitionFinancials(id);
+
+            return new JsonResult(_mapper.Map<IEnumerable<CompensationRequisitionModel>>(compReqFinancials));
```
asanchezr, line 226: could you please check that this is the correct mapping?
I would think that **PimsCompReqFinancial** which is returned by "GetCompensationRequisitionFinancials()" should be matched to the model **CompensationFinancialModel**

Instead I see here that it is being mapped to **CompensationRequisitionModel** - does this make sense here?

asanchezr: Also, I was expecting to see the ACQ controller method that returned this info before this PR to be removed since it would be replaced by this new endpoint - right? But I do not see a change here that removes that endpoint from the ACQ controller. 

I'm referring to this: 
image

asanchezr (approved):
Approved but see comments

source/frontend/src/models/generate/acquisition/GenerateCompensationPayee.ts
eddherrera: Updated

source/backend/dal/Repositories/CompensationRequisitionRepository.cs
eddherrera: agree

source/frontend/src/models/generate/CompensationRequisition/GenerateCompReqFileLease.ts
eddherrera: Updated

source/backend/api/Areas/CompensationRequisition/Controllers/CompensationRequisitionController.cs
eddherrera: removed and updated.

PR #4306, opened 2024-09-06
PSP-9164 : FT:Reg:Lease&Licence File: In the update mode if user deselects the project that has been already selected and try to save the form, system is still displaying associated selected product in the view/edit screen

eddherrera:


source/frontend/src/features/leases/add/LeaseDetailSubForm.tsx
```diff
@@ -119,7 +119,15 @@ export const LeaseDetailSubForm: React.FunctionComponent<ILeaseDetailsSubFormPro
   return (
     <Section header="Original Agreement">
       <SectionField label="Ministry project" labelWidth="3">
-        <ProjectSelector field="project" />
+        <ProjectSelector
+          field="project"
+          onChange={(vals: IAutocompletePrediction[]) => {
+            onMinistryProjectSelected(vals);
```
devinleighsmith, line 125: This should make the entire useeffect on line 113 redundant should it not?

devinleighsmith (changes requested):
see comment

source/frontend/src/features/leases/add/LeaseDetailSubForm.tsx
eddherrera: let me check, I was following the same patter as Acquisition..

eddherrera: removed useEffect and re-tested.

asanchezr: hmm - I think there is a subtle difference here that will still need the useEffect. I believe that the **onChange** only triggers when you actually change the value of the dropdown.

Without the useEffect what happens when we load the update form for the first time and there is a project selected in the data that comes from the API? Without the useEffect the products dropdown wouldn't get populated until you change the parent project to another value. 

I'm speaking from reading the code. Could you check this @eddherrera? If that is confirmed we will need to restore the useEffect back

devinleighsmith (approved):

source/frontend/src/features/leases/add/LeaseDetailSubForm.tsx
eddherrera: @asanchezr  is right. I restored the last change and added a new unit test to account for it.

PR #4309, opened 2024-09-09
PSP-8684 | Updated Lease list report

FuriousLlama:


source/backend/api/Areas/Reports/Models/Lease/LeaseModel.cs
```diff
@@ -67,27 +99,15 @@ public class LeaseModel
 
         [DisplayName("Lease Purpose Type")]
```
eddherrera, line 100: Header would probably need an update as well. Maybe as a suggestion: "Lease Purpose Type(s)"

source/backend/tests/unit/api/Controllers/Reports/LeaseControllerTest.cs
```diff
@@ -1,11 +1,13 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using FluentAssertions;
 using MapsterMapper;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
```
eddherrera, line 10: nit: probably not needed in test

eddherrera (approved):

PR #4311, opened 2024-09-09
Update payee options model to correctly handle null/undefined.

devinleighsmith:


asanchezr (approved):

source/frontend/src/features/mapSideBar/acquisition/models/PayeeOptionModel.ts
```diff
@@ -257,7 +257,9 @@ export class PayeeOption {
 
     switch (model.lessorType.id) {
       case ApiGen_CodeTypes_LessorTypes.ORG:
-        payeeName = `${model.organization.name}, Inc. No. ${model.organization.incorporationNumber}`;
+        payeeName = `${model?.organization?.name ?? ''}, Inc. No. ${
+          model?.organization?.incorporationNumber ?? ''
+        }`;
```
FuriousLlama, line 262: nit having a formater similar to the one below would be quite helpful for orgs

FuriousLlama (approved):

PR #4317, opened 2024-09-10
PSP-8713: UI UX Clean Up - First Page Banner + Warning Information + Images

stairaku:


source/frontend/src/features/account/Login.test.tsx
```diff
@@ -144,7 +144,7 @@ describe('login', () => {
       'MOTI Property Information Management System (PIMS)',
     );
     expect(getAllByRole('heading')[1]).toHaveTextContent(
-      'PIMS enables you to view highways and properties owned by the Ministry of Transportation and Infrastructure',
+      'PIMS enables users to track and manage information relating to the property interests of the MOTI and BCTFA.',
```
asanchezr, line 147: Please fix this unit test - it is failing because the new login page is displaying this text in a paragraph (<p>) instead of a heading (<h6>). Using **getByText(...)** should fix it

asanchezr (approved):

source/frontend/src/features/account/Login.test.tsx
stairaku: Updated unit test as suggested.

PR #4321, opened 2024-09-10
psp-9173 correct issues with exports.

devinleighsmith:


eddherrera (approved):

source/backend/api/Areas/Reports/Mapping/Lease/LeaseMap.cs
```diff
@@ -25,7 +25,7 @@ private static void MapLease((Entity.PimsLeasePeriod period, Entity.PimsLease le
             var historicalString = src.property?.Property?.GetHistoricalNumbersAsString();
             var currentRenewal = src.lease.PimsLeaseRenewals.FirstOrDefault(renewal => renewal != null && renewal.IsExercised == true && DateTime.Now > renewal.CommencementDt && DateTime.Now <= renewal.ExpiryDt);
 
-            var additionalRenewalsCount = src.lease.PimsLeaseRenewals.FirstOrDefault(r => r.LeaseRenewalId == currentRenewal.LeaseRenewalId) == null ? src.lease.PimsLeaseRenewals.Count : src.lease.PimsLeaseRenewals.Count - 1;
+            var additionalRenewalsCount = src.lease.PimsLeaseRenewals.FirstOrDefault(r => r.LeaseRenewalId == currentRenewal?.LeaseRenewalId) == null ? src.lease.PimsLeaseRenewals.Count : src.lease.PimsLeaseRenewals.Count - 1;
```
devinleighsmith, line 28: currentRenewal is populated by FirstOrDefault on line 26, meaning it can be null, so dereferencing without a null check will cause this to fail.

source/backend/api/Services/CompReqFinancialService.cs
```diff
@@ -61,11 +61,11 @@ public IEnumerable<PimsCompReqFinancial> SearchCompensationRequisitionFinancials
             // When a contractor exports data, they should only see data for files they have been assigned to.
             if (pimsUser.IsContractor)
             {
-                return allMatchingFinancials.Where(f => f.CompensationRequisition.AcquisitionFile.PimsAcquisitionFileTeams.Any(afp => afp.PersonId == pimsUser.PersonId));
+                return allMatchingFinancials.Where(f => f.CompensationRequisition?.AcquisitionFile?.PimsAcquisitionFileTeams.Any(afp => afp.PersonId == pimsUser.PersonId) == true);
```
devinleighsmith, line 64: we now have compensations that have no acquisition, so dereferencing like this breaks.

devinleighsmith: logged https://jira.th.gov.bc.ca/browse/PSP-9180 to ask how contractors should be able to access lease export data.

devinleighsmith: (since there is no lease team corresponding to the acquisition team).

PR #4334, opened 2024-09-13
Psp 6333 - run keycloak sync in openshift

devinleighsmith:


.github/workflows/ci-cd-pims-dev.yml
```diff
@@ -176,6 +150,25 @@ jobs:
...
+    steps:
+      - name: Checkout Source Code
+        uses: actions/checkout@v4
+      - name: Login to OpenShift
+        uses: redhat-actions/oc-login@v1
+        with:
+          openshift_server_url: ${{ env.OPENSHIFT_SERVER }}
+          openshift_token: ${{ env.OPENSHIFT_TOKEN }}
+          insecure_skip_tls_verify: true
+          namespace: $NAMESPACE_OVERRIDE
```
asanchezr, line 167: I see `$NAMESPACE_OVERRIDE` being referenced here and the other GH actions but I don't see it defined anywhere. Could you please indicate where is this value coming from?

asanchezr (approved):
See comments

PR #4338, opened 2024-09-16
PSP-8320 Update Lease button location 

asanchezr:
lease-details-generate


devinleighsmith:
The icon from the mockup doesn't match - please check with Ana if that was intentional

devinleighsmith (approved):
approved, see comment

asanchezr:
> The icon from the mockup doesn't match - please check with Ana if that was intentional

Yes, we did discuss it. She used an old icon from somewhere else and actually prefers the new icon. I told her to update the mockup

PR #4340, opened 2024-09-16
PSP-9119: UI UX Clean Up - Bomb and Toasts relocation

stairaku:
bomb

image


source/frontend/src/assets/scss/_variables.scss
```diff
@@ -2,4 +2,4 @@
 
 $header-height: 7.2rem;
 $mapfilter-height: 6rem;
-$footer-height: 3.4rem;
+$footer-height: 4.4rem;
```
devinleighsmith, line 5: I don't think this is what Ana wanted, please check with her.

source/frontend/src/components/layout/Header/Header.tsx
```diff
@@ -71,24 +49,11 @@ export const Header = () => {
         <VerticalBar />
       </div>
       {keycloak.obj.authenticated && <UserProfile />}
-      <Nav className="other">
-        {errors && errors.length > 0 ? (
-          <FaBomb size={30} className="errors" onClick={handleShow} />
-        ) : null}
-      </Nav>
-      <ErrorModal errors={errors} show={show} setShow={setShow}></ErrorModal>
+      <div className="other"></div>
```
devinleighsmith, line 52: This should probably be removed entirely if unused.

devinleighsmith (changes requested):
see comment. Also, do we need the empty header anymore?

source/frontend/src/assets/scss/_variables.scss
stairaku: I checked with her and she says that is ok to increase the footer height as long as the bomb icon is as the size as the help icon. I increased a little bit more than that since the icon looked too small. I showed to her and she was good with the changes.

image

source/frontend/src/components/layout/Header/Header.tsx
stairaku: I inserted that div to keep the same dimensions on the header. Without it, the user's name goes too much to the right. Let me know your thought, otherwise I can insert the css from "other" to the parent element?

stairaku:
> see comment. Also, do we need the empty header anymore?

I think maybe not. Will work on this one after I finish testing tasks, is that ok?

devinleighsmith (approved):

PR #4345, opened 2024-09-17
psp-9148 Add consultation outcomes and business logic.

devinleighsmith:


source/backend/entities/PimsBaseContext.cs
```diff
@@ -1698,6 +1704,38 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
 
+        modelBuilder.Entity<PimsConsultationOutcomeType>(entity =>
+        {
+            entity.HasKey(e => e.ConsultationOutcomeTypeCode).HasName("OUTCMT_PK");
+
+            entity.ToTable("PIMS_CONSULTATION_OUTCOME_TYPE", tb =>
+                {
+                    tb.HasComment("Description of the consultation outcome type for a lease or license.");
+                    tb.HasTrigger("PIMS_OUTCMT_I_S_I_TR");
+                    tb.HasTrigger("PIMS_OUTCMT_I_S_U_TR");
```
FuriousLlama, line 1715: why is the context being updated?

source/backend/entities/ef/PimsConsultationOutcomeType.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.ComponentModel.DataAnnotations.Schema;
+using Microsoft.EntityFrameworkCore;
+
+namespace Pims.Dal.Entities;
+
+/// <summary>
+/// Description of the consultation outcome type for a lease or license.
+/// </summary>
+[Table("PIMS_CONSULTATION_OUTCOME_TYPE")]
+public partial class PimsConsultationOutcomeType
```
FuriousLlama, line 13: Should this have been part of the scaffold?

source/backend/tests/unit/api/Controllers/Leases/ConsultationControllerTest.cs
```diff
@@ -0,0 +1,204 @@
...
+            // Arrange
+            var consultation = new PimsLeaseConsultation() { LeaseId = 1 };
+
+            this._service.Setup(m => m.GetConsultationById(It.IsAny<long>())).Returns(consultation);
+
+            // Act
+            var result = this._controller.GetLeaseConsultationById(1, 1);
+
+            // Assert
+            this._service.Verify(m => m.GetConsultationById(It.IsAny<long>()), Times.Once());
```
FuriousLlama, line 55: not a big fan of these tests. It leaks implementation details about what the controller does internally and does not provide a validation of anything other than proxying the call

source/backend/entities/PimsBaseContext.cs
devinleighsmith: Oh, this was branched from before the scaffold was merged, I'll rebase.

source/backend/tests/unit/api/Controllers/Leases/ConsultationControllerTest.cs
devinleighsmith: This is our pattern thus far for these tests, which we have gone through together as a group. Recommend we bring this up for review in our dev meeting, as nearly all of our controllers are tested this way.

source/frontend/src/features/mapSideBar/lease/tabs/consultations/edit/EditConsultationYupSchema.ts
```diff
@@ -1,14 +1,26 @@
 /* eslint-disable no-template-curly-in-string */
 import * as Yup from 'yup';
 
+import { ApiGen_CodeTypes_ConsultationOutcomeTypes } from './../../../../../../models/api/generated/ApiGen_CodeTypes_ConsultationOutcomeTypes';
```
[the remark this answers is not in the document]
devinleighsmith: hmm, I use a plugin to generate these, I wonder if I can configure it to use absolute paths by default.

FuriousLlama (approved):

asanchezr (approved):

PR #4348, opened 2024-09-18
psp-8335 | Pims properties boundary highlayer 

FuriousLlama:


source/frontend/src/components/maps/leaflet/Layers/useConfiguredMapLayers.tsx
```diff
@@ -19,9 +20,9 @@ export const useConfiguredMapLayers = () => {
 
     return {
       ...parent,
-      nodes: allNodes?.map((node: any, index) => ({
+      nodes: allNodes?.map((node: ILayerItem, index) => ({
         ...node,
-        zIndex: (parentIndex + 1) * index,
+        zIndex: node.zIndexAbsolute === true ? node.zIndex : (parentIndex + 1) * index,
```
FuriousLlama, line 25: I was unsure about having this as just node.zIndex, but I presume the index calculation was added to solve some previous issue. If that does not ring a bell I can simplify it and remove the zIndexAbsolute property

source/frontend/src/components/maps/leaflet/Control/LayersControl/LayersMenu.tsx
```diff
@@ -83,21 +83,54 @@ const LayerColor = styled.div<{ color: string }>`
...
+    const parentLayer = values.layers[index];
+    const activeChildren = parentLayer?.nodes?.filter(x => x?.on === true);
+    let isActive = false;
+    let isPartialActive = false;
+
+    if (activeChildren?.length > 0) {
+      isActive = true;
+      if (parentLayer?.nodes?.length !== activeChildren?.length) {
+        isPartialActive = true;
+      } else {
```
devinleighsmith, line 108: this seems redundant given that isPartialActive is set every time this useEffect runs.

source/frontend/src/components/maps/leaflet/Control/LayersControl/LayersMenu.tsx
```diff
@@ -83,21 +83,54 @@ const LayerColor = styled.div<{ color: string }>`
...
+        isPartialActive = false;
+      }
+    } else {
+      isActive = false;
```
devinleighsmith, line 112: same here, these values are already false when this is called.

source/frontend/src/components/maps/leaflet/Control/LayersControl/LayersMenu.tsx
```diff
@@ -83,21 +83,54 @@ const LayerColor = styled.div<{ color: string }>`
...
+    }
+
+    setFieldValue(`layers[${index}].on`, isActive);
+
+    if (parentChekRef.current) {
+      parentChekRef.current.indeterminate = isPartialActive;
+    }
+  }, [index, setFieldValue, values]);
+
+  const parentChekRef = useRef<HTMLInputElement>();
```
devinleighsmith, line 123: nit: move declaration above useage.

source/frontend/src/components/maps/leaflet/Layers/useConfiguredMapLayers.tsx
devinleighsmith: I'm pretty sure this logic predates PSP (ie. is from CITZ pims). I suppose this makes sense if the layers are ordered into groups by priority, which isn't really our case. However, we also don't have a defined priority for our existing layers - which sounds like a requirements gap. 

Logged: https://jira.th.gov.bc.ca/browse/PSP-9212 for the above.

devinleighsmith (approved):

source/frontend/src/components/maps/leaflet/Control/LayersControl/LayersMenu.tsx
FuriousLlama: that is true, I think it improves readability. but I can remove it.

PR #4352, opened 2024-09-19
PSP-8442 : UI UX Clean Up - arrow down - size and position

eddherrera:


source/frontend/src/components/common/form/YesNoSelect.tsx
```diff
@@ -1,4 +1,7 @@
+import './Select.scss';
+
 import cx from 'classnames';
+import classNames from 'classnames';
```
asanchezr, line 4: We already have a "**classnames**" import here - please remove duplicate import

asanchezr (approved):
See comments

source/frontend/src/components/common/form/YesNoSelect.tsx
eddherrera: I tried. this was the only way to import it.

PR #4355, opened 2024-09-20
PSP-7885 Improve interaction - with advanced search by retaining results

asanchezr:
Screenshot 2024-09-19 211316

----

https://github.com/user-attachments/assets/b2bfe2b3-1ccf-4ca1-ab78-d2aa511788d2

### DIFF
--- a/source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
+++ b/source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
@@ -346,7 +346,7 @@ const advancedFilterSideBarStates = {
     closed: {
       on: {
         TOGGLE_FILTER: {
-          target: 'filtering',
+          target: 'mapFilterOpened',
         },
         TOGGLE_LAYERS: {
           target: 'layerControl',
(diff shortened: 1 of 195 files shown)


source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -366,13 +366,14 @@ const advancedFilterSideBarStates = {
         },
       },
     },
-    filtering: {
-      entry: [send({ type: 'REFRESH_PROPERTIES', searchCriteria: defaultPropertyFilter })],
-      exit: [send({ type: 'REFRESH_PROPERTIES' })],
+    mapFilterOpened: {
+      entry: [
+        send({ type: 'REFRESH_PROPERTIES', searchCriteria: defaultPropertyFilter }),
```
devinleighsmith, line 371: Hmm, not sure if this makes sense anymore. Right now, every time I open the sidebar the property refresh event fires, even if I don't change anything within the advanced filter. Given that I haven't changed the filter, I'm not sure why a refresh would be necessary.

devinleighsmith (changes requested):
see comment on refresh. Previous pattern of refreshing properties whenever map filter opened no longer seems appropriate.

asanchezr:
> see comment on refresh. Previous pattern of refreshing properties whenever map filter opened no longer seems appropriate.

Updated. Please re-review

devinleighsmith (approved):

PR #4356, opened 2024-09-20
PSP-9152 : Expand documents tab to improve readability

eddherrera:
image


asanchezr (approved):
Approved but would have loved to see unit tests added that assert that **mapMachine.setFullWidthSidebar** has been called when switching to the relevant tabs

PR #4366, opened 2024-09-28
PSP-9214 - L/L Update payment details export to include payment notes

asanchezr:
lease-payment-notes-export


eddherrera (approved):

source/frontend/src/features/leases/reports/aggregated/ExportAggregatedLeasesContainer.test.tsx
```diff
@@ -60,22 +59,18 @@ describe('PaymentsContainer component', () => {
   it('makes a get request for an aggregated report', async () => {
     const {
       component: { getByTitle },
-    } = await setup();
+    } = setup();
```
eddherrera, line 62: Not quite sure what the value on removing the async is but result it's the same.

asanchezr: We usually use **renderAsync** when the component is a container with several **useEffect()** that we want to wait to finish before we test the component. This is not the case here so I simplified the tests by removing non-needed async/await

PR #4383, opened 2024-10-08
PSP-7786 Parcel Boundary - SRW delineation

asanchezr:
This is a story slated for next sprint, but based on the results of my spike and conversations with BAs - we felt it was useful to put it in front of users this sprint. 

TL; DR - upon my research I found that parcel map already provides a way to display SRW information via a _**CQL filter**_. Since we don't want to change the existing parcel map boundaries, I added another entry in the layer list that displays the filtered data for SRW on top of the regular parcel information.

I was also able to reverse-engineer the matching style (dotted red line) that the users were requesting for SRWs from the existing WMS service. All this have been demoed to BAs and they agree to put this forward.

image


devinleighsmith (approved):

devinleighsmith:
looks good to me, nice work!

PR #4388, opened 2024-10-10
Psp 9316 - hotfix for prod document issues introduced by streaming.

devinleighsmith:


source/frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -95,9 +95,8 @@ export const DocumentDetailContainer: React.FunctionComponent<
           const mayanMetadataResult = metadataResponse.payload?.results;
 
           let mayanFileId: number | undefined = undefined;
-          if (exists(mayanMetadataResult) && mayanMetadataResult.length > 0) {
-            const document = mayanMetadataResult[0].document;
-            mayanFileId = document?.file_latest?.id;
+          if (exists(detailResponse)) {
```
devinleighsmith, line 98: This meant that if a document had not metadata, the mayanFileId would never be fetched, and the file_latest api would be used instead. This is likely why this bypassed testing.

source/frontend/src/features/documents/DocumentPreviewContainer.tsx
```diff
@@ -166,9 +166,7 @@ export const DocumentPreviewContainer: FC<
         pages={documentPages}
         resetDocumentPreview={resetDocumentPreview}
         handleDownload={() => {
-          downloadWrappedDocumentFileLatest(previewDocumentId).then(file =>
-            createFileDownload(file),
-          );
+          streamDocumentFileLatest(previewDocumentId).then(response => showFile(response));
```
devinleighsmith, line 169: the preview should use the file streamer as well - or large file downloads will fail.

source/frontend/src/hooks/pims-api/useApiDocuments.ts
```diff
@@ -77,10 +77,14 @@ export const useApiDocuments = () => {
         ),
 
       streamDocumentFileApiCall: (mayanDocumentId: number, mayanFileId: number) =>
-        api.get<File>(`/documents/storage/${mayanDocumentId}/stream/${mayanFileId}`),
+        api.get<File>(`/documents/storage/${mayanDocumentId}/files/${mayanFileId}/stream`, {
```
devinleighsmith, line 80: as noted by eduardo, this url was incorrect.

source/frontend/src/hooks/pims-api/useApiDocuments.ts
```diff
@@ -77,10 +77,14 @@ export const useApiDocuments = () => {
...
+          responseType: 'blob',
```
devinleighsmith, line 81: Streaming text documents and some other formats works fine without responseType: 'blob'. however, pdfs and word documents do not.

asanchezr (approved):

PR #4403, opened 2024-10-16
PSP-9328 : Update document type

eddherrera:
image


source/frontend/src/features/documents/ComposedDocument.ts
```diff
@@ -235,6 +237,7 @@ export class DocumentUpdateFormData {
     return {
       documentId: this.documentId,
       mayanDocumentId: this.mayanDocumentId,
+      documentTypeId: stringToNumber(this.documentTypeId),
```
FuriousLlama, line 240: if empty i think this will set the document typeid to zero. If that happens, will the update be successful on the backend?

source/frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -78,6 +85,85 @@ export const DocumentDetailContainer: React.FunctionComponent<
     }
   };
 
+  const onUpdateDocument = async (updateRequest: ApiGen_Requests_DocumentUpdateRequest) => {
+    if (props.pimsDocument.id) {
+      const result = await updateDocument(props.pimsDocument.id, updateRequest);
+      result && props.onUpdateSuccess();
```
FuriousLlama, line 91: safer to use if(exists(result))

FuriousLlama: or checking by the contents of the result as well. result.data.id

source/frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
```diff
@@ -140,37 +231,43 @@ export const DocumentDetailContainer: React.FunctionComponent<
...
-      const result = await updateDocument(props.pimsDocument.id, updateRequest);
-      result && props.onUpdateSuccess();
+  useEffect(() => {
+    if (isEditable) {
+      fetchDocumentTypes();
     }
-  };
+  }, [fetchDocumentTypes, isEditable]);
 
-  if (isEditable === false) {
```
FuriousLlama, line 152: nit. Looks like this switched the internal component. Was that for readability? not an issue, but it makes the git diff harder to assess

FuriousLlama (approved):

source/frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
eddherrera: readability.  since 'isEditable' is already a boolean.

source/frontend/src/features/documents/ComposedDocument.ts
eddherrera: true, it does fail. I've added validations to mark as required.

source/frontend/src/features/documents/documentDetail/DocumentDetailContainer.tsx
eddherrera: updated

PR #4422, opened 2024-10-29
PSP-9089 Warn user if they are about to lose data when cancelling "Add Document"

asanchezr:
https://github.com/user-attachments/assets/41444fd1-e8b2-4375-b36a-b992e2c6608b


devinleighsmith (approved):
approved, but see comment.

PR #4427, opened 2024-10-31
split off geoserver functionality into a microservice - update docker…

devinleighsmith:
… and ci/cd to support.


devinleighsmith:
Please don't review until after 5.6 functionality is complete

.github/workflows/ci-cd-pims-dev.yml
```diff
@@ -26,13 +26,10 @@ env:
   TAG_PROD: "prod"
 
 on:
-  pull_request_target:
-    branches: [dev]
-    types: [closed]
```
asanchezr, line 31: I like this change but it will lead to a change in our process. From now on, developers will be responsible to trigger the CI build for DEV environment themselves, right? 
We should talk about how would that look like within the team

docker-compose.yml
```diff
@@ -67,6 +69,29 @@ services:
...
+      context: source/backend
+      dockerfile: Dockerfile.proxy
+      args:
+        BUILD_CONFIGURATION: Debug
+    env_file: source/backend/proxy/.env
+    environment:
+      - ASPNETCORE_ENVIRONMENT=Docker
+    ports:
+      - ${API_HTTPS_PORT:-5003}:443
+      - ${API_HTTP_PORT:-5002}:8080
```
asanchezr, line 89: I know we currently don't use docker for the API container but these env vars should be different than the one for the API (ie PROXY_HTTPS_PORT) or else we risk creating containers that open the same port on the host which will fail.

This also applies to new microservices that are added to this file (eg: for document service it could be DOC_HTTPS_PORT, etc)

source/backend/.s2i/environment
```diff
@@ -1 +0,0 @@
-DOTNET_STARTUP_PROJECT=api/Pims.Api.csproj
```
asanchezr, line 1: I think this file is used as part of openshift building .NET
Did you test this change in openshift?

source/backend/Dockerfile.proxy
```diff
@@ -0,0 +1,32 @@
...
+FROM build AS publish
+RUN dotnet publish "Pims.Proxy.csproj" -c "$BUILD_CONFIGURATION" -o /app/publish
+
+# Runtime image
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app/publish .
+COPY entrypoint.sh .
+RUN chmod +x /app/entrypoint.proxy.sh
+ENTRYPOINT ["/app/entrypoint.proxy.sh"]
```
asanchezr, line 32: Can you check that we are copying the right file?

```Dockerfile
COPY entrypoint.proxy.sh . 
# instead of entrypoint.sh which does not exist on this folder?

```

source/backend/api/Helpers/ConfigurationExtensions.cs
```diff
@@ -23,23 +23,5 @@ public static PimsOptions GeneratePimsOptions(this IConfiguration configuration)
...
-            return new JsonSerializerOptions()
-            {
-                DefaultIgnoreCondition = (!string.IsNullOrWhiteSpace(configuration["Serialization:Json:IgnoreNullValues"]) && bool.Parse(configuration["Serialization:Json:IgnoreNullValues"])) ? JsonIgnoreCondition.WhenWritingNull : JsonIgnoreCondition.Never,
-                PropertyNameCaseInsensitive = !string.IsNullOrWhiteSpace(configuration["Serialization:Json:PropertyNameCaseInsensitive"]) && bool.Parse(configuration["Serialization:Json:PropertyNameCaseInsensitive"]),
-                PropertyNamingPolicy = configuration["Serialization:Json:PropertyNamingPolicy"] == "CamelCase" ? JsonNamingPolicy.CamelCase : null,
-                WriteIndented = !string.IsNullOrWhiteSpace(configuration["Serialization:Json:WriteIndented"]) && bool.Parse(configuration["Serialization:Json:WriteIndented"]),
-                Converters = { new JsonStringEnumMemberConverter(JsonNamingPolicy.CamelCase) },
-                ReferenceHandler = ReferenceHandler.IgnoreCycles,
-            };
-        }
```
asanchezr, line 43: I'm guessing this was dead-code?

source/backend/core/Extensions/EnumExtensions.cs
```diff
@@ -1,4 +1,6 @@
+using Pims.Core.Security;
```
asanchezr, line 1: nit: github is complaining here about linting. perhaps need to sort the using statements

asanchezr (approved):

PR #4433, opened 2024-11-01
PSP-8759 : Creating and Editing Property Records – Property Address F…

eddherrera:
…ields

image


devinleighsmith (changes requested):
With this change, I'm guessing the "view" version of the address should also display the province, the way it does with the country.

eddherrera:
> With this change, I'm guessing the "view" version of the address should also display the province, the way it does with the country.

Province is already being displayed. I don't have a requirement for the 'Country' on the view.

devinleighsmith (approved):

PR #4434, opened 2024-11-01
psp-9436, require the user to change the advanced filter for it to be…

devinleighsmith:
… applied.

requires tests.


source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -372,7 +384,12 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
     <MapStateMachineContext.Provider
       value={{
         mapSideBarViewState: state.context.mapSideBarState,
-        isShowingSearchBar: !state.context.mapSideBarState.isOpen && !state.context.isFiltering,
+        isShowingSearchBar:
+          !state.context.mapSideBarState.isOpen &&
```
devinleighsmith, line 388: We want to hide the search bar if either the side bar or advanced filter is displayed, or if there is a non-default advanced filter.

devinleighsmith:
The intent behind the original story, was that there is a default filter in effect on the PIMS map at all times. If that default filter is changed, the filter icon should update to display that a non-default filter is in effect.

This required that we track the state of the advanced filter globally, with the global having a default value. This actually simplifies the point cluster logic as we are always filtering and not conditionally filtering using the pims visible properties list.

The other thing I added in, was to filter any properties that have no visible marker to display on the map, I did this to stop inaccurately cluster counts, which is more noticeable on your local, if you zoom in on clusters with no markers, the cluster count goes down but nothing gets displayed.

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -384,20 +401,22 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
...
         mapSearchCriteria: state.context.searchCriteria,
+        advancedSearchCriteria: state.context.advancedSearchCriteria,
         mapFeatureData: state.context.mapFeatureData,
         filePropertyLocations: state.context.filePropertyLocations,
         pendingFitBounds: state.matches({ mapVisible: { mapRequest: 'pendingFitBounds' } }),
         requestedFitBounds: state.context.requestedFitBounds,
         isSelecting: state.matches({ mapVisible: { featureView: 'selecting' } }),
         isRepositioning: isRepositioning,
         selectingComponentId: state.context.selectingComponentId,
-        isFiltering: state.context.isFiltering,
```
devinleighsmith, line 394: it is slightly harder to determine if we are filtering now, as in some ways we are "always" filtering. isFiltering is now true if we are using a non-default filter. Perhaps isFiltering could use a rename?

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -384,20 +401,22 @@ export const MapStateMachineProvider: React.FC<React.PropsWithChildren<unknown>>
...
+        isFiltering: !dequal(state.context.advancedSearchCriteria, new PropertyFilterFormModel()),
         isShowingMapFilter: isShowingMapFilter,
         isShowingMapLayers: isShowingMapLayers,
         activeLayers: state.context.activeLayers,
         activePimsPropertyIds: state.context.activePimsPropertyIds,
         showDisposed: state.context.showDisposed,
         showRetired: state.context.showRetired,
+        isMapVisible: state.matches({ mapVisible: {} }),
```
devinleighsmith, line 419: required so that the MapContainer doesn't sent the list of visible map properties before the map is loaded, meaning the state machine ignores the event.

source/frontend/src/components/maps/leaflet/Control/AdvancedFilter/FilterContentContainer.tsx
```diff
@@ -12,42 +10,27 @@ export interface IFilterContentContainerProps {
 }
 
 export const FilterContentContainer: React.FC<IFilterContentContainerProps> = ({ View }) => {
-  const { isFiltering, setVisiblePimsProperties, setShowDisposed, setShowRetired, resetMapFilter } =
-    useMapStateMachine();
-
-  const { getMatchingProperties } = usePimsPropertyRepository();
```
devinleighsmith, line 18: this request logic needed to be moved out of here as this is a conditionally rendered component, and therefore will not run during map load. It needs to run during map load since the default filter takes effect immediately now.

source/frontend/src/components/maps/leaflet/Control/AdvancedFilter/FilterContentContainer.tsx
```diff
@@ -12,42 +10,27 @@ export interface IFilterContentContainerProps {
...
   // Only render if the map state is filtering.
-  if (isFiltering) {
-    return (
-      <View
-        onChange={onChange}
-        onReset={resetMapFilter}
-        isLoading={getMatchingProperties.loading}
-      />
-    );
+  if (isShowingMapFilter) {
```
devinleighsmith, line 32: It makes more sense to show this view based on this boolean, not based on whether an advanced filter is active.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -100,38 +101,23 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
 
   const pimsLocationFeatures: FeatureCollection<Geometry, PIMS_Property_Location_View> =
     useMemo(() => {
-      if (mapMachine.isFiltering && mapMachine.mapFeatureData.pimsLocationFeatures !== null) {
```
devinleighsmith, line 103: don't need this if/else anymore because the list of activePimsPropertyIds is always active now.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -100,38 +101,23 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
+        filteredFeatures = filteredFeatures.filter(x => !x.properties.IS_RETIRED);
+      }
 
-          return {
-            type: mapMachine.mapFeatureData.pimsLocationFeatures.type,
-            features: filteredFeatures,
-          };
-        }
+      // Do not cluster any points that do not have markers on the map.
+      const displayableFeatures = filteredFeatures.filter(f => exists(getMarkerIcon(f, false)));
```
devinleighsmith, line 113: as discussed in my summary, this prevents "inaccurate" cluster counts.

asanchezr:
this PR is showing a lot of conflicts - might be worth to rebase the branch

asanchezr (approved):

PR #4436, opened 2024-11-01
PSP-8720 UI-UX cleanup - Disabled Buttons

asanchezr:
## Examples
example-01
example-02
example-03

### DIFF
--- a/source/frontend/src/components/maps/MapSearch.tsx
+++ b/source/frontend/src/components/maps/MapSearch.tsx
@@ -67,7 +67,6 @@ const StyledFilterContainer = styled(Container)`
       border-width: 0.2rem;
     }
     .btn-primary {
-      color: white;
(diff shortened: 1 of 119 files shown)


source/frontend/src/components/maps/MapSearch.tsx
```diff
@@ -67,7 +67,6 @@ const StyledFilterContainer = styled(Container)`
       border-width: 0.2rem;
     }
     .btn-primary {
-      color: white;
```
devinleighsmith, line 70: Is this intentional?

asanchezr: yes - this would interfere with the "disabled" state of the map search button (second screenshot I showed)

devinleighsmith (approved):

PR #4448, opened 2024-11-05
psp-9466 ensure that the full pims location data is included in the m…

devinleighsmith:
…ap marker click event - since the minimal layer does not have all necessary properties.


asanchezr:
@devinleighsmith - I believe I already approved and merged this fix yesterday. Please check https://github.com/bcgov/PSP/pull/4445

PR #4453, opened 2024-11-06
psp-9457 update layer popup flyout to display research/other files co…

devinleighsmith:
…rrectly for disposed/retired.

### DIFF
--- a/source/frontend/src/components/maps/leaflet/LayerPopup/LayerPopupView.test.tsx
+++ b/source/frontend/src/components/maps/leaflet/LayerPopup/LayerPopupView.test.tsx
@@ -247,7 +247,7 @@ describe('LayerPopupView component', () => {
     });
 
     it('only shows research file option if file disposed or retired', async () => {
-      const { getByTestId, getByText, queryByText } = setup({
+      const { getByTestId, getByText } = setup({
         layerPopup: {
           latlng: undefined,
           layers: [],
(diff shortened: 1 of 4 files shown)


source/frontend/src/components/maps/leaflet/LayerPopup/LayerPopupView.test.tsx
```diff
@@ -227,6 +227,111 @@ describe('LayerPopupView component', () => {
...
+      const ellipsis = getByTestId('fly-out-ellipsis');
+      await act(async () => userEvent.click(ellipsis));
+      const link = getByText('Research File');
+      expect(queryByText('Acquisition File')).toBeNull();
+      expect(queryByText('Disposition File')).toBeNull();
+      expect(queryByText('Lease/Licence File')).toBeNull();
+      expect(link).toBeVisible();
+    });
+
+    it('only shows research file option if file disposed or retired', async () => {
```
asanchezr, line 249: These two tests have the same title/description so not sure which one is which - please update to reflect actual test case

source/frontend/src/components/maps/leaflet/LayerPopup/LayerPopupView.test.tsx
```diff
@@ -227,6 +227,111 @@ describe('LayerPopupView component', () => {
...
+        },
+
+        claims: [Claims.RESEARCH_ADD],
+      });
+      const ellipsis = getByTestId('fly-out-ellipsis');
+      await act(async () => userEvent.click(ellipsis));
+      const link = getByText('Research File');
+      expect(queryByText('Acquisition File')).toBeNull();
+      expect(queryByText('Disposition File')).toBeNull();
+      expect(queryByText('Lease/Licence File')).toBeNull();
```
asanchezr, line 331: question - according to the test title this unit test is to show all create file options for non-retired properties but here we are asserting that the create Acquisition File is not present. 
Is this intended?

asanchezr (approved):
Looks good but see comments re: tests

PR #4457, opened 2024-11-07
PSP-9173 | Fix lease payment reports

FuriousLlama:
Updated repository to only retrieve the necessary information for a report


source/backend/dal/Repositories/LeasePaymentRepository.cs
```diff
@@ -47,40 +47,10 @@ public IEnumerable<PimsLeasePayment> GetAllTracking(DateTime startDate, DateTime
...
-                .Include(p => p.LeasePeriod)
-                    .ThenInclude(t => t.Lease)
-                    .ThenInclude(p => p.PimsPropertyLeases)
-                    .ThenInclude(p => p.Property)
-                    .ThenInclude(p => p.PimsHistoricalFileNumbers)
-                    .ThenInclude(h => h.HistoricalFileNumberTypeCodeNavigation)
-                .Include(p => p.LeasePeriod)
-                    .ThenInclude(t => t.Lease)
-                    .ThenInclude(p => p.RegionCodeNavigation)
-                .Include(p => p.LeasePeriod)
```
asanchezr, line 59: This whole query is a "tracking" query - but it is only used to generate a report (which is read-only). 
Any considerations to add "AsNotTracking" to the query so EF is not tracking these entities wastefully?

asanchezr (approved):
See comments re: EF query

PR #4458, opened 2024-11-07
PSP-9493 | Added purpose on edit and view for document pages

FuriousLlama:


source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
```diff
@@ -58,6 +60,36 @@ export const DocumentDetailForm: React.FunctionComponent<
     [props.documentTypes],
   );
 
+  const [documentTypePurpose, setDocumentTypePurpose] = useState(null);
+
+  const matchDocumentType = useCallback(
+    (documentTypeId: number) => {
+      const purpose = props.documentTypes.find(
```
devinleighsmith, line 67: Hmm, you need to do a type conversion of documentTypeId even though it is already typed as a number?

FuriousLlama: good catch, I think I reused the function from a place where the doctype was a string (the form)

source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
```diff
@@ -58,6 +60,36 @@ export const DocumentDetailForm: React.FunctionComponent<
...
+        x => x.id === Number(documentTypeId),
+      )?.documentTypePurpose;
+
+      setDocumentTypePurpose(purpose);
```
devinleighsmith, line 71: this will result in string | undefined | null I believe, do you want string | null here instead?

source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
```diff
@@ -58,6 +60,36 @@ export const DocumentDetailForm: React.FunctionComponent<
...
+    },
+    [props.documentTypes],
+  );
+
+  useEffect(() => {
```
devinleighsmith, line 76: hmm, I'm not sure why this one is here, since this is a ref and won't trigger useEffect updates. Why is this necessary when the event handler on line 81 exists? Is this just a useEffect that is intended to fire once during init? If so, I'm not sure if a useEffect is necessary.

source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
```diff
@@ -105,13 +137,18 @@ export const DocumentDetailForm: React.FunctionComponent<
...
-                        onChange={props.onDocumentTypeChange}
+                        onChange={onDocumentTypeChange}
                         disabled={
                           documentTypeOptions.length === 1 ||
                           props.relationshipType === ApiGen_CodeTypes_DocumentRelationType.Templates
                         }
                       />
                     </SectionField>
+                    {exists(documentTypePurpose) && (
+                      <SectionField label={null}>
```
devinleighsmith, line 148: why label={null} instead of just no declaration?

devinleighsmith (approved):
approved with minor comments.

source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
FuriousLlama: Yes, its for the first render, otherwise the purpose is not set from the other function

source/frontend/src/features/documents/documentDetail/DocumentDetailForm.tsx
FuriousLlama: thats the way we have right now. The label is required on the section field. I like having no label instead of a required null

PR #4480, opened 2024-11-15
PSP-9452: Split map search for PID PIN

asanchezr:
map-search

listview-search

### DIFF
--- a/source/backend/dal/Helpers/Extensions/PropertyViewExtensions.cs
+++ b/source/backend/dal/Helpers/Extensions/PropertyViewExtensions.cs
@@ -6,8 +6,6 @@ using LinqKit;
 using Microsoft.EntityFrameworkCore;
 using Pims.Core.Extensions;
 using Pims.Dal.Entities;
-using Pims.Core.Security;
-using Entity = Pims.Dal.Entities;
 
 namespace Pims.Dal.Helpers.Extensions
 {
@@ -24,7 +22,7 @@ namespace Pims.Dal.Helpers.Extensions
         /// <param name="user"></param>
         /// <param name="filter"></param>
         /// <returns></returns>
-        public static IQueryable<Entity.PimsPropertyVw> GeneratePropertyQuery(this PimsContext context, ClaimsPrincipal user, Entity.Models.PropertyFilter filter)
+        public static IQueryable<PimsPropertyVw> GeneratePropertyQuery(this PimsContext context, ClaimsPrincipal user, Entities.Models.PropertyFilter filter)
         {
             ArgumentNullException.ThrowIfNull(filter, nameof(filter));
             ArgumentNullException.ThrowIfNull(user, nameof(user));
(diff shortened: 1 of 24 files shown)


source/backend/dal/Helpers/Extensions/PropertyViewExtensions.cs
```diff
@@ -54,22 +51,27 @@ public static class PropertyViewExtensions
...
+            {
+                // note - 2 part search required. all matches found by removing leading 0's, then matches filtered in subsequent step. This is because EF core does not support an lpad method.
+                Regex nonInteger = new Regex("[^\\d]");
+                var formattedPid = Convert.ToInt32(nonInteger.Replace(filter.Pid, string.Empty)).ToString();
+                predicateBuilder = predicateBuilder.And(p => EF.Functions.Like(p.Pid.ToString(), $"%{formattedPid}%"));
+            }
+
+            if (!string.IsNullOrWhiteSpace(filter.Pin))
             {
                 // note - 2 part search required. all matches found by removing leading 0's, then matches filtered in subsequent step. This is because EF core does not support an lpad method.
```
devinleighsmith, line 71: this comment is no longer applicable here, correct?

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -56,25 +56,32 @@ public int Count()
...
 
-            if (!string.IsNullOrWhiteSpace(filter.PinOrPid))
+            if (!string.IsNullOrWhiteSpace(filter.Pid))
+            {
+                Regex nonInteger = new Regex("[^\\d]");
+                var formattedPid = nonInteger.Replace(filter.Pid, string.Empty);
+                items = items.Where(i => i.Pid.ToString().PadLeft(9, '0').Contains(formattedPid)).ToArray();
+            }
+
+            if (!string.IsNullOrWhiteSpace(filter.Pin))
```
devinleighsmith, line 80: is this necessary? it seems like prior to this there was some lpadding happening for the pid/pin, but it seems like a 2 part search isn't required for pin.

source/backend/tests/core/PimsTestContext.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.Extensions.Options;
+using Pims.Dal;
+using Pims.Dal.Entities;
+
+namespace Pims.Core.Test
+{
+    /// <summary>
+    /// PimsTestContext class, provides a data context to manage the datasource for the PIMS application (specific to UNIT TESTS).
+    /// </summary>
+    public class PimsTestContext : PimsContext
```
devinleighsmith, line 13: thanks for this.

devinleighsmith (approved):

source/backend/dal/Helpers/Extensions/PropertyViewExtensions.cs
asanchezr: Not sure if it applies to be honest. I think you implemented the original logic and I didn't want to break it.
If you say that the padding applies to PID only then I will remove the 2-part search from the PIN option (leaving PID as-is)

source/backend/dal/Repositories/PropertyRepository.cs
asanchezr: I see your point - I will remove the 2-part search for PIN (and leave PID as-is)

PR #4481, opened 2024-11-15
PSP-9447 :  UI UX Clean Up - Map Modal location (Tracy)

eddherrera:


source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -150,12 +154,20 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
...
       mapMachineProcessFlyTo();
     }
   }, [isMapReady, hasPendingFlyTo, requestedFlyTo, mapMachineProcessFlyTo]);
 
+  useEffect(() => {
+    if (hasPendingCenterTo && isMapReady && requestedCenterTo.location) {
+      mapRef.current?.setView(requestedCenterTo.location);
+    }
+    mapMachineProcessCenterTo();
+  }, [hasPendingCenterTo, isMapReady, mapMachineProcessCenterTo, requestedCenterTo.location]);
```
FuriousLlama, line 169: Is the setview not the same as flyto without the animation?

source/frontend/src/components/maps/leaflet/Markers/SingleMarker.tsx
```diff
@@ -83,6 +83,7 @@ const SinglePropertyMarker: React.FC<React.PropsWithChildren<SinglePropertyMarke
         fullyAttributedFeature: pointFeature.properties,
       });
     }
+    mapMachine.requestCenterToLocation(latlng);
```
FuriousLlama, line 86: why not use the flyto here?

eddherrera: 'SetView' is better suited to accomplish what is needed in this case sin pan-zoom is not required for this scenario only centering.

source/frontend/src/components/maps/MapLeafletView.tsx
eddherrera: They are similar in the sense they both center the map to a location, but 'FlyTo' is better suited for changing the map location while doing pan-zoom. Based on the documentation 'SetView' is better suited for this case.

devinleighsmith:
@eddherrera unfortunately Ana didn't update the story for this one after answering one of my comments - You'll need to check with her if this approach is approved. The approach she approved is the use of "autopan" on the popup, not necessarily centering the map when showing the popup.

eddherrera:
> @eddherrera unfortunately Ana didn't update the story for this one after answering one of my comments - You'll need to check with her if this approach is approved. The approach she approved is the use of "autopan" on the popup, not necessarily centering the map when showing the popup.

Gotcha, I'll review with her and modify accordingly.

source/frontend/src/components/maps/MapLeafletView.tsx
eddherrera: I don't need to modify the current zoom level for the new requirement. Additionally there's no easy way to use the 'flyTo' method to only center as per the reequipment. It would be more risky to modify the current 'flyTo' method to  calculate the current bounds.

eddherrera:
@devinleighsmith The current implementation  was reviewed with Ana, and she also verified with Tracy.

devinleighsmith (approved):

PR #4482, opened 2024-11-18
IS93 Scaffold

FuriousLlama:


devinleighsmith (approved):

FuriousLlama:
> @FuriousLlama in your last PR all of the concurrencyControlNumbers were changed to long? (from long) should that not be reversed here?

That is correct. However, the last pr was not merged, so no need to change that here.

PR #4483, opened 2024-11-18
psp-9500 Scheduler microservice

devinleighsmith:


source/backend/scheduler/.editorconfig
FuriousLlama, line 1: we should reuse he same editorconfig for all c# files

FuriousLlama (approved):

PR #4484, opened 2024-11-18
HOTFIX: Set the map bounds correctly - on startup.

devinleighsmith:


source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -87,6 +87,14 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
   const requestedFlyTo = mapMachine.requestedFlyTo;
   const mapMachineProcessFlyTo = mapMachine.processFlyTo;
 
+  // Set the bounds when the map is ready. Not called from existing handleMapCreated as that function is called every time a state change occurs.
+  useEffect(() => {
+    const bounds = mapRef?.current?.getBounds();
+    if (exists(bounds)) {
```
FuriousLlama, line 93: should it check if the isMapReady is true?

asanchezr (approved):

PR #4494, opened 2024-11-21
psp-9446 allow historical numbers to be collapsed in two dimensions.

devinleighsmith:


asanchezr (approved):

asanchezr:
Need to update snapshots

PR #4497, opened 2024-11-21
PSP-9133 : Implement check/warning on frontend that backend and datab…

eddherrera:
…ase versions are as expected.

image


source/backend/dal/Repositories/SystemConstantRepository.cs
```diff
@@ -35,6 +35,12 @@ public IEnumerable<PimsStaticVariable> GetAll()
         {
             return this.Context.PimsStaticVariables.AsNoTracking().ToArray();
         }
+
+        public string GetDataBaseVersion()
+        {
+            return this.Context.PimsStaticVariables.Where(x => x.StaticVariableName == "DBVERSION")
+                .AsNoTracking().FirstOrDefault().StaticVariableValue ?? string.Empty;
```
asanchezr, line 42: There is potential for a null-pointer exception in this code if the database didn't have an entry with key "DBVERSION" in the static variables table. Granted, we would never see this as we populate that table but better avoid the NPE if possible.

If no DBVERSION value is in the table, then **FirstOrDefault()** returns null and the call to get the **.StaticVariableValue** would throw an exception. We would just need to replace with:

```cs
...AsNoTracking().FirstOrDefault()?.StaticVariable ?? string.Empty
```

source/frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -27,11 +32,78 @@ export const ApiVersionInfo = () => {
...
+  const frontEndDBVersion = findDBVersion(frontEndVersion);
+
+  const apiVersionMissmatch = version?.informationalVersion !== frontEndVersion;
+  const dbVersionMissmatch = version?.dbVersion !== frontEndDBVersion;
+
+  const versionMissmatchMsg = (apiMissmatch: boolean, dbMissmatch: boolean): string => {
+    let msg = '';
+    if (apiMissmatch || dbMissmatch) {
+      msg = msg.concat(
+        `Warning: There is a version missmatch with the backend.
```
asanchezr, line 59: typo: missmatch -> mismatch

source/frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -27,11 +32,78 @@ export const ApiVersionInfo = () => {
...
+  display: flex;
+  gap: 10px;
+  flex-grow: 1;
+  flex-direction: row;
+  justify-content: flex-end;
+  align-items: center;
+`;
+
+export const VersionMissmatchDiv = styled(InlineFlexDiv)`
+  text-transform: uppercase;
```
asanchezr, line 100: I don't think this style is needed. We don't want the warning message in all caps

asanchezr (approved):
See comments

source/frontend/src/components/common/ApiVersionInfo.tsx
eddherrera: removed

source/frontend/src/components/common/ApiVersionInfo.tsx
eddherrera: Updated

PR #4500, opened 2024-11-23
PSP-9486 UI UX Property Information - moving PIN information to the left

asanchezr:
capture

### DIFF
--- a/source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
+++ b/source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
@@ -30,17 +30,17 @@ export interface IMotiInventoryHeaderProps {
 }
 
 export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderProps> = props => {
-  const pid = pidFormatter(props.composedProperty.pid);
-
   const parcelMapData = props.composedProperty.parcelMapFeatureCollection;
   const geoserverMapData = props.composedProperty.geoserverFeatureCollection;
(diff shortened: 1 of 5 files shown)


source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
```diff
@@ -81,10 +81,18 @@ export const MotiInventoryHeader: React.FunctionComponent<IMotiInventoryHeaderPr
       <LoadingBackdrop show={isLoading} parentScreen={true} />
       <Row className="no-gutters">
         <Col xs="7">
-          <HeaderField label="Civic Address:" labelWidth={'3'} contentWidth="9">
+          <Row className="justify-content-start">
+            <HeaderLabelCol label="PID #:" labelWidth="3" />
+            <HeaderContentCol>{pid}</HeaderContentCol>
+            <HeaderLabelCol label="PIN #:" />
+            <HeaderContentCol>
+              <span className="pl-2">{pin}</span>
```
devinleighsmith, line 89: what does the wrapping span provide?

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/property/MotiInventoryHeader.tsx
asanchezr: the span is there so we can space the PIN value a but from the pin label. I cannot add a classname to HeaderContentCol because it is already a styled components with its own styles (and it is re-used on other places)

PR #4503, opened 2024-11-27
PSP-9561 Update release GHA to support deployments

asanchezr:


.github/workflows/uat_pre_release_hotfix_argo.yml
```diff
@@ -0,0 +1,272 @@
...
+        with:
+          openshift_server_url: ${{ env.OPENSHIFT_SERVER }}
+          openshift_token: ${{ env.OPENSHIFT_TOKEN }}
+          insecure_skip_tls_verify: true
+          namespace: 3cd915-tools
+      - name: create all PSP build configurations for branch
+        shell: bash
+        run: |
+          oc process -f ./openshift/s2i/nginx-runtime/nginx-runtime.yaml -p GIT_REF=${{github.event.inputs.HOTFIX_BRANCH}} -p OUTPUT_IMAGE_TAG="latest-${{github.event.inputs.HOTFIX_BRANCH}}" | oc create --selector ci_cd=true -f -
+          oc process -f ./openshift/4.0/templates/api/build.yaml -p GIT_REF=${{github.event.inputs.HOTFIX_BRANCH}} -p OUTPUT_IMAGE_TAG="latest-${{github.event.inputs.HOTFIX_BRANCH}}" | oc create --selector ci_cd=true -f -
```
devinleighsmith, line 70: We would also need a create builds for the new microservices for this to work (or there won't be a correctly named build for the proxy to run). Options here are:
1. add a build.yaml for the proxy and then add a section here to create that build, same as the others.
2. disable this GHA for now, as we always hold UAT until after release these days.

.github/workflows/uat_hotfix_argo.yml
```diff
@@ -0,0 +1,222 @@
...
+        uses: redhat-actions/oc-login@v1
+        with:
+          openshift_server_url: ${{ env.OPENSHIFT_SERVER }}
+          openshift_token: ${{ env.OPENSHIFT_TOKEN }}
+          insecure_skip_tls_verify: true
+          namespace: ${{ env.OPENSHIFT_TOOLS_NAMESPACE }}
+      - name: Call script to build frontend (pims-app and pims-app-base)
+        run: |
+          ./openshift/4.0/player.sh build app-base -apply
+          ./openshift/4.0/player.sh build app -apply
```
devinleighsmith, line 65: You'll need to build the proxy as well (that is in my PR for the scheduler microservice).

.github/workflows/uat_hotfix_argo.yml
```diff
@@ -0,0 +1,222 @@
...
+          openshift_server_url: ${{ env.OPENSHIFT_SERVER }}
+          openshift_token: ${{ env.OPENSHIFT_TOKEN }}
+          insecure_skip_tls_verify: true
+          namespace: 3cd915-tools
+      - name: tag uat image such that it can be promoted to prod
+        shell: bash
+        run: |
+          VERSION=$(make version)
+          oc tag pims-app:uat pims-app:v${VERSION}-master
+          oc tag pims-api:uat pims-api:v${VERSION}-master
```
devinleighsmith, line 204: also need a tag here for the proxy.

.github/workflows/uat_pre_release_hotfix_argo.yml
asanchezr: This action is for the weird case of a UAT hotfix where we are holding UAT to release to PROD  (I think).
The most used UAT hotfix action is the other one (named uat-hotfix.yaml). I think Devin knows the difference.
This would need to be modified in the near future

devinleighsmith (changes requested):
see comments on uat_pre_release_hotfix_argo - you'll need to choose option 1, 2 or a third I haven't thought of.

.github/workflows/uat_pre_release_hotfix_argo.yml
asanchezr: I think I am going to disable this GHA for now and log tech debt to address this. Otherwise the rest of the changes would be waiting for that. As you said - this is not often used.

devinleighsmith (approved):

.github/workflows/uat_hotfix_argo.yml
asanchezr: The proxy DOES get built in the next step (bundled with the API build). This step is for the frontend. Did I miss something else?

.github/workflows/uat_hotfix_argo.yml
asanchezr: Good catch. Fixed there and in the retag GHA from test to uat

asanchezr: Also needed to add a tag for mayan here

PR #4529, opened 2024-12-12
Psp 9512 jobs for upload, polling, retries

devinleighsmith:


source/backend/api/Areas/Documents/DocumentQueueController.cs
```diff
@@ -33,32 +42,155 @@ public class DocumentQueueController : ControllerBase
...
+            var documentQueueModels = _mapper.Map<List<DocumentQueueModel>>(queuedDocuments);
             return new JsonResult(documentQueueModels);
         }
 
+        /// <summary>
+        /// Get Document Queue item via id.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("{documentQueueId:long}")]
+        [HasPermission(Permissions.SystemAdmin)]
```
asanchezr, line 175: general question about permissions - I see all queue endpoints require sysadmin permissions. Is this intended? How will the frontend connect to this new API for non-admin users?

devinleighsmith: the frontend won't connect to this api, at least, not directly. The documents will come in from the frontend via the document controller, which will handle adding the document queue entity. This controller that allows direct interaction with the document queue is really only used by the scheduler.

asanchezr (approved):

PR #4532, opened 2024-12-16
psp-9457 leases allow disposed properties to be added.

devinleighsmith:


source/backend/api/Services/LeaseService.cs
```diff
@@ -548,8 +542,11 @@ private PimsLease AssociatePropertyLeases(PimsLease lease, IEnumerable<UserOverr
             {
                 PimsProperty property = propertyLease.Property;
                 var existingPropertyLeases = _propertyLeaseRepository.GetAllByPropertyId(property.PropertyId);
+                var propertyWithAssociations = _propertyRepository.GetAllAssociationsById(property.PropertyId);
                 var isPropertyOnOtherLease = existingPropertyLeases.Any(p => p.LeaseId != lease.Internal_Id);
                 var isPropertyOnThisLease = existingPropertyLeases.Any(p => p.LeaseId == lease.Internal_Id);
+                var isDisposedOrRetired = existingPropertyLeases.Any(p => p.Property.IsRetired.HasValue)
+                    || propertyWithAssociations?.PimsDispositionFileProperties?.Any(d => d.DispositionFile.DispositionFileStatusTypeCode == DispositionFileStatusTypes.COMPLETE.ToString()) == true;
```
asanchezr, line 549: question: the old logic had this `x.Property.IsRetired.HasValue && x.Property.IsRetired.Value` to account for NULLABLE retired flag to have a value (and that the value is **true**).

However the new/fixed logic is only checking that "IsRetired" is not null. Shouldn't we also check that it is true as well below?
`existingPropertyLeases.Any(p => p.Property.IsRetired.HasValue)`

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
```diff
@@ -479,10 +536,114 @@ public void UpdateProperties_WithRetiredProperty_Should_Fail()
...
+            // Arrange
+            var lease = EntityHelper.CreateLease(1);
+
+            var service = this.CreateLeaseService(Permissions.LeaseEdit, Permissions.PropertyAdd, Permissions.PropertyView);
+            var leaseRepository = this._helper.GetService<Mock<ILeaseRepository>>();
+            var propertyLeaseRepository = this._helper.GetService<Mock<IPropertyLeaseRepository>>();
+            var propertyRepository = this._helper.GetService<Mock<IPropertyRepository>>();
+            var userRepository = this._helper.GetService<Mock<IUserRepository>>();
+
+            PimsProperty disposedProperty = new PimsProperty()
```
asanchezr, line 639: nit: disposedProperty -> retiredProperty

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
```diff
@@ -523,6 +684,42 @@ public void UpdateProperties_MatchProperties_Success()
             propertyService.Verify(x => x.UpdateLocation(It.IsAny<PimsProperty>(), ref It.Ref<PimsProperty>.IsAny, It.IsAny<IEnumerable<UserOverrideCode>>()), Times.Once);
         }
 
+        [Fact]
+        public void UpdateProperties_MatchProperties_Success_NoInternalId()
+        {
+            // Arrange
+            var lease = EntityHelper.CreateLease(1);
```
asanchezr, line 691: I believe this test was already added in your other PR. Not sure if it would end up in a duplicate test when both PRs get merged

asanchezr (approved):
See comments

PR #4535, opened 2024-12-17
PSP-9513 : Change document workflow - support asynchronous upload

eddherrera:


source/backend/api/Services/FormDocumentService.cs
```diff
@@ -60,6 +60,47 @@ public IList<PimsFormType> GetFormDocumentTypes(string formTypeCode)
             return new List<PimsFormType>() { _formTypeRepository.GetByFormTypeCode(formTypeCode) };
         }
 
+        public async Task UploadFormDocumentTemplate(string formTypeCode, DocumentUploadRequest uploadRequest)
+        {
+            throw new NotImplementedException();
```
devinleighsmith, line 65: Can you explain what is going on here? generally we do no leave blocks of commented-out code in the application.

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
-                    ResearchFileId = researchFileId,
-                    DocumentId = uploadResult.Document.Id,
-                };
-                newResearchFileDocument = researchFileDocumentRepository.AddResearch(newResearchFileDocument);
-                researchFileDocumentRepository.CommitTransaction();
-
-                relationshipResponse.DocumentRelationship = mapper.Map<DocumentRelationshipModel>(newResearchFileDocument);
-            }
+                DocumentId = pimsDocument.DocumentId,
+                AcquisitionFileDocumentId = newAcquisitionDocument.AcquisitionFileDocumentId,
```
devinleighsmith, line 119: You can take this line out, as we are removing all file foreign key relationships from the document queue table. Same goes for each of the file types.

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
-            DocumentUploadRelationshipResponse relationshipResponse = new DocumentUploadRelationshipResponse()
+            // Step 2 - Create the File Document
+            PimsAcquisitionFileDocument newAcquisitionDocument = new()
             {
-                UploadResponse = uploadResult,
+                AcquisitionFileId = acquisitionFileId,
+                DocumentId = pimsDocument.DocumentId,
             };
+            _acquisitionFileDocumentRepository.AddAcquisition(newAcquisitionDocument);
+            _documentQueueRepository.SaveChanges();
```
devinleighsmith, line 113: I don't think you want to call save changes between repository transactions, I'm guessing all of this work should be done in a single transaction to allow rollback if a later transaction fails.

source/backend/api/Helpers/Extensions/DocumentUploadRequestExtension.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using Pims.Api.Models.Requests.Document.Upload;
+using Pims.Core.Api.Exceptions;
+
+namespace Pims.Api.Helpers.Extensions
+{
+    public static class DocumentUploadRequestExtension
+    {
+        public static void ThrowInvalidFileSize(this DocumentUploadRequest documentUploadRequest)
```
devinleighsmith, line 9: Hmm, I'm curious about what the benefit of moving this to an extension method is, can you provide some background?

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
-                UploadResponse = uploadResult,
+                ProjectId = projectId,
+                DocumentId = pimsDocument.DocumentId,
             };
+            _projectRepository.AddProjectDocument(newFileDocument);
+            _documentQueueRepository.SaveChanges();
 
-            // Throw an error if Mayan returns a null document. This means it wasn't able to store it.
-            ValidateDocumentUploadResponse(uploadResult);
+            throw new NotImplementedException();
```
devinleighsmith, line 193: @eddherrera what is going on here? why can't we save project documents?

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
             }
+
+            return result;
         }
 
-        private static void ValidateZeroLengthFile(DocumentUploadRequest uploadRequest)
+        private PimsDocument RemoveDocumentMayanID(PimsDocument doc)
         {
-            if (uploadRequest.File is not null && uploadRequest.File.Length == 0)
+            // Update the Mayan ID from PIMS, this will leave a trace
```
devinleighsmith, line 568: I don't think this method executed alone will leave a trace.

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
             Logger.LogInformation("Deleting PIMS document for single research file");
             User.ThrowIfNotAllAuthorized(Permissions.DocumentDelete, Permissions.ResearchFileEdit);
 
-            var relationshipCount = _documentRepository.DocumentRelationshipCount(researchFileDocument.DocumentId);
-            if (relationshipCount == 1)
-            {
-                return await documentService.DeleteDocumentAsync(researchFileDocument.Document);
-            }
-            else
+            var result = new ExternalResponse<string>() { Status = ExternalResponseStatus.NotExecuted };
```
devinleighsmith, line 339: hmm, if a status of NotExecuted is returned in the case of a document with no mayanId, does that cause problems on the frontend?

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -89,333 +91,509 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
+            return _documentRepository.Update(doc, false);
+        }
+
+        private void DeleteQueuedDocumentItem(long documentId)
+        {
+            var documentQueuedItem = _documentQueueRepository.GetByDocumentId(documentId);
+            if (documentQueuedItem.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.PENDING.ToString()
+                || documentQueuedItem.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.PROCESSING.ToString())
+            {
+                throw new BadRequestException("Doucment in process can not be deleted");
```
devinleighsmith, line 579: *Document

source/backend/apimodels/Models/Concepts/Document/DocumentMap.cs
```diff
@@ -14,6 +15,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.DocumentType, src => src.DocumentType)
                 .Map(dest => dest.StatusTypeCode, src => src.DocumentStatusTypeCodeNavigation)
                 .Map(dest => dest.FileName, src => src.FileName)
+                .Map(dest => dest.DocumentQueueStatusTypeCode, src => src.PimsDocumentQueues.Count > 0 ? src.PimsDocumentQueues.FirstOrDefault().DocumentQueueStatusTypeCodeNavigation : null)
```
devinleighsmith, line 18: This is an interesting one, as generally we try to pass concepts as-is without merging entity fields. Thoughts @FuriousLlama ?

source/backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -35,7 +35,7 @@ public DocumentRepository(
         #region Methods
 
         /// <summary>
-        /// Get the document from the database based on document id.
+        /// Get the document from the database based on document id.z
```
devinleighsmith, line 38: z?

eddherrera:
@devinleighsmith I am changing this one to a draft since it seems it would still need adjustments with the pending changes coming to IS96 DB Schema. I'll address most of the comments now.

devinleighsmith:
@eddherrera I've done some local testing and I don't see the processing or error states (of the document queue) being shown on the frontend

devinleighsmith:
@eddherrera I don't think any of my comments need to wait for db changes, they relate to changing the logic to better fit both the current and the upcoming schema. There should be no upcoming schema that you need to wait for (in regards to my changes at least).

source/frontend/src/components/common/buttons/RefreshButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Button displaying a refresh/recycle icon, used to reset form data.
+ * @param param0
+ */
+const RefreshButton: React.FC<React.PropsWithChildren<ButtonProps>> = ({ ...props }) => {
+  return (
+    <TooltipWrapper tooltipId="btn-refresh-tooltip" tooltip="Refresh">
+      <Button
+        data-testid="refresh-button"
```
devinleighsmith, line 15: given that this button will be placed inside other screens, please make this id/testid/title more specific

source/frontend/src/features/mapSideBar/acquisition/models/PayeeOptionModel.ts
```diff
@@ -102,7 +102,6 @@ export class PayeeOption {
       agreementDate: null,
       expropriationNoticeServedDate: null,
       expropriationVestingDate: null,
-      advancedPaymentServedDate: null,
```
devinleighsmith, line 105: hmm, there seems to be some unrelated changes in this PR, thoughts?

devinleighsmith:
@eddherrera it seems that right now, when not metadata is included with the document, the word "null" is being written into the DOCUMENT_METADATA column.

source/backend/api/Services/FormDocumentService.cs
eddherrera: removed, this was part of draft.

source/backend/api/Services/DocumentFileService.cs
eddherrera: removed

source/backend/api/Services/DocumentFileService.cs
eddherrera: this is still a single transaction, SaveChanges() won't commit any work and it will rollback if the transaction fails.

source/backend/api/Helpers/Extensions/DocumentUploadRequestExtension.cs
eddherrera: this private method was first defined on a different Service and since I needed it on another it made sense to put it as an Extension method at that time. At this point if it's only used in one service I could make it back to a private method.

source/backend/api/Services/DocumentFileService.cs
eddherrera: This was related to a column missing in the Queue Table but that has been fixed now. All good now.

source/backend/api/Services/DocumentFileService.cs
eddherrera: making sure that the mayan Id is deleted is the calling to mayan api is successful and making sure that mayan id is null after that so it does not point to a deleted file. trace is just the history table

source/backend/api/Services/DocumentFileService.cs
eddherrera: no issues.

source/backend/dal/Repositories/DocumentRepository.cs
eddherrera: removed

source/frontend/src/components/common/buttons/RefreshButton.tsx
eddherrera: updated

source/frontend/src/features/mapSideBar/acquisition/models/PayeeOptionModel.ts
eddherrera: needed to compile, not sure who's fixing those?

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
```diff
@@ -112,46 +132,103 @@ const renderActions = (
...
-              icon={<FaEye size={24} id={`document-view-${index}`} title="document view details" />}
-              onClick={() => original?.id && onViewDetails(DocumentRow.toApi(original))}
-            ></Button>
-          </Col>
-        )}
-        {hasClaim(Claims.DOCUMENT_DELETE) && (
-          <Col>
+
+    const documentInError =
+      (original.mayanDocumentId === null || original.mayanDocumentId === -1) &&
```
devinleighsmith, line 137: This logic appears inconsistent across this file. Also, I believe you can use isValidId to be consistent with the rest of our code base.

devinleighsmith: In my testing, mayanId is being converted from null to 0 on the backend, so none of these icons are being displayed.

devinleighsmith: Also, it looks like some of these boolean statements can be collapsed to make them more readable (ie: original.queueStatusTypeCode && original.queueStatusTypeCode.id === ApiGen_CodeTypes_DocumentQueueStatusTypes.PENDING
is equivalent to original?.queueStatusTypeCode === ApiGen_CodeTypes_DocumentQueueStatusTypes.PENDING

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
```diff
@@ -112,46 +132,103 @@ const renderActions = (
...
+            icon={<FaEye size={21} id={`document-view-${index}`} title="document view details" />}
+            onClick={() => original?.id && onViewDetails(DocumentRow.toApi(original))}
+          ></Button>
+        )}
+
+        {hasClaim(Claims.DOCUMENT_DELETE) && (
+          <StyledRemoveLinkButton
+            data-testid="document-delete-button"
+            icon={<FaTrash size={21} id={`document-delete-${index}`} title="document delete" />}
+            onClick={() => original?.id && onDelete(DocumentRow.toApi(original))}
```
devinleighsmith, line 204: I'd prefer to not display the button if there is no id, instead of displaying a button that does nothing when clicked.

source/backend/api/Services/DocumentFileService.cs
devinleighsmith: at the very least you don't need to call savechanges twice, have a look at that method.

source/backend/dal/Repositories/DocumentQueueRepository.cs
```diff
@@ -32,6 +34,36 @@ public DocumentQueueRepository(
...
+        }
+
+        /// <summary>
+        /// Find Queue Item for a Document.
+        /// </summary>
+        /// <param name="documentId"></param>
+        /// <returns></returns>
+        public PimsDocumentQueue GetByDocumentId(long documentId)
+        {
+            return Context.PimsDocumentQueues.Where(x => x.DocumentId == documentId).FirstOrDefault();
```
devinleighsmith, line 64: should this be asNoTracking?

source/backend/tests/unit/api/Services/DocumentFileServiceTest.cs
```diff
@@ -288,48 +290,48 @@ public void UploadDocumentAsync_Disposition_ShouldThrowException_NotAuthorized()
...
-                DocumentStatusCode = "DocumentStatus",
-            };
-
-            await service.UploadProjectDocumentAsync(1, uploadRequest);
-
-            // Assert
-            documentService.Verify(x => x.UploadDocumentAsync(It.IsAny<DocumentUploadRequest>()), Times.Once);
-            projectRepository.Verify(x => x.AddProjectDocument(It.IsAny<PimsProjectDocument>()), Times.Once);
-        }
+        //[Fact]
```
devinleighsmith, line 300: @eddherrera why is this commented out?

source/frontend/src/components/common/buttons/RefreshButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+const RefreshButton: React.FC<React.PropsWithChildren<IRefreshButtonProps>> = ({
+  toolId,
+  toolText,
+  refreshButtonId,
+  refreshButtonTestId,
+  refreshButtonTitle,
+  ...props
+}) => {
+  const refreshButtonIdValue = refreshButtonId ?? 'refresh-button';
```
devinleighsmith, line 32: please set the defaults such that the dom won't throw an error due to duplicate ids.

source/frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
```diff
@@ -125,7 +125,7 @@ export const DocumentUploadContainer = forwardRef<
       await Promise.all(uploadDocumentTasks);
     setIsUploading(false);
     if (tasksResult !== undefined) {
-      const batchResults = tasksResult.map<BatchUploadResponseModel>((r, index) => {
+      const batchResults = tasksResult.map((r, index) => {
```
devinleighsmith, line 128: Why remove the type? typings are nice!

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
```diff
@@ -112,46 +132,103 @@ const renderActions = (
...
+        original.queueStatusTypeCode.id === ApiGen_CodeTypes_DocumentQueueStatusTypes.MAYAN_ERROR);
+
+    const documentProcessing =
+      (original.mayanDocumentId === null || original.mayanDocumentId < 1) &&
+      original.queueStatusTypeCode &&
+      (original.queueStatusTypeCode.id === ApiGen_CodeTypes_DocumentQueueStatusTypes.PENDING ||
+        original.queueStatusTypeCode.id === ApiGen_CodeTypes_DocumentQueueStatusTypes.PROCESSING);
+
+    const canViewDocument =
+      (original.mayanDocumentId &&
```
devinleighsmith, line 149: the mayan document id being present isn't sufficient to download a document if it is still being processed - need to check for success.

source/backend/api/Services/DocumentFileService.cs
eddherrera: not sure what you mean, savechanges is called just 3 times, one for the document, for the file-document and for the que queue document.

source/backend/dal/Repositories/DocumentQueueRepository.cs
eddherrera: Needed to update the mayanId to NULL after successful delete call to Mayan API.

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
eddherrera: A little lost here, the only way this displays is when the document has been processed as either success or failure. if it fails it deletes everything so they can retry.

source/frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
eddherrera: I'll try to put it back, not quite the type I am expecting since this one is tightly coupled to Mayan responses.

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
eddherrera: I wouldn't expect a MayanId if it's still being processed, is this not the case?

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
devinleighsmith: right now when the button is checked you ensure that original?.id is set before you fire the call to onDelete. I'd rather you either remove that check, or hide the entire button if original?.id is not set.

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
devinleighsmith: Mayan ids are returned from mayan immediately so that you can track the progress of the document as it is processed. documents may have mayan ids but no file contents while they are being processed.

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
eddherrera: Updated

source/backend/tests/unit/api/Services/DocumentFileServiceTest.cs
eddherrera: Updated

source/frontend/src/features/documents/list/DocumentResults/DocumentResultsColumns.tsx
eddherrera: UIpdated

source/frontend/src/features/documents/documentUpload/DocumentUploadContainer.tsx
eddherrera: updated

eddherrera:
@devinleighsmith  please review comments and bugs are addressed

devinleighsmith (approved):

PR #4539, opened 2024-12-18
PSP-9570 Property Information leaflet should open up when a property is part of the PIMS Inventory regardless if it has a PID or not

asanchezr:
https://github.com/user-attachments/assets/7023593e-5ba0-4236-b43b-f010dab9b544


source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.tsx
```diff
@@ -42,7 +43,13 @@ export const MotiInventoryContainer: React.FunctionComponent<
   const composedPropertyState = useComposedProperties({
     id: props.id,
     pid:
-      props?.pid === undefined || props?.pid === '' || isNaN(+props.pid) ? undefined : +props.pid,
+      props?.pid === undefined || props?.pid === '' || isNaN(Number(props.pid))
+        ? undefined
+        : Number(props.pid),
+    pin:
```
devinleighsmith, line 49: would it make sense to use the pinparser here?

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/property/MotiInventoryContainer.tsx
asanchezr: yes, makes sense. Will make the change.

PR #4545, opened 2024-12-20
Add alter ups to change the name of the PIMS_ROLEs for lease/license

devinleighsmith:


asanchezr (approved):
Looks good but I wonder if the same changes should be added to the BUILD folder and to the corresponding PSP_PIMS_LATEST folder as well. 
Otherwise we risk loosing those changes when Doug creates a new migration

PR #4546, opened 2024-12-20
PSP-9211 and PSP-9719: UI/UX Clean up: Map Icons Replacement and PIMS File header colors

stairaku:
image
image


asanchezr (approved):
Approved but you will need to fix the frontend unit tests

PR #4548, opened 2024-12-21
PSP-9740 HOTFIX: User is not able to update the Acquisition Files

asanchezr:
The root cause was a previous attempt to fix a UTC date, which resulted in this error.
Fixed by improving how we assign default "current date" in the backend.

image


source/frontend/src/features/mapSideBar/acquisition/tabs/fileDetails/update/models.ts
```diff
@@ -121,7 +120,7 @@ export class UpdateAcquisitionSummaryFormModel
     newForm.legacyFileNumber = model.legacyFileNumber ?? undefined;
     newForm.fileName = model.fileName || '';
     newForm.rowVersion = model.rowVersion ?? undefined;
-    newForm.assignedDate = model.assignedDate ? formatUTCDateTime(model.assignedDate) : '';
```
eddherrera, line 124: isn't his going to bring back the issue of the time zone?

asanchezr: nope because if you see the changes in the backend will generate the current date in the server timezone (DateTime.Today) - we have configured our API servers to run in PST timezone so we don't get the issue anymore.
The original issue happened because when this field is not populated, the DB was assigning the current UTC date.
By having the backend put a value always (either what comes from the frontend or the current date (non-UTC)) we fix this and do not need to format as UTC in the frontend.

This is consistent with the other dates you see on the acquisition details screen - none of those are formatted as UTC.

eddherrera (approved):

PR #4552, opened 2024-12-24
PSP-9471: UI UX Clean Up - Contact – Manage Contacts: Allow to save contacts that are set as Inactive without Contact Methods

stairaku:


source/frontend/src/features/contacts/contact/edit/Person/UpdatePersonForm.test.tsx
```diff
@@ -13,7 +13,15 @@ import { ApiGen_Concepts_Organization } from '@/models/api/generated/ApiGen_Conc
 import { ApiGen_Concepts_Person } from '@/models/api/generated/ApiGen_Concepts_Person';
 import { ApiGen_Concepts_PersonAddress } from '@/models/api/generated/ApiGen_Concepts_PersonAddress';
 import { lookupCodesSlice } from '@/store/slices/lookupCodes';
-import { act, fillInput, render, RenderOptions, userEvent } from '@/utils/test-utils';
+import {
+  act,
+  fillInput,
+  findByText,
+  queryByText,
```
asanchezr, line 20: not sure we are using these 2 imports in the test file here...

stairaku: Let me check and will update my PR

source/frontend/src/features/contacts/contact/utils/contactUtils.ts
```diff
@@ -23,9 +23,14 @@ export const onValidateOrganization = (
   const errors = {} as any;
   try {
     // combine yup schema validation with custom rules
-    if (!hasEmail(values) && !hasPhoneNumber(values) && !hasAddress(values)) {
+    if (
+      !hasEmail(values) &&
+      !hasPhoneNumber(values) &&
+      !hasAddress(values) &&
+      values.isDisabled === false
```
asanchezr, line 30: the `validatePerson()` method below has the following logic:
```tsx
if (
   ... &&
   (values.isDisabled === 'false' || values.isDisabled === false)
) {
...
}
```

But the `validateOrganization()` has a different logic: `values.isDisabled === false`
wouldn't it make sense to have the same check for disabled for person and organization?

stairaku: I agree, but I'll have to do a refactor on the logic we have now on contacts. I'll give it a go and update my PR.

source/frontend/src/features/contacts/contact/edit/Organization/UpdateOrganizationForm.tsx
```diff
@@ -83,12 +83,16 @@ const UpdateOrganization: React.FC<FormikProps<IEditableOrganizationForm>> = ({
...
   const organizationId = getIn(values, 'id');
   const persons = getIn(values, 'persons') as Partial<ApiGen_Concepts_Person>[];
+  const isOrganizationDisabled = getIn(values, 'isDisabled');
 
   const onCancel = () => {
     history.push(`/contact/O${organizationId}`);
   };
 
   const isContactMethodInvalid = useMemo(() => {
+    if (isOrganizationDisabled === 'true') {
```
eddherrera, line 93: Feels weird that this should be a boolean but the rigth hand operator is a string?

source/frontend/src/features/contacts/contact/edit/Person/UpdatePersonForm.tsx
```diff
@@ -90,6 +91,9 @@ const UpdatePersonComponent: React.FC<
   };
 
   const isContactMethodInvalid = useMemo(() => {
+    if (isPersonDisabled === 'true') {
```
eddherrera, line 94: one is a boolean, the right hand is a string. probably need to check the right type?

source/frontend/src/features/contacts/contact/edit/Organization/UpdateOrganizationForm.tsx
asanchezr: This is because Formik values for dropdowns (SELECT input field) are always strings. We have several utility methods to convert from "stringy" booleans to "real" booleans

source/frontend/src/features/contacts/formModels.ts
```diff
@@ -183,7 +183,7 @@ export class IEditableOrganizationForm {
   alias?: string;
   incorporationNumber?: string;
   comment?: string;
-  isDisabled: boolean;
+  isDisabled: string | boolean = false;
```
asanchezr, line 186: This field should **always** be string. Yes, it will contain string values of "true" or "false" but those are string representations of boolean values. We need this field to be a string because it is used in Formik to render the "disabled" dropdown (SELECT input) which always takes string values (so no boolean values are accepted for select)

The pattern we follow is to use the utility method **booleanToString()** to convert from the API value (which is boolean) into the string representation that the dropdown can render. We do this in the **fromApi()** method.

Then in the corresponding **toApi()** method we call the "inverse" function - **stringToBoolean()** to convert the string representation from formik into the API boolean value.

From what I see the code in contact is doing this half-way. The "toApi" is correctly converting from string to boolean. BUT the "fromApi" is missing the conversion from a boolean. The solution is to make "isDisabled" here a string and follow the pattern from "UpdatePropertyDetailsFormModel.ts" below:

image

source/frontend/src/features/contacts/contact/utils/contactUtils.ts
```diff
@@ -23,9 +23,14 @@ export const onValidateOrganization = (
   const errors = {} as any;
   try {
     // combine yup schema validation with custom rules
-    if (!hasEmail(values) && !hasPhoneNumber(values) && !hasAddress(values)) {
+    if (
+      !hasEmail(values) &&
+      !hasPhoneNumber(values) &&
+      !hasAddress(values) &&
+      (values.isDisabled === 'false' || values.isDisabled === false)
+    ) {
```
asanchezr, line 31: if you follow the pattern I described in my other comment, then here you would only need to check for the string "false"

source/frontend/src/features/contacts/contact/utils/contactUtils.ts
```diff
@@ -40,9 +45,14 @@ export const onValidatePerson = (values: IEditablePersonForm, otherCountryId?: s
...
```
asanchezr, line 53: same comment here re: checking for the string representation value

asanchezr (approved):
Please see comments

source/frontend/src/features/contacts/contact/edit/Organization/UpdateOrganizationForm.tsx
stairaku: let me try with with booleans. I see the tranform method when it is saved and send to the API but not before that.

source/frontend/src/features/contacts/formModels.ts
stairaku: Thank you for the comments and lead me to a better solution. I'm updating this PR with the suggested changes. Thank you!!!!

PR #4555, opened 2024-12-24
psp-9256 allow users added to a project to view an acquisition file.

devinleighsmith:
tests are coming in a future PR.


source/backend/apimodels/Models/Concepts/Project/ProjectPersonMap.cs
```diff
@@ -0,0 +1,27 @@
...
+                .Map(dest => dest.PersonId, src => src.PersonId)
+                .Map(dest => dest.Project, src => src.Project)
+                .Inherits<Entity.IBaseEntity, BaseConcurrentModel>();
+
+            config.NewConfig<ProjectPersonModel, Entity.PimsProjectPerson>()
+                .Map(dest => dest.ProjectPersonId, src => src.Id)
+                .Map(dest => dest.ProjectId, src => src.ProjectId)
+                .Map(dest => dest.PersonId, src => src.PersonId)
+                .Map(dest => dest.Person, src => src.Person)
+                .Inherits<BaseConcurrentModel, Entity.IBaseEntity>();
```
asanchezr, line 24: question - I believe our pattern when we map models into EF entities is that we keep the ID but do not map child entities to avoid accidental updates to the children - right?

I see here that we correctly map the PersonId so that we can point to a new existing person but I also see the mapping of `.Map(dest => dest.Person, src => src.Person)` - shouldn't we take this out here?

source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
```diff
@@ -132,6 +136,9 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
   const fetchAcquisitionFile = useCallback(async () => {
     const retrieved = await retrieveAcquisitionFile(acquisitionFileId);
     if (exists(retrieved)) {
+      if (isValidId(retrieved.projectId)) {
+        retrieved.project = await getProjectFunction(retrieved.projectId);
+      }
```
asanchezr, line 141: Not to do this now - but this could would benefit from a `Promise.all(...)` approach to fetch the related entities. Right now we wait for each request to complete in sequence. Putting a TODO comment for now would be fine with me

source/frontend/src/features/mapSideBar/acquisition/add/models.ts
```diff
@@ -75,7 +75,7 @@ export class AcquisitionForm implements WithAcquisitionTeam, WithAcquisitionOwne
       acquisitionTypeCode: toTypeCodeNullable(this.acquisitionType),
       regionCode: toTypeCodeNullable(Number(this.region)),
       projectId: isValidId(this.project?.id) ? this.project!.id : null,
-      productId: this.product !== '' ? Number(this.product) : null,
+      productId: isValidId(Number(this.product)) ? Number(this.product) : null,
```
asanchezr, line 78: thanks for doing this!

PR #4586, opened 2025-01-09
PSP-8850 : PIMS performance analysis - DEV

eddherrera:


devinleighsmith:
@eddherrera While I appreciate this PR, the task was primarily about identifying performance issues triaging, and creating follow up stories. During our Friday meeting, please take some time to go through your results for group discussion (apologies if this was already your plan).

source/backend/api/Services/DocumentService.cs
```diff
@@ -421,15 +425,24 @@ public async Task<ExternalResponse<string>> DeleteMayanStorageDocumentAsync(long
...
             {
-                throw GetMayanResponseError(result.Message);
+                ExternalResponse<QueryResponse<Models.Mayan.Document.DocumentTypeModel>> result = await documentStorageRepository.TryGetDocumentTypesAsync(ordering, page, pageSize);
+                if (result.Status != ExternalResponseStatus.Success)
+                {
+                    throw GetMayanResponseError(result.Message);
+                }
+
+                MemoryCacheEntryOptions cacheOptions = new();
+                cacheOptions.SetSlidingExpiration(TimeSpan.FromHours(2));
```
devinleighsmith, line 440: For this change, let's make this time controllable by configuration instead of hardcoding it.

devinleighsmith:
@eddherrera this one still has an outstanding comment.

devinleighsmith (approved):

devinleighsmith (changes requested):
I think, as written, this is what will happen with our existing release process.
1) we start our release GHA
2) the images are tagged and the deployments restart.
3) the mayan document types will be cached.
4) the GHA will run the document type update.

So we will always need to restart the api again to see this change. Does that seem correct?

PR #4591, opened 2025-01-13
PSP-9794 : Acquisition Files - Team Members: Select contact button lo…

eddherrera:
…oks misaligned when chosen Acquisition Team is and organization with more than 1 individual contact attach to it


source/frontend/src/components/Table/Table.scss
```diff
@@ -39,6 +39,19 @@
...
+          word-break: break-all;
+
+          span {
+            display: contents;
+            text-overflow: ellipsis;
+            overflow: hidden;
+            display: -webkit-box;
+            -webkit-line-clamp: 2;
+            line-clamp: 2;
+            -webkit-box-orient: vertical;
```
asanchezr, line 52: did we check this styling? a couple things I see:

- there are 2 declarations of `"display: contents"` and `"display: -webkit-box"` - seems odd
- `-webkit-box-orient: vertical` is deprecated and should not be used. According to https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient

asanchezr (approved):
Please see comments regarding deprecated CSS rules

source/frontend/src/components/Table/Table.scss
eddherrera: Updated, also I don't believe box-orient and webkit-box-orient are the same.

PR #4593, opened 2025-01-14
PSP-9630: Historical File on Header - ability to hide Historical File Numbers when headers is expanded

stairaku:
Additionally to implement the UI/UX solution suggested by Ana, I also added a max-height on the header to avoid the Property Details Header to expand and occupy all the screen height.

Screenshot 2025-01-13 162059


source/frontend/src/components/common/ExpandableTextList.tsx
```diff
@@ -25,12 +26,41 @@ export function ExpandableTextList<T>({
...
   hideText,
 }: IExpandableTextListProps<T>) {
   const [isExpanded, setIsExpanded] = useState(false);
-  const displayedItemsLength = !isExpanded ? maxCollapsedLength ?? items.length : items.length;
+  const [displayedItemsLength, setDisplayedItemsLength] = useState(
+    !isExpanded ? maxCollapsedLength : displayedPerChunk ?? items.length,
+  );
   const displayedItems = items.slice(0, displayedItemsLength);
+
+  console.log(items.length - displayedItemsLength);
```
asanchezr, line 39: please remove console log here

source/frontend/src/components/common/ExpandableTextList.tsx
```diff
@@ -25,12 +26,41 @@ export function ExpandableTextList<T>({
...
   maxCollapsedLength,
+  displayedPerChunk,
   moreText,
   hideText,
 }: IExpandableTextListProps<T>) {
   const [isExpanded, setIsExpanded] = useState(false);
-  const displayedItemsLength = !isExpanded ? maxCollapsedLength ?? items.length : items.length;
+  const [displayedItemsLength, setDisplayedItemsLength] = useState(
+    !isExpanded ? maxCollapsedLength : displayedPerChunk ?? items.length,
+  );
```
asanchezr, line 36: in the old logic `"maxCollapsedLength ?? items.length"` was protecting against maxCollapsedLength being **undefined/null**. In the new logic I don't see the same. Shouldn't we keep that logic in the new one?

stairaku: Yes, will make those changes

asanchezr (approved):

PR #4597, opened 2025-01-15
Psp 9066 - allow multiple payees on acq comp reqs.

devinleighsmith:


source/frontend/vite.config.ts
```diff
@@ -55,7 +55,6 @@ export default defineConfig({
         },
       },
     },
-    pool: 'vmThreads',
```
devinleighsmith, line 58: Note, this seems to be the root cause of the console logs not being visible in the vite test terminal.

devinleighsmith:
Note that there are some follow-up activities for this one, which I will log:
1) The analysis for this story was not finished prior to the start time, and so the desired format of multiple payees in the generated H120 is not known. Currently just appending with ", " but that may not be desired.
2) Analysis was not completed on whether leases need multiple payees (stakeholders) if that is the case, we would like lump lease stakeholders into the payee object instead of having two separate many-to-many relationships.

source/frontend/src/features/mapSideBar/acquisition/models/PayeeOptionModel.ts
```diff
@@ -57,98 +61,56 @@ export class PayeeOption {
...
-        }
+        compReqPayeeModel.acquisitionOwnerId = this.api_id;
         break;
     }
 
-    return compensationModel;
+    return compReqPayeeModel;
   }
 
   private static truncateName(name: string): string {
```
asanchezr, line 116: **nit:** I've seen a couple of duplicates of this function. Perhaps they could be moved to a common location?

source/frontend/src/features/mapSideBar/compensation/detail/CompensationRequisitionDetailContainer.test.tsx
```diff
@@ -40,6 +40,7 @@ vi.mock('@/hooks/repositories/useRequisitionCompensationRepository', () => ({
   useCompensationRequisitionRepository: () => {
     return {
       getCompensationRequisitionProperties: mockGetCompReqPropertiesApi,
+      getCompensationRequisitionPayees: mockGetCompReqPropertiesApi,
```
asanchezr, line 43: getCompensationRequisitionPayees is using the same mocked function as getCompensationRequisitionProperties - please add separate mock for it

source/frontend/src/features/mapSideBar/compensation/detail/CompensationRequisitionDetailView.tsx
```diff
@@ -346,30 +348,30 @@ export const CompensationRequisitionDetailView: React.FunctionComponent<
...
-                <FaExternalLinkAlt className="ml-2" size="1rem" />
-              </StyledLink>
-            )}
-
-            {!payeeDetails?.contactEnabled && <label>{payeeDetails?.displayName ?? ''}</label>}
-            {payeeDetails?.isPaymentInTrust && <label>, in trust</label>}
-          </StyledPayeeDisplayName>
+        <SectionField label="Payee(s)" labelWidth="4">
+          {payeeDetails.map(payeeDetail => (
+            <StyledPayeeDisplayName key={`compenations-payee-${payeeDetail.compReqPayeeId}`}>
```
asanchezr, line 353: nit: `compenations-payee-....` -> `compensation-payee...`

source/frontend/src/features/mapSideBar/compensation/update/UpdateCompensationRequisitionForm.test.tsx
```diff
@@ -17,6 +18,7 @@ import {
   act,
   fakeText,
   fireEvent,
+  prettyDOM,
```
asanchezr, line 21: I don't see this referenced anywhere in the test file - can be safely removed

asanchezr (approved):
See comments

PR #4604, opened 2025-01-20
PSP-9837 : UI UX Clean Up - Manage Tables - Create Buttons - move to the top

eddherrera:


source/frontend/src/components/contact/ContactManagerModal.tsx
```diff
@@ -36,7 +36,6 @@ export const ContactManagerModal: React.FunctionComponent<
           <ContactManagerView
             setSelectedRows={props.setSelectedRows}
             selectedRows={props.selectedRows}
-            showAddButton={props.showAddButton}
```
asanchezr, line 39: **showAddButton** should be removed from ContactManagerModal props as it is just passing it down to the view and the view no longer needs it

asanchezr (approved):
see comments

source/frontend/src/components/contact/ContactManagerModal.tsx
eddherrera: removed

PR #4605, opened 2025-01-21
Use named hangfire server and queue - to assist with debugging.

devinleighsmith:


devinleighsmith:
Currently the server name is randomly generated and the queue name is always "default" this made debugging an issue with an errant replicationcontroller difficult.

asanchezr (approved):

PR #4611, opened 2025-01-23
Psp 9893 mayan document queue corrections.

devinleighsmith:


source/backend/api/Services/DocumentQueueService.cs
```diff
@@ -217,14 +217,6 @@ public async Task<PimsDocumentQueue> Upload(PimsDocumentQueue documentQueue)
...
 
-            // if the document queued for upload is already in an error state, update the retries.
-            if (databaseDocumentQueue.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.PIMS_ERROR.ToString() || databaseDocumentQueue.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.MAYAN_ERROR.ToString())
-            {
-                this.Logger.LogDebug("Document Queue {documentQueueId}, previously errored, retrying", documentQueue.DocumentQueueId);
-                databaseDocumentQueue.DocProcessRetries += 1;
-                databaseDocumentQueue.DocProcessEndDt = null;
-            }
-
             bool isValid = ValidateQueuedDocument(databaseDocumentQueue, documentQueue);
```
devinleighsmith, line 220: the above logic needs to run after this check - to avoid the processDocRetries from being changed.

source/backend/api/Services/DocumentQueueService.cs
```diff
@@ -234,6 +226,14 @@ public async Task<PimsDocumentQueue> Upload(PimsDocumentQueue documentQueue)
                 return databaseDocumentQueue;
             }
 
+            // if the document queued for upload is already in an error state, update the retries.
+            if (databaseDocumentQueue.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.PIMS_ERROR.ToString() || databaseDocumentQueue.DocumentQueueStatusTypeCode == DocumentQueueStatusTypes.MAYAN_ERROR.ToString())
+            {
+                this.Logger.LogDebug("Document Queue {documentQueueId}, previously errored, retrying", documentQueue.DocumentQueueId);
+                databaseDocumentQueue.DocProcessRetries = ++databaseDocumentQueue.DocProcessRetries ?? 1;
```
devinleighsmith, line 233: using += does not work on a null value.

source/backend/api/Services/DocumentQueueService.cs
```diff
@@ -243,12 +243,6 @@ public async Task<PimsDocumentQueue> Upload(PimsDocumentQueue documentQueue)
                 UpdateDocumentQueueStatus(databaseDocumentQueue, DocumentQueueStatusTypes.PIMS_ERROR);
                 return databaseDocumentQueue;
             }
-            else if (relatedDocument?.MayanId != null && relatedDocument?.MayanId > 0)
```
devinleighsmith, line 246: This is incorrect, in addition to having a mayanId, we need to wait for the filedetail to be available.

asanchezr (approved):

asanchezr:
@devinleighsmith - this doesn't look right...
image

PR #4616, opened 2025-01-25
PSP-9762: Update and creation of new automated test cases

stairaku:


testing/PIMS.Tests.Automation/Classes/AcquisitionFile.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace PIMS.Tests.Automation.Classes
+﻿using System.Drawing;
+
```
asanchezr, line 2: I usually just approve these PRs but I have a question - do you really need **System.Drawing** here?
It seems like this dependency might be added by accident

asanchezr (approved):
See comments

testing/PIMS.Tests.Automation/Classes/AcquisitionFile.cs
stairaku: No, will delete it for the next PR. Thanks for catching that up

PR #4621, opened 2025-01-30
PSP-9905 Error thrown when opening compensation requisition with legacy payee/null payee

asanchezr:
Fixed various null-pointer exceptions and faulty logic when showing/editing compensation requisitions due to having all payee keys set as null. 
Added unit tests to cover those code paths


source/frontend/src/features/mapSideBar/acquisition/models/PayeeOptionModel.ts
```diff
@@ -70,18 +70,20 @@ export class PayeeOption {
...
+    if (isValidId(compReqPayee.acquisitionOwnerId)) {
       payee = PayeeOption.createOwner(compReqPayee.acquisitionOwner, compReqPayee);
-    } else if (compReqPayee.acquisitionFileTeamId) {
+    } else if (isValidId(compReqPayee.acquisitionFileTeamId)) {
       payee = PayeeOption.createTeamMember(compReqPayee.acquisitionFileTeam, compReqPayee);
-    } else if (compReqPayee.interestHolderId) {
+    } else if (isValidId(compReqPayee.interestHolderId)) {
       payee = PayeeOption.createInterestHolder(compReqPayee.interestHolder, compReqPayee);
     }
-    payee.rowVersion = compReqPayee.rowVersion;
```
asanchezr, line 80: this was one of the sources of NPE - when all payees are null then this variable is null so it cannot take an assignment of "rowVersion"

source/frontend/src/features/mapSideBar/compensation/update/UpdateCompensationRequisitionForm.tsx
```diff
@@ -264,7 +264,9 @@ const UpdateCompensationRequisitionForm: React.FC<CompensationRequisitionFormPro
                       selectFunction={(optionPayees, selectedPayees) =>
                         optionPayees.filter(payee =>
                           selectedPayees?.find(
-                            selectedPayee => selectedPayee.api_id === payee.api_id,
```
asanchezr, line 267: The old logic was selecting some payees when they shouldn't. If you had an owner and an interest holder that had the same id then the multiselect would show **both** when one of them was selected. This happens because the same Id number (ie 1) can be used in different tables that feed into the payees multi-select.

Fixed by also checking for the type of payee (Owner, InterestHolder, etc) in addition to the entity key

source/frontend/src/features/mapSideBar/compensation/detail/CompensationRequisitionDetailView.tsx
```diff
@@ -84,28 +72,34 @@ export const CompensationRequisitionDetailView: React.FunctionComponent<
...
+            currentPayeeDetail = PayeeDetail.createFromPerson(
+              currentPayee.acquisitionFileTeam.person,
+            );
+          } else if (isValidId(currentPayee?.acquisitionFileTeam?.organizationId)) {
+            currentPayeeDetail = PayeeDetail.createFromOrganization(
+              currentPayee.acquisitionFileTeam.organization,
+            );
           }
         }
         currentPayeeDetail.compReqPayeeId = currentPayee.compReqPayeeId;
```
asanchezr, line 105: Another source of NPE if all payees where null

source/frontend/src/features/mapSideBar/compensation/models/CompensationRequisitionFormModel.ts
```diff
@@ -58,7 +58,10 @@ export class CompensationRequisitionFormModel {
 
   toApi(): ApiGen_Concepts_CompensationRequisition {
     const compReqPayees =
-      this.payees?.map<ApiGen_Concepts_CompReqPayee>(formPayee => formPayee.toApi()) ?? [];
+      this.payees
+        ?.map<ApiGen_Concepts_CompReqPayee>(formPayee => formPayee.toApi())
+        ?.filter(exists) ?? [];
+
```
asanchezr, line 64: We don't want to save `CompReqPayee` where all the payees are null - so we filter those records before calling the server

source/frontend/src/features/mapSideBar/compensation/update/UpdateCompensationRequisitionContainer.test.tsx
```diff
@@ -16,79 +20,54 @@ import {
...
-    return {
-      getAcquisitionOwners: getMockRepositoryObj(),
-      getAcquisitionFileSolicitors: {
-        execute: vi.fn(),
-        loading: false,
-      },
-      getAcquisitionFileRepresentatives: {
-        execute: vi.fn(),
-        loading: false,
-      },
```
asanchezr, line 45: These methods no longer exist - needed some cleanup

devinleighsmith (approved):
confirmed locally - approved.

PR #4628, opened 2025-01-31
opinionated cleanup of dead code and duplicated style code.

devinleighsmith:


source/frontend/src/features/disposition/list/styles.ts
```diff
@@ -1,47 +0,0 @@
-import { Col } from 'react-bootstrap';
-import styled from 'styled-components';
-
-import { InlineForm, InlineInput } from '@/components/common/form/styles';
-import { Scrollable as ScrollableBase } from '@/components/common/Scrollable/Scrollable';
-
-export const ListPage = styled.div`
```
devinleighsmith, line 7: This is the most opinionated part of this change - we essentially had this exact file copied to every list view in the system. I've removed these duplicates and replaced with a single set of common styles for list views.

asanchezr (approved):

FuriousLlama (approved):
Looks great. Thank you for doing this cleanup, less code with same functionality is awesome

PR #4642, opened 2025-02-06
PSP-9891 : L/L File - Ability to Add a Renewal Commencement Date Prio…

eddherrera:
…r to Expiry Date of Previous  Term


source/backend/tests/unit/api/Services/LeaseServiceTest.cs
```diff
@@ -15,6 +15,8 @@
 using Pims.Dal.Repositories;
 using Pims.Core.Security;
 using Xunit;
+using static Pims.Dal.Entities.PimsLeaseStatusType;
+using DocumentFormat.OpenXml.Vml.Spreadsheet;
```
asanchezr, line 19: couple questions:
* DocumentFormat.OpenXml.Vml.Spreadsheet seems like we wouldn't need that in tests...
* First time I see "using static" - had to look it up. Not opposed to it but would like we discuss as a team pros vs cons

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
```diff
@@ -684,6 +686,100 @@ public void UpdateProperties_MatchProperties_Success()
...
+            // Arrange
+            var lease = EntityHelper.CreateLease(1);
+            lease.OrigExpiryDate = new DateTime(2025, 1, 31);
+            lease.LeaseStatusTypeCode = PimsLeaseStatusTypes.ACTIVE;
+            lease.LeaseStatusTypeCodeNavigation = new PimsLeaseStatusType()
+            {
+                Id = PimsLeaseStatusTypes.ACTIVE,
+            };
+            lease.OrigStartDate = new DateTime(2024, 1, 1);
+            lease.OrigExpiryDate = new DateTime(2025, 1, 31);
```
asanchezr, line 701: `lease.OrigExpiryDate` was already given a value in line 694. Please remove one of them - no need to have this assignment duplicated

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
```diff
@@ -684,6 +686,100 @@ public void UpdateProperties_MatchProperties_Success()
...
```
asanchezr, line 748: same comment about assigning "OrigExpiryDate" twice

asanchezr (approved):
See comments

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
eddherrera: removed

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
eddherrera: removed

source/backend/tests/unit/api/Services/LeaseServiceTest.cs
eddherrera: removed

PR #4644, opened 2025-02-08
Psp 9824 changes to final property states.

devinleighsmith:


source/frontend/src/features/mapSideBar/compensation/list/CompensationListView.tsx
```diff
@@ -119,7 +100,7 @@ export const CompensationListView: React.FunctionComponent<ICompensationListView
             title="Compensation Requisitions"
             addButtonText="Add Requisition"
             addButtonIcon={<FaPlus size={'2rem'} />}
-            onAdd={onAdd}
+            onAdd={statusUpdateSolver?.canEditOrDeleteCompensation() ? onAdd : null}
```
asanchezr, line 103: on other components we show a tooltip that they cannot "Add" but I don't see it here. Just confirming this is expected or whether the tooltip should be displayed here...

source/frontend/src/features/mapSideBar/disposition/tabs/fileDetails/detail/DispositionStatusUpdateSolver.ts
```diff
@@ -1,7 +1,8 @@
+import { IUpdateChecklistStrategy } from '@/features/mapSideBar/compensation/models/IUpdateChecklistStrategy';
```
asanchezr, line 1: **nit:** this file doesn't seem exclusive to compensations - would prefer to see this moved to a common folder instead of being under compensations folder

asanchezr (approved):
See comments

PR #4647, opened 2025-02-11
psp-9813 ensure error messages from backend display properly.

devinleighsmith:


devinleighsmith:
Note, here is the corresponding logic that prevents those details from displaying:

`        /// <summary>
        /// Creates a new instance of an ErrorResponseModel object, initializes with specified arguments.
        /// </summary>
        /// <param name="environment"></param>
        /// <param name="ex"></param>
        /// <param name="message"></param>
        /// <param name="details"></param>
        /// <param name="errorCode"></param>
        public ErrorResponseModel(IWebHostEnvironment environment, Exception ex, string message = null, string details = null, string errorCode = null)
        {
            var showError = !environment.IsProduction();
            this.Error = showError ? ex.Message : message;
            this.Type = ex.GetType().Name;
            this.Details = showError ? details ?? ex.GetAllMessages() : null;
            this.StackTrace = showError ? ex.StackTrace : null;
            this.ErrorCode = errorCode;
        }`

note that showError is always false in production or production-like environments.

eddherrera (approved):

asanchezr (approved):

PR #4648, opened 2025-02-11
Psp 9940 - correct issue with map display when jumping to boundaries

devinleighsmith:


source/frontend/src/components/common/mapFSM/machineDefinition/mapMachine.ts
```diff
@@ -168,9 +168,7 @@ const mapRequestStates = {
                 context?.mapFeatureData?.pimsLocationFeatures?.features ?? [];
 
               // business logic, if there are file properties, use those, otherwise, zoom to a single feature if there is only one, or all features if there are more than one.
-              if (context.filePropertyLocations.length > 0) {
```
devinleighsmith, line 171: unfortunately this logic doesn't work - as this method has no context who called it. As a result, efforts to send a fileProperties of [], result in the else clause of this function to be executed.

suggested solution - keep the file boundary logic separate from the map load/ map search logic.

asanchezr (approved):

PR #4661, opened 2025-02-18
psp-10029 use debounce, memo to limit renders of clusters/markers. Update cluster library.

devinleighsmith:


source/frontend/src/components/maps/hooks/useSupercluster.ts
```diff
@@ -46,7 +56,7 @@ const useSupercluster = <
      * greater than 100 (avoid perfomrance issues when rendering all points)
      */
     if (bounds && (options?.enableClustering || points.length > 100)) {
-      setClusters(superclusterRef.current.getClusters(bounds, zoomInt));
```
devinleighsmith, line 49: Due to useEffects and other prop changes, this method would be called 3-4 times per user interaction prior to this change.

source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -252,122 +259,135 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
-                    className: `marker-cluster marker-cluster-${sizeClass}`,
-                    iconSize: [40, 40],
-                  })
-                }
-              />
-            );
-          } else {
-            const clusterFeature = cluster as PointFeature<
+  const mapMarkerClickFn = mapMachine.mapMarkerClick;
+  const renderedPoints = useMemo(() => {
```
devinleighsmith, line 263: In normal situations, react handles the prevention of unnecessary renders to the render tree. However, it may not be able to do so here due to interactions with leaflet - as a result, this render method was greatly slowing down the overall map render time.

devinleighsmith: (without this memo, this render method is called 4-6 times per user action)

source/frontend/package.json
```diff
@@ -51,7 +51,7 @@
     "redux-thunk": "2.3.0",
     "retry-axios": "2.4.0",
     "styled-components": "5.3.6",
-    "supercluster": "7.1.3",
+    "supercluster": "8.0.1",
```
devinleighsmith, line 54: tested, has better getcluster performance.

FuriousLlama (approved):
glad you had the time to look into this! thank you

asanchezr (approved):

PR #4666, opened 2025-02-20
psp-9528 double click to zoom the map.

devinleighsmith:


FuriousLlama (approved):

source/frontend/src/features/properties/map/MapContainer.test.tsx
```diff
@@ -242,15 +247,18 @@ describe('MapContainer', () => {
     history.push('/mapview');
   });
 
-  afterEach(cleanup);
+  afterEach(() => {
+    vi.useRealTimers();
+    cleanup;
```
asanchezr, line 252: you need to call the function "cleanup()" vs "cleanup"

source/frontend/src/features/properties/map/MapContainer.test.tsx
```diff
@@ -262,13 +270,7 @@ describe('MapContainer', () => {
     expect(basemapToggle).toHaveAttribute('src', '/streets.jpg');
   });
 
-  it('shows the current map scale', async () => {
-    await setup();
-    expect(document.querySelector('.leaflet-control-scale')).toBeVisible();
-    expect(document.querySelector('.leaflet-control-scale-line')).toHaveTextContent(/100 km/i);
-  });
-
```
asanchezr, line 270: Why did we remove this test? was it moved to another file? I think it is asserting that the new scale widget is present and showing proper information

asanchezr (approved):

source/frontend/src/features/properties/map/MapContainer.test.tsx
devinleighsmith: merge issue, will correct.

PR #4673, opened 2025-03-03
Lease and License updates

FuriousLlama:
image


source/frontend/src/features/leases/detail/LeasePages/details/PropertyInformation.tsx
```diff
@@ -57,10 +44,7 @@ export const PropertyInformation: React.FunctionComponent<
       </SectionField>
       {!hideAddress ? (
         <SectionField label="Address" labelWidth="3" className="py-2">
-          <AddressSubForm
-            nameSpace={withNameSpace(nameSpace, 'property.address')}
-            disabled={disabled}
-          />
+          <>{property.property.address.streetAddress3}</>
```
devinleighsmith, line 47: This doesn't seem right, the form used to display the address subform with a number of address fields but now we only display streetAddress3?

source/frontend/src/features/mapSideBar/shared/fileMenuView.tsx
devinleighsmith, line 1: Nit, this file does not match our naming conventions for tsx files.

source/frontend/src/features/mapSideBar/shared/sidebarPathSolver.ts
```diff
@@ -0,0 +1,152 @@
...
+  showFilePropertyDetail: (
+    fileType: string,
+    fileId: number,
+    filePropertyId: number,
+    detailType: string,
+    replace: boolean,
+  ) => void;
+}
+
+export type IPathSolver = () => IPathSolverMethods;
```
devinleighsmith, line 23: I like this, it is nice to lay out these common *file routing patterns in a shared location.
Perhaps we should avoid the *solver syntax - as so far that has been used to denote a file that calculates authorization.

devinleighsmith:
The title of the lease details appears to have changed:
new
image
old
image

source/frontend/src/features/mapSideBar/shared/sidebarPathSolver.ts
FuriousLlama: thats fair. Thoughts on a better name?

FuriousLlama:
> The title of the lease details appears to have changed: new image old image

Yes, this is on purpose. The lease business mentioned that the title was different and Original Agreement is the one they wanted.

devinleighsmith:
@FuriousLlama I retested this now that you fixed the issue with the service - however, when one of the business rules on the property fails, no error is displayed on the frontend (actually, there is no visual indication of anything occurring at all).

source/frontend/src/features/mapSideBar/shared/sidebarPathSolver.ts
devinleighsmith: maybe sidebarPathGenerator?

source/frontend/src/features/leases/detail/LeasePages/details/PropertyInformation.tsx
FuriousLlama: Good catch, I was testing the address and stashed the view. I've updated it now to use the new address view

source/backend/api/Services/LeaseService.cs
```diff
@@ -278,8 +278,52 @@ public PimsLease Update(PimsLease lease, IEnumerable<UserOverrideCode> userOverr
             ValidateNewTotalAllowableCompensation(lease.LeaseId, lease.TotalAllowableCompensation);
 
             _leaseRepository.Update(lease, false);
+
+            _leaseRepository.UpdateLeaseRenewals(lease.Internal_Id, lease.ConcurrencyControlNumber, lease.PimsLeaseRenewals);
+
+            List<PimsPropertyLease> differenceSet = currentFileProperties.Where(x => !lease.PimsPropertyLeases.Any(y => y.Internal_Id == x.Internal_Id)).ToList();
```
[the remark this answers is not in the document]
FuriousLlama: That is correct. I think this comment is on an older version.

FuriousLlama:
> @FuriousLlama I retested this now that you fixed the issue with the service - however, when one of the business rules on the property fails, no error is displayed on the frontend (actually, there is no visual indication of anything occurring at all).

I have updated the logic on the frontend to handle the errors from the backend. Please give it another try.

devinleighsmith:
> > @FuriousLlama I retested this now that you fixed the issue with the service - however, when one of the business rules on the property fails, no error is displayed on the frontend (actually, there is no visual indication of anything occurring at all).
> 
> I have updated the logic on the frontend to handle the errors from the backend. Please give it another try.

@FuriousLlama something strange going on, getting a compilation error from the LeaseContainer.tsx

FuriousLlama:
> > > @FuriousLlama I retested this now that you fixed the issue with the service - however, when one of the business rules on the property fails, no error is displayed on the frontend (actually, there is no visual indication of anything occurring at all).
> > 
> > 
> > I have updated the logic on the frontend to handle the errors from the backend. Please give it another try.
> 
> @FuriousLlama something strange going on, getting a compilation error from the LeaseContainer.tsx

The case was off and git didn't detect the change. Should work now

devinleighsmith (approved):

PR #4677, opened 2025-03-05
PSP-9843 : SPIKE: UI UX Clean Up - Remove the success (green) toasts

eddherrera:


devinleighsmith (changes requested):
Please do a code search for toast.success and see if any of those messages should be removed as well.

eddherrera:
> Please do a code search for toast.success and see if any of those messages should be removed as well.

updated

devinleighsmith (approved):

PR #4688, opened 2025-03-10
psp-10148 | Lease Deposit display fixes

FuriousLlama:


devinleighsmith:
This PR seems to be logged against the wrong jira - this is the description of 10148 [HOTFIX]-FT:Reg:Lease & Licence: User is not able to update Receivable lease files

devinleighsmith:
This PR seems to be mostly display changes, is this really a hotfix?

source/frontend/src/features/notes/list/NoteListView.tsx
```diff
@@ -93,70 +92,66 @@ export const NoteListView: React.FunctionComponent<React.PropsWithChildren<INote
   };
 
   return (
-    <ListPage>
```
devinleighsmith, line 96: Hmm, is this related to the other deposit changes? Is there another jira issue that this is in reference to?

devinleighsmith:
Perhaps unrelated to these changes, but this would be an argument for making this a hotfix:

When I fill out all required fields on a deposit return, but leave the rest blank, I get this error:
image

Would you mind fixing that here? Happy to treat as a hotfix with that correction included.

source/frontend/src/features/notes/list/NoteListView.tsx
FuriousLlama: Not an issue with deposits. It's a padding issue with the notes in all screens that use the widget. Figure it would be ok to piggy back it with the UI changes.

devinleighsmith (approved):

PR #4694, opened 2025-03-12
Psp 10010 - placeholder endpoint for LTSA integration

devinleighsmith:


source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
```diff
@@ -66,7 +66,7 @@ public async Task<IActionResult> GetSupportedDocumentTypes()
         [ProducesResponseType(typeof(ExternalResponse<string>), 200)]
         [SwaggerOperation(Tags = new[] { "document-generation" })]
         [TypeFilter(typeof(NullJsonResultFilter))]
-        public async Task<IActionResult> UploadTemplate([FromForm] IFormFile file)
+        public async Task<IActionResult> UploadTemplate(IFormFile file)
```
devinleighsmith, line 69: recommended by swagger - currently blocking swagger from displaying

source/backend/api/Areas/Reports/Controllers/LeaseController.cs
```diff
@@ -209,7 +209,7 @@ public IActionResult ExportLeasePayments(int fiscalYearStart)
         /// <param name="filter"></param>
         /// <param name="all"></param>
         /// <returns></returns>
-        public IEnumerable<LeaseModel> GetCrossJoinLeases(Lease.Models.Search.LeaseFilterModel filter, bool all = false)
+        private IEnumerable<LeaseModel> GetCrossJoinLeases(Lease.Models.Search.LeaseFilterModel filter, bool all = false)
```
devinleighsmith, line 212: currently blocking swagger compilation.

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -130,6 +141,59 @@ public async Task<IActionResult> PostLtsaFields(string pid)
...
+                "Request received by Controller: {Controller}, Action: {ControllerAction}, User: {User}, DateTime: {DateTime}",
+                nameof(DispositionController),
+                nameof(LtsaController),
+                _user.GetUsername(),
+                DateTime.Now);
+
+            _logger.LogDebug(ownershipFile.Serialize());
+
+            // TODO: placeholder endpoint to facilitate LTSA integration.
+            return new OkResult();
```
devinleighsmith, line 168: For now I just want an endpoint that accepts the correct data, and authenticates/authorizes an LTSA service account. Logic can wait until they have confirmed this format works for them.

source/backend/api/Areas/DocumentGeneration/DocumentGenerationController.cs
FuriousLlama: our swagger needs a lot of help

source/backend/core/Security/Permissions.cs
```diff
@@ -200,5 +200,17 @@ public enum Permissions
 
         [Display(GroupName = "disposition", Name = "disposition-delete", Description = "Ability to delete Disposition Files.")]
         DispositionDelete = 65,
+
+        [Display(GroupName = "ltsaOwnership", Name = "ltsa-ownership-view", Description = "Ability to view LTSA ownership information.")]
```
FuriousLlama, line 204: q: why LTSA ownership vs LTSA integration or LTSA Service?

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -130,6 +141,59 @@ public async Task<IActionResult> PostLtsaFields(string pid)
...
+        /// </summary>
+        /// <param name="ownershipFile">the file containing the list of pids.</param>
+        /// <returns>The orders created within LTSA.</returns>
+        [HttpPut("ownership")]
+        [Produces("application/json")]
+        [ProducesResponseType(200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "tools-ltsa" })]
+        [HasPermission(Permissions.LtsaOwnershipEdit)]
+        public async Task<IActionResult> PutBctfaOwnership(IFormFile ownershipFile)
```
FuriousLlama, line 156: did they say anything about being able to do a call with JSON instead of a file?

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
```diff
@@ -130,6 +141,59 @@ public async Task<IActionResult> PostLtsaFields(string pid)
...
+        /// <param name="ownershipData">the json structure containing the list of pids.</param>
+        /// <returns>The orders created within LTSA.</returns>
+        [HttpPut("ownership/list")]
+        [Consumes("application/json")]
+        [Produces("application/json")]
+        [ProducesResponseType(200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "tools-ltsa" })]
+        [HasPermission(Permissions.LtsaOwnershipEdit)]
+        public async Task<IActionResult> PutBctfaOwnershipJson([FromBody] Model.BctfaOwnershipList ownershipData)
```
FuriousLlama, line 183: Oh, I think I get it. One is for the JSON version of it

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
devinleighsmith: Not yet, this is the first time they would offer what they call a "data product" as a direct rest api call. In the past they have only supported SFTP uploads.

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
devinleighsmith: Yes, I wanted to offer both, since they haven't decided upon an implementation on their end yet.

source/backend/api/Areas/Tools/Controllers/LtsaController.cs
asanchezr: My only comment here would be that we are collocating "two concerns" of the application/API. One concerns deals with "us" talking to their (LTSA) service, and the other concern (the new code added) is allowing "them" to talk to us.
I think I would prefer to separate the two concerns into their own controllers + services. The reason for me is that it looks simple now but it the future this controller might grow and the "in" functionality will get mixed with the "out" functions and will be hard to determine which is which.

I understand it is all "about" LTSA but generically this endpoint could be called by "anyone" with the right claims. What do you think? I'd prefer something like a new controller **LtsaDataIngestionController**  (not set on naming, can be anything else)

source/database/mssql/scripts/dbscripts/PSP_PIMS_LATEST/Alter Up/76_DML_PIMS_CLAIMS.sql
```diff
@@ -0,0 +1,30 @@
...
+DECLARE @dbClaimsUserGuid uniqueidentifier = null;
+DECLARE @appClaimsUserId nvarchar(30) = N'SEED';
+DECLARE @appClaimsUserGuid uniqueidentifier = '1b93f614-91da-4b32-b36e-bd2c6ebd12e2';
+DECLARE @appClaimsUserDirectory nvarchar(30) = N'';
+DECLARE @seedClaimsTime DateTime = GETDATE();
+
+INSERT [dbo].[PIMS_CLAIM]
+    ([CLAIM_ID], [CLAIM_UID], [NAME], [KEYCLOAK_ROLE_ID], [DESCRIPTION], [IS_DISABLED], [CONCURRENCY_CONTROL_NUMBER], [APP_CREATE_TIMESTAMP], [APP_CREATE_USERID], [APP_CREATE_USER_GUID], [APP_CREATE_USER_DIRECTORY], [DB_CREATE_TIMESTAMP], [DB_CREATE_USERID], [APP_LAST_UPDATE_TIMESTAMP], [APP_LAST_UPDATE_USERID], [APP_LAST_UPDATE_USER_GUID], [APP_LAST_UPDATE_USER_DIRECTORY], [DB_LAST_UPDATE_TIMESTAMP], [DB_LAST_UPDATE_USERID])
+VALUES
+    (65, N'db03a38d-4eb4-41b1-b4f4-31e3355453f7', N'ltsa-ownership-view', N'db03a38d-4eb4-41b1-b4f4-31e3355453f7', N'Ability to view LTSA ownership information.', 0, 1, @seedClaimsTime, @appClaimsUserId, @appClaimsUserGuid, @appClaimsUserDirectory, @seedClaimsTime, @dbClaimsUserId, @seedClaimsTime, @appClaimsUserId, @appClaimsUserGuid, @appClaimsUserDirectory, @seedClaimsTime, @dbClaimsUserId);
```
asanchezr, line 13: My comment for this (and the other SQL files) is that we should make sure that Doug does not override these in upcoming DB schemas. We've seen this happen in the past.

asanchezr (approved):

PR #4696, opened 2025-03-12
HOTIFX: psp-10149 ensure that user overrides are displayed by L&L property ed…

devinleighsmith:
…its.

### DIFF
--- a/source/frontend/src/features/leases/shared/propertyPicker/LeaseUpdatePropertySelector.tsx
+++ b/source/frontend/src/features/leases/shared/propertyPicker/LeaseUpdatePropertySelector.tsx
@@ -45,7 +45,6 @@ export const LeaseUpdatePropertySelector: React.FunctionComponent<
   LeaseUpdatePropertySelectorProp
 > = ({ lease }) => {
   const [showSaveConfirmModal, setShowSaveConfirmModal] = useState<boolean>(false);
-  const [showAssociatedEntityWarning, setShowAssociatedEntityWarning] = useState<boolean>(false);
 
   const formikRef = useRef<FormikProps<LeaseFormModel>>(null);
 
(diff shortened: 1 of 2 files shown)


source/frontend/src/features/leases/shared/propertyPicker/LeaseUpdatePropertySelector.tsx
```diff
@@ -258,42 +265,40 @@ export const LeaseUpdatePropertySelector: React.FunctionComponent<
...
+            }
+            formikRef.current?.resetForm();
+            await getCompleteLease();
+            mapMachine.refreshMapProperties();
+            pathSolver.showFile('lease', lease.id);
+          }
+        });
+      },
+      [],
+      (axiosError: AxiosError<IApiError>) => {
```
asanchezr, line 287: question: the old logic would show a custom modal with text: 
> This property can not be removed from the file. This property is related to one or more entities in the file, only properties that are not linked to any entities in the file can be removed.

But the hotfix removes the only call to `setShowAssociatedEntityWarning(true);` - shouldn't we remove the custom modal as well? Also how is this change still supporting that flow? Are doing the check in the backend?

FuriousLlama (approved):

source/frontend/src/features/leases/shared/propertyPicker/LeaseUpdatePropertySelector.tsx
devinleighsmith: Removed the modal. This error appears to be legacy and is only thrown in one area of the backend - when a payee is removed from a lease. Also, this error was not working before as the axiosError.code would never be 409.

asanchezr (approved):

PR #4697, opened 2025-03-12
HOTFIX psp-10139 | Lease header historical file fix

FuriousLlama:


devinleighsmith (changes requested):
I get an infinite render loop if I try and edit a lease with these changes.

FuriousLlama:
> I get an infinite render loop if I try and edit a lease with these changes.

Updated the hook. please re-test

devinleighsmith:
@FuriousLlama when saving a property, I see the lease and all sub-entities being requested 4 times now. I checked this against TEST, which is requesting the lease information twice for the same workflow.

Ideal case would be to only request the lease once after an update. However, since that issue existed prior to this hotfix, I'd be fine if you were able to fix this such that only 2 requests are made to match current state.

devinleighsmith:
As discussed, we are re-classifying the bug this PR relates to as not a hotfix. @FuriousLlama , please make sure you have a story or task to get the other half of your lease changes merged, so we can track that work.

PR #4711, opened 2025-03-21
Psp 10010 - service/repository implementation.

devinleighsmith:


source/backend/dal/PIMSContextFactory.cs
```diff
@@ -1,104 +0,0 @@
...
-using System.Data.SqlClient;
-using System.Diagnostics.CodeAnalysis;
-using System.IO;
-using System.Text.Json;
-using System.Text.Json.Serialization;
-using Microsoft.EntityFrameworkCore;
-using Microsoft.EntityFrameworkCore.Design;
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Logging;
-using Pims.Core.Extensions;
```
asanchezr, line 11: Just checking that this file is safe to delete...

source/backend/dal/Repositories/BasePimsRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Pims.Dal.Repositories
+{
+    /// <summary>
+    /// BaseService abstract class, provides a generic service layer to perform CRUD operations on the datasource.
+    /// </summary>
+    public abstract class BasePimsRepository : IRepository
+    {
+        #region Variables
```
asanchezr, line 12: I thought we already had a "base" repository class to provide common functionality. Why create a new one? If more common functions are needed, why not implement them in the existing base class?

image

source/backend/entities/PimsBaseContext.cs
```diff
@@ -9686,6 +9787,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             .HasMin(1L)
             .HasMax(2147483647L);
         modelBuilder.HasSequence("PIMS_COMP_REQ_ACQ_PAYEE_ID_SEQ")
+            .StartsAt(60L)
```
asanchezr, line 9790: this doesn't seem right...

source/backend/entities/PimsBaseContext.cs
```diff
@@ -9698,6 +9800,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             .HasMin(1L)
             .HasMax(2147483647L);
         modelBuilder.HasSequence("PIMS_COMP_REQ_LEASE_PAYEE_ID_SEQ")
+            .StartsAt(52L)
```
asanchezr, line 9803: this doesn't seem right...

source/backend/api/Services/BctfaOwnershipService.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            try
+            {
+                var intArray = values.Select(int.Parse).ToArray();
+                return intArray;
+            }
+            catch (FormatException)
+            {
+                throw new InvalidDataException("File contains non-integer values.");
+            }
```
asanchezr, line 63: I haven't seen the format of this CSV file but if I'm reading this logic correctly, it:

- splits the file content into **multiple** lines and stores it into the `lines` variable (line 46)
- then, it takes **only the first line** and parses the comma-separated values from that single line and that is what we return

Is that correct? Wouldn't the provided CSV have multiple rows/lines of comma-separated values or will it always be a single row/line?

devinleighsmith: The structure of a csv should be one row for a unique "column" in a database, so a database export of a single column of a database should be a single row of comma separated values.

However, given that there is some risk of csv being sent in with different formatting, I'll change this to be more generic in how it would handle a incorrectly formatted csv.

asanchezr: I asked chatgpt how a CSV of a one column should look like and the answer is consistent with my how I was imagining. But I see you modified the code to cover for the case of a file with single values each in it's own line so I think we are good

image

source/backend/dal/Repositories/PmbcBctfaPidRepository.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                this.Context.PmbcBctfaPids.AddRange(pids);
+            }
+        }
+
+        public void UpdateRange(IEnumerable<PmbcBctfaPid> pids)
+        {
+            if (pids != null && pids.Any())
+            {
+                this.Context.UpdateRange(pids);
```
asanchezr, line 49: **nit:** I'm sure this works fine but in line 41 we are explicit on the table we want to affect change, ie `Context.PmbcBctfaPids.AddRange(...)`

Perhaps we should follow that pattern here: `Context.PmbcBctfaPids.UpdateRange(...)`

asanchezr (approved):

PR #4713, opened 2025-03-21
PSP-7828 : PIMS - Session Time Out Setting 

eddherrera:


asanchezr (approved):

source/frontend/src/tenants/config/defaultTenant.ts
```diff
@@ -88,7 +88,7 @@ export const defaultTenant: ITenantConfig2 = {
     },
   },
   idlePromptTimeout: 15,
-  idleTimeout: 15,
+  idleTimeout: 60,
```
devinleighsmith, line 91: Taking the story literally, the timeout should be set for 60m:

Allow users to remain logged into the application interface despite being idle for 60 minutes.

However, our logic kicks a user out after both the idleprompttimeout and idletimeout have elapsed, with your change a total of 75 minutes.

eddherrera: Updated

PR #4715, opened 2025-03-22
PSP-10016 Implement browser telemetry

asanchezr:
image

image


source/frontend/vite.config.ts
```diff
@@ -70,7 +70,7 @@ export default defineConfig({
         "base-uri 'self'; \
          default-src 'self'; \
          script-src 'self' blob: 'sha256-8ZgGo/nOlaDknQkDUYiedLuFRSGJwIz6LAzsOrNxhmU='; \
-         connect-src 'self' https://www.arcgis.com/sharing/rest/ https://tiles.arcgis.com/ https://maps.gov.bc.ca/arcgis/rest/ https://server.arcgisonline.com/ArcGIS/rest/ https://dev.loginproxy.gov.bc.ca/ https://dev-pims.th.gov.bc.ca/api/ https://openmaps.gov.bc.ca/ https://delivery.apps.gov.bc.ca/ https://apps.gov.bc.ca/ https://maps.th.gov.bc.ca/ https://gov.bc.ca/ https://dev-maps.th.gov.bc.ca/; \
+         connect-src 'self' http://localhost:*/ https://collector-3cd915-dev.apps.silver.devops.gov.bc.ca/ https://www.arcgis.com/sharing/rest/ https://tiles.arcgis.com/ https://maps.gov.bc.ca/arcgis/rest/ https://server.arcgisonline.com/ArcGIS/rest/ https://dev.loginproxy.gov.bc.ca/ https://dev-pims.th.gov.bc.ca/api/ https://openmaps.gov.bc.ca/ https://delivery.apps.gov.bc.ca/ https://apps.gov.bc.ca/ https://maps.th.gov.bc.ca/ https://gov.bc.ca/ https://dev-maps.th.gov.bc.ca/; \
```
asanchezr, line 73: We need to add **localhost** to be able to access all the monitoring containers (open-telemetry-collector, jaeger) locally. We don't need to add localhost to the live envs CSP

tools/prometheus/prometheus.yml
```diff
@@ -1,6 +1,6 @@
 global:
-  scrape_interval: 10s
-  scrape_timeout: 10s
+  scrape_interval: 5s
+  scrape_timeout: 5s
```
asanchezr, line 3: this configuration is only used in local containers so it is okay to increase the frequency at which local prometheus "scrapes" for metrics from our local otel-collector container.

In live envs, this is outside of our control. OpenShift has a Prometheus operator installed that handles it automatically based on container metadata that we already provide

tools/prometheus/prometheus.yml
```diff
@@ -9,15 +9,21 @@ alerting:
...
     static_configs:
       - targets:
           - host.docker.internal:5000
+  - job_name: otel-collector
+    honor_timestamps: true
+    metrics_path: /metrics
+    scheme: http
+    static_configs:
+      - targets:
+          - otel-collector:9464
```
asanchezr, line 29: We need to instruct our local prometheus container to scrape the metrics generated by otel-collector (from the traces that are sent to it by our React frontend). 

The following image shows a diagram of our setup for local monitoring:

otel-collector-jaeger-prometheus-architecture-2

tools/otel-collector/otel-config.yaml
```diff
@@ -0,0 +1,74 @@
...
+    logs:
+      receivers: [otlp]
+      processors: [memory_limiter, batch]
+      exporters: [debug, otlp/jaeger]
+    traces:
+      receivers: [otlp]
+      exporters: [debug, otlp/jaeger, spanmetrics]
+    metrics/spanmetrics:
+      receivers: [spanmetrics]
+      exporters: [debug, prometheus]
```
asanchezr, line 19: this is the YAML config file for the local open-telemetry collector. A similar YAML exists in Argo for the live environments

tools/jaeger/jaeger-config.yaml
```diff
@@ -0,0 +1,82 @@
...
+service:
+  extensions: [jaeger_storage, jaeger_query, healthcheckv2]
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [jaeger_storage_exporter, spanmetrics]
+    metrics/spanmetrics:
+      receivers: [spanmetrics]
+      exporters: [prometheus]
```
asanchezr, line 12: YAML config file for Jaeger local container. Jaeger receives the **traces** (forwarded by otel-collector) and also reads metrics generated by Prometheus to show a UI where you can search/list the traces and also show metrics like the ones below:
pims-RED-metrics

Makefile
```diff
@@ -180,7 +180,11 @@ refresh: | down build up ## Recreates local docker environment
 .PHONY: infra
 infra: ## Starts infrastructure containers (e.g. database, geoserver). Useful for local debugging
 	@echo "$(P) Starting up infrastructure containers..."
-	@"$(MAKE)" start n="database geoserver grafana prometheus proxy"
+	@"$(MAKE)" start n="database geoserver proxy"
+
+monitoring-up: ## Calls the docker compose up for the monitoring/telemetry images
+	@echo "$(P) Create or start mayan-edms system"
```
devinleighsmith, line 186: hmm, this echo doesn't seem correct.

asanchezr: you're right - copy pasta. will fix

source/frontend/src/index.tsx
```diff
@@ -41,19 +44,44 @@ async function prepare() {
...
+
+  // get telemetry configuration from tenant json
+  if (tenant?.telemetry?.enabled) {
+    const config: TelemetryConfig = {
+      name: tenant?.telemetry?.serviceName ?? 'frontend',
+      appVersion: import.meta.env.VITE_PACKAGE_VERSION ?? '',
+      environment: tenant?.telemetry?.environment || 'local',
+      otlpEndpoint: tenant?.telemetry?.endpoint || '',
+      debug: tenant?.telemetry?.debug ?? false,
+      exportInterval: tenant?.telemetry?.exportInterval ?? 30_000,
```
devinleighsmith, line 71: 30_000?

asanchezr: this is the same as 30000 ms (or 30 seconds). If you are asking about the numeric separator - they are supported by JS
https://backbencher-dev.medium.com/numeric-separators-in-javascript-29b154a0297a

```js
console.log(1000_000_000 === 1000000000); // true
```

source/frontend/src/telemetry/metrics.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const meter = metrics.getMeter(NETWORK_METER);
+
+  // measure the time it takes to make an XHR request
+  const timeSpentMetric = meter.createHistogram(METRIC_HTTP_CLIENT_REQUEST_DURATION, {
+    description: 'The duration of an outgoing HTTP request.',
+    unit: 's',
+    advice: {
+      explicitBucketBoundaries: [
+        0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10,
```
devinleighsmith, line 28: would a comment help for this? also, should this be externally configurable?

source/frontend/src/telemetry/metrics.ts
```diff
@@ -0,0 +1,82 @@
...
+  // Use PerformanceObserver Browser API to track these things
+  // https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver
+  const observer = new PerformanceObserver(list => {
+    const entries = list.getEntriesByType('resource');
+    entries.forEach((entry: PerformanceResourceTiming) => {
+      if (
+        'initiatorType' in entry &&
+        (entry.initiatorType === 'xmlhttprequest' || entry.initiatorType === 'fetch')
+      ) {
+        const uri = new URL(entry.name);
```
devinleighsmith, line 48: presumably entry.name is always valued?

source/frontend/src/telemetry/metrics.ts
asanchezr: we could make this configurable - sure. This is lifted from the open-telemetry documentation as the suggested buckets for the histogram distribution. These are the buckets that count how many requests fall in those boundaries (the above is in seconds). We can move that to the tenant.json where we already have some other config

FYI: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation

image

source/frontend/src/telemetry/metrics.ts
asanchezr: I can add a null check to be more defensive there but according to https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/name this is meant to be defined

source/frontend/src/telemetry/users/UserAPI.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function getUserDetailsFromKeycloakToken(userInfo?: KeycloakTokenParsed): TraceUser {
+  const displayName = userInfo?.display_name;
+  const fullName =
+    exists(userInfo?.given_name) && exists(userInfo?.family_name)
+      ? `${userInfo?.given_name} ${userInfo?.family_name}`
+      : 'default';
+
+  return {
+    displayName: displayName ?? fullName,
```
devinleighsmith, line 60: would like to see client_roles as well if possible

asanchezr: sure - good idea

devinleighsmith (approved):

PR #4720, opened 2025-03-28
PSP-10196 : Remove on-prem sonarqube, and all git action references.

eddherrera:


asanchezr (approved):
Thanks for removing these

PR #4725, opened 2025-04-01
PSP-10208: UI UX Clean Up - Larger header

stairaku:
Changes made were discussed with Ana.
image
image


source/frontend/src/components/common/styles.ts
```diff
@@ -144,9 +144,19 @@ export const StyledSectionParagraph = styled.p`
 `;
 
 export const TrayHeader = styled(H1)`
-  padding: 1rem;
-  background-color: ${props => props.theme.bcTokens.surfaceColorPrimaryButtonDefault};
```
devinleighsmith, line 148: You've (probably by accident) removed the background color of all trayheaders in the application. This is fine in the case comp reqs, as you've added the background color to the TrayHeaderContent manually there, but not fine for other headers, like management, that aren't wrapped by a TrayHeaderContent:

image

stairaku: Sorry about that! I just updated my code to include the expected background.

devinleighsmith: No need to apologize, code reviews aren't about laying blame, just about trying to catch issues before they are released and need to be bugfixed :-)

devinleighsmith (approved):

PR #4729, opened 2025-04-02
PSP-9933  Acquisition - Add a "Physical file details" text field

asanchezr:
acq_update_form


devinleighsmith (changes requested):
I don't see the physical file details on the create form - from the confluence it appears that field should be included there.

asanchezr:
> I don't see the physical file details on the create form - from the confluence it appears that field should be included there.

yes, working in that now. From the initial story it wasn't clear but will add it

devinleighsmith (approved):

PR #4730, opened 2025-04-02
Psp 10174 - when searching by plan number - show both PIMS and PMBC results simultaneously.

devinleighsmith:


source/backend/api/Areas/Tools/Controllers/BctfaOwnershipController.cs
```diff
@@ -58,7 +58,7 @@ public BctfaOwnershipController(ILogger<BctfaOwnershipController> logger, Claims
         [ProducesResponseType(200)]
         [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "tools-ltsa" })]
-        [HasPermission(Permissions.BctfaOwnershipEdit)]
+        [HasPermission(Permissions.PropertyAdd)]
```
asanchezr, line 61: the other endpoint in this controller is using the old Bctfa claim. Shouldn't that one be updated as well?

More generally - do we need the Bctfa specific claims when we are planning on reusing the PropertyAdd claim?

asanchezr (approved):
See comments

source/backend/api/Areas/Tools/Controllers/BctfaOwnershipController.cs
devinleighsmith: My mistake, changed this one for testing, will change it back.

PR #4734, opened 2025-04-03
psp-10008 allow lat/lng search on property picker.

devinleighsmith:


source/frontend/src/features/properties/filter/CoordinateSearch/CoordinateSearchForm.tsx
```diff
@@ -1,5 +1,6 @@
-import { Col, Container, Form } from 'react-bootstrap';
+import { Col, Row } from 'react-bootstrap';
 import styled from 'styled-components';
+import { clsx } from 'yet-another-react-lightbox';
```
asanchezr, line 3: I'm not sure this is the proper import. I believe we use `import clsx from 'classnames';`

asanchezr (approved):

source/frontend/src/setupTests.ts
```diff
@@ -18,6 +18,7 @@ import { cleanup } from './utils/test-utils';
 if (typeof global.URL.createObjectURL === 'undefined') {
   global.URL.createObjectURL = vi.fn();
 }
+global.innerWidth = 1920;
```
asanchezr, line 21: ?

asanchezr (approved):

PR #4738, opened 2025-04-04
Psp 10088 geographic search

devinleighsmith:
this has been confirmed with Tracy - there is no alternative for searching abstract locations that are free.


source/frontend/package.json
```diff
@@ -16,6 +16,7 @@
     "convert": "4.4.2",
     "dequal": "2.0.2",
     "esri-leaflet": "3.0.14",
+    "dms-conversion": "3.1.3",
```
asanchezr, line 19: do we need this extra library? I believe we implemented the DML -> Lat/Long logic internally but perhaps there is value in replacing that with a library? I ask because I did not see the original DML logic removed in favor of this library

asanchezr (approved):

source/frontend/src/features/properties/filter/GeographicNameInput.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+}));
+vi.mock('@/hooks/useGeographicNamesRepository');
+
+const mockSearchName = {
+  execute: vi.fn(),
+  error: null,
+  loading: false,
+  status: null,
+  response: null,
+};
```
asanchezr, line 23: **nit:** we have a utility function to reduce boilerplate like this - `getMockRepositoryObj()` 
perhaps we could use instead?

source/frontend/src/features/properties/filter/GeographicNameInput.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+  setFieldValue: vi.fn(),
+  values: { testField: '' },
+} as unknown as ReturnType<typeof useFormikContext>);
+
+describe('GeographicNameInput', () => {
+  const setup = (props = {}) => {
+    return render(
+      <TestCommonWrapper>
+        <GeographicNameInput field="testField" placeholder="Search..." {...props} />
+      </TestCommonWrapper>,
```
asanchezr, line 40: nit - I believe the common `render` function already wraps the component with TestCommonWrapper - not sure why would need to wrap it twice

PR #4744, opened 2025-04-07
PSP-9900 : Notification services

eddherrera:


source/frontend/src/components/layout/Healthcheck/HealthcheckView.test.tsx
```diff
@@ -15,6 +15,7 @@ describe('Healthcheck View component', () => {
     const utils = render(
       <HealthcheckView
         {...renderOptions.props}
+        systemChecked={renderOptions.props?.systemChecked ?? true}
```
asanchezr, line 18: why default to **true** here? it would seem like the default would be "system check not yet performed", ie "false" by default

asanchezr (approved):
see comments

source/frontend/src/components/layout/Healthcheck/HealthcheckView.test.tsx
eddherrera: The entirety of the View only displays when both are true. Making it false wouldn't display anything.

eddherrera: Actually, the other one, systemDegraded should be true.

PR #4745, opened 2025-04-09
PSP-10279 : FT:LTSA Service degradation notification is popping up on the screen even after service is looking fine

eddherrera:


source/frontend/src/components/layout/Healthcheck/HealthcheckContainer.test.tsx
```diff
@@ -120,4 +120,11 @@ describe('Healthcheck container', () => {
 
     expect(updateHealthcheckResult).toHaveBeenCalledWith(true);
   });
+
+  it(`sets the right amount of healtcheck issues`, async () => {
+    const { } = await setup();
```
asanchezr, line 125: nit: no need for the empty `const {}` here - just do `await setup()`

source/frontend/src/components/layout/Healthcheck/HealthcheckContainer.tsx
```diff
@@ -30,39 +30,63 @@ export const HealthcheckContainer: React.FunctionComponent<IHealthcheckContainer
   const keycloak = useKeycloakWrapper();
 
   const checkExternalSystemStatus = useCallback(
-    (response: IHealthcheckResponse, service: HealthcheckMessagesTypesEnum) => {
+    (
+      response: IHealthcheckResponse,
+      service: HealthcheckMessagesTypesEnum,
+      issues: IHealthCheckIssue[],
+    ) => {
```
asanchezr, line 37: nit: I know this PR is mostly about fixing bugs but I see some inconsistent naming here. We should pick either "IHealthcheck..." or "IHealthCheck..." - I personally prefer the 2nd option. But regardless of which option you pick, we should use a consistent naming

asanchezr (approved):

source/frontend/src/components/layout/Healthcheck/HealthcheckContainer.test.tsx
eddherrera: updated

PR #4748, opened 2025-04-10
psp-10193 do not allow adding a property via click when there are multiple results.

devinleighsmith:


devinleighsmith:
@FuriousLlama I'm not sure how this'll interact with your changes, please advise.

PR #4756, opened 2025-04-15
use "sub" instead of idir_user_guid to identify users.

devinleighsmith:
see corresponding jira for details.

### DIFF
--- a/source/backend/api/Startup.cs
+++ b/source/backend/api/Startup.cs
@@ -201,7 +201,6 @@ namespace Pims.Api
                     {
                         OnTokenValidated = context =>
                         {
-                            JsonWebToken token = context.SecurityToken as JsonWebToken;
                             return Task.CompletedTask;
                         },
                         OnAuthenticationFailed = context =>
--- a/source/backend/api/Startup.cs
+++ b/source/backend/api/Startup.cs
@@ -181,7 +181,6 @@ namespace Pims.Api
                     options.Audience = Configuration["Keycloak:Audience"];
                     options.SaveToken = true;
                     options.UseSecurityTokenValidators = true;
-                    options.MapInboundClaims = true;
                     options.MapInboundClaims = false;
                     options.TokenValidationParameters = new Microsoft.IdentityModel.Tokens.TokenValidationParameters()
                     {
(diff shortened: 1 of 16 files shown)


source/frontend/src/components/common/form/UserRegionSelect/UserRegionSelectContainer.tsx
```diff
@@ -1,3 +1,4 @@
+import { first } from 'lodash';
```
asanchezr, line 1: nit: use `import first from 'lodash/first'` to improve tree-shaking

source/backend/api/Startup.cs
```diff
@@ -173,16 +175,21 @@ public void ConfigureServices(IServiceCollection services)
...
                 {
-                    var key = Encoding.ASCII.GetBytes(Configuration["Keycloak:Secret"]);
+                    var key = Encoding.ASCII.GetBytes(Env.GetString("KEYCLOAK__SECRET"));
                     options.RequireHttpsMetadata = false;
                     options.Authority = Configuration["OpenIdConnect:Authority"];
                     options.Audience = Configuration["Keycloak:Audience"];
                     options.SaveToken = true;
+                    options.UseSecurityTokenValidators = true;
+                    options.MapInboundClaims = true;
+                    options.MapInboundClaims = false;
```
asanchezr, line 185: this doesn't look right...
`x = true` followed by `x = false`?

asanchezr (approved):

source/backend/api/Startup.cs
```diff
@@ -173,16 +175,21 @@ public void ConfigureServices(IServiceCollection services)
                 })
                 .AddJwtBearer(options =>
                 {
-                    var key = Encoding.ASCII.GetBytes(Configuration["Keycloak:Secret"]);
+                    var key = Encoding.ASCII.GetBytes(Env.GetString("KEYCLOAK__SECRET"));
```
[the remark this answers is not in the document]
devinleighsmith: This is actually a sonar recommendation.
we use DotNetEnv to load env vars, and those env vars are loaded into configuration if they are formatted like OUTER__INNER__JSON__OBJECT.
Sonar recommends against using configuration, because someone could mistakenly copy a secret directly into the configuration (appsettings) which is why we have this:

"Keycloak": {
  "Authority": "https://loginproxy.gov.bc.ca/auth/realms/standard",
  "Audience": "property-services-project-api-4380",
  "ValidAudience": "property-services-project-app-4379",
  "Secret": "[USE SECRETS]",

loading from the env var directly, instead of the env var through the configuration object, is safer.

PR #4766, opened 2025-04-23
PSP-10098 Track keycloak events with browser telemetry

asanchezr:
image


source/frontend/public/tenants/tenant.json
```diff
@@ -48,14 +48,5 @@
...
   "doubleClickInterval": 250,
-  "telemetry": {
-    "enabled": true,
-    "debug": false,
-    "environment": "local",
-    "serviceName": "frontend",
-    "endpoint": "https://collector-3cd915-dev.apps.silver.devops.gov.bc.ca",
-    "exportInterval": 10000,
-    "histogramBuckets": [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
-  },
```
asanchezr, line 59: I had to move the telemetry config out from the tenant.json file because I noticed issues with the telemetry setup running multiple times - the first render will use the "defaultTenant" and then when tenant.json is fetched via XHR then the context is updated and a re-render happens.

This was messing up the telemetry code so I moved it to a different pattern of injecting config values, which is:
- In development (ie **npm start** in our locals) it uses the .env files to load the values
- In production (ie **npm run build**) it relies on a **config.js** file that is loaded in index.html. The file is empty by default on our source code but it gets overridden with a config-map in Openshift environments.

See: https://www.justinpolidori.it/posts/20210913_containerize_fe_react/

source/frontend/src/index.tsx
```diff
@@ -42,9 +45,8 @@ async function prepare() {
   return Promise.resolve();
 }
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-//@ts-ignore
```
asanchezr, line 46: this is no longer needed - we are now using the proper typing

source/frontend/src/utils/getKeycloakEventHandler.ts
```diff
@@ -15,24 +24,39 @@ const getKeycloakEventHandler = (keycloak: Keycloak, onRefresh: () => void) => {
...
+        // store the currently logged user so that telemetry spans can be traced back to user actions
+        const userDetails = getUserDetailsFromKeycloakToken(keycloak.tokenParsed);
+        user.getUserManager().setUser(userDetails);
+        SpanEnrichment.enrichWithKeycloakToken(attributes, keycloak);
+      } else if (eventType === 'onAuthRefreshSuccess') {
+        onRefresh();
+        store.dispatch(saveJwt(keycloak.token ?? ''));
+        SpanEnrichment.enrichWithKeycloakToken(attributes, keycloak);
+      } else if (eventType === 'onAuthLogout' || eventType === 'onTokenExpired') {
+        store.dispatch(clearJwt());
```
devinleighsmith, line 46: Can we get something added to span in this case? These two eventTypes are likely what a user will see when they are forcibly logged out.

tools/cicd/scripts/gen-env-files.sh
```diff
@@ -105,7 +105,15 @@ else
...
-CHOKIDAR_USEPOLLING=true" >> ./source/frontend/.env
+CHOKIDAR_USEPOLLING=true
+VITE_SITEMINDER_LOGOUT_URL=https://logontest.gov.bc.ca/clp-cgi/logoff.cgi
+VITE_TELEMERY_ENABLED=false
+VITE_TELEMERY_DEBUG=false
+VITE_TELEMERY_ENVIRONMENT=local
+VITE_TELEMERY_SERVICE_NAME=frontend
+VITE_TELEMERY_URL=http://localhost:4318
+VITE_TELEMERY_EXPORT_INTERVAL=10000
+VITE_TELEMERY_HISTOGRAM_BUCKETS=[0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]" >> ./source/frontend/.env
```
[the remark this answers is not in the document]
devinleighsmith: please add this to gen-env-files.sh

devinleighsmith:
Please simulate the results of a logout event (my recommendation would be to log into a non-dev environment while logged into your local pims, this will invalidate your dev token). and show what details are captured by the span? I want to confirm that this will give us everything we'd need to debug.

source/frontend/src/utils/getKeycloakEventHandler.ts
asanchezr: We are logging the event name when all the events happen (in line 56, outside of all the if/then/elses). For this event in particular what extra would you want to log? the token expiration date? open to ideas

devinleighsmith: I was thinking the error/error_description

devinleighsmith (approved):

PR #4768, opened 2025-04-24
Test hotfixes

devinleighsmith:
[PSP-10327](https://moti-imb.atlassian.net/browse/PSP-10327)
[PSP-10283](https://moti-imb.atlassian.net/browse/PSP-10283)

### DIFF
--- a/source/frontend/package.json
+++ b/source/frontend/package.json
@@ -29,7 +29,7 @@
     "is-absolute-url": "3.0.3",
     "js-file-download": "0.4.12",
     "keycloak-js": "26.0.0",
-    "leaflet": "1.9.4",
+    "leaflet": "1.9.2",
     "lodash": "4.17.21",
     "maplibre-gl": "4.7.1",
     "moment": "2.29.4",
--- a/source/frontend/package.json
+++ b/source/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "5.9.0-102.46",
+  "version": "5.9.1-102.46",
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "1.0.1",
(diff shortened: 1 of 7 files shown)


FuriousLlama (approved):

source/frontend/package.json
```diff
@@ -29,7 +29,7 @@
     "is-absolute-url": "3.0.3",
     "js-file-download": "0.4.12",
     "keycloak-js": "26.0.0",
-    "leaflet": "1.9.2",
+    "leaflet": "1.9.4",
```
asanchezr, line 32: I think this is fine but any change in dependencies might introduce further errors/regressions. Would prefer to do this outside of horfixes (unless the fix is to update the library...)

source/frontend/src/components/common/mapFSM/useMapSearch.tsx
```diff
@@ -132,48 +132,49 @@ export const useMapSearch = () => {
...
+              }
+            : emptyPimsLocationFeatureCollection,
+          pimsBoundaryFeatures: emptyPimsBoundaryFeatureCollection,
+          fullyAttributedFeatures: exists(attributedFeatures)
+            ? {
+                type: attributedFeatures.type,
+                bbox: attributedFeatures.bbox,
+                features: validPmbcFeatures,
+              }
+            : emptyPmbcFeatureCollection,
```
asanchezr, line 169: hmm - this logic seems a bit wonky. On line 150 we calculate **validPmbcFeatures** to be the subset of features with valid geometry (good!). But then here we check that the original **attributedFeatures** exists (not the validPmbcFeatures array) and then we assign the filtered one?

By looking at the pattern for pimsLocationFeature, I would expect that we check for `exists(validPmbFeatures)` on line 163. Am I missing something?

asanchezr (approved):

PR #4779, opened 2025-05-01
Psp 10329 fix

devinleighsmith:


source/frontend/src/features/properties/map/MapContainer.tsx
```diff
@@ -75,7 +74,7 @@ const MapContainer: React.FC<React.PropsWithChildren<MapContainerProps>> = () =>
         <CompensationRequisitionRouter setShowActionBar={setShowActionBar} />
         <PropertyActivityRouter setShowActionBar={setShowActionBar} />
       </SideBarContextProvider>
-      {!(showActionBar || mapSideBarViewState?.isFullWidth) && (
```
devinleighsmith, line 78: @FuriousLlama can you please validate this? I wasn't able to see the base layer toggle after removing this, but I don't know the exact repro steps.

devinleighsmith:
Confirmed, the second part of the OR is not required anymore to prevent the layer square from appearing.

asanchezr (approved):

PR #4787, opened 2025-05-06
psp-10349 | common property fixes

FuriousLlama:
image

### DIFF
--- a/source/backend/api/Services/AcquisitionFileService.cs
+++ b/source/backend/api/Services/AcquisitionFileService.cs
@@ -690,6 +690,27 @@ namespace Pims.Api.Services
                         acquisitionProperty.Property = _propertyService.PopulateNewProperty(acquisitionProperty.Property);
                     }
                 }
+                else if(!string.IsNullOrWhiteSpace(acquisitionProperty.Property.SurveyPlanNumber))
+                {
+                    var plan = acquisitionProperty.Property.SurveyPlanNumber;
+                    try
+                    {
+                        var foundProperty = _propertyRepository.GetWithOnlyPlan(plan);
+                        if (foundProperty.IsRetired.HasValue && foundProperty.IsRetired.Value)
+                        {
+                            throw new BusinessRuleViolationException("Retired property can not be selected.");
+                        }
+
+                        acquisitionProperty.PropertyId = foundProperty.Internal_Id;
+                        _propertyService.UpdateLocation(acquisitionProperty.Property, ref foundProperty, userOverrideCodes);
+                        acquisitionProperty.Property = foundProperty;
+                    }
+                    catch (KeyNotFoundException)
+                    {
+                        _logger.LogDebug("Adding new property with no pid or pin. plan:{plan}", plan);
+                        acquisitionProperty.Property = _propertyService.PopulateNewProperty(acquisitionProperty.Property);
+                    }
+                }
                 else
                 {
                     _logger.LogDebug("Adding new property without a pid");
--- a/source/backend/dal/Repositories/PropertyRepository.cs
+++ b/source/backend/dal/Repositories/PropertyRepository.cs
@@ -257,6 +257,47 @@ namespace Pims.Dal.Repositories
         }
 
         /// <summary>
+        /// Get the properties that dont hace PIN or PID but have a plan#.
+        /// </summary>
+        /// <param name="plan"></param>
+        /// <returns></returns>
+        public PimsProperty GetWithOnlyPlan(string plan, bool includeRetired = false)
+        {
+            this.User.ThrowIfNotAllAuthorized(Permissions.PropertyView);
+
+            var query = Context.PimsProperties.AsNoTracking();
+
+            if (!includeRetired)
+            {
+                query = query.Where(r => !r.IsRetired.HasValue || (r.IsRetired.HasValue && !r.IsRetired.Value));
+            }
+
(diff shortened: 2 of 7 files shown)


source/frontend/src/components/maps/leaflet/LayerPopup/MultiplePropertyPopupView.tsx
```diff
@@ -41,21 +44,27 @@ export const MultiplePropertyPopupView: React.FC<
...
-          feature: x,
-        };
-      })
-      .sort((a, b) => {
-        if (a.isStrataLot === b.isStrataLot) return 0;
-        if (a.isStrataLot) return -1;
-        if (b.isStrataLot) return 1;
-        return 0;
-      }) ?? [];
+  const groupedFeatures = chain(featureDataset?.parcelFeatures)
```
devinleighsmith, line 47: Might need a comment here for the grouping->ungrouping mechanism.

Question, if there are multiple plan numbers, wouldn't this result in some strata lots not being sorted to the top? That is a functional change - not sure if that is what was desired. Is the desire here to show that x properties all have the same plan? If so, maybe the plan should be displayed in addition to the pid/pin.

As an example, I think this change would result in something like:
Common Property VIS8102
125-521-123
125-521-122
125-521-121
125-521-121
Common Property VIS56215
512-516-231

whereas previously this would have displayed (without the grouped sort):
Common Property VIS8102
Common Property VIS56215
125-521-123
125-521-122
125-521-121
125-521-121
512-516-231

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -256,6 +256,47 @@ public PimsProperty GetByPin(int pin, bool includeRetired = false)
                 .FirstOrDefault(p => p.Pin == pin) ?? throw new KeyNotFoundException();
         }
 
+        /// <summary>
+        /// Get the properties that dont hace PIN or PID but have a plan#.
```
devinleighsmith, line 260: nit: hace -> have.

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -256,6 +256,47 @@ public PimsProperty GetByPin(int pin, bool includeRetired = false)
...
+        /// </summary>
+        /// <param name="plan"></param>
```
devinleighsmith, line 262: nit: missing param.

source/backend/api/Services/AcquisitionFileService.cs
```diff
@@ -690,6 +690,27 @@ private void MatchProperties(PimsAcquisitionFile acquisitionFile, IEnumerable<Us
                         acquisitionProperty.Property = _propertyService.PopulateNewProperty(acquisitionProperty.Property);
                     }
                 }
+                else if(!string.IsNullOrWhiteSpace(acquisitionProperty.Property.SurveyPlanNumber))
```
devinleighsmith, line 693: Ok, do we know if this is just an acquisition - specific requirement? What happens if I try and add a common property to another file type?

source/backend/dal/Repositories/PropertyRepository.cs
```diff
@@ -256,6 +256,47 @@ public PimsProperty GetByPin(int pin, bool includeRetired = false)
...
+                .Include(p => p.VolumeUnitTypeCodeNavigation)
+                .Include(p => p.Address)
+                    .ThenInclude(a => a.RegionCodeNavigation)
+                .Include(p => p.Address)
+                    .ThenInclude(a => a.DistrictCodeNavigation)
+                .Include(p => p.Address)
+                    .ThenInclude(a => a.ProvinceState)
+                .Include(p => p.Address)
+                    .ThenInclude(a => a.Country)
+                .FirstOrDefault(p => p.SurveyPlanNumber == plan && p.Pin == null && p.Pid == null) ?? throw new KeyNotFoundException();
```
devinleighsmith, line 297: We're pretty confident on this I take it? In the long run we may want to back this up with some documentation from LTSA - on both issues: 

1. there is only one non-pid/pin common property for plan number
2. common property can be identified as a parcel with no pid/pin but with a plan (and ends with an 'S')

source/frontend/src/components/maps/leaflet/LayerPopup/MultiplePropertyPopupView.tsx
FuriousLlama: Yes, that's correct. I was surprised by a single location having multiple Common Properties. I figured that showing the properties in a grouped manner makes more sense than just putting the common strata on top.

source/backend/api/Services/AcquisitionFileService.cs
FuriousLlama: For now is only in acquisition; I presume all others might need it, but for now the common properties are only used by the business for acquisition files.

source/backend/dal/Repositories/PropertyRepository.cs
FuriousLlama: Yes, AFAIK, that is the way we would know if it's the same property. On your second point, we need to have that on the front end too, but I am unsure if its a hotfix or something we can add later.

devinleighsmith: To me it seems like a 5.10 change.

devinleighsmith (approved):

PR #4788, opened 2025-05-07
PSP-8629 create management activity

asanchezr:
image


source/frontend/src/features/mapSideBar/management/tabs/ManagementFileTabs.tsx
```diff
@@ -52,7 +54,13 @@ export const ManagementFileTabs: React.FC<IManagementFileTabsProps> = ({
     name: 'File Details',
   });
 
-  if (managementFile?.id && hasClaim(Claims.DOCUMENT_VIEW)) {
+  tabViews.push({
```
devinleighsmith, line 57: Does this view depend on a management file id the way documents and notes does?

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/InvoiceForm.tsx
```diff
@@ -0,0 +1,152 @@
...
+export interface IInvoiceForm {
+  namespace: string;
+  index: number;
+  formikProps: FormikProps<ManagementActivityFormModel>;
+  onDelete: () => void;
+  gstConstant: number;
+  pstConstant: number;
+}
+
+export const InvoiceForm: React.FunctionComponent<IInvoiceForm> = ({
```
devinleighsmith, line 23: Are there any differences between this form and the ad-hoc invoices form, besides the typing of the model?

asanchezr: There are some difference in the main form (ManagementActivityForm) but the Invoice forms are the same. I was undecided whether to point to the ad-hoc ones or create a copy of them (which is what I did). I could go either way. Both invoice forms are the same but I tried to keep both "branches" separate. Otherwise the code is somewhat intertwined.

Again, I would be fine pointing to the other form but wanted to keep it separate in case they need to diverge later

source/frontend/src/features/mapSideBar/management/tabs/ManagementFileTabs.tsx
asanchezr: good question. I was following the same pattern of the management summary tab which doesn't check for a valid management file ID but I can make the change here to be more defensive

asanchezr: Made the change to check for management file id

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ContactListForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+  };
+
+  const contacts = getIn(formikProps.values, field) as IContactSearchResult[];
+  return (
+    <FieldArray name={field}>
+      {arrayHelpers => (
+        <>
+          {contacts.map((_, index) => (
+            <Row key={`${field}.${index}`} className="no-gutters pr-4 mr-2">
```
devinleighsmith, line 39: this is probably copied, but is still "new" code, therefore, we shouldn't be using indexes to build keys, prefer a unique identifier and fall back to index if unavailable.

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ManagementActivityEditContainer.tsx
```diff
@@ -0,0 +1,97 @@
...
+  if (!isValidId(file?.id) && fileLoading === false) {
+    throw new Error('Unable to determine id of current file.');
+  }
+
+  const castedFile = file as unknown as ApiGen_Concepts_ManagementFile;
+
+  // Load the subtypes
+  const fetchSubtypes = useCallback(async () => {
+    const retrieved = await getSubtypes();
+    if (retrieved !== undefined) {
```
devinleighsmith, line 45: exists(retrieved)

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -0,0 +1,224 @@
...
+      serviceProviderOrgId: this.serviceProvider?.organizationId ?? null,
+      serviceProviderOrg: null,
+      serviceProviderPersonId: this.serviceProvider?.personId ?? null,
+      serviceProviderPerson: null,
+
+      involvedParties: this.involvedParties
+        .filter(exists)
+        .map<ApiGen_Concepts_PropertyActivityInvolvedParty>(x => {
+          return {
+            id: 0,
```
devinleighsmith, line 61: hmm, not sure about this, I think this will may ef to add/remove an entity every time this is saved even if no changes are made to the involved parties.

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -0,0 +1,224 @@
...
+
+export class ManagementActivityFormModel {
+  activityTypeCode = '';
+  activitySubtypeCode = '';
+  activityStatusCode = '';
+  requestedDate = '';
+  completionDate = '';
+  description = '';
+  requestedSource = '';
+  ministryContacts: (IContactSearchResult | null)[] = [null];
```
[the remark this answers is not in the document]
asanchezr: this is copied from the ad-hoc activity model code. I believe is to have an "empty" line in the various contacts here:
image

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
asanchezr: This was also lifted from the original activity implementation. We can enhance it for sure but it works for creation. For update (another upcoming story) I agree would be better to change

asanchezr: I will change it to use the existing ID (if is there, otherwise use 0)

asanchezr: After looking into this. in the **fromApi()** method we "drop" those IDs because we convert to an IContactSearchResult so there is no ID to use back in the **toApi** method.

Given that this code was already in place in ad-hoc activity I would lean to leave it as-is and discuss a better pattern in our dev meeting

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ContactListForm.tsx
asanchezr: Updated

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ManagementActivityEditContainer.tsx
asanchezr: Updated

devinleighsmith (approved):

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
devinleighsmith: That seems fine - my comment about copying code, is that copied code needs to follow our current standards, not the standards that code was developed with. We don't want to pervade negative patterns from old code if there is a viable solution.

PR #4795, opened 2025-05-09
psp-10456 updates to strata common property calculation.

devinleighsmith:


devinleighsmith:
see psp-10455 for rules.

source/frontend/src/utils/propertyUtils.ts
```diff
@@ -121,15 +122,20 @@ export function formatApiPropertyManagementLease(
...
+export function isStrataCommonProperty(
   feature: Feature<Geometry, PMBC_FullyAttributed_Feature_Properties> | undefined | null,
 ) {
   if (!exists(feature)) {
     return false;
   }
+
+  const planNumber = feature.properties.PLAN_NUMBER;
+  const nonNumericPrefix = firstOrNull(planNumber?.match(/^\D+/)); // Extract non-numeric prefix
+  const isStrataCommonPropertyPrefix = nonNumericPrefix?.slice(-1) === 'S'; // Check if the last character is 'S' PSP-10455
```
asanchezr, line 134: nit: I'd prefer if we were a bit more defensive and check for "S" case-insensitively. Just in case some data might not be in all caps

asanchezr (approved):

PR #4802, opened 2025-05-13
PSP-10287: Clean Up - Date Pop-up - uncover

stairaku:
image


source/frontend/src/features/leases/add/LeaseDetailSubForm.tsx
```diff
@@ -207,6 +207,7 @@ export const LeaseDetailSubForm: React.FunctionComponent<ILeaseDetailsSubFormPro
           formikProps={formikProps}
           field="startDate"
           required={statusTypeCode === ApiGen_CodeTypes_LeaseStatusTypes.ACTIVE}
+          popperPlacement="top-end"
```
asanchezr, line 210: Question: I think this story calls for placing the calendar popup on the top-end for all date fields. Did you check across all forms in PIMS that they don't need updates like the ones here? 
An alternative would be to make "top-end" the default value of the "popperPlacement" property so all datepickers support it by default

stairaku: @asanchezr, thanks for the heads up. I changed the code to apply the changes at the Datepicker object level instead.

source/frontend/src/components/common/form/FastDatePicker.tsx
```diff
@@ -130,6 +133,7 @@ const FormikDatePicker: FunctionComponent<React.PropsWithChildren<FastDatePicker
         maxDate={maxDate}
         {...rest}
         popperModifiers={popperModifiers}
+        popperPlacement={popperPlacement !== undefined ? popperPlacement : 'top-end'}
```
asanchezr, line 136: nit: we use the "exists" function to check for null or undefined. Please update this like so:

```tsx
popperPlacement={exists(popperPlacement) ? popperPlacement : 'top-end'}
```

asanchezr (approved):

PR #4803, opened 2025-05-13
PSP-10349 | Fixed properties not having a region if search results were more than 15

FuriousLlama:


source/frontend/src/components/propertySelector/search/PropertySearchSelectorFormView.tsx
```diff
@@ -62,34 +61,29 @@ export const PropertySearchSelectorFormView: React.FunctionComponent<
   }
 
   return (
-    <>
-      <Section header={undefined}>
-        <Styled.H3>Search for a property</Styled.H3>
```
FuriousLlama, line 67: Layout fixes. This allows the search and the result section to look cohesive.

source/frontend/src/components/propertySelector/search/PropertySelectorSearchContainer.tsx
```diff
@@ -80,63 +84,86 @@ export const PropertySelectorSearchContainer: React.FC<IPropertySelectorSearchCo
...
+        const latLngKey = propertyToLatLngKey(foundProperty);
+        if (exists(regionDistrictResults[latLngKey])) {
+          foundProperty.regionFeature = regionDistrictResults[latLngKey].regionResult;
+          foundProperty.districtFeature = regionDistrictResults[latLngKey].districtResult;
+        }
+
+        /*
+        *  TODO: this is always false. This section needs to be analized. Otherwise, both the PIMS and Address search results are not being used at all
+        * PSP-140476
+        if (exists(foundProperty?.pimsFeature)) {
```
FuriousLlama, line 146: Note that foundProperty.pimsFeature has not been initialized at all. Line 136 creates an empty pimsFeature

source/frontend/src/components/propertySelector/search/PropertySelectorSearchContainer.tsx
```diff
@@ -80,63 +84,86 @@ export const PropertySelectorSearchContainer: React.FC<IPropertySelectorSearchCo
...
-        });
-        setSearchResults(locations);
-      } else {
-        const locations = result?.features?.map(p => featureToLocationFeatureDataset(p));
-        setSearchResults(locations ?? []);
+      // match the region and district for all found properties
+      const dataset: SelectedFeatureDataset[] = result.features.map(p =>
+        featureToLocationFeatureDataset(p),
+      );
+      const regionDistrictTasks = getRegionAndDisctricts(dataset, findRegion, findDistrict);
```
devinleighsmith, line 119: Disctricts

source/frontend/src/components/propertySelector/search/PropertySelectorSearchContainer.tsx
```diff
@@ -174,9 +201,22 @@ export const PropertySelectorSearchContainer: React.FC<IPropertySelectorSearchCo
       });
 
       // match the region and district for all found properties
-      await Promise.all(
-        propertyResults.map(p => matchRegionAndDistrict(p, findRegion, findDistrict)),
-      );
+      const regionDistrictTasks = getRegionAndDisctricts(propertyResults, findRegion, findDistrict);
+
+      const regionDistrictResults: Dictionary<RegionDistrictResult> = {};
+      for (const latLngKey in regionDistrictTasks) {
```
devinleighsmith, line 207: This syntax is easier to read/more concise in my OP: 
dictionary.forEach((value, key)

devinleighsmith: This may be premature optimization, but this implementation seems to loop through the list of properties 4 times do to this processing - how does this perform on large lists of properties?

FuriousLlama:
> I tested this out, and the majority of the requests failed with: Err: insufficient resources - when dealing with a large number of properties. I think that is equivalent to current state, thus the approval, but it doesn't seem like this code really works when there are > 15 results. We will likely need to rewrite this.

Thats odd. I tested it with about 100 and it performed quite well. Might be a good idea to find the largest number of concurrent requests we can work with.

PR #4805, opened 2025-05-13
psp-10350 Backend | Property documents

FuriousLlama:


source/backend/dal/Repositories/Interfaces/ILeaseRepository.cs
```diff
@@ -26,14 +26,8 @@ public interface ILeaseRepository : IRepository<PimsLease>
 
         Paged<PimsLease> GetPage(LeaseFilter filter, HashSet<short> regions);
 
-        IList<PimsLeaseDocument> GetAllLeaseDocuments(long leaseId);
```
FuriousLlama, line 29: cleaned up repositories

source/backend/api/Services/DocumentFileService.cs
```diff
@@ -74,376 +77,143 @@ public IList<T> GetFileDocuments<T>(FileType fileType, long fileId)
...
-                    throw new BadRequestException("FileT type not valid to get documents.");
+                    throw new BadRequestException("File type not valid to get documents.");
             }
         }
 
         public async Task UploadAcquisitionDocument(long acquisitionFileId, DocumentUploadRequest uploadRequest)
         {
             Logger.LogInformation("Uploading document for single acquisition file");
             User.ThrowIfNotAllAuthorized(Permissions.DocumentAdd, Permissions.AcquisitionFileEdit);
-            ValidateDocumentUpload(uploadRequest);
```
FuriousLlama, line 105: All of the calls were doing the same thing. The new generic method handles it and these ones maintain logs and permisions

source/backend/entities/ef/PimsAccessRequestHist.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
asanchezr, line 1: could you remove this changes from the PR? These files are auto-generated and will be overridden when the next scaffold is run

FuriousLlama: good call. I've regenerated the entities to clear the changes

asanchezr (approved):

PR #4807, opened 2025-05-15
PSP-10345 Edit management activity

asanchezr:


source/backend/api/Services/ManagementFileService.cs
```diff
@@ -249,6 +252,39 @@ public Paged<PimsManagementFile> GetPage(ManagementFilter filter)
...
+            throw new BadRequestException("Activity with the given id does not match the management file id");
+        }
+
+        public PimsPropertyActivity UpdateActivity(long managementFileId, long activityId, PimsPropertyActivity propertyActivity)
+        {
+            _logger.LogInformation("Updating property Activity with Id: {ActivityId}", activityId);
+            _user.ThrowIfNotAuthorized(Permissions.ManagementEdit);
+            _user.ThrowIfNotAuthorized(Permissions.PropertyEdit);
+
+            if (!propertyActivity.ManagementFileId.HasValue || propertyActivity.ManagementFileId != managementFileId || propertyActivity.Internal_Id != activityId)
```
FuriousLlama, line 277: The managementFileId value is not used in the service call. If it is only being used for validation, let's move it to the controller given that one is responsible of validating the request.

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ManagementActivityEditContainer.tsx
```diff
@@ -61,19 +105,19 @@ export const ManagementActivityEditContainer: React.FunctionComponent<
...
-      // TODO: implement update activity flow
+      result = await updateManagementActivity(managementFileId, model);
     } else {
       result = await addManagementActivity(managementFileId, model);
     }
 
-    if (exists(result)) {
+    if (exists(result) && isValidId(managementFileId)) {
       setStaleLastUpdatedBy(true);
       history.push(`/mapview/sidebar/management/${managementFileId}/activities/${result.id}`);
```
FuriousLlama, line 115: Instead of hardcoding the path here, lets use the sidebarPathGenerator to create the routing path

source/backend/api/Services/ManagementFileService.cs
asanchezr: This is following the same patterns as **PropertyService.UpdateActivity()** for ad-hoc activities. I still see value in validating inputs at the service level to make the code more defensive (in the same way that we validate user permissions at the service level instead at the controller level). But I would be okay moving it to the controller id others feel like we should do that. Let's discuss this in our dev meeting later...

image

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ManagementActivityEditContainer.tsx
asanchezr: This is the first time I've heard of the "sidebarPathGenerator" - If this is something you created feel free to present it on our dev meeting so I can learn about it

asanchezr: Also - we have a ton of "hardcoded" routes. If you created something more generic we should address all paths in a tech-debt task, not just this PR

FuriousLlama: we talked about this some time ago. Without it, maintaining the routes is quite work-intensive.

source/backend/api/Services/ManagementFileService.cs
FuriousLlama: Yes, we should talk about this.

source/frontend/src/features/mapSideBar/router/ManagementFileActivityRouter.tsx
```diff
@@ -40,16 +40,40 @@ export const ManagementFileActivityRouter: React.FunctionComponent<
           <ManagementActivityEditContainer
             managementFileId={Number(match.params.managementFileId)}
             onClose={() => {
-              const parentPath = match?.url.substring(0, match?.url.lastIndexOf('/')) || '/';
+              const parentPath = generatePath(
```
devinleighsmith, line 43: I'm guessing that these need to be updated as well based on our friday conversation

asanchezr:
> Currently getting:
> 
> Uncaught Error: Unable to determine id of current file. at ManagementActivityEditContainer (ManagementActivityEditContainer.tsx:50:11)
> 
> removing that if statement/throw I was able to test this.

I need to rebase and fix all merge conflicts before this is ready for review

PR #4809, opened 2025-05-15
Psp 10346 - View management activities

devinleighsmith:


source/backend/api/Services/PropertyService.cs
```diff
@@ -281,6 +281,14 @@ public IList<PimsPropertyActivity> GetActivities(long propertyId)
             return _propertyActivityRepository.GetActivitiesByProperty(propertyId);
         }
 
+        public IList<PimsPropertyActivity> GetFileActivities(long managementFileId)
+        {
+            _logger.LogInformation("Getting property management activities for management file with id {managementFileId}", managementFileId);
+            _user.ThrowIfNotAuthorized(Permissions.ManagementView, Permissions.PropertyView);
```
asanchezr, line 287: I believe in the past we learnt that using multiple permissions here will check for "any" so there is potential that a user who has a subset of the claims above would be able to call this method.

Please call `_user.ThrowIfNotAllAuthorized()` instead (or add permission checks in separate lines)

source/backend/api/Services/PropertyService.cs
```diff
@@ -296,6 +304,21 @@ public PimsPropertyActivity GetActivity(long propertyId, long activityId)
             throw new BadRequestException("Activity with the given id does not match the property id");
         }
 
+        public PimsPropertyActivity GetFileActivity(long managementFileId, long activityId)
+        {
+            _logger.LogInformation("Retrieving single property Activity...");
+            _user.ThrowIfNotAuthorized(Permissions.ManagementView, Permissions.PropertyView);
```
asanchezr, line 310: _user.ThrowIfNotAllAuthorized()

source/backend/api/Services/PropertyService.cs
```diff
@@ -329,6 +352,29 @@ public PimsPropertyActivity UpdateActivity(long propertyId, long activityId, Pim
             return propertyActivityResult;
         }
 
+        public bool DeleteFileActivity(long managementFileId, long activityId)
+        {
+            _logger.LogInformation("Deleting Management Activity with id {activityId}", activityId);
+            _user.ThrowIfNotAuthorized(Permissions.ManagementDelete, Permissions.PropertyEdit);
```
asanchezr, line 358: _user.ThrowIfNotAllAuthorized()

source/backend/dal/Repositories/PropertyActivityRepository.cs
```diff
@@ -142,6 +160,29 @@ public bool TryDelete(long activityId)
...
+        /// </summary>
+        /// <param name="activityId"></param>
+        /// <returns>Boolean of deletion sucess.</returns>
+        public bool TryDeleteFile(long activityId, long managementFileId)
+        {
+            // There may be zero to many of these relationships.
+            Context.PimsPropPropActivities.ForEach(pp =>
+            {
+                Context.PimsPropPropActivities.Remove(pp);
+            });
```
asanchezr, line 174: hmm - this sounds dangerous. If I'm reading this right, it would delete **ALL** property-activity associations (not just the ones between the desired activity and its properties

Maybe something along the lines of:
```suggestion
            Context.PimsPropPropActivities.Where(x => x.PimsPropertyActivityId == activityId).ForEach(pp =>
            {
                Context.PimsPropPropActivities.Remove(pp);
            });
```

source/frontend/src/features/mapSideBar/management/tabs/activities/ActivitiesTab.tsx
```diff
@@ -43,6 +46,10 @@ export const ActivitiesTab: React.FunctionComponent<IActivitiesTabProps> = ({ ma
           <strong>Activity documentation:</strong> You can attach a document after creating the
           activity. Create, then edit and attach a file if needed.
         </p>
+        <ManagementFileActivitiesListContainer
+          View={ManagementActivitiesListView}
+          managementFileId={managementFile.id}
+        ></ManagementFileActivitiesListContainer>
```
asanchezr, line 52: Please check with @eddherrera because I thought he has the story about displaying a list of activities?
https://moti-imb.atlassian.net/browse/PSP-8837

It would make sense to include in this PR but please make sure he is not working on the same thing

asanchezr (changes requested):
Please fix the delete of all property-activity relationships regardless of activityId

source/backend/api/Services/PropertyService.cs
```diff
@@ -296,6 +304,21 @@ public PimsPropertyActivity GetActivity(long propertyId, long activityId)
...
         }
 
+        public PimsPropertyActivity GetFileActivity(long managementFileId, long activityId)
+        {
+            _logger.LogInformation("Retrieving single property Activity...");
+            _user.ThrowIfNotAuthorized(Permissions.ManagementView, Permissions.PropertyView);
+
+            var propertyActivity = _propertyActivityRepository.GetActivity(activityId);
+
+            if (propertyActivity.ManagementFileId == managementFileId)
```
FuriousLlama, line 314: Given that the managementFileId is only used to validate the request, let's remove it from this method and move the check to the controller.

source/backend/dal/Repositories/PropertyActivityRepository.cs
```diff
@@ -142,6 +160,29 @@ public bool TryDelete(long activityId)
...
+
+            var propertyActivity = Context.PimsPropertyActivities.FirstOrDefault(x => x.PimsPropertyActivityId == activityId && x.ManagementFileId == managementFileId);
+            if(propertyActivity is null)
```
FuriousLlama, line 177: Is this extra validation necessary? It would make sense to me if the repository deletes the activity with the given id, without considering the management file

source/backend/dal/Repositories/PropertyActivityRepository.cs
```diff
@@ -142,6 +160,29 @@ public bool TryDelete(long activityId)
             return deletedSuccessfully;
         }
 
+        /// <summary>
+        /// Delete an activity, and all property-activity relationships.
+        /// </summary>
+        /// <param name="activityId"></param>
+        /// <returns>Boolean of deletion sucess.</returns>
+        public bool TryDeleteFile(long activityId, long managementFileId)
```
FuriousLlama, line 168: This method looks off. The name implies that the file is being deleted, but the implementation deletes the activity instead. Is it supposed to be a "Delete All activities and then the file"?

source/frontend/src/features/mapSideBar/management/tabs/activities/detail/FileActivityDetailView.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <StyledFormWrapper>
+              <StyledSummarySection>
+                <LoadingBackdrop show={props.loading} />
+                <StyledEditWrapper className="mr-3 my-1">
+                  {hasClaim(Claims.MANAGEMENT_EDIT) && (
+                    <EditButton
+                      title="Edit File Property Activity"
+                      onClick={() => {
+                        history.push(
+                          `/mapview/sidebar/management/${props.managementId}/activities/${props.activity?.id}/edit`,
```
FuriousLlama, line 86: better to use the sidebarPathGenerator instead of hardcododing the url

source/frontend/src/features/mapSideBar/management/tabs/activities/list/ManagementFileActivitiesListContainer.tsx
```diff
@@ -0,0 +1,85 @@
...
+      setPropertyActivities([...response.map(x => PropertyActivityRow.fromApi(x))]);
+    }
+  }, [getActivities, isMounted, managementFileId]);
+
+  const onDelete = useCallback(
+    async (activityId: number) => {
+      const result = await deleteActivity(managementFileId, activityId);
+      if (result === true) {
+        fetchPropertyActivities();
+        history.push(`/mapview/sidebar/management/${managementFileId}/activities`);
```
FuriousLlama, line 42: same here. We need to move away from hardcoding the navigation at the container lavel.

source/backend/dal/Repositories/PropertyActivityRepository.cs
devinleighsmith: Some context required here, but the general pattern of this file (and others) is just:
propertyActivityRepository.TryDelete();
what is this deleting? presumably a property activity. Following this pattern, propertyActivityRepository.TryDeleteFile() is trying to remove a property activity related to a file.

Perhaps TryDeleteByFile?

source/backend/dal/Repositories/PropertyActivityRepository.cs
devinleighsmith: The idea with this repository method is that it should only be used on activities that were created with management files. This is because deleting a management file activity is different from deleting an ad-hoc activity (above).

devinleighsmith:
@FuriousLlama @asanchezr requested updates made.

source/backend/api/Services/PropertyService.cs
asanchezr: I still see `_user.ThrowIfNotAuthorized()` with multiple permissions which only checks the user has any (not all) of the permissions - please change to `ThrowIfNotAllAuthorized()`

source/backend/api/Services/PropertyService.cs
asanchezr: I thought we agreed to move this logic to the controller and make the service "trust" that the supplied entities were valid....

source/backend/api/Services/PropertyService.cs
```diff
@@ -284,6 +307,29 @@ public PimsPropertyActivity UpdateActivity(long propertyId, long activityId, Pim
...
+        {
+            _logger.LogInformation("Deleting Management Activity with id {activityId}", activityId);
+            _user.ThrowIfNotAuthorized(Permissions.ManagementDelete, Permissions.PropertyEdit);
+
+            var propertyManagementActivity = _propertyActivityRepository.GetActivity(activityId);
+
+            if(propertyManagementActivity.ManagementFileId != managementFileId)
+            {
+                throw new BadRequestException("Activity with the given id does not match the management file id");
+            }
```
asanchezr, line 320: we agreed to move this model check to the calling controller

source/backend/dal/Repositories/Interfaces/IPropertyActivityRepository.cs
```diff
@@ -12,10 +12,14 @@ public interface IPropertyActivityRepository : IRepository<PimsPropertyActivity>
...
 
+        IList<PimsPropertyActivity> GetActivitiesByManagementFile(long managementFileId);
+
         PimsPropertyActivity Create(PimsPropertyActivity propertyActivity);
 
         PimsPropertyActivity Update(PimsPropertyActivity propertyActivity);
 
         bool TryDelete(long activityId);
+
+        bool TryDeleteFile(long activityId, long managementFileId);
```
asanchezr, line 23: I believe we were going to call this "TryDeleteFileActivity" to avoid confusion on whether this method deletes the whole file (which it shouldn't)

source/backend/api/Services/PropertyService.cs
devinleighsmith: wrong commit, sorry, up to date now.

asanchezr (approved):

PR #4813, opened 2025-05-22
PSP-10352 Property Notes

asanchezr:
image

This PR also refactors **NoteService** and **EntityNoteRepository** to reduce code duplication:
- Created common base class for **PimsNoteRelationship**
- Cleaned up the note repositories to be generic (ie ProjectNoteRepo, AcquisitionFileNoteRepo, etc)
- Moved mapping logic (entity -> model) out from service and into the calling controller. For some reason the NoteService was dealing with model to entity conversion (which is against our patterns)


source/backend/api/Areas/Notes/Controllers/NoteController.cs
```diff
@@ -49,18 +57,56 @@ public NoteController(INoteService noteService, IMapper mapper)
...
+            switch (type)
+            {
+                case NoteType.Acquisition_File:
+                    var acquisitionFileNoteEntity = _mapper.Map<PimsAcquisitionFileNote>(model);
+                    var acquisitionFileNote = _noteService.AddAcquisitionFileNote(acquisitionFileNoteEntity);
+                    return new JsonResult(_mapper.Map<EntityNoteModel>(acquisitionFileNote));
+                case NoteType.Disposition_File:
+                    var dispositionFileNoteEntity = _mapper.Map<PimsDispositionFileNote>(model);
+                    var dispositionFileNote = _noteService.AddDispositionFileNote(dispositionFileNoteEntity);
+                    return new JsonResult(_mapper.Map<EntityNoteModel>(dispositionFileNote));
```
asanchezr, line 86: This whole logic was being handled by the service (including mapping the model to entities). It is now moved up to the controller who is responsible for data marshalling/conversion. The service now only knows about Pims entities

source/backend/api/Services/NoteService.cs
```diff
@@ -18,116 +14,134 @@ namespace Pims.Api.Services
...
         private readonly INoteRepository _noteRepository;
-        private readonly IEntityNoteRepository _entityNoteRepository;
-        private readonly IMapper _mapper;
+        private readonly INoteRelationshipRepository<PimsAcquisitionFileNote> _acquisitionFileNoteRepository;
+        private readonly INoteRelationshipRepository<PimsDispositionFileNote> _dispositionFileNoteRepository;
+        private readonly INoteRelationshipRepository<PimsLeaseNote> _leaseNoteRepository;
+        private readonly INoteRelationshipRepository<PimsManagementFileNote> _managementFileNoteRepository;
+        private readonly INoteRelationshipRepository<PimsProjectNote> _projectNoteRepository;
+        private readonly INoteRelationshipRepository<PimsPropertyNote> _propertyNoteRepository;
+        private readonly INoteRelationshipRepository<PimsResearchFileNote> _researchFileNoteRepository;
```
asanchezr, line 23: Replaced **EntityNoteRepository** (which was growing and growing) with individual repos for each type of note-relationship (project-note, acquisitionfile-note, etc)

This is consistent with the work Manuel did on the Property Documents refactor

source/backend/api/Services/NoteService.cs
```diff
@@ -18,116 +14,134 @@ namespace Pims.Api.Services
...
+            ClaimsPrincipal user,
+            ILogger<NoteService> logger,
+            INoteRepository noteRepository,
+            INoteRelationshipRepository<PimsAcquisitionFileNote> acquisitionFileNoteRepository,
+            INoteRelationshipRepository<PimsDispositionFileNote> dispositionFileNoteRepository,
+            INoteRelationshipRepository<PimsLeaseNote> leaseNoteRepository,
+            INoteRelationshipRepository<PimsManagementFileNote> managementFileNoteRepository,
+            INoteRelationshipRepository<PimsProjectNote> projectNoteRepository,
+            INoteRelationshipRepository<PimsPropertyNote> propertyNoteRepository,
+            INoteRelationshipRepository<PimsResearchFileNote> researchFileNoteRepository)
```
asanchezr, line 48: Cleaned up the note repos

source/backend/api/Services/NoteService.cs
```diff
@@ -165,27 +180,37 @@ public bool DeleteNote(NoteType type, long noteId, bool commitTransaction = true
...
 
+        private static T AddNoteRelationship<T>(T noteRelationship, INoteRelationshipRepository<T> noteRelationshipRepository)
+            where T : PimsNoteRelationship, new()
+        {
+            T newEntity = noteRelationshipRepository.AddNoteRelationship(noteRelationship);
+            noteRelationshipRepository.CommitTransaction();
+
+            return newEntity;
+        }
+
```
asanchezr, line 213: Generic function to add note-relationships (property-note, acquisition-note, etc)

source/backend/entities/Partials/PimsNoteRelationship.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public abstract class PimsNoteRelationship : StandardIdentityBaseAppEntity<long>
+    {
+        [NotMapped]
+        public abstract long ParentId { get; set; }
+
+        [NotMapped]
+        public abstract long InternalNoteId { get; set; }
+    }
+}
```
asanchezr, line 13: New base class for all note-relationships

source/backend/api/Services/NoteService.cs
```diff
@@ -18,116 +14,134 @@ namespace Pims.Api.Services
...
         }
 
-        public EntityNoteModel Add(NoteType type, EntityNoteModel model)
+        public PimsAcquisitionFileNote AddAcquisitionFileNote(PimsAcquisitionFileNote acquisitionFileNote)
         {
-            Logger.LogInformation("Adding note with type {Type} and model {Model}", type, model);
-            model.ThrowIfNull(nameof(model));
+            acquisitionFileNote.ThrowIfNull(nameof(acquisitionFileNote));
+            Logger.LogInformation("Adding note for acquisition file with Id: {Id}", acquisitionFileNote.ParentId);
             User.ThrowIfNotAuthorized(Permissions.NoteAdd);
```
devinleighsmith, line 73: This seems like a miss/limitation of the original implementation, but given the update I think it would make sense that the user has at least the corresponding file view claim, perhaps the edit claim (may be worth validating on the frontend if users without edit can add notes/documents to files.

asanchezr: I think you are right - according to the "original" notes implementation way then https://moti-imb.atlassian.net/browse/PSP-3693 it says:
> Precondition
> - User should have appropriate permissions (update) the parent screen
> - User should have appropriate permissions (add) the notes

I can apply the required view + update permissions on the backend but I worry that will break notes across the application. I feel we need to address this separately.

I also checked and both notes and documents tabs show even if you don't have EDIT permission on a file (you do need READ permission to see the file and its tabs). So both notes and documents suffer from the same "you can add them without edit permission on the parent" - so another argument to create a new story to fix both. 

Thoughts on that?

devinleighsmith:
Looks good, I like the changes, much neater.

source/backend/api/Services/NoteService.cs
devinleighsmith: I tested this, and currently the frontend is displaying no add/edit functionality on the frontend for read-only users - so I don't see any regression in functionality if we start enforcing this on the backend as well. I'm not seeing much risk here as a result, just correct enforcement of claims - perhaps I'm missing something.

I understand that you are running a bit behind on your stories this sprint however, so I'll leave it up to you if you want to log a ticket for this or fix it with this PR.

devinleighsmith (approved):

source/backend/api/Services/NoteService.cs
FuriousLlama: nit: the new() might not be necessary

FuriousLlama (approved):

FuriousLlama:
Great work adding the functionality and cleaning up the note stack.

PR #4833, opened 2025-06-02
PSP-8964: Fixing inserted bug

stairaku:
Inserting the suggested changes to fix the issue. Additionally, changing the multiselect height to match select and input elements' height.

image
image
image
image


source/frontend/src/components/common/form/Multiselect.scss
```diff
@@ -7,3 +7,16 @@
...
+  padding-top: 0rem !important;
+  padding-bottom: 0rem !important;
+}
+
+.search-wrapper input {
+  margin-top: 5px !important;
+}
+
+.chip {
+  margin-top: 5px;
```
asanchezr, line 21: we try to use **rem** instead of px in new CSS code - please change to **0.5rem** (1rem = 10px)

source/frontend/src/components/common/form/Select.scss
```diff
@@ -2,7 +2,7 @@
 
 .form-select {
   align-items: center;
-  height: 40px;
+  height: 38px;
```
asanchezr, line 5: **nit:** please use REM here

asanchezr (approved):

PR #4837, opened 2025-06-05
PSP-10342 PIMS File’s properties show polygons differently

asanchezr:
https://github.com/user-attachments/assets/c0915ba4-16d1-479a-bda3-0aaf42964eb5


source/frontend/src/components/maps/leaflet/Layers/PointClusterer.tsx
```diff
@@ -89,16 +87,6 @@ export const PointClusterer: React.FC<React.PropsWithChildren<PointClustererProp
...
-  const draftPoints = useMemo<LatLngLiteral[]>(() => {
-    return mapMachine.filePropertyLocations.map(x => {
-      // The values on the feature are rounded to the 4th decimal. Do the same to the draft points.
-      return {
-        lat: x.lat,
-        lng: x.lng,
-      };
-    });
-  }, [mapMachine.filePropertyLocations]);
-
```
asanchezr, line 101: Refactored the code related to "draftPoints" - ie the file-property markers into its own class (FilePropertiesLayer). 
It simplifies a bit the PointClusterer and moves out the logic to highlight file-properties into its own component

PR #4839, opened 2025-06-06
PSP-10492 | Moved the search bar into the right side bar

FuriousLlama:
image


source/frontend/src/components/maps/leaflet/Control/Search/SearchView.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const pimsGroupedFeatures = chain(props.searchResult?.pimsLocationFeatures.features)
+    .groupBy(feature => feature?.properties?.SURVEY_PLAN_NUMBER)
+    .map(
+      planGroup =>
+        planGroup
+          ?.map<PropertyProjection<PIMS_Property_Location_View>>(x => ({
+            pid: x.properties.PID_PADDED,
+            pin: exists(x.properties.PIN) ? String(x.properties.PIN) : null,
+            isStrataLot: false,
```
devinleighsmith, line 65: Am I missing something here? it seems like you are setting isStrataLot to false and then sort by is isStrataLot, isn't that a no-op?

PR #4854, opened 2025-06-18
Psp 10371

devinleighsmith:


source/frontend/src/features/mapSideBar/management/ManagementContainer.tsx
```diff
@@ -134,9 +134,10 @@ export const ManagementContainer: React.FunctionComponent<IManagementContainerPr
 
   const close = useCallback(() => onClose && onClose(), [onClose]);
 
-  const navigateToMenuRoute = (selectedIndex: number) => {
-    const route = selectedIndex === 0 ? '' : `/property/${selectedIndex}`;
-    history.push(`${stripTrailingSlash(match.url)}${route}`);
+  const navigateToMenuRoute = (propertyId: number) => {
+    const menuIndex = managementFile.fileProperties.findIndex(x => x.id === propertyId);
+
+    history.push(`${stripTrailingSlash(match.url)}/property/${menuIndex + 1}`);
```
FuriousLlama, line 140: you could use the path generator instead

source/frontend/src/components/common/mapFSM/MapStateMachineContext.tsx
```diff
@@ -42,7 +43,7 @@ export interface IMapStateMachineContext {
   isLoading: boolean;
   mapSearchCriteria: IPropertyFilter | null;
   mapFeatureData: MapFeatureData;
-  filePropertyLocations: LatLngLiteral[];
+  filePropertyLocations: IMapProperty[];
```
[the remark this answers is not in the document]
devinleighsmith: I'll add the new type, will increase the scope of my changes, but will also help work towards removing IMapProperty.

source/frontend/src/features/mapSideBar/shared/sidebarPathGenerator.ts
```diff
@@ -174,7 +174,7 @@ const usePathGenerator: IPathGenerator = () => {
     editDetails,
     addDetail,
     editProperties,
-    showFilePropertyIndex,
+    showFilePropertyId: showFilePropertyIndex,
```
asanchezr, line 177: nit: I would consider renaming the internal function as well to match the new name (also the parameters in the existing function say "menuIndex" - where in new code it would be filePropertyId

source/frontend/src/utils/mapPropertyUtils.ts
```diff
@@ -172,6 +172,30 @@ export const getFeatureBoundedCenter = (feature: Feature<Geometry, GeoJsonProper
...
+  return {
+    propertyId: fileProperty.propertyId,
+    pid: fileProperty.property?.pid?.toString(),
+    pin: fileProperty.property?.pin?.toString(),
+    latitude: fileProperty.property?.location?.coordinate.y,
+    longitude: fileProperty.property?.location?.coordinate.x,
+    fileLocation: {
+      lat: fileProperty.property?.location?.coordinate.y,
+      lng: fileProperty.property?.location?.coordinate.x,
+    },
```
asanchezr, line 185: just want to double check - I believe existing logic for "fileLocation" is to have the potentially different lat/long for a property in a file (regardless of original lat/long at property level). As such Concept_FileProperty has its own **location** property. 

I think the IMapProperty mapLocation should be assigned the **fileProperty.location** (instead of **fileProperty.property?.location**)

Could you please check with current implementation?

asanchezr (approved):
Approved with minor comments. Also please rebase and fix merge conflicts

PR #4859, opened 2025-06-19
PSP-10589 : FT-REG: Management Files Activity - Properties at the Act…

eddherrera:
…ivity Details level are not saved properly


source/frontend/src/features/mapSideBar/management/tabs/activities/edit/ManagementActivityEditContainer.tsx
```diff
@@ -61,17 +62,31 @@ export const ManagementActivityEditContainer: React.FunctionComponent<
...
+        return {
+          id: x.id,
+          fileId: castedFile.id,
+          propertyName: x.propertyName,
+          location: x.location,
+          displayOrder: x.displayOrder,
+          property: x.property,
+          propertyId: x.propertyId,
+        } as ApiGen_Concepts_FileProperty;
+      });
```
asanchezr, line 82: I like this approach but have a couple of minor comments. I'd like to see a null check before using `castedFile.fileProperties` to defend against NPE. Something like `(castedFile.fileProperties ?? []).map(...)`

The second comment is why we have to "force" the typing with "as ApiGen_Concepts_FileProperty;" - we usually do that when there are additional properties we don't want to populate but they are present in the type. 
If you specified all the properties then I don't think we need the "as ..." - thoughts?

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -82,17 +85,29 @@ export class ManagementActivityFormModel {
             ...getEmptyBaseAudit(0),
           };
         }),
-      activityProperties: this.activityProperties
-        .filter(exists)
-        .map<ApiGen_Concepts_PropertyActivityProperty>(x => x.toApi()),
+      activityProperties: this.selectedProperties.map(x => {
+        const matched =
+          this.activityProperties.filter(y => y.propertyId === x.propertyId)[0] ?? null;
```
asanchezr, line 90: it seems you are trying to find a single element in the array - why not use `Array.find()` instead of `filter[0]`?

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -128,6 +143,21 @@ export class ManagementActivityFormModel {
     formModel.activityProperties =
       model.activityProperties?.map(p => ActivityPropertyFormModel.fromApi(p)) ?? [];
 
+    formModel.selectedProperties = model.activityProperties.map(x => {
+      const matchProperty =
+        fileProperties.filter(
+          y => y.fileId === formModel.managementFileId && y.propertyId === x.propertyId,
+        )[0] ?? null;
```
asanchezr, line 150: same comment re: `Array.find()`

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -82,17 +85,29 @@ export class ManagementActivityFormModel {
...
+        const matched =
+          this.activityProperties.filter(y => y.propertyId === x.propertyId)[0] ?? null;
+
+        return {
+          id: matched ? matched.id : 0,
+          propertyActivityId: this.id ?? 0,
+          propertyId: x.propertyId,
+          property: null,
+          rowVersion: matched.rowVersion ?? 0,
+        } as ApiGen_Concepts_PropertyActivityProperty;
```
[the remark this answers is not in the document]
eddherrera: not needed but I would have to assign all properties which are not needed since they only need the Foreign keys from the selected array. 
These are not the same, the selection of properties comes from a FileProperty[].  The mapping of models is  PropertyActivityProperty => FileProperty[] => PropertyActivityProperty[].

source/frontend/src/features/mapSideBar/management/tabs/activities/edit/models.ts
```diff
@@ -128,6 +143,21 @@ export class ManagementActivityFormModel {
...
+          y => y.fileId === formModel.managementFileId && y.propertyId === x.propertyId,
+        )[0] ?? null;
+
+      return {
+        id: matchProperty ? matchProperty.id : 0,
+        fileId: formModel.managementFileId,
+        propertyId: x.propertyId,
+        property: x.property,
+        rowVersion: matchProperty?.rowVersion ?? 0,
+      } as ApiGen_Concepts_ManagementFileProperty;
```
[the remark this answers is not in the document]
eddherrera: not needed but I would have to assign all properties which are not needed since they only need the Foreign keys from the selected array.
These are not the same, the selection of properties comes from a FileProperty[]. The mapping of models is PropertyActivityProperty => FileProperty[] => PropertyActivityProperty[].

asanchezr (approved):

PR #4863, opened 2025-06-20
PSP-10592 Add new layers: Parcel by Class, Parcel by Owner Type

asanchezr:
# Parcels by Class
pmbc_parcel_by_class

# Parcels by Owner Type
pmbc_parcel_by_owner_type


devinleighsmith (approved):

asanchezr:
> @asanchezr can you log a placeholder story to add legends for these layers? There are so many details and colours available here, which is great, but unless the user is familiar with parcel map I'm not sure how they would know the meanings behind that.

Agreed - for context there are a number of other layers with multiple colors as well (mostly from Tantalis crown lands). I recall we discussed perhaps having a link or a target off the layer list to show legends. Will need to work with Jordan on UI/UX

PR #4864, opened 2025-06-24
PSP-10568 | Added buttons to zoom into properties from files

FuriousLlama:
image


source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -169,13 +169,13 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
...
-        mapRef?.current?.flyToBounds(requestedFlyTo.bounds, { animate: false });
+        mapRef?.current?.flyToBounds(requestedFlyTo.bounds, { animate: true });
       }
       if (requestedFlyTo.location !== null) {
         mapRef?.current?.flyTo(requestedFlyTo.location, MAP_MAX_ZOOM, {
-          animate: false,
+          animate: true,
         });
-        mapRef?.current?.panTo(requestedFlyTo.location);
+        //mapRef?.current?.panTo(requestedFlyTo.location);
```
asanchezr, line 178: if this is no longer needed - please remove instead of commenting out

source/frontend/src/features/mapSideBar/shared/FileMenuView.tsx
```diff
@@ -53,44 +96,56 @@ const FileMenuView: React.FunctionComponent<React.PropsWithChildren<IFileMenuPro
...
               <Col xs="auto" className="pr-2">
-                <StyledIconWrapper className={cx({ selected: currentPropertyIndex === index })}>
+                <StyledIconWrapper className={cx({ selected: isCurrentProperty })}>
                   {index + 1}
                 </StyledIconWrapper>
               </Col>
-              <Col>
-                {currentPropertyIndex === index ? (
+              <StyledCol
+                onClick={() => onPropertyClick(index, property.id)}
```
asanchezr, line 135: nit: why not move the onclick handler to the LinkButton below? Feels odd to have an onclick handler that doesn't fire when the span is rendered - `isCurrentProperty === true`

source/frontend/src/features/mapSideBar/shared/FileMenuView.tsx
```diff
@@ -38,6 +48,39 @@ const FileMenuView: React.FunctionComponent<React.PropsWithChildren<IFileMenuPro
   // respect the order of properties as set by the user creating the file
   const sortedProperties = sortFileProperties(file?.fileProperties ?? []);
   const isSummary = useMemo(() => !exists(currentPropertyIndex), [currentPropertyIndex]);
+  const mapMachine = useMapStateMachine();
+
+  const fitBoundaries = () => {
+    const fileProperties = file.fileProperties;
```
asanchezr, line 54: nit please change to `file?.fileProperties` to avoid null errors (like the rest of the code here)

source/frontend/src/features/mapSideBar/shared/FileMenuView.tsx
FuriousLlama: This is to match the previous way the tests were selecting for the class and the testid. I can change it needed.

devinleighsmith:
I'm not in love with how the buttons look with a long property name.
However, I think it is fair to say that is outside the scope of your changes, since the line just extends into the next component. I'll log a small followup for this:

image

devinleighsmith:
The fit boundaries button only seems to work on properties that existed before the edit button was clicked. ie:
go to a file with two properties

1. click the "fit map to the file boundaries" button, works as expected.
2. add a new property that is outside of the boundary from #1
3. when clicking the "fit map to the file boundaries" button, the map zooms to only show the original properties, and ignores the new property.

devinleighsmith:
given that this has been merged, will log as a bug.

PR #4878, opened 2025-07-02
document queue refactor - resiliency update

devinleighsmith:
Major changes:

- Abort a new batch upload if one or more documents is still processing from the previous batch (scheduler)
- Use a distributed lock on the document queue update to avoid trigger-based deadlocking issues.
- Refactor main document queue service to simplify methods.


source/backend/dal/Repositories/DocumentQueueRepository.cs
```diff
@@ -1,13 +1,17 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Security.Claims;
+using DocumentFormat.OpenXml.Office2010.Excel;
```
asanchezr, line 1: This import doesn't look relevant - please remove

PR #4880, opened 2025-07-03
psp-10487 address major memory useage outliers.

devinleighsmith:


source/backend/api/Pims.Api.csproj
```diff
@@ -7,6 +7,8 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <ProjectGuid>{16BC0468-78F6-4C91-87DA-7403C919E646}</ProjectGuid>
     <TargetFramework>net8.0</TargetFramework>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
```
devinleighsmith, line 10: The default GC setting is actually "workstation" which doesn't seem appropriate for our case. See here: 
https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/workstation-server-gc

source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -257,8 +257,9 @@ public async Task<ExternalResponse<DocumentDetailModel>> TryUploadDocumentAsync(
             string authenticationToken = await _authRepository.GetTokenAsync();
 
             byte[] fileData;
-            using var byteReader = new BinaryReader(file.OpenReadStream());
-            fileData = byteReader.ReadBytes((int)file.OpenReadStream().Length);
```
devinleighsmith, line 261: This opens an extra (untracked) readstream just to get the stream length.

source/backend/core.api/Middleware/LogResponseMiddleware.cs
```diff
@@ -62,11 +62,12 @@ private async Task LogResponse(HttpContext context)
 
             await _next(context);
 
-            context.Response.Body.Seek(0, SeekOrigin.Begin);
-            using var reader = new StreamReader(context.Response.Body);
             string body = null;
-            if (reader.BaseStream.Length < maxStreamLength)
+            if (context.Response.ContentLength < maxStreamLength)
```
devinleighsmith, line 66: similar story here - larger responses end up persisting in memory after the request is handled in these logging processes.

asanchezr (approved):

PR #4896, opened 2025-07-09
do not copy stream, just apply directly to output content - to reduce…

devinleighsmith:
… memory footprint.

### DIFF
--- a/source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
+++ b/source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
@@ -256,19 +256,14 @@ namespace Pims.Api.Repositories.Mayan
             _logger.LogDebug("Uploading document {documentType}...", documentType);
             string authenticationToken = await _authRepository.GetTokenAsync();
 
-            byte[] fileData;
             using Stream stream = file.OpenReadStream();
-            using var byteReader = new BinaryReader(stream);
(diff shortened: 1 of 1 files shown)


source/backend/api/Repositories/Mayan/MayanDocumentRepository.cs
```diff
@@ -256,19 +256,14 @@ public async Task<ExternalResponse<DocumentDetailModel>> TryUploadDocumentAsync(
             _logger.LogDebug("Uploading document {documentType}...", documentType);
             string authenticationToken = await _authRepository.GetTokenAsync();
 
-            byte[] fileData;
             using Stream stream = file.OpenReadStream();
-            using var byteReader = new BinaryReader(stream);
```
devinleighsmith, line 261: even though this wasn't leaking memory, it did double the memory footprint while processing a file.

source/backend/scheduler/Startup.cs
```diff
@@ -267,6 +267,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddHangfireServer(options =>
             {
                 options.Queues = hangfireOptions.Queues;
+                options.WorkerCount = 1;
```
devinleighsmith, line 270: Runs the scheduled document tasks serially, for now. This prevents very large files from being uploaded twice before PIMS can set the processing flag due to long clamav processing time.

asanchezr (approved):

PR #4899, opened 2025-07-09
Allow admins to change a file to an active status.

devinleighsmith:


source/backend/api/Services/ManagementFileService.cs
```diff
@@ -289,7 +289,7 @@ private void ValidateStatus(ManagementFileStatusTypes? currentManagementStatus,
             }
 
             // Not Editable -- Status must be updated first
-            if (!_managementStatusSolver.CanEditDetails(currentManagementStatus))
```
devinleighsmith, line 292: This was throwing this error for admins too - preventing admins from changing back to an active state.
Should only prevent inactive state -> inactive state (for admins and non-admins).

eddherrera (approved):

PR #4900, opened 2025-07-09
psp-10559 | Property Quick information

FuriousLlama:
image


source/frontend/src/components/maps/leaflet/Control/Search/PropertyQuickInfoContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const hasMultipleProperties = useMemo(
+    () => mapLocationFeatureDataset?.parcelFeatures?.length > 1,
+    [mapLocationFeatureDataset?.parcelFeatures?.length],
+  );
+
+  const getOwnerInfo = useCallback(
+    async (pid: string) => {
+      if (isValidString(pid)) {
+        const ltsaOrders = await getLtsaExecute(pid);
+        console.log(ltsaOrders.titleOrders);
```
devinleighsmith, line 40: a couple console.logs left in this PR.

source/frontend/src/components/maps/leaflet/Control/Search/PropertyQuickInfoContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { useMapStateMachine } from '@/components/common/mapFSM/MapStateMachineContext';
+import { SectionField } from '@/components/common/Section/SectionField';
+import TooltipWrapper from '@/components/common/TooltipWrapper';
+import usePathGenerator from '@/features/mapSideBar/shared/sidebarPathGenerator';
+import { useLtsa } from '@/hooks/useLtsa';
+import { exists, firstOrNull, isValidString, pidFormatter } from '@/utils';
+import { formatNames } from '@/utils/personUtils';
+
+export const PropertyQuickInfoContainer: React.FC<React.PropsWithChildren> = () => {
+  const [ownerNames, setOwnerNames] = useState('');
```
devinleighsmith, line 16: Arguable here, but I prefer to store objects in their "raw" unformatted state instead of storing formatted objects intended for display - that way multiple consumers can access the raw data and format as needed. There may not be a requirement for multiple consumers here however.

source/frontend/src/components/maps/leaflet/Control/Search/PropertyQuickInfoContainer.tsx
devinleighsmith: technically you could have ismounted checks here, but not a pattern we've been following consistently.

source/frontend/src/components/maps/leaflet/Control/Search/PropertyQuickInfoContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+          </Row>
+          <SectionField label={'Owners'} labelWidth={{ xs: 12 }}>
+            {ownerNames ?? '-'}
+          </SectionField>
+          <SectionField label={'Legal Description'} labelWidth={{ xs: 12 }}>
+            {locationInfo?.LEGAL_DESCRIPTION ?? '-'}
+          </SectionField>
+        </StyledInfoWrapper>
+      )}
+      {hasMultipleProperties && <div className="pt-8">Multiple properties found</div>}
```
devinleighsmith, line 135: Hmm, how are we handling multiple properties for this component? does the map popup still exist for this use case maybe?

devinleighsmith: Nvm, I see that logic was removed. Should I log a story to re-introduce that logic with the new quick-info workflow?

devinleighsmith:
Well, this ended up being nice and simple, just maybe need a few follow ups to make this functional for all of our workflows. Great start!

devinleighsmith (approved):

source/frontend/src/components/maps/leaflet/Control/Search/PropertyQuickInfoContainer.tsx
FuriousLlama: we dont have a solution for the multi properties atm

PR #4907, opened 2025-07-14
PSP-10336: List-view Tables – File Status Field Default Setting

stairaku:
image
image
image
image
image


asanchezr (approved):
Thanks for fixing this! Great work

PR #4914, opened 2025-07-15
use generated codeql file.

devinleighsmith:


devinleighsmith:
Turns out we were using codeql - just with default settings.
Default settings don't correctly support forked prs: https://github.com/github/codeql-action/issues/2136

This should correct that.

PR #4921, opened 2025-07-25
PSP-10560 unit tests

asanchezr:


source/frontend/src/features/properties/worklist/context/WorklistContext.tsx
```diff
@@ -10,6 +10,12 @@ import {
 
 import { ParcelFeature } from '../models';
 
+export interface IWorklistNotifier {
+  error: (msg: string) => void;
+  success: (msg: string) => void;
+  warn: (msg: string) => void;
+}
+
```
FuriousLlama, line 18: this is is changing more than unittests. Is it relevant to this pr?

asanchezr: This is relevant because we need dependency injection in tests. I tried directly mocking **react-toastifier** and it didn't work in tests. It is good practice anyway to pass as property - which already defaults to react-toastifier in the runtime code but can be overriden in tests with `vi.fn()`

FuriousLlama (approved):

PR #4926, opened 2025-07-28
PSP-10707 : Property information edit screen is still displaying comments field

eddherrera:


source/frontend/src/features/mapSideBar/property/tabs/propertyDetails/update/UpdatePropertyDetailsForm.tsx
```diff
@@ -362,10 +362,6 @@ export const UpdatePropertyDetailsForm: React.FunctionComponent<
           </SectionField>
         )}
       </Section>
-
-      <Section header="Comments">
-        <TextArea field="notes" rows={4} />
-      </Section>
```
asanchezr, line 368: thanks for removing the UI but I would expect also modifications to the model that gets submitted to the server? ie **UpdatePropertyDetailsFormModel**

asanchezr (approved):

PR #4929, opened 2025-07-30
PSP-10639: Changes on frontend to add data-testids and first Playwright project commit

stairaku:


stairaku:
I get one error from SonarCube but it is complaining because the login page uses the word "password" on one of the inputs. Unfortunately, we cannot change anything from the login page, right?

PR #4933, opened 2025-08-05
PSP-10695: Management Tab Activities List Section:  update Activities List Section Heading and the “Add an Activity” Button Wording

stairaku:
image
image
image


source/frontend/src/components/common/SectionListHeader.tsx
```diff
@@ -47,7 +47,7 @@ export const SectionListHeader: React.FunctionComponent<
 };
 
 const StyledRow = styled(Row)`
-  display: grid;
+  display: flex;
   grid-template-columns: 1fr 0.5fr;
```
asanchezr, line 51: I'm fine with the change from grid to flex. However, the next line "grid-template-column" only applies to grids - wither restore the display grid or delete this line. Also please make sure that the layout is not broken with this change. thanks!

asanchezr (approved):
See comments

PR #4934, opened 2025-08-05
PSP-10709: (POC) Hide Advanced Filter

stairaku:
image


source/frontend/src/components/maps/MapLeafletView.tsx
```diff
@@ -274,10 +273,10 @@ const MapLeafletView: React.FC<React.PropsWithChildren<MapLeafletViewProps>> = (
         <LegendControl />
         <ZoomOutButton />
         <ScaleControl position="bottomleft" metric={true} imperial={false} />
-        <AdvancedFilterButton
+        {/* <AdvancedFilterButton
           onToggle={mapMachine.toggleMapFilterDisplay}
           active={mapMachine.isFiltering}
-        />
+        /> */}
```
asanchezr, line 279: I'd rather remove the code instead of commenting it out - we have git history if we ever want to bring it back

asanchezr (approved):

PR #4937, opened 2025-08-06
PSP-8338: UI UX Clean Up - Verbiage - MOTI to MOTT

stairaku:


.github/workflows/deploy-prod-start-argo.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           namespace: ${{ env.OPENSHIFT_TOOLS_NAMESPACE }}
       - name: Display the maintenance page instead of the app
         run: |
-          ./tools/cicd/maintenance/maintenance.sh prod on
+          APPLICATION_NAME=pims-app-prod ./tools/cicd/maintenance/maintenance.sh prod on
```
asanchezr, line 71: this seems odd - did you actually make this change or perhaps you created your local branch from **test** instead of **dev** ? I ask because I see weird commit history in this PR as well...

Image

stairaku: hmmm .... I created this branch from dev :( will try to take out that file from my PR

PR #4977, opened 2025-08-26
PSP-10792 : [HOTFIX]FT: Results are not populated with Search by plan  type

eddherrera:


source/frontend/src/components/maps/leaflet/Layers/HighwayParcelsLayer.tsx
```diff
@@ -12,10 +12,10 @@ export const HighwayParcelsLayer: React.FunctionComponent = () => {
 
   // For now, lat/long properties in the highway list will not display on the map
   // Ignore properties without a valid boundary
-  const validParcels = useMemo<Feature<Geometry, ISS_ProvincialPublicHighway>[]>(
-    () => (mapFeatureData?.highwayPlanFeatures.features ?? []).filter(p => exists(p?.geometry)),
-    [mapFeatureData?.highwayPlanFeatures.features],
-  );
+  const validParcels = useMemo<Feature<Geometry, ISS_ProvincialPublicHighway>[]>(() => {
+    const features = mapFeatureData?.highwayPlanFeatures.features ?? [];
```
FuriousLlama, line 16: If highwayPlanFeatures in null or undefined the '.features' part will still fail. I think that  is the current issue atm

source/frontend/src/components/maps/leaflet/Layers/HighwayParcelsLayer.tsx
```diff
@@ -13,7 +13,7 @@ export const HighwayParcelsLayer: React.FunctionComponent = () => {
   // For now, lat/long properties in the highway list will not display on the map
   // Ignore properties without a valid boundary
   const validParcels = useMemo<Feature<Geometry, ISS_ProvincialPublicHighway>[]>(() => {
-    const features = mapFeatureData?.highwayPlanFeatures.features ?? [];
+    const features = mapFeatureData?.highwayPlanFeatures?.features ?? [];
     return features.filter(p => exists(p?.geometry)) ?? [];
   }, [mapFeatureData]);
```
asanchezr, line 18: I still think you should revert to the old implementation. This memo will only be recomputed when the **whole** `mapFeatureData` object changes. I would think that we want to re-run this memo when `mapFeatureData?.highwayPlanFeatures?.features` changes

eddherrera: Features most likely wouldn't change on itself, but I don't mind.

eddherrera: Updated with @asanchezr suggestions.

asanchezr (approved):

PR #4983, opened 2025-08-27
PSP-10750: List view search - open result in new tab 

stairaku:
image
image
image
image
image
image
image


source/frontend/src/features/acquisition/list/AcquisitionSearchResults/columns.tsx
```diff
@@ -38,8 +39,14 @@ export const columns: ColumnWithProps<AcquisitionSearchResultModel>[] = [
       const { hasClaim } = useKeycloakWrapper();
       if (hasClaim(Claims.ACQUISITION_VIEW)) {
         return (
-          <Link to={`/mapview/sidebar/acquisition/${props.row.original.id}`}>
+          <Link
+            to={`/mapview/sidebar/acquisition/${props.row.original.id}`}
+            target="_blank"
+            rel="noopener noreferrer"
+            style={{ display: 'flex', alignItems: 'center', gap: '8px' }}
+          >
```
asanchezr, line 47: **nit:** we usually avoid inline styles when we can. I would suggest using styles-component instead

```tsx
const StyledLink = styled(Link)`
  display: flex:
  ...
`;
```

source/frontend/src/features/disposition/list/DispositionSearchResults/columns.tsx
```diff
@@ -35,8 +36,14 @@ export const columns: ColumnWithProps<DispositionSearchResultModel>[] = [
...
         return (
-          <Link to={`/mapview/sidebar/disposition/${props.row.original.id}`}>
+          <Link
+            to={`/mapview/sidebar/disposition/${props.row.original.id}`}
+            target="_blank"
+            rel="noopener noreferrer"
+            style={{ display: 'flex', alignItems: 'center', gap: '8px' }}
+          >
             D-{props.row.original.fileNumber}
+            <FaExternalLinkAlt />
```
asanchezr, line 46: There is a level of repetition for all these Link component. I wonder if it would make sense to create a reusable component for "ExternalLinks" such as:

```tsx
import { Link } from 'react-router-dom';
import { FaExternalLinkAlt } from 'react-icons/fa';

export interface ExternalLinkProps {
  to: string;
  children: React.ReactNode;
  className?: string;
  target?: string;
  rel?: string;
}

const StyledLink = styled(Link)`
  display: flex;
  align-items: center;
  gap: 8px;
  text-decoration: none;
  color: inherit;

  &:hover {
    text-decoration: underline;
  }
`;

export const ExternalLink: React.FC<ExternalLinkProps> = ({
  to,
  children,
  className,
  target = '_blank',
  rel = 'noopener noreferrer',
}) => {
  return (
    <StyledLink to={to} target={target} rel={rel} className={className}>
      {children}
      <FaExternalLinkAlt />
    </StyledLink>
  );
};
```

Usage:

```tsx
<ExternalLink to={`/mapview/sidebar/lease/${props.row.original.id}`}>
  {props.row.original.lFileNo}
</ExternalLink>

```

asanchezr (approved):
See comments

PR #4999, opened 2025-09-10
psp-10818 allow users to add property to an open file (new, existing).

devinleighsmith:


asanchezr (approved):
Looks good to me but please rebase and ensure this works well with Manuel's changes to new workflow creation

PR #5004, opened 2025-09-17
PSP-10834 : Finalize "add property to file logic"

eddherrera:
image


source/frontend/src/features/mapSideBar/shared/update/properties/UpdateProperties.tsx
```diff
@@ -166,6 +182,34 @@ export const UpdateProperties: React.FunctionComponent<IUpdatePropertiesProps> =
...
+    mapLocationFeatureDataset?.pimsFeatures,
+    mapLocationFeatureDataset?.regionFeature,
+    mapLocationFeatureDataset?.districtFeature,
+    mapLocationFeatureDataset?.municipalityFeatures,
+  ]);
+
+  const handleAddToSelection = useCallback(() => {
+    prepareForCreation([selectedFeatureDataset]);
+  }, [prepareForCreation, selectedFeatureDataset]);
+
```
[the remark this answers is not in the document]
eddherrera: Updated

source/frontend/src/features/mapSideBar/disposition/form/DispositionPropertiesSubForm.tsx
```diff
@@ -22,7 +30,55 @@ export interface DispositionPropertiesSubFormProps {
 const DispositionPropertiesSubForm: React.FunctionComponent<DispositionPropertiesSubFormProps> = ({
   formikProps,
 }) => {
-  const mapMachine = useMapStateMachine();
+  const localRef = useRef<FormikProps<DispositionFormModel>>(null);
```
asanchezr, line 33: This "localRef" gets created with null value and later in the useEffect we check to see if it not null. When is this assigned a value? I'm not sure I follow the code here

source/frontend/src/features/mapSideBar/disposition/form/DispositionPropertiesSubForm.tsx
```diff
@@ -33,12 +89,23 @@ const DispositionPropertiesSubForm: React.FunctionComponent<DispositionPropertie
...
     }
   };
 
+  const handleAddToSelection = useCallback(() => {
+    prepareForCreation([selectedFeatureDataset]);
+  }, [prepareForCreation, selectedFeatureDataset]);
+
+  useEffect(() => {
+    if (exists(localRef.current) && propertyForms.length > 0) {
+      addPropertiesToCurrentFile(localRef, 'fileProperties', propertyForms, noop);
```
asanchezr, line 102: I think the problem here might stem from the fact that "addPropertiesToCurrentFile" function is expecting a formikRef but this component receives the FormikProps directly so perhaps we would need another "addProperties" function that takes formikprops instead of a ref?

source/frontend/src/features/mapSideBar/disposition/form/DispositionPropertiesSubForm.tsx
eddherrera: I removed the null, which as you said didn't quite made sense.

PR #5010, opened 2025-09-23
PSP-10860 : FT:AcquisitionFile:  New workflow is not populated under "Properties to include in this file" on the create screen of Acq file

eddherrera:


eddherrera:
> @eddherrera unrelated, but wasn't the "new workflow" section supposed to be expanded by default for this release?

Good catch. Added and updated.

PR #5013, opened 2025-09-25
PSP-10831 Activity list view invoice report

asanchezr:
activities_invoices_report


source/backend/api/Areas/Reports/Models/Management/ManagementActivityInvoicesReportModel.cs
```diff
@@ -0,0 +1,181 @@
...
+        private static string GetNullableDate(DateOnly? dateTime)
+        {
+            return dateTime.HasValue ? dateTime.Value.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture) : string.Empty;
+        }
+
+        private static string GetPropertiesAsString(PimsManagementActivity activity)
+        {
+            if (activity.ManagementFile is not null && activity.ManagementFile.PimsManagementFileProperties is not null)
+            {
+                return string.Join("|", activity.ManagementFile.PimsManagementFileProperties
```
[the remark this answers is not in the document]
asanchezr: I'm fine taking out the "old behaviour" and keeping it only with the activity-selected properties.  will make the change for this report and the other one. 
The story is a bit ambiguous in that it calls for:

> o.      Property (PID# and street address) PIMS_MANAGEMENT_FILE_PROPERTY (roll up if multiple, separated by '|') 

so it lends to try to get the property info from the file. But I will make the change to only show the ones selected for the activity

source/frontend/src/features/management/activities/list/ManagementActivitiesListView.tsx
```diff
@@ -83,6 +92,21 @@ export const ManagementActivitiesListView: React.FC<unknown> = () => {
...
+  useEffect(() => {
+    if (invoiceStatus === 204) {
+      setModalContent({
+        variant: 'error',
+        title: 'Warning',
+        message: 'There is no data for the input parameters you entered.',
+        okButtonText: 'Close',
+        handleOk: () => setDisplayModal(false),
+      });
+      setDisplayModal(true);
```
devinleighsmith, line 104: in my test of this logic, this modal was not able to be dismissed. Also, the warning variant may be more appropriate here. Let me know if you need reproduction steps.

asanchezr: hmm - I'll look into it. I copied this from another export logic in another list view

devinleighsmith:
I see some commits on the PR but I'm still seeing the issue with the popup not being dismissible - I'll keep an eye here for more updates.

asanchezr:
> I see some commits on the PR but I'm still seeing the issue with the popup not being dismissible - I'll keep an eye here for more updates.

PR #5027, opened 2025-10-06
PSP-10639: Smoke test on playwright + adding data-testid on UI elements

stairaku:


asanchezr (approved):
Approved but please fix the CodeQL warnings about unused imports like the one shown below

Image

PR #5201, opened 2026-02-10
PSP-8129: For each renewal, make the expiry date optional.

stairaku:


source/backend/api/Services/LeaseService.cs
```diff
@@ -541,13 +541,13 @@ private static void ValidateRenewalDates(PimsLease lease, PimsLease currentLease
...
                 if (renewal.IsExercised == true)
                 {
-                    if (renewal.CommencementDt.HasValue && renewal.ExpiryDt.HasValue)
+                    if(!renewal.CommencementDt.HasValue)
                     {
-                        renewalDates.Add(new Tuple<long, DateTime, DateTime>(renewal.LeaseRenewalId, renewal.CommencementDt.Value, renewal.ExpiryDt.Value));
+                        throw new BusinessRuleViolationException("Exercised renewals must have a commencement date");
                     }
-                    else
+                    else if (renewal.CommencementDt.HasValue && renewal.ExpiryDt.HasValue)
```
devinleighsmith: Does this work? I feel like if a user enters just a commencement date that isn't going to get saved.

devinleighsmith (approved):

PR #5262, opened 2026-03-23
PSP-9438 : System is throwing error message when user deletes the tenant contact  as that contact is already associated with payee compensation

eddherrera:


source/backend/api/Services/LeaseService.cs
```diff
@@ -265,18 +268,16 @@ public PimsLease Update(PimsLease lease, IEnumerable<UserOverrideCode> userOverr
...
             _leaseRepository.Update(lease, false);
-
             _leaseRepository.UpdateLeaseRenewals(lease.Internal_Id, lease.ConcurrencyControlNumber, lease.PimsLeaseRenewals);
-
             _leaseRepository.CommitTransaction();
 
             return _leaseRepository.GetNoTracking(lease.LeaseId);
         }
 
+
```
asanchezr: nit: please remove extra line break

source/backend/api/Services/LeaseService.cs
```diff
@@ -530,6 +531,30 @@ public IEnumerable<PimsLeaseLicenseTeam> GetTeamMembers()
...
+        {
+            if ((currentLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.PYBLBCTFA.ToString() || currentLease.LeasePayRvblTypeCode.ToString() == LeasePaymentReceivableTypes.PYBLMOTI.ToString())
+                && updatedLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.RCVBL.ToString())
+            {
+                if (currentLease.PimsCompensationRequisitions.Count > 0 || currentLease.PimsLeaseStakeholders.Count > 0)
+                {
+                    throw new BusinessRuleViolationException("This file has active Payees or Compensation Requisitions which should be removed prior to switching to a 'Receivable' Lease.");
+                }
+            }
+            else if (currentLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.RCVBL.ToString() && currentLease.PimsLeaseStakeholders.Count > 0)
```
asanchezr: I believe the **else if** branch doesn't check if the type is actually changing...

The second condition checks if the current type is RCVBL and there are stakeholders, but it never verifies that the updated type is actually being switched to a Payable type. This means the exception would fire even if the user is just saving the lease without changing the type at all, as long as it's Receivable and has stakeholders. Could you please take a look?

A suggestion to simplify the logic could be to store the state change in a couple of boolean variables to make it clearer to understand perhaps?

```cs
    bool isChangingToReceivable =
        (currentLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.PYBLBCTFA.ToString() ||
         currentLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.PYBLMOTI.ToString()) &&
        updatedLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.RCVBL.ToString();

    bool isChangingToPayable =
        currentLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.RCVBL.ToString() &&
        (updatedLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.PYBLBCTFA.ToString() ||
         updatedLease.LeasePayRvblTypeCode == LeasePaymentReceivableTypes.PYBLMOTI.ToString());

    if (isChangingToReceivable)
    {
       // throw #1
    } else if (isChangingToPayable)
    {
        // throw #2
    }
```

source/backend/tests/dal/Repositories/LeaseRepositoryTest.cs
```diff
@@ -1,7 +1,4 @@
-using System;
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
-using System.Linq;
+using ClosedXML.Excel;
```
asanchezr: nit: do we really need this import?

asanchezr (approved):
Looks good but please address comments

source/backend/api/Services/LeaseService.cs
eddherrera: the check for change is outside this method.

PR #5268, opened 2026-03-26
psp-11362 limit document global search visibility based on file/regio…

devinleighsmith:
…n/contractor permissions.


source/backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -310,6 +298,172 @@ private IQueryable<PimsDocument> GetCommonQueryDeep(DocumentSearchFilterModel fi
...
+
+            if (excludeOrphans)
+            {
+                predicate = predicate.And(x => x.PimsAcquisitionFileDocuments.Count > 0 || x.PimsDispositionFileDocuments.Count > 0 || x.PimsLeaseDocuments.Count > 0
+                                            || x.PimsManagementFileDocuments.Count > 0 || x.PimsMgmtActivityDocuments.Count > 0 || x.PimsProjectDocuments.Count > 0
+                                            || x.PimsPropertyDocuments.Count > 0 || x.PimsResearchFileDocuments.Count > 0);
+            }
+
+            return predicate;
+        }
```
asanchezr, line 471: Thanks for refactoring this logic to its own method but I think the logic is a bit different than the original code. We have two boolean values **excludeTemplates** and **excludeOrphans**. 

In the old code, you could have excludeTemplates == false (not running that branch of code) but still be able to excludeOrphans.

This new method aborts early if excludeTemplates is ever false, so no chance for excludeOrphans to run unless excludeTemplates is true

source/backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -310,6 +298,172 @@ private IQueryable<PimsDocument> GetCommonQueryDeep(DocumentSearchFilterModel fi
...
+            {
+                targetAccessPredicate = targetAccessPredicate.Or(PropertyAccessPredicate());
+            }
+
+            return targetAccessPredicate;
+        }
+
+        private ExpressionStarter<PimsDocument> AcquisitionAccessPredicate(HashSet<short> userRegions, long? contractorPersonId)
+        {
+            _ = Context;
```
asanchezr: sorry - why do we need this here? I don't see any usage of ```_```

```cs
_ = Context;
```

source/backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -310,6 +298,172 @@ private IQueryable<PimsDocument> GetCommonQueryDeep(DocumentSearchFilterModel fi
...
+            return PredicateBuilder.New<PimsDocument>(doc =>
+                doc.PimsAcquisitionFileDocuments.FirstOrDefault(rel =>
+                    (userRegions.Contains(rel.AcquisitionFile.RegionCode) || rel.AcquisitionFile.RegionCode == 4)
+                    && (!contractorPersonId.HasValue
+                        || rel.AcquisitionFile.PimsAcquisitionFileTeams.Any(t => t.PersonId == contractorPersonId)
+                        || (rel.AcquisitionFile.Project != null
+                            && rel.AcquisitionFile.Project.PimsProjectPeople.Any(p => p.PersonId == contractorPersonId))))
```
asanchezr: This logic looks good but the query on line 273 is not bringing any data under the document -> AcquisitionFile. I mean the query is pulling the following:

```cs
var query = Context.PimsDocuments.AsNoTracking()
                .Include(x => x.PimsAcquisitionFileDocuments)
                    .ThenInclude(y => y.AcquisitionFile)
```
Yet this extra predicates check for `doc.PimsAcquisitionFileDocuments[].AcquisitionFile.PimsAcquisitionFileTeams[]` and `doc.PimsAcquisitionFileDocuments[].AcquisitionFile.Project.PimsProjectPeople[]` - wouldn't the checks below be working against "empty" data? 

Did you test that you wouldn't need to modify the original query to also pull the inner data via `ThenInclude()`?

```cs
// something like this
var query = Context.PimsDocuments.AsNoTracking()
                .Include(x => x.PimsAcquisitionFileDocuments)
                    .ThenInclude(y => y.AcquisitionFile)
                        .ThenInclude(y => y.Project)
                            .ThenInclude(y => y.PimsProjectPeople)
```

source/backend/dal/Repositories/DocumentRepository.cs
```diff
@@ -310,6 +298,172 @@ private IQueryable<PimsDocument> GetCommonQueryDeep(DocumentSearchFilterModel fi
...
+                    || rel.DispositionFile.PimsDispositionFileTeams.Any(t => t.PersonId == contractorPersonId)))
+                != null);
+        }
+
+        private ExpressionStarter<PimsDocument> LeaseAccessPredicate(HashSet<short> userRegions)
+        {
+            _ = Context;
+            return PredicateBuilder.New<PimsDocument>(doc => doc.PimsLeaseDocuments.FirstOrDefault(rel =>
+                !rel.Lease.RegionCode.HasValue || userRegions.Contains(rel.Lease.RegionCode.Value) || rel.Lease.RegionCode == 4) != null);
+        }
```
asanchezr: Lease predicate is missing the check for Lease Team member and Lease Project team member...

Image

asanchezr (changes requested):
See comments

source/backend/dal/Repositories/DocumentRepository.cs
devinleighsmith: This was an automated fix I applied, but I agree that the solution doesn't make sense, I'll make these methods static instead.

source/backend/dal/Repositories/DocumentRepository.cs
devinleighsmith: good find.

source/backend/dal/Repositories/DocumentRepository.cs
devinleighsmith: huh, I thought that for leases we weren't using the team for access restrictions, but checking the code I can see I'm wrong, will update.

source/backend/dal/Repositories/DocumentRepository.cs
devinleighsmith: lazy loading affects the structure of data returned by an ef query, not the ef query's ability to filter data. I've confirmed that I'm able to filter by project team membership without having an explicit include.

asanchezr (approved):

PR #5272, opened 2026-04-01
PSP-11331: Backend implementation for Notifications

areyeslo:
- Add CRUD operations for notifications.
- Add validation that only one file id and sub-id is added during creation:
```
---------------------------------------------------------------------------
| Subtype                     | File                                      |
---------------------------------------------------------------------------
| AGREEMENT_ID                | ACQUISITION_FILE_ID or DISPOSITION_FILE_ID |
| LEASE_CONSULTATION_ID       | LEASE_ID                                  |
| LEASE_RENEWAL_ID            | LEASE_ID                                  |
| NOTICE_OF_CLAIM_ID          | ACQUISITION_FILE_ID or MANAGEMENT_FILE_ID |
| INSURANCE_ID                | LEASE_ID                                  |
| EXPROP_OWNER_HISTORY_ID     | ACQUISITION_FILE_ID                       |
| TAKE_ID                     | ACQUISITION_FILE_ID                       |
---------------------------------------------------------------------------
``` 
- Add privacy, only the notification user can retrieve, update or delete their own notification .
-  Validate that sub-ID is assigned to FileID.


asanchezr:
Changed the base of this PR to be **notifications** - we don't want to release this just yet

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        /// <param name="dbContext"></param>
+        /// <param name="user"></param>
+        /// <param name="logger"></param>
+        /// <param name="sequenceRepository"></param>
+        /// <param name="mapper"></param>
+        public NotificationRepository(PimsContext dbContext, ClaimsPrincipal user, ILogger<NotificationRepository> logger, ISequenceRepository sequenceRepository, IMapper mapper)
+            : base(dbContext, user, logger)
+        {
+        }
```
asanchezr: `ISequenceRepository sequenceRepository` and `IMapper mapper` parameters are not used - please remove

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        /// <summary>
+        /// Retrieves all notifications for a given user.
+        /// </summary>
+        /// <param name="userId"></param>
+        /// <returns></returns>
+        public IEnumerable<PimsNotification> GetByUser(long userId)
+        {
+            using var scope = Logger.QueryScope();
+            return Context.PimsNotifications
```
asanchezr, line 40: We use no-tracking queries to avoid EF having to load these entities in the memory context/tracker. This is our pattern given that we do not do any post-processing with these entities that need to be tracked by EF - we just serialize to JSON and return back to the frontend.


```suggestion
            return Context.PimsNotifications.AsNoTracking()
```

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        /// <summary>
+        /// Retrieves a notification by its id.
+        /// </summary>
+        /// <param name="notificationId"></param>
+        /// <returns></returns>
+        public PimsNotification GetById(long notificationId)
+        {
+            using var scope = Logger.QueryScope();
+            return Context.PimsNotifications
```
asanchezr, line 54:
```suggestion
            return Context.PimsNotifications.AsNoTracking()
```

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,182 @@
...
+                .FirstOrDefault(n => n.NotificationId == notificationId);
```
asanchezr: Another pattern we use in "GetOneById" methods is to throw **KeyNotFoundException** when no entity is found. This is handled in our ASP middleware to return a known error to the frontend. Please make the change


```suggestion
                .FirstOrDefault(n => n.NotificationId == notificationId) ?? throw new KeyNotFoundException();
```

source/backend/dal/Repositories/UserRepository.cs
```diff
@@ -279,6 +279,30 @@ public Paged<PimsUser> GetAllByFilter(UserFilter filter = null)
...
+
+        /// <summary>
+        /// Get the user with the specified 'username'.
+        /// </summary>
+        /// <param name="username"></param>
+        /// <returns></returns>
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        public PimsUser GetByUsername(string username)
+        {
+            User.ThrowIfNotAuthorized(Permissions.AdminUsers);
```
asanchezr: Please remove this permission check. We don't enforce permissions at the repo/DB access layer. We have moved all checks to the "Service" layer. I understand that this class is quite old and other methods might have the check but new methods should follow current practices. 

Also the permission check for "SysAdmin" here (or at the service-level) does not make sense because it is called in the Notifications code which is accessible to every user (not only to sys-admins)

We might have to discuss how to secure this call so you can only call it for your "own" user - not sure how to address that yet

asanchezr: Take a look at **GetUserInfoByKeycloakUserId** in which no permissions are checked (because that method is called by services extensively in various areas of the application)

source/backend/entities/Partials/Notification.cs
```diff
@@ -0,0 +1,13 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsNotification partial class for custom logic and inheritance.
```
asanchezr: nit: please provide a more relevant description of PimsNotification here (see other classes in the folder for examples)

The description generated from EF on the actual entity is acceptable as well, something like
> Describes the user notification and related field type (e.g. acquisition, disposition, management) provided to the user.

source/backend/entities/Partials/NotificationUser.cs
```diff
@@ -0,0 +1,13 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsNotificationUser partial class for custom logic and inheritance.
```
asanchezr: nit: please update class description:
> Catalogs the notification sent to a user.  Multiple notifications may be sent to a user.

source/backend/entities/Partials/NotificationUserOutput.cs
```diff
@@ -0,0 +1,13 @@
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// PimsNotificationUserOutput partial class for custom logic and inheritance.
```
asanchezr:
```suggestion
    /// PimsNotificationUserOutput describes the details of the notification sent the user including the sent date and read date.
```

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="notification">The notification to update.</param>
+        /// <param name="userId">The ID of the user associated with the notification.</param>
+        /// <returns>The updated notification.</returns>
+        /// <exception cref="KeyNotFoundException">Thrown when the notification with the specified ID is not found.</exception>
+        public PimsNotification Update(PimsNotification notification, long userId)
+        {
+            notification.ThrowIfNull(nameof(notification));
+
+            var existing = this.Context.PimsNotifications.FirstOrDefault(n => n.NotificationId == notification.NotificationId) ?? throw new KeyNotFoundException($"Notification {notification.NotificationId} not found");
+            Context.Entry(existing).CurrentValues.SetValues(notification);
```
asanchezr: This is risky in my opinion because `.CurrentValues.SetValues(...)` lets you change **all properties** wholesale - which could lead to issues/bugs where a notification for leaseID = 1 is updated to target managementFileID = 999

In my mind this should not be supported. Once a notification is created for a "file + sub-entity" combination - it should not be allowed to be changed. 

The only relevant properties that should be editable are the NOTIFICATION_TRIGGER_DATE and NOTIFICATION_MESSAGE.

PropertyRepository.Update(...) shows an example of how we deal with this by basically:

```cs
public PimsProperty Update(PimsProperty property
{
  var existingProperty = Context.PimsProperties
                .Include(p => p.Address)
                .FirstOrDefault(p => p.PropertyId == propertyId) ?? throw new KeyNotFoundException();

  // ignore a number of properties that we don't the frontend to override
  property.AddressId = existingProperty.AddressId;
  property.PropertyDataSourceEffectiveDate = existingProperty.PropertyDataSourceEffectiveDate;
  property.PropertyDataSourceTypeCode = existingProperty.PropertyDataSourceTypeCode;
  property.IsRetired = existingProperty.IsRetired;
  ...
}
```

asanchezr: I'm thinking that given that we would only want to update two properties of the "existing" notification (triggerDate and notifMessage) it would be better to: 

```cs
var existing = this.Context.PimsNotifications.FirstOrDefault(n => n.NotificationId == notification.NotificationId) ?? throw new KeyNotFoundException($"Notification {notification.NotificationId} not found");

existing.NotificationTriggerDate = notification.NotificationTriggerDate;
existing.NotificationMessage = notification.NotificationMessage;

Context.Notifications.Update(existing);
```

source/backend/api/Helpers/Validators/AgreementValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            if (notification.AgreementId.HasValue && notification.AcquisitionFileId.HasValue)
+            {
+                var agreement = _context.PimsAgreements
+                    .FirstOrDefault(a => a.AgreementId == notification.AgreementId.Value) ?? throw new InvalidOperationException("Agreement not found.");
+                if (agreement.AcquisitionFileId != notification.AcquisitionFileId)
+                {
+                    throw new InvalidOperationException("Agreement's AcquisitionFileId does not match the notification's AcquisitionFileId.");
+                }
+            }
```
asanchezr, line 32: I believe Agreements can belong to either an **Acquisition** or **Disposition** File. I correctly see here the check for a parent Acquisition File, but I do not see the corresponding check for when an agreement belongs to a Disposition File - please add

source/backend/api/Helpers/Validators/AgreementValidator.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Linq;
+using Pims.Dal;
+using Pims.Dal.Entities;
+
+namespace Pims.Api.Helpers.Validators
+{
+    public class AgreementValidator : INotificationSubtypeValidator
```
asanchezr, line 8: Super nice to see individual validators for each kind of "notification rule"!

source/backend/api/Helpers/Validators/NoticeOfClaimValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            if (notification.NoticeOfClaimId.HasValue && notification.AcquisitionFileId.HasValue)
+            {
+                var claim = _context.PimsNoticeOfClaims
+                    .FirstOrDefault(noc => noc.NoticeOfClaimId == notification.NoticeOfClaimId.Value) ?? throw new InvalidOperationException("Notice of Claim not found.");
+                if (claim.AcquisitionFileId != notification.AcquisitionFileId)
+                {
+                    throw new InvalidOperationException("Notice of Claim's AcquisitionFileId does not match the notification's AcquisitionFileId.");
+                }
+            }
```
asanchezr, line 32: Please add corresponding check for parent Management File

source/backend/api/Services/NotificationService.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public PimsNotification Update(PimsNotification notification, string username)
+        {
+            _logger.LogInformation("Updating notification with id {NotificationId} for user {Username}", notification.NotificationId, username);
+
+            _user.ThrowIfNotAllAuthorized(Permissions.NotificationEdit);
+            notification.ThrowIfNull(nameof(notification));
+            ValidateVersion(notification.NotificationId, notification.ConcurrencyControlNumber);
+
+            var user = _userRepository.GetByUsername(username) ?? throw new KeyNotFoundException($"User '{username}' not found in PIMS_USERS.");
```
asanchezr: Please use the common private method:

```suggestion
            var user = GetUserByUsername(username);
```

source/backend/apimodels/Models/Concepts/Notification/NotificationAccessResponse.cs
```diff
@@ -0,0 +1,11 @@
+using Pims.Dal.Entities;
+
+namespace Pims.Api.Models.Concepts.Notification
+{
+    public class NotificationAccessResponse
```
asanchezr, line 5: nit: Please update the this file **name** - Notification(A)ccessResponse.cs (missing the "A")

source/backend/apimodels/Models/Concepts/Notification/NotificationUserOutputModel.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public long NotificationUserId { get; set; }
+
+        public string NotificationOutputTypeCode { get; set; }
+
+        public DateTime? NotificationSentDt { get; set; }
+
+        public DateTime? NotificationReadDt { get; set; }
+
+        public short? NotificationRetryCnt { get; set; }
```
asanchezr, line 18: nit: no need to change these but we usually don't shorten the property names on our API models (unlike the DB generated entities which follow the DB naming convention). ie We would usually name these:
* NotificationSentDate
* NotificationReadDate
* NotificationRetryCount

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        /// <summary>
+        /// Retrieves all notifications for a given user.
+        /// </summary>
+        /// <param name="userId"></param>
+        /// <returns></returns>
+        public IEnumerable<PimsNotification> GetByUser(long userId)
+        {
+            using var scope = Logger.QueryScope();
+            var response = Context.PimsNotifications
```
asanchezr: Please use "AsNoTracking()" to all the queries in this repository
```suggestion
            var response = Context.PimsNotifications.AsNoTracking()
```

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,194 @@
...
+            Logger.LogInformation("Adding notification to context: {@Notification}", notification);
+            Context.PimsNotifications.Add(notification);
+            Context.SaveChanges();
+
+            // Add the notification user
+            var notificationUser = new PimsNotificationUser
+            {
+                NotificationId = notification.NotificationId,
+                UserId = userId,
+            };
```
asanchezr, line 89: This method calls SaveChanges() three times and manually manages foreign key IDs between saves. This is unnecessary with EF Core, which can resolve relationships automatically when you build the full object graph before persisting.

The method saves **PimsNotification** first to get its ID, then saves **PimsNotificationUser** to get its ID, and only then can it attach the two **PimsNotificationUserOutput** records. This results in 3 round-trips to the database and would leave the data in a partially-saved state if any intermediate step throws. This represents a silent data integrity risk since there's no transaction wrapping the whole operation.

I'd suggest to refactor this method to use EF Core object graph insertion. Build a single PimsNotification object with its children nested via navigation properties, then call Context.Add() once and Context.SaveChanges() once. EF Core will resolve all the IDs in the correct order within a single transaction:

Something like:

```cs
public PimsNotification Add(PimsNotification notification, long userId)
{
    using var scope = Logger.QueryScope();
    notification.ThrowIfNull(nameof(notification));

    notification.PimsNotificationUsers = new List<PimsNotificationUser>
    {
        new PimsNotificationUser
        {
            UserId = userId,
            PimsNotificationUserOutputs = new List<PimsNotificationUserOutput>
            {
                new PimsNotificationUserOutput
                {
                    NotificationOutputTypeCode = NotificationOutputTypeCode.EMAIL.ToString(),
                },
                new PimsNotificationUserOutput
                {
                    NotificationOutputTypeCode = NotificationOutputTypeCode.PIMS.ToString(),
                },
            }
        }
    };

    Logger.LogInformation("Adding notification to context: {@Notification}", notification);
    Context.Add(notification);
    Context.SaveChanges();

    return notification;
}
```

source/backend/dal/Repositories/NotificationRepository.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// <summary>
+        /// Deletes a notification for a given user.
+        /// </summary>
+        /// <param name="notificationId">The ID of the notification to delete.</param>
+        /// <param name="userId">The ID of the user associated with the notification.</param>
+        /// <returns>True if the notification was deleted successfully, otherwise false.</returns>
+        /// <exception cref="KeyNotFoundException">Thrown when the notification with the specified ID is not found.</exception>
+        public bool Delete(long notificationId, long userId)
+        {
+            var notification = Context.PimsNotifications.FirstOrDefault(n => n.NotificationId == notificationId) ?? throw new KeyNotFoundException($"Notification {notificationId} not found");
```
asanchezr: Please refactor this method to follow the same pattern as "TryDelete" in **CompensationRequisitionRepository**

The current method issues multiple separate queries to fetch the notification, its user links, and their outputs. TryDeleteCompensation instead fetches the root entity with all children eagerly loaded in a single query using Include() and AsNoTracking()

```cs
var deletedEntity = Context.PimsNotifications
    .Include(n => n.PimsNotificationUsers)
        .ThenInclude(nu => nu.PimsNotificationUserOutputs)
    .AsNoTracking()
    .FirstOrDefault(n => n.NotificationId == notificationId && 
                         n.PimsNotificationUsers.Any(nu => nu.UserId == userId));
```

Use stub entity removal inside foreach loops
```cs
if (deletedEntity != null)
{
    foreach (var notificationUser in deletedEntity.PimsNotificationUsers
                 .Where(nu => nu.UserId == userId))
    {
        foreach (var output in notificationUser.PimsNotificationUserOutputs)
        {
            Context.PimsNotificationUserOutputs.Remove(
                new PimsNotificationUserOutput
                {
                    NotificationUserOutputId = output.NotificationUserOutputId,
                });
        }
        Context.PimsNotificationUsers.Remove(
            new PimsNotificationUser
            {
                NotificationUserId = notificationUser.NotificationUserId,
            });
    }

    // Only remove the parent if no other users are linked
    bool hasOtherLinks = deletedEntity.PimsNotificationUsers.Any(nu => nu.UserId != userId);
    if (!hasOtherLinks)
    {
        ... // Delete PimsNotification
    }

    Context.SaveChanges();
    return true;
}
return false;
```

Also please do not throw the KeyNotFound exception on deletes. We return false when nothing was found to delete

asanchezr (changes requested):
Please see comments

source/backend/apimodels/Models/Concepts/Notification/NotificationUserModel.cs
```diff
@@ -0,0 +1,13 @@
+using Pims.Api.Models.Base;
+
+namespace Pims.Api.Areas.Notification.Models
```
asanchezr: Change namespace to be consistent with other concept models

```suggestion
namespace Pims.Api.Models.Concepts.Notification
```

source/backend/apimodels/Models/Concepts/Notification/NotificationUserOutputModel.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using Pims.Api.Models.Base;
+
+namespace Pims.Api.Areas.Notification.Models
```
asanchezr: Change namespace to be consistent with other concept models

```suggestion
namespace Pims.Api.Models.Concepts.Notification
```

source/backend/apimodels/Models/Concepts/Notification/NotificationUserOutputModel.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public string NotificationOutputTypeCode { get; set; }
+
+        public DateTime? NotificationSentDt { get; set; }
+
+        public DateTime? NotificationReadDt { get; set; }
+
+        public short? NotificationRetryCnt { get; set; }
+
+        public string? NotificationErrorReason { get; set; }
```
asanchezr, line 20: I don't think we need the nullable on the string here. Please change with to plain **string**

source/backend/apimodels/Models/Concepts/Notification/NotificationUserOutputModel.cs
areyeslo: DB entities have these fields defined. I'd prefer to keep them to be consistent.

source/backend/dal/Repositories/NotificationRepository.cs
asanchezr: I still see the Update method rely on `.CurrentValues.SetValues(...)` which is dangerous as I mentioned above

asanchezr (approved):

PR #5277, opened 2026-04-08
PSP-11333: When saving a property that is in the MOTT Inventory but lacks of spatial data, the property is saved within the file but it doesn't navigate back to the View form

areyeslo:
When the user adds a property with no boundary or location, they are prompted with the modal “User Override Required.” When the user accepts, the navigation goes to Update Property instead of the file details.
image

The fix allows the user to navigate to file details instead of having to click on Cancel button to go tack to file details.
image

[Jira](https://moti-imb.atlassian.net/browse/PSP-11333)

### DIFF
--- a/source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
+++ b/source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
@@ -263,6 +263,9 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
     await fetchLastUpdatedBy();
     mapMachine.refreshMapProperties();
     setIsEditing(false);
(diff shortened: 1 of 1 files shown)


source/frontend/src/features/mapSideBar/acquisition/AcquisitionContainer.tsx
```diff
@@ -263,6 +263,9 @@ export const AcquisitionContainer: React.FunctionComponent<IAcquisitionContainer
     await fetchLastUpdatedBy();
     mapMachine.refreshMapProperties();
     setIsEditing(false);
+    if (acquisitionFileId) {
```
asanchezr: nit: please use `isValidId()` to check acquisitionFileId

asanchezr (approved):

PR #5280, opened 2026-04-10
PSP-11411: [HOTFIX] Acquisition Files - When update information on any tab from an Acquisition File, PIMS navigates back to FIle Details instead of staying at the chosen tab - view form

areyeslo:
The previous [PR](https://github.com/bcgov/PSP/pull/5277) added the logic to navigate to fileDetails in `onSuccess` function. However, this is shared among the child components in the Container. This caused that when saving changes in Stakeholder the navigation sent the user to fileDetails.

The change needs to be apply only for updating properties in the container and after `onSuccess` execution to add the customized logic to redirect the user to fileDetails. During my review I also found ResearchContainer presented the same issue when properties with no location are added. The other containers include the logic added to Acquisition and Research.

[Jira](https://moti-imb.atlassian.net/browse/PSP-11411)

### DIFF
--- a/source/frontend/package.json
+++ b/source/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "6.1.1-119.16",
+  "version": "6.1.0-119.16",
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "1.0.1",
--- a/source/frontend/package.json
+++ b/source/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "6.1.0-119.16",
+  "version": "6.1.1-119.16",
   "private": true,
   "dependencies": {
     "@bcgov/bc-sans": "1.0.1",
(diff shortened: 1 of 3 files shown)


source/frontend/src/features/mapSideBar/research/ResearchContainer.tsx
```diff
@@ -240,6 +240,9 @@ export const ResearchContainer: React.FunctionComponent<IResearchContainerProps>
           userOverrideCodes,
         ).then(response => {
           onSuccess();
+          if (isValidId(response?.id)) {
+            pathGenerator.showFile('acquisition', response.id);
+          }
```
asanchezr, line 245: This is the RESEARCH container so I would expect the pathGenerator to go research file details (not acquisition)

source/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frontend",
-  "version": "6.1.0-119.16",
+  "version": "6.1.1-119.16",
```
asanchezr: I already incremented this number in my changes, please remove this change. Additionally, the hotfix number needs to be updated also in the Pims.Api.csproj

asanchezr (approved):
Please see comments to make minor changes

PR #5289, opened 2026-04-21
PSP-11443 : File/Project list views do not filter by the users regions for the initial search

eddherrera:


source/frontend/src/features/acquisition/list/AcquisitionListView.tsx
```diff
@@ -142,6 +135,22 @@ export const AcquisitionListView: React.FunctionComponent<
     formattedGuid && retrieveUserInfo(formattedGuid);
   }, [formattedGuid, retrieveUserInfo]);
 
+  useEffect(() => {
+    if (userRegionsOptions === null && retrieveUserInfoResponse && pimsRegionsTypes) {
```
asanchezr: nit" use **exists()** for null checks

devinleighsmith: this block seems to be repeated a few times, could this be placed in a reusable method?

devinleighsmith: we normally use exists(x) for these sorts of checks.

source/frontend/src/features/acquisition/list/AcquisitionListView.tsx
```diff
@@ -142,6 +135,22 @@ export const AcquisitionListView: React.FunctionComponent<
...
+        retrieveUserInfoResponse?.userRegions?.map(x => x.regionCode.toString()) ?? [];
+
+      const userRegionsOptions: MultiSelectOption[] =
+        pimsRegionsTypes
+          .filter(opt => userRegionsIds?.includes(opt.code))
+          .map<MultiSelectOption>(x => {
+            return { id: x.code as string, text: x.label };
+          }) ?? [];
+      setUserRegionsOptions(userRegionsOptions ?? []);
+      setFilter(new AcquisitionFilterModel(userRegionsOptions).toApi());
```
devinleighsmith, line 150: this seems to be the only difference between duplicated sections.

asanchezr (approved):

PR #5291, opened 2026-04-21
Psp 11434

devinleighsmith:
Tested in dev environment. Not intended for production use.

Does a best-effort clean of data created by test user(s) specified in smoke tests.

Does not delete entities where sub-entities have been manually added by other users.

Database scripts generated by AI.

Deletion does not affect PIMS_USER* or PIMS_PERSON* tables required for testing.

Deletion scripts run by openshift jobs due to firewall restrictions.

Cleans history as well as data rows - as stated, not for production.


.github/workflows/smoke-test.yml
```diff
@@ -36,5 +57,48 @@ jobs:
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps
 
-      - name: 🚀 Run Cucumber Smoke Tests
-        run: npx cucumber-js --tags @smokeTest
```
asanchezr, line 40: why do the playwright tests were removed? Do they not work currently?

devinleighsmith: just removed for test purposes, since these aren't running cleanly atm.

PR #5304, opened 2026-05-06
PSP-11311 : Notification hangfire job for PIMS

eddherrera:


source/backend/api/Areas/Admin/Controllers/NotificationUserController.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Pims.Api.Areas.Notification.Controllers
+{
+    [Authorize]
+    [ApiController]
+    [ApiVersion("1.0")]
+    [Area("user-notifications")]
+    [Route("v{version:apiVersion}/[area]")]
+    [Route("[area]")]
+    public class NotificationUserController : ControllerBase
```
asanchezr, line 26: This is not the notification controller for "users" but instead is an admin-only API that requires system-admin access to modify notifications for **any user**. For that reason, I would prefer it it were named something like NotificationAdminController (or similar) - thoughts?

source/backend/api/Areas/Notification/Controllers/NotificationUserController.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public NotificationUserController(INotificationUserService notificationService, IMapper mapper, ILogger<NotificationController> logger)
+        {
+            _notificationUserService = notificationService;
+            _mapper = mapper;
+            _logger = logger;
+        }
+
+        [HttpPost("search")]
+        [HasPermission(Permissions.SystemAdmin)]
```
asanchezr: instead of setting permissions on every endpoint ,please put the permission enforcement on the whole class at the top. All these endpoints are admin-only

source/backend/api/Services/NotificationUserService.cs
```diff
@@ -0,0 +1,168 @@
...
+using Pims.Core.Api.Services;
+using Pims.Core.Extensions;
+using Pims.Core.Security;
+using Pims.Dal.Entities;
+using Pims.Dal.Helpers.Extensions;
+using Pims.Dal.Repositories;
+
+namespace Pims.Api.Services
+{
+    public class NotificationUserService : BaseService, INotificationUserService
```
asanchezr, line 22: For the same reason I mentioned in the controller: NotificationAdminService would make more sense here

source/backend/apimodels/CodeTypes/CotactMethodTypes.cs
```diff
@@ -0,0 +1,30 @@
+using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
+
+namespace Pims.Api.Models.CodeTypes
+{
+    [JsonConverter(typeof(JsonStringEnumMemberConverter))]
+    public enum CotactMethodTypes
```
asanchezr: fix typo CotactMethodTypes -> ContactMethodTypes

asanchezr: File should be renamed as well

source/backend/dal/Repositories/NotificationUserRepository.cs
```diff
@@ -0,0 +1,137 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using Pims.Api.Models.Models.Concepts.Notification;
+using Pims.Core.Extensions;
+using Pims.Dal.Entities;
+using Pims.Dal.Helpers.Extensions;
+
+namespace Pims.Dal.Repositories
+{
+    public class NotificationUserRepository : BaseRepository<PimsNotificationUserOutput>, INotificationUserRepository
```
asanchezr: Name is a bit confusing NotificationUserRepository but it deals with mostly outputs, What do you think about NotificationOutputRepository?

source/backend/dal/Repositories/UserRepository.cs
```diff
@@ -13,6 +10,10 @@
 using Pims.Dal.Entities.Models;
 using Pims.Dal.Exceptions;
 using Pims.Dal.Helpers.Extensions;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
```
asanchezr: I am not sure if you are using an autoformatter for C# but I believe the system imports go at the top of the list. Can you check your formatting options?

source/backend/api/Areas/Notification/Controllers/NotificationUserController.cs
```diff
@@ -0,0 +1,81 @@
...
+using Pims.Core.Json;
+using Pims.Core.Security;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace Pims.Api.Areas.Notification.Controllers
+{
+    [Authorize]
+    [ApiController]
+    [ApiVersion("1.0")]
+    [Area("user-notifications")]
```
asanchezr: I think these endpoints should be under"/admin" to let the proper user notifications (the ones scoped to the current user calling the API) to use the "user-notification" routes. I fear that if we don't separate the admin endpoints from the user endpoints we will introduce bugs

source/backend/scheduler/Repositories/PimsNotificationUserRepository.cs
```diff
@@ -0,0 +1,68 @@
...
+            _authRepository = authRepository;
+            _configuration = configuration;
+        }
+
+        public async Task<ExternalResponse<List<NotificationUserOutputModel>>> SearchUserNotificationsAsync(NotificationUserSearchFilterModel filter)
+        {
+            _logger.LogDebug("Getting filtered list of  {filter}", filter);
+
+            string authenticationToken = await _authRepository.RequestAccessToken();
+            Uri endpoint = new($"{_configuration.CurrentValue.Uri}/user-notifications/search");
```
asanchezr: in the controller I recommended urls with "admin" in it to signal that this is an internal/admin-only API

source/backend/api/Areas/Admin/Controllers/NotificationUserController.cs
eddherrera: I see your point. I analyzed and found that we have an "Area" for admin functionality, I'll relocate the controller endpoints to that 'Admin' Area. 
For the name of the controller I am a little hesitant since the core of this controller is operate towards the users and the way we named the tables and definitions works better to understand how the layers connect.

source/backend/api/Areas/Notification/Controllers/NotificationUserController.cs
eddherrera: Updated

source/backend/api/Services/NotificationUserService.cs
eddherrera: I would rather keep it this way to make it clear that these are the actual notifications to the user.

source/backend/apimodels/CodeTypes/CotactMethodTypes.cs
eddherrera: Updated

source/backend/dal/Repositories/NotificationUserRepository.cs
eddherrera: sure, this works.

source/backend/dal/Repositories/UserRepository.cs
eddherrera: it's a Visual Studio Professional thing, it's a pain when using the "remove and sort"...

source/backend/api/Areas/Notification/Controllers/NotificationUserController.cs
eddherrera: done

source/backend/scheduler/Repositories/PimsNotificationUserRepository.cs
eddherrera: done

asanchezr (approved):

PR #5321, opened 2026-05-28
PSP-11192 : DEV: When a user doesn't have a specific region on their access but chooses on a map a property from that region and creates a file, the backend gives a 403 error.

eddherrera:


asanchezr (approved):
LGTM

PR #5324, opened 2026-06-02
Automation - Research files

stairaku:


testing/playwright-testing/tests/smoke-tests/research.spec.ts
```diff
@@ -0,0 +1,100 @@
+import test, { BrowserContext, expect, Page } from "@playwright/test";
+import { ResearchCreatePage } from "../../pages/research/research-create.page";
+import { ResearchListPage } from "../../pages/research/research-list.page";
+
+let context: BrowserContext;
+let page: Page;
+let researchCreatePage: ResearchCreatePage;
+let researchListPage: ResearchListPage;
+
+test.describe.serial("Research File modal", () => {
```
eddherrera: in this case since these are test on different pages, there's no need for one to wait for the other. We could put the test to run in parallel OR we could create one new .spec file for each test.

testing/playwright-testing/tests/smoke-tests/research.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    await expect(researchListPage.researchTableOrderUpdatedDate).toBeVisible();
+    await expect(researchListPage.researchTableStatusHeader).toBeVisible();
+    await expect(researchListPage.researchTableOrderStatus).toBeVisible();
+
+    await expect(researchListPage.researchTableEntriesSpan).toBeVisible();
+    await expect(researchListPage.researchTablePagination10).toBeVisible();
+    await expect(researchListPage.researchTableNextPageButton).toBeVisible();
+    await expect(researchListPage.researchTable1stPageButton).toBeVisible();
+
+    await expect(researchListPage.getResearchListTotal).toBeGreaterThan(0);
```
eddherrera: Try adding an assertion that WHEN the user clicks on the "+ Create Research button" THEN the create page is loaded ( just that the URL matches '/research/new'

testing/playwright-testing/pages/research/research-list.page.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  async getResearchListTotal(): Promise<number> {
+    return await this.researchTableContent.count();
+  }
+
+  async createNewResearchClick() {
+    await this.researchNewButton.click();
+  }
+
+  async changeToNewTab(): Promise<Page> {
```
eddherrera: Please just move this to the Test. This is the kind of code that goes into a Test fixture. 
This way we keep the POM clean.

eddherrera (approved):

PR #5328, opened 2026-06-05
PSP-9839 : Check file type on document upload form (frontend)

eddherrera:
image

### DIFF
--- a/source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
+++ b/source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
@@ -1,6 +1,7 @@
 import * as Yup from 'yup';
 
 import ValidDocumentExtensions from '@/constants/ValidDocumentExtensions';
+import { exists } from '@/utils';
 
 export const getDocumentMetadataYupSchema = (maxDocumentCount: number) => {
   return Yup.object().shape({
@@ -10,9 +11,11 @@ export const getDocumentMetadataYupSchema = (maxDocumentCount: number) => {
       .of(
         Yup.object().shape({
           documentTypeId: Yup.string().required('Document type is required'),
-          file: Yup.mixed().test('fileExtension', 'File type not supported!', value => {
+          file: Yup.mixed<File>().test('fileExtension', 'File type not supported!', value => {
             // If no file is uploaded, let 'required' handle it (or return true if optional)
-            if (!value) return true;
+            if (!exists(value)) {
+              return true;
+            }
 
             const fileName = value.name;
             if (typeof fileName !== 'string') return false;
--- a/source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
+++ b/source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
@@ -1,5 +1,7 @@
 import * as Yup from 'yup';
 
+import ValidDocumentExtensions from '@/constants/ValidDocumentExtensions';
+
 export const getDocumentMetadataYupSchema = (maxDocumentCount: number) => {
   return Yup.object().shape({
     documents: Yup.array()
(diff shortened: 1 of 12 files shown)


source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
```diff
@@ -8,6 +10,16 @@ export const getDocumentMetadataYupSchema = (maxDocumentCount: number) => {
       .of(
         Yup.object().shape({
           documentTypeId: Yup.string().required('Document type is required'),
+          file: Yup.mixed().test('fileExtension', 'File type not supported!', value => {
+            // If no file is uploaded, let 'required' handle it (or return true if optional)
+            if (!value) return true;
```
asanchezr: nit: we always use braces, even for single lines. That has been the pattern so far so prefer to be consistent


```suggestion
            if (!value) {
              return true;
             }
```

asanchezr: Also use **!exists(value)**

source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
```diff
@@ -8,6 +10,16 @@ export const getDocumentMetadataYupSchema = (maxDocumentCount: number) => {
       .of(
         Yup.object().shape({
           documentTypeId: Yup.string().required('Document type is required'),
+          file: Yup.mixed().test('fileExtension', 'File type not supported!', value => {
```
asanchezr: The typing can be improved here to avoid bugs due to "value" being of type **any**


```suggestion
          file: Yup.mixed<File>().test('fileExtension', 'File type not supported!', value => {
```

source/frontend/src/features/documents/documentUpload/SelectedDocumentRow.tsx
```diff
@@ -74,6 +81,49 @@ export const SelectedDocumentRow: React.FunctionComponent<ISelectedDocumentRowPr
...
+      <Section
+        header={
+          <div style={{ fontSize: '1.6rem' }}>
+            <Row className={clsx('no-gutters')}>
+              <Col>
+                <span>File {index + 1}:</span>
+                <span className="ml-4" style={{ color: 'red' }}>
+                  {truncate(document.file.name, { length: 50 })}
+                </span>
+                <FaTimes className="ml-2" size="1.6rem" color={theme.css.pimsRed80} />
```
asanchezr: For consistency, use the same icon and color specified in the **UploadResultView** for upload errors

`FaTimesCircle` (instead of `FaTimes`) and color `theme.bcTokens.iconsColorDanger` instead of pimsRed80

Same with the error text below - it should not be hardcoded to `style={{ color: 'red' }}` - we have the colorDanger for that

asanchezr (changes requested):
The UI doesn't match the mockups for multiple incorrect files uploaded. Mockup 2 reflects this state by only displaying "File type not supported" only once. 

Also there is a divider between incorrect files that doesn't look good (and do not match the mockups). I think the failure rows should not be a Section component, just plain divs

Image

source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
eddherrera: Updated

source/frontend/src/features/documents/DocumentMetadataYupSchema.tsx
eddherrera: Updated

source/frontend/src/features/documents/documentUpload/SelectedDocumentRow.tsx
eddherrera: updated.

asanchezr (approved):

PR #5346, opened 2026-06-26
PSP-11665: DA Task Modify PIMS_CONTACT_MGR_VW to include join to PIMS…

eddherrera:


source/database/mssql/scripts/dbscripts/PSP_PIMS_S122_00/Alter Down/010_PSP_PIMS_Alter_Down.sql
```diff
@@ -0,0 +1,133 @@
+-- Script generated by Aqua Data Studio Schema Synchronization for MS SQL Server 2016 on Wed Apr 29 19:28:29 PDT 2026
+-- Execute this script on:
+-- 		PSP_PIMS_S121_00/dbo - This database/schema will be modified
+-- to synchronize it with MS SQL Server 2016:
+-- 		PSP_PIMS_S120_00/dbo
```
asanchezr, line 5: nit: I know these are just comments but I believe they should state that DB is going from 122 -> 121 (not 121 -> 120)

asanchezr (approved):

PR #5347, opened 2026-06-29
PSP-11710 : PIMS Notifications Panel not Displaying Reminders  (BVT O…

eddherrera:
…bservations) - HOTFIX

### DIFF
--- a/source/backend/dal/Repositories/NotificationUserOutputRepository.cs
+++ b/source/backend/dal/Repositories/NotificationUserOutputRepository.cs
@@ -1,8 +1,9 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using System.Threading.Tasks;
 using LinqKit;
+using Medallion.Threading;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using Pims.Api.Models.Models.Concepts.Notification;
@@ -14,12 +15,16 @@ namespace Pims.Dal.Repositories
 {
     public class NotificationUserOutputRepository : BaseRepository<PimsNotificationUserOutput>, INotificationUserOutputRepository
     {
+        private readonly IDistributedLockProvider _synchronizationProvider;
+
         public NotificationUserOutputRepository(
             PimsContext dbContext,
             ClaimsPrincipal user,
-            ILogger<DocumentRepository> logger)
+            IDistributedLockProvider synchronizationProvider,
+            ILogger<NotificationUserOutputRepository> logger)
             : base(dbContext, user, logger)
         {
+            _synchronizationProvider = synchronizationProvider;
         }
 
         public PimsNotificationUserOutput GetById(long notificationUserOutputId)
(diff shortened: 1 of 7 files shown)


source/backend/dal/Repositories/NotificationUserOutputRepository.cs
```diff
@@ -75,25 +80,21 @@ public PimsNotificationUserOutput GetById(long notificationUserOutputId)
                 .FirstOrDefault(x => x.NotificationUserOutputId == notificationUserOutputId) ?? throw new KeyNotFoundException();
         }
 
-        public PimsNotificationUserOutput Update(PimsNotificationUserOutput userNotification)
+        public async Task<PimsNotificationUserOutput> Update(PimsNotificationUserOutput userNotification)
```
asanchezr, line 83: nit: Naming conventions suggest to name async methods XYZAsync to let consumers know it is async. Can be done next sprint if no time now

https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model#naming-convention

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
```diff
@@ -75,25 +80,21 @@ public PimsNotificationUserOutput GetById(long notificationUserOutputId)
...
         {
-            try
+            var @lock = _synchronizationProvider.CreateLock("NotificationUserOutputLock");
```
asanchezr, line 85: A comment here goes a long way in explaining future developers why we put a lock in place here:


```suggestion
            // Use a distributed lock to ensure that only one process can update the notification output table at a time (prevents deadlocks from history triggers).
            var @lock = _synchronizationProvider.CreateLock("NotificationUserOutputLock");
```

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
```diff
@@ -75,25 +80,21 @@ public PimsNotificationUserOutput GetById(long notificationUserOutputId)
...
+            await using (await @lock.AcquireAsync())
             {
-                using var scope = Logger.QueryScope();
                 userNotification.ThrowIfNull(nameof(userNotification));
 
                 var existingUserNotification = Context.PimsNotificationUserOutputs
                     .FirstOrDefault(x => x.NotificationUserOutputId == userNotification.NotificationUserOutputId);
 
                 Context.Entry(existingUserNotification).CurrentValues.SetValues(userNotification);
+                Context.Update(existingUserNotification);
```
asanchezr, line 94: I'm not sure if need the extra Update here since the line above updates the values via Context.Entry().CurrentValues

Also I see in the document queue they force the save in scope of the lock like so:

```cs
Context.SaveChanges(); // Force changes to be saved here, in the scope of the lock.
```

I would assume we need it here too?

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
```diff
@@ -122,7 +123,7 @@ public IEnumerable<PimsNotificationUserOutput> GetAllByFilter(NotificationUserSe
 
             if(filter.NotificationTriggerDate is not null)
             {
-                predicateBuilder.And(x => x.NotificationUser.Notification.NotificationTriggerDate == filter.NotificationTriggerDate);
+                predicateBuilder.And(x => x.NotificationUser.Notification.NotificationTriggerDate <= filter.NotificationTriggerDate);
```
asanchezr, line 126: thank you for fixing this, but I think this query need an extra condition to exclude PRIOR notifications that have already been sent (sent_date !== null)


```suggestion
                predicateBuilder.And(x => x.NotificationUser.Notification.NotificationTriggerDate <= filter.NotificationTriggerDate).And(x => x.NotificationSentDt == null);
```

asanchezr (approved):
Approved but see comments

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
asanchezr: Otherwise we risk the query returning ALL outputs from this day on (including all previously sent)

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
eddherrera: tried it, didnt work. not needed.

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
eddherrera: NotificationSentDate == null is already in the predicate builder.

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
eddherrera: Not always if code if self explanatory.

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
eddherrera: Should do on a another version.

source/backend/dal/Repositories/NotificationUserOutputRepository.cs
asanchezr: I see

PR #5354, opened 2026-07-09
PSP-11664 Remove Contractor Permissions for Specified Region

asanchezr:


eddherrera (approved):
Although the code looks good I am moderately concerned about the fact that methods at the repository level became heavily dependent and tightly coupled on a "UserContext" which may not always be the case when have other services making calls to the API. 
I see other changes not related to the scope of the ticket like 'Appsettings' and 'project' settings and changes on ports configurations that would have been better on a sepparate PR. 
And the last thing would be the TICKET number in comments, which I believe we don't have a convention and would be nice to agree upon. 
Please take a note on the project settings changes and communicate to us on the Dev meeting.

PR #5356, opened 2026-07-13
PSP-11432: Integration(smoke) tests operational in GHAs

eddherrera:


eddherrera:
I've commented some of the testing, will review later and update with a future PR.

testing/pims-integration/tests/e2e/research/research.spec.ts
```diff
@@ -4,7 +4,7 @@ import { DocumentsListPage } from '../../../pages/documents/documents-list.page'
 
 import { ResearchViewFileDetails } from '../../../pages/research/research-view-file-details.page';
 import { DocumentUploadModalPage } from '../../../pages/documents/document-upload-modal.page';
-import { generateFileName, normalize, formatApiDate, formatApiBoolean } from '../../../utils/utils';
+import { generateFileName, normalize, nullableBooleanToYesNoString } from '../../../utils/utils';
```
asanchezr, line 7: The convention in PIMS frontend is to use "absolute import paths" for clarity - I see this is not applied in this project. While not related to this PR I would like that it gets changed to absolute paths for import, ie `import {..} from @src/utils/utils` instead of `../../../utils`

testing/pims-integration/tests/e2e/research/research.spec.ts
```diff
@@ -62,36 +66,36 @@ test.describe('Research Files feature', () => {
...
-      for (const date of datesToCompare) {
-        const uiValue = await researchViewDetails.getFieldValueByLabel(date.label);
-        const changedFormat = formatApiDate(date.apiValue);
-        expect(normalize(uiValue)).toBe(changedFormat);
-      }
+      // const fieldsToCompare = [
+      //   { label: 'Ministry project', apiValue: apiFeatureFileJson.researchFileProjects[0] },
+      //   { label: 'Road name', apiValue: apiFeatureFileJson.roadName },
+      //   { label: 'Road alias', apiValue: apiFeatureFileJson.roadAlias },
+      //   {
```
asanchezr, line 73: I'm ignoring this in the review because you said it will be fixed in a subsequent PR

testing/pims-integration/utils/utils.ts
```diff
@@ -30,10 +31,7 @@ export const formatApiDate = (apiDate: string | null | undefined): string => {
...
 
-export const formatApiBoolean = (value: boolean | null | undefined): string => {
-  if (value == null) {
-    return '';
-  }
-
-  return value ? 'Yes' : 'No';
-};
+export function nullableBooleanToYesNoString(value?: boolean | null): string {
+  if (value === true) return 'Yes';
```
asanchezr, line 35: We have been following the convention of always adding opening and closing braces `{` in TS code (even for one-liners like the above). 

Alternatively this could be refactored into a ternary operation

```ts
return value === true ? 'Yes' : 'No'
```

.github/workflows/ci-cd-pims-e2e-test.yml
```diff
@@ -0,0 +1,57 @@
...
+        - name: Checkout Source Code
+          uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #6.0.3
+
+        - name: Setup Node
+          uses: actions/setup-node@v4
+          with:
+            node-version: "20"
+
+        - name: Install dependencies
+          run: npm install
```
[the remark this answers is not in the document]
asanchezr: I agree with this warning - we should do `npm ci` against a verified list of dependencies. Npm install could lead to security issues or compatibility bugs

.github/workflows/ci-cd-pims-e2e-test.yml
```diff
@@ -0,0 +1,57 @@
...
+          working-directory: testing/pims-integration
+
+        - name: Run Playwright Smoke tests
+          run: npx playwright test --project=pims-e2e
```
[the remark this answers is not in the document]
asanchezr: Also agree with this warning - we should not use NPX on our CI builds. Npx will pull the latest version and install all of its dependencies (risky for security reasons). In my opinion, playwright should be installed via `npm ci` and then run locally once installed (vs using npx)

.github/workflows/ci-cd-pims-smoke-test.yml
```diff
@@ -0,0 +1,57 @@
...
+      container:
+        image: mcr.microsoft.com/playwright:v1.61.0-noble
+        options: --user 1001
+
+      steps:
+        - name: Checkout Source Code
+          uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #6.0.3
+
+        - name: Setup Node
+          uses: actions/setup-node@v4
```
asanchezr: recommended best practice is to pin the actions version to the SHA (like you did with checkout a few lines before).

Something like the following (we do this in other workflows in this repo):

```suggestion
          uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```

.github/workflows/ci-cd-pims-smoke-test.yml
```diff
@@ -0,0 +1,57 @@
...
+            node-version: "20"
+
+        - name: Install dependencies
+          run: npm install
+          working-directory: testing/pims-integration
+
+        - name: Run Playwright Smoke tests
+          run: npx playwright test --project=smoke-test
+          working-directory: testing/pims-integration
+        - uses: actions/upload-artifact@v5
```
asanchezr: same comment here re: action version pinning

asanchezr (changes requested):
See comments

.github/workflows/ci-cd-pims-smoke-test.yml
eddherrera: Updated

.github/workflows/ci-cd-pims-smoke-test.yml
eddherrera: Updated

.github/workflows/ci-cd-pims-e2e-test.yml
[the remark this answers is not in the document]
eddherrera: this command is not for 'install' not quite sure what the problem is.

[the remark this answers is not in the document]
asanchezr: NPX **does** install whatever library you call globally (in this case playwright). It also installs the latest version (that could be vulnerable and the team did not review - or any of its dependencies). It also runs "pre-install" scripts within that library that could be compromised. 

More details: https://github.com/bcgov/PSP/security/code-scanning/6892

> When installing dependencies, some package managers will automatically execute scripts distributed along with the source code of packages. Post-install scripts, for example, are a common way to execute malicious code at install time whenever a package is compromised.
>
> Why is this an issue?
When package managers execute installation scripts, they run arbitrary code distributed with third-party packages. A compromised package can use this mechanism to execute malicious code on the build system, potentially stealing credentials, injecting backdoors, or otherwise compromising the supply chain.
>
>What is the potential impact?
If a dependency is compromised and its scripts are executed, an attacker can run arbitrary code with the permissions of the process performing the installation. This can lead to credential theft from the build environment, introduction of backdoors into the application, or lateral movement within CI/CD infrastructure.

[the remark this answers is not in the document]
asanchezr: Also, by adding playwright to the devDependencies on package.json we control which version to install via `npm ci` - that way is safer than using NPX

Source: https://github.com/bcgov/PSP/security/code-scanning/6893

> Why is this an issue?
Without a committed lock file, dependency resolution runs anew on every fresh build, always selecting the latest version available that satisfies the declared range. This means different developers and CI systems may silently use different dependency versions, leaving the project exposed if a malicious version of a dependency is published within the allowed range.
>
> What is the potential impact?
If an attacker publishes a malicious version of a dependency within your declared version range, an unlocked project may automatically adopt it on the next fresh build. This can result in arbitrary code execution, data exfiltration, or backdoors being silently introduced into your application as part of a supply chain attack.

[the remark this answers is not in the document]
eddherrera: my point is that there's no other command in Playwright to run tests  without "npx"

testing/pims-integration/tests/e2e/research/research.spec.ts
eddherrera: will fix later, we need to have a word on "npm registry"

.github/workflows/ci-cd-pims-e2e-test.yml
```diff
@@ -0,0 +1,57 @@
...
+          required: true
+          type: choice
+          options:
+            - test
+          default: 'test'
+    workflow_call:
+        inputs:
+          target_env:
+            required: true
+            type: string
```
asanchezr, line 17: I am going to approve this PR in the interest of moving forward and testing but the two new workflows are exactly the same content (except for the npm run command). We have tons of duplicated code already and it is making our github actions hard to maintain. I recommend these two are refactored into a single one that takes the npm command to run as an input parameter

Something like below:

```suggestion
    workflow_call:
        inputs:
          target_env:
            required: true
            type: string
          run_command:
            required: true
            type: string
            default: 'e2e-tests'
```

.github/workflows/ci-cd-pims-e2e-test.yml
```diff
@@ -0,0 +1,57 @@
+name: CI-CD PIMS e2e test
+
+on:
+    workflow_dispatch:
```
asanchezr, line 4: Given that these actions are meant to be run automatically but also on demand - I recommend we introduced concurrency configuration to abort any running action if a new run is triggered

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/retag-dev-to-test.yml
```diff
@@ -353,3 +353,19 @@ jobs:
...
+    name: PIMS e2e smoke tests
+    needs: ci-cd-end-notification
+    uses: ./.github/workflows/ci-cd-pims-smoke-test.yml
+    with:
+      target_env: test
+    secrets: inherit
+
+  run-e2e-integration-tests:
+    name: PIMS e2e integration tests
+    needs: run-e2e-smoke-tests
```
asanchezr, line 367: Here we are waiting for smoke-tests to finish before we run the full e2e suite. Wouldn't it be better to have both jobs run in parallel? This setup will add to the wait time of each TEST CI run

asanchezr (approved):
I'm approving this in the interest of getting e2e tests going but I have left comments that should be addressed in future PRs.

We also agreed to use source aliases for ApiGen models in **tsconfig.json** - I did not see those changes here

.github/workflows/retag-dev-to-test.yml
eddherrera: I am willing to short circuit the e2e if the smoke tests fails. If something goes wrong it will go wrong faster.

.github/workflows/ci-cd-pims-e2e-test.yml
eddherrera: I'll dig into it, but I would rather keep human input out as much as possible.

.github/workflows/ci-cd-pims-e2e-test.yml
eddherrera: I am not so concerned about it since deployments to test are few and far between.

.github/workflows/ci-cd-pims-e2e-test.yml
asanchezr: That is why I provided a default value in my sample. You could also make it a "choice" to avoid manual entry errors

PR #5360, opened 2026-07-20
PSP-11769: HOTFIX: When a Contractor is assigned to a Project or a Team but no longer to a Region, the file doesn't appear on the list but they can still access by copying the URL to a Browser.

asanchezr:


source/backend/api/Helpers/Extensions/AcquisitionFileExtensions.cs
```diff
@@ -62,5 +52,40 @@ public static void ThrowContractorLegacyFileForbidden(this PimsAcquisitionFile a
...
+            {
+                return false;
+            }
+
+            if (pimsUser.IsContractor)
+            {
+                var onTeamOrProject = acquisitionFile.PimsAcquisitionFileTeams.Any(x => x.PersonId == pimsUser.PersonId)
+                    || (project != null && project.PimsProjectPeople.Any(x => x.PersonId == pimsUser.PersonId));
+
+                return onTeamOrProject && acquisitionFile.IsUserAssignedToAcquisitionFileRegion(pimsUser);
```
eddherrera: Suggestion: return onTeamOrProject && pimsUser.PimsRegionUsers.Any(ur => ur.RegionCode == acquisitionFile.RegionCode);

Unless 'IsUserAssignedToAcquisitionFileRegion' is called outside of the class it should be removed since it's making all the checks already made and it can be simplified as described above.

source/backend/api/Helpers/Extensions/AcquisitionFileExtensions.cs
```diff
@@ -62,5 +52,40 @@ public static void ThrowContractorLegacyFileForbidden(this PimsAcquisitionFile a
...
+                return pimsUser.PimsRegionUsers.Any(ur => ur.RegionCode == acquisitionFile.RegionCode);
+            }
+
+            // Regular (non-contractor) users are not bound by region access here.
+            return true;
+        }
+
+        public static bool HasAccessToFile(this PimsAcquisitionFile acquisitionFile, PimsUser pimsUser, PimsProject project)
+        {
+            if (acquisitionFile is null || pimsUser is null)
```
eddherrera, line 58: We are already throwing an exception if either of these two is null. Checks are not needed.

source/backend/api/Helpers/Extensions/AcquisitionFileExtensions.cs
```diff
@@ -62,5 +52,40 @@ public static void ThrowContractorLegacyFileForbidden(this PimsAcquisitionFile a
...
+            if (pimsUser.IsContractor)
+            {
+                return pimsUser.PimsRegionUsers.Any(ur => ur.RegionCode == acquisitionFile.RegionCode);
+            }
+
+            // Regular (non-contractor) users are not bound by region access here.
+            return true;
+        }
+
+        public static bool HasAccessToFile(this PimsAcquisitionFile acquisitionFile, PimsUser pimsUser, PimsProject project)
```
eddherrera, line 56: if This extension method is not called outside the class it should be 'private'

source/backend/api/Helpers/Extensions/DispositionFileExtensions.cs
```diff
@@ -10,32 +10,67 @@ namespace Pims.Api.Helpers.Extensions
...
+            PimsProject project = dispositionFile.ProjectId.HasValue ? projectRepository.TryGet(dispositionFile.ProjectId.Value) : null;
 
-            if (pimsUser?.IsContractor == true && !dispositionFile.PimsDispositionFileTeams.Any(x => x.PersonId == pimsUser.PersonId))
+            if (!dispositionFile.HasAccessToFile(pimsUser, project))
             {
                 throw new ContractorNotInTeamException("Contractors cannot remove themselves from a Disposition file. Please contact the admin at pims@gov.bc.ca");
             }
         }
+
+        public static bool IsUserAssignedToDispositionFileRegion(this PimsDispositionFile dispositionFile, PimsUser pimsUser)
```
eddherrera: Probably not needed unless there's a reference outside of this class.

source/backend/api/Helpers/Extensions/DispositionFileExtensions.cs
```diff
@@ -10,32 +10,67 @@ namespace Pims.Api.Helpers.Extensions
...
+            if (pimsUser.IsContractor)
+            {
+                return pimsUser.PimsRegionUsers.Any(ur => ur.RegionCode == dispositionFile.RegionCode);
+            }
+
+            // Regular (non-contractor) users are not bound by region access here.
+            return true;
+        }
+
+        public static bool HasAccessToFile(this PimsDispositionFile dispositionFile, PimsUser pimsUser, PimsProject project)
```
eddherrera, line 41: Probably a private static method, not an extension method, unless called outside of this class.

source/backend/api/Helpers/Extensions/AcquisitionFileExtensions.cs
asanchezr: We are throwing a null-check in the Throw version - this one is also public but does not throw so we still need to check the parameters

source/backend/api/Helpers/Extensions/DispositionFileExtensions.cs
asanchezr: This is called in the PrincipalExtensions file so yes it is called outside of this file

eddherrera (approved):

source/backend/api/Helpers/Extensions/AcquisitionFileExtensions.cs
asanchezr: Updated

PR #5364, opened 2026-07-22
PSP-10909, PSP-11666: Changes on Document Types and Metadata types for Leases and Dispositions.

stairaku:


eddherrera:
It looks good but I am concerned about the changes on the casing changes... is that something that Dar requested?

eddherrera (approved):

PR #5365, opened 2026-07-22
PSP-11432 : Patch Playwright

eddherrera:


stairaku (approved):

testing/pims-integration/package.json
```diff
@@ -19,7 +20,7 @@
   "type": "commonjs",
   "devDependencies": {
     "@eslint/js": "9.39.4",
-    "@playwright/test": "1.60.0",
+    "@playwright/test": "^1.61.1",
```
asanchezr: I'm okay with updating the version but please pin it to exact version (remove the ^)

eddherrera: Updated

PR #5368, opened 2026-07-23
PSP-11432 : e2e patch

eddherrera:


.github/workflows/ci-cd-pims-smoke-test.yml
```diff
@@ -31,7 +32,7 @@ jobs:
       runs-on: ubuntu-22.04
       container:
         image: mcr.microsoft.com/playwright:v1.61.0-noble
-        options: --user 1001
+        options: --user 1001 --ipc=host
```
asanchezr: Google says that the ipc option is risky for security. do we really need it?

asanchezr (approved):
Approved but see comments re: ipc flag

PR #5370, opened 2026-07-27
PSP-11568 : Versioning Strategy for PIMS (Maintenance Mode)

eddherrera:


source/frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -1,75 +1,9 @@
...
-  }
-
-  const parsedValue = (() => {
-    if (typeof rawValue === 'string') {
-      if (!rawValue.trim()) {
-        return null;
-      }
-
-      try {
-        return JSON.parse(rawValue) as unknown;
```
asanchezr, line 31: I am okay to remove the warning here but we will need to think about a strategy to alert users when the API and DB are not compatible: I don't think we will run into this scenario but we need a plan for it

source/frontend/src/components/common/ApiVersionInfo.test.tsx
```diff
@@ -63,81 +63,7 @@ describe('ApiVersionInfo suite', () => {
...
-        dbVersion: '114.00',
-      },
-    } as any);
-
-    const { queryByTestId } = setup();
-    await waitForEffects();
-
-    const element = queryByTestId(`version-mismatch-warning`);
-    expect(element).not.toBeInTheDocument();
+    expect(element).toHaveTextContent('v11.1.1');
```
asanchezr: I see that you are dropping the build number from the "publicly displayed version" in PIMS. I'm not sure that would help users report bugs in an specific build and then be reassured that it was fixed (by seeing an incremented build number).

For debugging purposes build numbers are critical because they tell us when the bug/issue was occurring

source/frontend/src/components/common/ApiVersionInfo.tsx
```diff
@@ -100,76 +28,18 @@ export const ApiVersionInfo = () => {
...
-      );
-    }
-
-    return msg;
-  };
+  const frontEndVersionSegments = frontEndVersion?.split('.') ?? [];
+  const displayFrontEndVersion =
+    frontEndVersionSegments.length === 4
+      ? frontEndVersionSegments.slice(0, 3).join('.')
+      : frontEndVersion ?? '';
```
asanchezr: I already mentioned in the test portion that I think we should keep the build number visible, 
Dropping this number would require confirmation with Dar + Cara. I'll remove the make changes flag if they approve it

tools/cicd/build/bump-version.js
```diff
@@ -73,31 +67,29 @@ function updateCsproj(csproj) {
...
 function isValidVersion(version) {
   if (!version) {
     return false;
   }
-  return /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)-(0|[1-9]\d*)\.(0|[1-9]\d*)$/.test(version);
+  return /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*).(0|[1-9]\d*)$/.test(version);
 }
 
 function parse(version) {
   const [semVer, metadata] = version.split('-');
```
asanchezr: we are dropping the "-" component from the version so I think we can remove this split statement

asanchezr (changes requested):
See comments re: dropping build number from public version displayed in PIMS

source/frontend/src/components/common/ApiVersionInfo.test.tsx
eddherrera: You are missing the point. The users don't see build numbers, they just see a release.

tools/cicd/build/bump-version.js
eddherrera: True, let me fix it.

source/frontend/src/components/common/ApiVersionInfo.tsx
eddherrera: They don't need the build number for release. Devs need it for deployment.

asanchezr (approved):

PR #5372, opened 2026-07-29
PSP-11716 : TECH DEBT: Update GitHub actions to Node 24 - Part II

eddherrera:


.github/workflows/ci-cd-pims-dev.yml
```diff
@@ -54,7 +56,7 @@ jobs:
       - name: Checkout Source Code
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #6.0.3
       - name: Login to OpenShift
-        uses: redhat-actions/oc-login@5eb45e848b168b6bf6b8fe7f1561003c12e3c99d # v1.3
+        uses: redhat-actions/oc-login@f19c5f4f98c725396a5eeb265b1338333634828a # v2.0
```
asanchezr: It seems they released a new version to fix crashes in the action. Please update the SHA to the 2.0.1 version

https://github.com/redhat-actions/oc-login/releases/tag/v2.0.1

Fixes
* Replaced @vercel/ncc with esbuild — ncc cannot bundle ESM-only @actions/core v3 and @actions/exec v3 dependencies, producing a broken bundle that crashes at runtime with Cannot find module '@actions/core'

asanchezr (changes requested):

.github/workflows/ci-cd-pims-dev.yml
eddherrera: Updated

asanchezr (approved):
LGTM

PR #5373, opened 2026-07-31
PSP-11426: Proxy caddy updates

asanchezr:
* Fixed values in the proxy-caddy v2 BuildConfig (old values were not being allowed by OpenShift)
* Updated maintenance page from generic message to look more like PIMS

## Old page
maintenance-page-old

## New page

image


stairaku (approved):
Thank you Alejandro!

PR #5380, opened 2026-08-05
Improvements Automation

stairaku:


source/frontend/src/features/mapSideBar/property/tabs/propertyImprovements/list/PropertyImprovementsListView.tsx
```diff
@@ -46,7 +46,7 @@ export const PropertyImprovementsListView: React.FunctionComponent<
         }
       >
         {propertyImprovements?.map((improvement, index) => (
-          <StyledBorder key={improvement.id}>
+          <StyledBorder key={improvement.id} data-testid={`property-improvement-${index}`}>
```
eddherrera, line 49: Please use data-testid={`property-improvement-${improvement.id}`}

source/frontend/src/features/mapSideBar/shared/improvements/FilePropertiesImprovements/FilePropertiesImprovementsView.tsx
```diff
@@ -65,7 +65,7 @@ export const FilePropertiesImprovementsView: React.FunctionComponent<
         >
           {propertyImprovements.improvements.map(
             (improvement: ApiGen_Concepts_PropertyImprovement, index: number) => (
-              <div key={improvement.id}>
+              <div key={improvement.id} data-testid={`property-improvement-${index}`}>
```
eddherrera, line 68: Please use data-testid={property-improvement-${improvement.id}}

testing/pims-integration/pages/research/research-create.page.ts
```diff
@@ -71,8 +71,9 @@ export class ResearchCreatePage extends LayoutPage {
     await this.page.goto('/mapview/sidebar/research/new', { waitUntil: 'domcontentloaded' });
   }
 
-  async setResearchFilenameInput(fileName: string) {
-    await this.researchNameInput.fill(fileName);
+  async fillInField(elementId: string, content: string) {
```
eddherrera, line 74: I think we should keep the single responsibility method patter, instead of using parameters for the element.

testing/pims-integration/tests/e2e/research/research.spec.ts
```diff
@@ -53,46 +56,49 @@ test.describe('Research Files feature', () => {
...
+        },
+        { label: 'Requester', apiValue: apiFeatureFileJson.requestorPerson },
+      ];
+
+      const datesToCompare = [
+        { label: 'Request date', apiValue: apiFeatureFileJson.requestDate },
+        //{ label: 'Research completed on', apiValue: apiFeatureFileJson.researchCompletionDate },
+      ];
+
+      for (const field of fieldsToCompare) {
```
eddherrera, line 92: let's not use this patter of looping. would rather have it explicit expects for each.

testing/pims-integration/tests/e2e/research/research.spec.ts
```diff
@@ -53,46 +56,49 @@ test.describe('Research Files feature', () => {
...
+        const uiValue = await researchViewDetails.getFieldValueByLabel(field.label);
+        expect(normalize(uiValue)).toBe(normalize(field.apiValue));
+      }
+
+      for (const date of datesToCompare) {
+        const uiValue = await researchViewDetails.getFieldValueByLabel(date.label);
```
eddherrera, line 98: let's not use this patter of looping. would rather have it explicit expects for each.

testing/pims-integration/utils/utils.ts
```diff
@@ -31,7 +31,10 @@ export const formatApiDate = (apiDate: string | null | undefined): string => {
   });
 };
 
-export function nullableBooleanToYesNoString(value?: boolean | null): string {
-  if (value === true) return 'Yes';
-  return 'No';
-}
+export const formatApiBoolean = (value: boolean | null | undefined): string => {
+  if (value == null) {
```
eddherrera, line 35: This one needs more validations, and it should use the "===" operator.  you are missing the check for undefined. The only value that should return 'Yes' is 'true'. would rather leave it as it was before.

source/frontend/src/features/mapSideBar/property/tabs/propertyImprovements/list/PropertyImprovementsListView.tsx
stairaku: That makes so difficult to get the element .... the IDs are unique and doesn't allow to have the improvements in the order they appear, that's why I'm using the index.

PR #5384, opened 2026-08-11
PSP-11438: Contacts - PIMS users 

stairaku:
image
image


source/frontend/src/components/contact/ContactManagerView/ContactFilterComponent/ContactFilterComponent.tsx
```diff
@@ -1,29 +1,48 @@
...
+  ONLY_ORGANIZATIONS = 'organizations',
+  ONLY_PIMSUSERS = 'pimsusers',
+}
+
+const allContactTypes = [
+  RestrictContactType.ONLY_PIMSUSERS,
+  RestrictContactType.ONLY_INDIVIDUALS,
+  RestrictContactType.ONLY_ORGANIZATIONS,
+];
+
```
areyeslo: This should be moved to src/constants/Contacts.  If we change ContactFilterComponent would be only if business logic changed, but a change in constants or strings is not a valid case to manipulate this file. I would send these constants outside.

source/backend/dal/Constants/SearchContactTypes.cs
```diff
@@ -0,0 +1,9 @@
+namespace Pims.Dal.Constants
+{
+    public enum SearchContactTypes
+    {
+        PIMSUSERS,
+        PERSONS,
+        ORGANIZATIONS,
+    }
+}
```
areyeslo: That's very close:
```
using System.Runtime.Serialization;

namespace Pims.Dal.Constants
{
    public enum SearchContactTypes
    {
        [EnumMember(Value = "PIMSUSERS")]
        PIMSUSERS,
        [EnumMember(Value = "PERSONS")]
        PERSONS,
        [EnumMember(Value = "ORGANIZATIONS")]
        ORGANIZATIONS,
    }
}
```

source/frontend/src/components/contact/ContactManagerView/ContactFilterComponent/ContactFilterComponent.tsx
```diff
@@ -1,29 +1,48 @@
...
 export const defaultFilter: IContactFilter = {
   summary: '',
   municipality: '',
-  searchBy: 'all',
+  searchBy: allContactTypes,
```
areyeslo: would change to `defaultSearchBy ` with only PIMSUSERS.

stairaku: The ticket tries to implement the new PIMS User option for contact manager, however, if we put PIMS.USERS as the default, that would break the iteraction that we already have around the system since most of the cases, we need the ability to choose individuals and/or organizations. I believe there's another ticket to set PIMS.USERS as a restriction for certain contact manager cases.

source/backend/dal/Constants/SearchContactTypes.cs
stairaku: Should I move this document to src/constants then?

source/frontend/src/components/contact/ContactManagerView/ContactFilterComponent/ContactFilterComponent.tsx
```diff
@@ -72,59 +91,69 @@ export const ContactFilterComponent: React.FunctionComponent<
...
-                      <ActiveFilterCheck<IContactFilter>
-                        fieldName="activeContactsOnly"
-                        setFilter={setFilter}
-                      />
-                      <ActiveIndicator active size={16} />
-                      <span className="ml-1">Show active only</span>
-                    </FlexRowNoGap>
-                  )}
+                  <Row className="pl-5 small">
+                    <CheckGroup
```
areyeslo: Based on the story we would need to check that only Organizations are selected in the CheckGroup. Then, we would need to ensure this scenario, something similar to this under handleChange?
```
handleChange={searchByValues => {
    if (
      searchByValues.includes(RestrictContactType.ONLY_ORGANIZATIONS) &&
      searchByValues.length > 1
    ) {
      setFieldValue('searchBy', [RestrictContactType.ONLY_ORGANIZATIONS]);
    }
  }}

**Comment by areyeslo** [comment]
We should restrict the user to select only `Active` in the checkbox group because this would trigger this  GET request api/contacts/search?activeContactsOnly=true&municipality=&page=1&quantity=10&summary= without a searchBy:
```
{
    "error": "Argument must have a valid filter",
    "type": "ArgumentException",
    "details": "Argument must have a valid filter (Parameter \u0027filter\u0027) ",
}
```
We can have a fallback in the onChange to have always a value for searchBy?
```
if (searchByValues.length === 0) {
    const fallbackSearchBy = restrictContactType
      ? [restrictContactType]
      : [RestrictContactType.ONLY_PIMSUSERS];
    setFieldValue('searchBy', fallbackSearchBy);
    return;
  }
```

source/frontend/src/components/contact/ContactManagerView/ContactFilterComponent/ContactFilterComponent.tsx
eddherrera: Yes, default is not PimsUsers.

areyeslo (approved):
LGTM. Thank you for the changes.

